PR #45, opened 2023-09-14
feat: implement websocket

leanmendoza:


rust/decentraland-godot-lib/src/dcl/js/js_modules/ws.js
```diff
@@ -0,0 +1,137 @@
...
+        this.url = url
+        this.protocols = protocols
+        this.readyState = WebSocket.CONNECTING
+        this._internal_ws_id = Deno.core.ops.op_ws_create(
+            url, protocols ?? []
+        )
+
+        this.bufferedAmount = 0 // TODO: implement
+        this.extensions = "" // TODO: implement
+        this.protocol = "" // TODO: implement
```
kuruk-mm, line 64:
```suggestion
        this.protocol = "" // TODO: implement
        this.binaryType = "blob"
```

rust/decentraland-godot-lib/src/dcl/js/js_modules/ws.js
```diff
@@ -0,0 +1,137 @@
...
+                    if (typeof this.onclose === 'function') {
+                        this.onclose({ type: "close" })
+                    }
+                    break
+                }
+
+
+                if (data.binary_data) {
+                    if (typeof this.onmessage === 'function') {
+                        this.onmessage({ type: "binary", data: data.binary_data })
```
kuruk-mm, line 113: According to:

https://websockets.spec.whatwg.org/#ref-for-dom-websocket-binarytype%E2%91%A0
and
https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType

The binary data should be returned as a `Blob` (by default), and if you specify in `WebSocket.binaryType=arraybuffer` it should return type `ArrayBuffer`

I will guess that `data.binary_data` is `ArrayBuffer`, so here should be the correct code, converting to blob:
```suggestion
                        this.onmessage({ type: "binary", data: this.binaryType === 'blob' ? new Blobl([data.binary_data]) : data.binary_data })
```

kuruk-mm (changes requested):
According to the current [SceneRuntime](https://github.com/decentraland/scene-runtime/blob/2408927cc2f8164f3f7176bd2499723f39825a57/src/common/WebSocket.ts#L7-L27), the Explorer should only allows `wss://`. It would be great to maintain that limitation to have the same behaviour on all platforms (we should do the same for fetch, too).

Another thing that we can add later is the permissions for using WebSocket that are in the scene.json. I want to add a `TODO` for that.

leanmendoza:
> According to the current [SceneRuntime](https://github.com/decentraland/scene-runtime/blob/2408927cc2f8164f3f7176bd2499723f39825a57/src/common/WebSocket.ts#L7-L27), the Explorer should only allows `wss://`. It would be great to maintain that limitation to have the same behaviour on all platforms (we should do the same for fetch, too).
> 
> Another thing that we can add later is the permissions for using WebSocket that are in the scene.json. I want to add a `TODO` for that.

Thanks for the review. Nice catch, we're going to document that extra part of the runtime.

rust/decentraland-godot-lib/src/dcl/js/js_modules/ws.js
leanmendoza: We do not intend to implement the WebAPI here, the implementation adopted by the Scene Runtime (or as far the [exposed by the SDK7](https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/js-runtime/index.d.ts)) is a kind of subset of that. This clean implementation allows the protocol squad to bring more details to the [ADR-133](https://adr.decentraland.org/adr/ADR-133)  (right now with TODO in those parts).

kuruk-mm (dismissed):

PR #50, opened 2023-09-24
chore: copy rust lib only when it changes

kuruk-mm:


leanmendoza (approved):
Thank you for the PR :) Could you change the Spanish comments?

leanmendoza (approved):

PR #61, opened 2023-10-14
feat: add audio stream and fix only current parcel id can plays audio

leanmendoza:
- [x] fix: in the party-time scene the audio is played twice
- [x] apply and verify component properties
- [x] load local resources (videos)

### DIFF
--- a/rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
+++ b/rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
@@ -391,28 +391,24 @@ impl SceneManager {
                         (dirty_entities, dirty_lww_components, dirty_gos_components),
                         logs,
                         _,
+                        rpc_calls,
                     ) => {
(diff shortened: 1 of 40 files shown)


rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
```diff
@@ -243,6 +243,12 @@ impl SceneManager {
                     audio_source_node.bind_mut().set_dcl_enable(false);
                     audio_source_node.call("apply_audio_props".into(), &[false.to_variant()]);
                 }
+                for (_, audio_stream_node) in scene.audio_streams.iter_mut() {
+                    audio_stream_node.set_volume_db(-80.0);
```
kuruk-mm, line 247: can use create a function named `set_mute(true);` that sets the volume to -80 instead of using in this way?

kuruk-mm (approved):

kuruk-mm:
Great work! Everything looks great

PR #62, opened 2023-10-15
feat: add scenes rpcs thought tokio oneshot channels and change realm for restricted action api

kuruk-mm:
Extra BugFix for CameraModeArea: when you unload the scene while you're in a CameraModeArea, it triggers the exit of the area.


rust/decentraland-godot-lib/src/dcl/js/js_modules/RestrictedActions.js
```diff
@@ -1,7 +1,9 @@
 module.exports.movePlayerTo = async function (body) { return {} }
 module.exports.teleportTo = async function (body) { return {} }
 module.exports.triggerEmote = async function (body) { return {} }
-module.exports.changeRealm = async function (body) { return {} }
+module.exports.changeRealm = function (body) {
+    return Deno.core.ops.op_change_realm(body.realm, body.message);
+}
```
leanmendoza, line 6:
```suggestion
module.exports.changeRealm = async function (body) {
    const response = await Deno.core.ops.op_change_realm(body.realm, body.message)
    return response
}
```

It's not bad as you put, but I'd write more verbosely because as we're in JavaScript and not TypeScript.

leanmendoza (changes requested):
The RPC implementation is good, I'm not really sure about setting callbacks for the answers, maybe use directly with channels?

leanmendoza (approved):

PR #70, opened 2023-11-06
refactor: http request, remove content stages, and add content thread pool

kuruk-mm:


godot/src/logic/content_thread.gd
```diff
@@ -0,0 +1,504 @@
...
+	return pending_content.size()
+
+
+func _init(id: int, thread: Thread):
+	self.thread = thread
+	self.id = id
+
+
+func process(content_cache_map: Dictionary, wearable_cache_map: Dictionary):
+	http_requester.poll()
```
leanmendoza, line 34: I don't understand this, if it supposes to use co-routines below why it needs to call to poll after the entire resource fetching?

leanmendoza (approved):

godot/src/logic/content_thread.gd
kuruk-mm: You are right.
The `process` is not being awaited correctly. I changed the code of the process to:

```gdscript
func process(content_cache_map: Dictionary):  # not a coroutine
	_http_requester.poll()

	if _pending_content.is_empty():
		return

	var loading_content: Array[Dictionary] = []
	while _pending_content.size() > 0:
		loading_content.push_back(_pending_content.pop_front())

	for content in loading_content:
		# coroutine not awaited, it can process multiple times
		co_process_content(content, content_cache_map)
```

So now, it's spawns `co_process_content` as many times as needed in the thread, and they are processed using the `_http_requester`.

Issue #75, opened 2023-11-07
`Condition "det == 0" is true` at 81, -6


kuruk-mm:
The Godot Engine spams the error `Condition "det == 0" is true.` every frame.

This is happening in the Goerli Plaza Scene in 81, -6

Scene code:
https://github.com/decentraland/sdk7-goerli-plaza/tree/main/lazy-loading

Probably it's happening because it's using `scale(0,0,0)` to hide the scenes.

```typescript
export function hideScene(scene: Entity) {
  Transform.get(scene)
  Transform.createOrReplace(scene, { scale: Vector3.create(0, 0, 0) })
}

export function showScene(scene: Entity) {
  Transform.get(scene)
  Transform.createOrReplace(scene, { scale: Vector3.create(1, 1, 1) })
}
```

leanmendoza:
fixed https://github.com/decentraland/godot-explorer/pull/127

PR #77, opened 2023-11-08
fix: `condition "det == 0" is true`

kuruk-mm:
Fix https://github.com/decentraland/godot-explorer/issues/75

@leanmendoza too hacky? :innocent: 

Looks like it's a Godot bug.


leanmendoza:
tackled on https://github.com/decentraland/godot-explorer/pull/127
extended version of that condition check with each component

Issue #78, opened 2023-11-08
CameraModeArea no scene limits


kuruk-mm:
If you create a CameraModeArea that is bigger than the scene, it will not be limited, and it will affect other scenes.

PR #80, opened 2023-11-08
fix: CameraModeArea limits, they only works in the active scene

kuruk-mm:
feat: new on_active_scene component that emits a signal when the scenes becames active or non-active

Fix https://github.com/decentraland/godot-explorer/issues/78

# How to use it?
- Add the `OnActiveScene.tscn` to your scene.
- Connects the signal `on_scene_active` to your code.
- Do whatever you want, when the scene became active.

### DIFF
--- a/rust/decentraland-godot-lib/src/godot_classes/dcl_scene_node.rs
+++ b/rust/decentraland-godot-lib/src/godot_classes/dcl_scene_node.rs
@@ -1,0 +1,42 @@
+use godot::engine::Node3D;
+use godot::prelude::*;
+
+#[derive(GodotClass)]
+#[class(init, base=Node3D)]
+pub struct DclSceneNode {
+    scene_id: u32,
+
+    is_global: bool,
+
+    #[base]
+    _base: Base<Node3D>,
+}
+
+#[godot_api]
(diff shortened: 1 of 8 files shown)


rust/decentraland-godot-lib/src/godot_classes/dcl_scene_node.rs
```diff
@@ -0,0 +1,42 @@
+use godot::engine::Node3D;
+use godot::prelude::*;
+
+#[derive(GodotClass)]
+#[class(init, base=Node3D)]
+pub struct DclSceneNode {
+    scene_id: u32,
+
+    is_global: bool,
```
leanmendoza, line 9: This is a property of the scene, not of the entity. Could you move this to the SceneRoot and then go until that?

leanmendoza (changes requested):

rust/decentraland-godot-lib/src/godot_classes/dcl_scene_node.rs
leanmendoza: misunderstood, forget it

leanmendoza (approved):

PR #88, opened 2023-11-10
refactor: use new promises

kuruk-mm:
From https://github.com/kuruk-mm/gdscript-async-utils


leanmendoza (approved):
LGTM, change the `utils` folder with something more descriptive (maybe `async`, `promises`)

kuruk-mm:
> LGTM, change the `utils` folder with something more descriptive (maybe `async`, `promises`)

I would prefer to let the folder at it's. When I get the Asset approved in the Godot Asset Library, I'm going to add it as an `addon`.

Util folder can be useful for other usages.

PR #102, opened 2023-11-22
refactor: add gdscript linter and async-function-name linter

kuruk-mm:
- Use GDLinter https://github.com/Scony/godot-gdscript-toolkit/wiki/3.-Linter
- [Fork GDLinter](https://github.com/kuruk-mm/godot-gdscript-toolkit/commit/ba4428a140246c6bd0ab4eb8b547853223b88a28) and implement async-function-name, and apply to the project
- Added CONTRIBUTING.md section, with pre-hooks instructions and the script for it


godot/src/logic/realm.gd
```diff
@@ -19,6 +19,9 @@ static func dcl_world_url(dcl_name: String) -> String:
 
 
 static func ensure_ends_with_slash(str_param: String) -> String:
+	if str_param.is_empty():
```
leanmendoza, line 22: nice catch

godot/src/ui/components/backpack/backpack.gd
```diff
@@ -114,7 +118,7 @@ func show_wearables():
 		var wearable_item = wearable_item_instanceable.instantiate()
 		grid_container_wearables_list.add_child(wearable_item)
 		wearable_item.button_group = items_button_group
-		wearable_item.set_wearable(wearable_data[wearable_id])
+		wearable_item.async_set_wearable(wearable_data[wearable_id])
```
leanmendoza, line 121: is it okay this doesn't have an await?

leanmendoza (approved):

godot/src/ui/components/backpack/backpack.gd
kuruk-mm: Yes. It was the previous behaviour. Without a `await` it will forget the promise. So multiple `async_set_wearable` can be executed at the same time.

Maybe we should add an await. But if I added an await there, I would change the behaviour. So for that case, we must do it in another PR.

PR #104, opened 2023-11-23
test: add visual snapshots and comparison

leanmendoza:
This PR introduces a method in the `~system/Testing` API to generate a visual snapshot of the Scene and compare it with the existing one (if it applies).
The current method for comparison is to compute the grayscale from the NTSC standard of each image (the stored snapshot and the just taken one), calculate the absolute difference, and then compute a standard deviation.
The similarity is 1.0, completely similar to 0.0, completely different.

branch test/snapshot
https://github.com/sdilauro/dae-unit-tests


robtfm:
looks good, although the path being user:// is unix-centric.

if we had user agent in engine info (or some other way to determine the client) we could have the screenshots as scene content.

godot/src/logic/realm.gd
```diff
@@ -115,9 +115,11 @@ func async_set_realm(new_realm_string: String) -> void:
 			if parsed_urn != null:
 				realm_global_scene_urns.push_back(parsed_urn)
 
-		realm_city_loader_content_base_url = Realm.ensure_ends_with_slash(
-			configuration.get("cityLoaderContentServer", "")
-		)
+		realm_city_loader_content_base_url = configuration.get("cityLoaderContentServer", "")
```
leanmendoza, line 118: fix realm_city_loader_content_base_url always = "/"

godot/src/logic/content/content_thread.gd
```diff
@@ -270,7 +270,10 @@ func _async_process_loading_texture(
 
 		var content_result = await promise.async_awaiter()
 		if content_result is Promise.Error:
-			printerr("Failing on loading gltf ", url, " reason: ", content_result.get_error())
+			content_cache_map[file_hash]["promise"].call_deferred(
```
leanmendoza, line 273: fix wait for ever the texture fetch

rust/decentraland-godot-lib/src/dcl/js/testing.rs
```diff
@@ -0,0 +1,152 @@
...
+            if value.is_none() {
+                std::thread::sleep(std::time::Duration::from_millis(100));
+            } else {
+                break;
+            }
+        }
+        value.unwrap()
+    };
+
+    // let response = rx.blocking_recv();
```
kuruk-mm, line 134:
```suggestion
```

kuruk-mm (approved):
Looks great, minor reviews

PR #105, opened 2023-11-23
feat: web3 signin

leanmendoza:
The main goals of this PR is to add the web3 signing in, in this case through a brwoser login and sign a ephemeral identity.

Summary:
- Add the concept of RemoteWallet 
  - [X] test sign-in with Metamask
  - [X] test sign-in with Magic
- Add Ephemeral Identity in the equation of signed-login (livekit and ws-room) instead of direct signing
- Force being logged to connect a comms

Next steps:
- [ ] Improve React App (visually)
- [ ] Integrate Web3 scene apis
- [ ] Fetch/Deploy profile
- [ ] Known issue: the first time you authenticate in `decentraland-auth` web app, it fails, then when you retry, it works fine (and the next times also)


rust/decentraland-godot-lib/src/auth/with_browser_and_server.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+pub async fn get_account(
+    url_reporter: tokio::sync::mpsc::Sender<RemoteReportState>,
+) -> Result<(H160, u64), ()> {
+    let get_account_req_id = gen_id();
+    let server_endpoint = urlencoding::encode(AUTH_SERVER_ENDPOINT_URL);
+    let open_url: String = format!(
+        "{AUTH_FRONT_URL}get-account?id={get_account_req_id}&server-endpoint={server_endpoint}"
+    );
```
kuruk-mm, line 120: Can we parse using Url from hyper?

```rust
let mut url = Url::parse(base_url)?;

// Add parameters
{
    let mut query_pairs = url.query_pairs_mut();
    query_pairs.append_pair("key1", "value1");
    query_pairs.append_pair("key2", "value2");
    // Add more parameters as needed
} // The mutable borrow of url ends here
```

kuruk-mm (changes requested):
Great!

Just minor review

PR #106, opened 2023-11-23
feat: remove scenes with different scene_id when changing realm

kuruk-mm:
When changing the realm, only remove scenes that have different scene_id


kuruk-mm:
We decided with @leanmendoza that we're going to do this in the future with a better refactor.

PR #112, opened 2023-11-30
feat: godot 4.2

kuruk-mm:
Depends on: https://github.com/decentraland/godot-explorer/pull/111

Re-imported assets were required due to the following warning:

```
WARNING: At least one surface uses an old surface format and needs to be upgraded. The upgrade happens automatically at load time every time until the mesh is saved again or re-imported. Once saved (or re-imported), this mesh will be incompatible with earlier versions of Godot.
```


leanmendoza (approved):

.github/workflows/ci.yml
```diff
@@ -125,7 +125,7 @@ jobs:
   build:
     name: Build and test
     # Comment next line to test this job on PRs
-    if: ${{ github.ref == 'refs/heads/main' }} 
+    #if: ${{ github.ref == 'refs/heads/main' }}
```
leanmendoza, line 128:
```suggestion
    if: ${{ github.ref == 'refs/heads/main' }} 
```

PR #114, opened 2023-11-30
feat: debug panel with console

kuruk-mm:
- [x] Fix focus problems
- [x] Create right click dialog menu


leanmendoza (approved):
Looks good :) And the panel is excellent, you can add a screenshot in the PR description for nostalgia later haha
Two things: 
- Avoid using $ and % accesors directly in the methods. Instead, you can use `@onready` statement and then re-use it in the code. It's cleaner then to modify the node path
- No actionable now, but we can make custom controls with the predef config like focus mode inheriting the Button for example, and use CustomUiButton instead in every case

I left the approval anyway, you can merge it after the @onready thing

PR #168, opened 2024-01-16
fix: content loading related crashes (android)

leanmendoza:
# What?
Avoid duplicating to load the same resource more than once, with `ContentNotificator` applied to the downloads all the shared dependencies awaits the first request.
Use `tokio::fs` instead of `godot::FileAccess` (crashing when multiple file handlers in the same file)

- feat: add content notification to hold futures of downloads 
- fix: move to std::fs/tokio::fs instead of godot FileAccess

## Minor
- fix: now it waits until promises are resolved before take snapshot
- fix: tooltip when action=any
- chore: move PackedByteArray conversion from Vec to a bytes module  
- fix: race condition when `getUserData` is called after `onStart` but it's not resolved by the `async` engine
- chore: remove nop()
- mock ~system/CommunicationsController::sendBinary
- fix: remove error message when texture loading failed at first time (before ensuring is fetched)
- use tokio::fs to wrtie files

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -2,8 +2,8 @@ name: Android Build
 
 on:
   push:
-    # branches:
-    #   - main
+    branches:
+      - main
 
 jobs:
   build:
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -2,8 +2,8 @@ name: Android Build
 
 on:
   push:
-    branches:
-      - main
+    # branches:
+    #   - main
 
 jobs:
   build:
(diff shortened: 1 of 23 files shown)


.github/workflows/android-build.yml
```diff
@@ -2,8 +2,8 @@ name: Android Build
 
 on:
   push:
-    branches:
-      - main
+    # branches:
+    #   - main
```
kuruk-mm, line 6:
```suggestion
     branches:
       - main
```

kuruk-mm (approved):
Greatt!! Awesome :raised_hands:

PR #175, opened 2024-01-20
feat: loading screen

kuruk-mm:
- [x] Add loading screen
- [x] Add popup warning when the loading doesn´t change for 20 seconds and allow you to reload or hide the loading
- [x] Fix scene metadata loading async issue

Test popup warning:
In `res://src/ui/components/loading_screen/loading_screen_progress_logic.gd` change `set_physics_process(true)` to `set_physics_process(false)`. It the loading screen is going to stop to report progress. All the time at 0%. Wait 20 seconds


godot/src/ui/components/pagination/Pagination.gd
leanmendoza, line 1: filename => `pagination.gd`

leanmendoza (changes requested):
It looks good overall.
One thing: should also be shown when you teleport to a non-loaded scene.

PS: The `sdk-test-scenes.decentraland.zone` is always stuck at 20%, but I think it's unrelated.

leanmendoza (approved):

Issue #118, opened 2023-12-08
Implement `face_direction` in Tween Component


kuruk-mm:
The `face_direction` parameter from Tween Component for changing a transform to face the destination position is not implemented

Code example:
```typescript
const box = engine.addEntity()
Transform.create(box, {
  position: Vector3.create(4, 2, 4),
  scale: Vector3.create(0.25, 0.5, 1.0)
})
MeshRenderer.setBox(box)
MeshCollider.setBox(box)

Tween.create(box, {
  mode: Tween.Mode.Move({
    start: Vector3.create(4, 2, 4),
    end: Vector3.create(4, 2, 12),
    faceDirection: true
  }),
  duration: 2000,
  easingFunction: EasingFunction.EF_LINEAR,
})
```

kuruk-mm:
Fixed by https://github.com/decentraland/godot-explorer/pull/176

PR #176, opened 2024-01-22
implement face_direction(tween) and top/bottom to cylinder (mesh collider)

### LINKED ISSUE #139: Cylinder with top and bottom and top
Mesh Collider is not following the same as the Renderer for the Cylinder

leanmendoza:
Fixes #139 #118


kuruk-mm (approved):
Great! Tested and it's working!

Issue #149, opened 2024-01-11
Mobile Interactions (Primary/Secondary)


kuruk-mm:
- [x] Add primary/secondary (like E or F in Desktop) interactions when you focus an object that has those interactions. It should show buttons depending in which interaction it can play

PR #196, opened 2024-01-25
feat: mobile interactions with input actions

kuruk-mm:
Fix https://github.com/decentraland/godot-explorer/issues/149

- is_mobile moved to Rust DclGlobal.rs, and set from Global.gd
- Fix interactions bugs
- Add icons, and effect to pressing a interaction


godot/src/ui/components/pointer_tooltip/pointer_tooltip.gd
```diff
@@ -13,13 +13,17 @@ func set_pointer_data(interacts_array: Array):
 		child.queue_free()
 	var i = 0
 	for interact in interacts_array:
+		if i >= angles.size():
```
leanmendoza, line 16: Does this discard the remaining interactions?

godot/src/global.gd
```diff
@@ -30,11 +30,13 @@ var nft_frame_loader: NftFrameStyleLoader
 var standalone = false
 var dcl_android_plugin
 
-@onready var is_mobile = OS.has_feature("mobile")
-#@onready var is_mobile = true
-
 
 func _ready():
+	# First
+	_set_is_mobile(OS.has_feature("mobile"))
```
leanmendoza, line 36: could this be in the `init` function in Rust?

rust/decentraland-godot-lib/src/godot_classes/dcl_global.rs
```diff
@@ -34,11 +34,14 @@ mod android {
     }
 }
 
+const IS_MOBILE: bool = cfg!(target_os = "android");
```
leanmendoza, line 37: this won't be true for VR haha

leanmendoza (approved):
LGTM, just some comments.

rust/decentraland-godot-lib/src/godot_classes/dcl_global.rs
kuruk-mm: Haha, we can take care on that moment

godot/src/ui/components/pointer_tooltip/pointer_tooltip.gd
kuruk-mm: Yes. It avoids a crash. In the Foundation Client is crashing. And Godot was crashing too due to not having enough angles.
There is no limit to how many tooltip interactions you can have. 5 it's more than ok.

PR #199, opened 2024-01-26
fix: avatar attach collider and send raycast_hit in global pointer events

### LINKED ISSUE #55: Collision inside AvatarAttach and reparenting
When an entity is parenting by a `AvatarAttach`, the colliders inside that tree are disabled. This is restored when the `AvatarAttach` is removed.

The case when a collider is added or removed after the `AvatarAttach` is not covered. Race conditions can appear here.

Possible solution: create a flag in the `EntityNode` to know if the entity should disable collisions or not.

### LINKED ISSUE #192: Beer scene not working well

leanmendoza:
Fix #55 
Fix #192


godot/src/decentraland_components/avatar_attach.gd
```diff
@@ -12,30 +15,68 @@ extends Node
...
-	_set_children_physics(self.get_parent(), false)
+	for old_collider in _computed_colliders:
+		# Corner case when is moved from one avatar attach to other
+		if not is_parented_with_avatar_attach(old_collider):
+			old_collider.collision_layer = old_collider.get_meta("dcl_col")
+
+
+func is_parented_with_avatar_attach(_collider: Node) -> bool:
+	# TODO
+	return false
```
kuruk-mm, line 69:
```suggestion
func is_parented_with_avatar_attach(_collider: Node) -> bool:
	# TODO
	return false
```

what is the purpose of this func?

kuruk-mm (approved):
Looks great!

godot/src/decentraland_components/avatar_attach.gd
leanmendoza: it should return true when the collider is an nth-level child of some avatar attach.
the purpose of the usage above is to avoid removing collider (see comment)

PR #203, opened 2024-01-29
feat: new custom popup

sdilauro:
image
Design only, need to implement navigation and how to populate


kuruk-mm:
Not going to be used for now

PR #204, opened 2024-01-29
feat: add side-bar

sdilauro:
with methods like `.push(child)` and `.pop(child)` to navigation.
image


kuruk-mm:
Not going to be used for now

PR #206, opened 2024-01-30
feat: Add teleport (jump in scene preview) 

sdilauro:
Layout only. For include in a sidebar or popup
image


kuruk-mm:
Used in https://github.com/decentraland/godot-explorer/pull/211

Issue #197, opened 2024-01-25
UI Design Debt (Track)


kuruk-mm:
Topic to track the design debt on the UI

- [ ] Pointer interaction must have an icon that represents a pointer [current](https://github.com/decentraland/godot-explorer/assets/12563266/d57f426a-0ab7-48a3-a8b5-231f4ccb9848)
- [ ] The HUD should have two buttons to act as Primary/Secondary global actions

PR #211, opened 2024-01-31
feat: discover

sdilauro:
Layout only:

image


sdilauro:
Icons are provisory

kuruk-mm:
Generated tech debt:
`Round shader for textures on TextureRects for Discover mainly and other places`
added to: https://github.com/decentraland/godot-explorer/issues/197

godot/.godot/uid_cache.bin
leanmendoza, line 1: This should be not commited

godot/src/config/config_data.gd
```diff
@@ -116,6 +120,38 @@ var audio_general_volume: float = 100.0:
...
+		var place_realm = place.get("realm")
+		var place_position = place.get("position")
+		if place_realm == realm:
+			if Realm.is_genesis_city(realm):
+				if place_position == position:
+					to_remove.push_front(i)
+			else:
+				to_remove.push_front(i)
+
+	for i in to_remove:
```
leanmendoza, line 138: This works because of the way items are pushed into the vector.
Add a comment or use by_reference remove (`to_remove.push(_last_places[i])` and then use erase). Don't think the performance makes a big difference here

leanmendoza (approved):
Looks great :D Only two comments

PR #216, opened 2024-02-06
feat: opensea nft v2 and multiple image formats for fetch textures

kuruk-mm:
feat: add multiple image formats for fetch texture


rust/decentraland-godot-lib/src/utils/infer_mime.rs
```diff
@@ -0,0 +1,88 @@
+/// Returns whether a buffer is JPEG image data.
```
kuruk-mm, line 1: I tried using the `infer` crate. But some formats are missing (SVG, KTX, TGA).

leanmendoza (approved):

PR #217, opened 2024-02-07
feat: iterate settings and resolution

### LINKED ISSUE #156: Settings UI

### LINKED ISSUE #194: UI resizing and modes
Currently, the mode used to scale ui is holding a base resolution and aspect ratio and fits everything in that aspect.
- This can be very useful for the explorer UI itself, but it doesn't seem to be true for the scenes UIs
- There is an issue with the TextShape and the avatar label (both use Label3D) in different resolutions, there is no consistent resolution in this sense.

leanmendoza:
Fix #156 
Fix #194 

# What?
- Replace `canvas_item` mode by `disabled` mode and a manual set of UI scale. The UI Scale is worked as `ui_zoom` (to approximate the concept of zooming in a web browser). The maximum value is set as default on hDPI devices (so the behaviour is kept)
- Add AudioSettings => general volume
- Add `rendering_3d_scale` setting for low-end devices
- Remove window_size settings to simplify the logic
- Add `windowed` option in Desktop
- Move scene loading radius config to General


godot/src/config/config_data.gd
```diff
@@ -5,9 +5,9 @@ signal param_changed(param: ConfigParams, new_value)
 
 enum ConfigParams {
 	CONTENT_DIRECTORY,
-	RESOLUTION,
-	WINDOW_SIZE,
-	UI_SCALE,
+	WINDOWED,
+	UI_ZOOM,
+	RENDERING_3D_SCALE,
```
kuruk-mm, line 10: I would prefer "RESOLUTION_3D_SCALE", but up to you

kuruk-mm (approved):

PR #222, opened 2024-02-08
fix: target position for spawn points

kuruk-mm:
Sometimes the target_position is an array with just one element, and it was not considered in the first iteration of the spawn points.

This PR fix that.

### DIFF
--- a/godot/src/logic/scene_fetcher.gd
+++ b/godot/src/logic/scene_fetcher.gd
@@ -44,19 +44,17 @@ func _ready():
 	Global.loading_finished.connect(self.on_loading_finished)
 
 
-func get_target_position(target_position: Variant) -> int:
+func get_target_position(target_position: Variant) -> float:
 	if target_position is Array:
 		if target_position.is_empty():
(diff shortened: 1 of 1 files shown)


godot/src/logic/scene_fetcher.gd
```diff
@@ -44,6 +44,19 @@ func _ready():
...
+			return 0
+
+		if target_position.size() == 1:
+			return target_position[0]
+
+		# size >= 2
+		return randf_range(target_position[0], target_position[1])
+	return target_position
+
+
```
leanmendoza, line 59:
```suggestion
func get_target_position(target_position: Variant) -> float:
	if target_position is Array:
		if target_position.is_empty():
			return 0.0

		if target_position.size() == 1:
			return float(target_position[0])

		# size >= 2
		return randf_range(target_position[0], target_position[1])
	return float(target_position)


```

leanmendoza (changes requested):

leanmendoza (approved):

PR #257, opened 2024-02-13
[CONCEPT] feat: add profile with a front camera of your player

kuruk-mm:
This is just a concept. We need to decide if we want it.

No performance issues found in Mobile.

https://github.com/decentraland/godot-explorer/assets/12563266/27519989-4a01-4a19-923f-d82c01307dad


kuruk-mm:
Not going to be implemented

PR #260, opened 2024-02-16
feat: send emote data to scenes and avatar shape fixes

leanmendoza:
# What?
- emotes triggered by primary player and comms are sent to the scenes
- fix avatar shape
- fix preview hot-reload
- avatar_shape can send emotes by appending AvatarEmoteCommand
- fix shared animation node state (it was replicating the current emote)


kuruk-mm (approved):
Great!

Issue #293, opened 2024-02-20
(82, -5) mirror-puzzle


sdilauro:
- [ ] Mirrors don't have rotation or push interactions in Godot.
- [ ] We can't test the light ray without mirror interaction.

Godot:
![image](https://github.com/decentraland/godot-explorer/assets/22157752/c503900e-589f-4746-a41e-d69e8e647c55)

Foundation:
![image](https://github.com/decentraland/godot-explorer/assets/22157752/c29315d3-825e-4f28-81e6-2079aa1bbe56)

Issue #63, opened 2023-10-19
Pointer event distance


leanmendoza:
Distance is being ignored to compute pointer events tooltips and effect.

PR #358, opened 2024-03-04
fix: pointer events, and add mesh name to the raycast result

kuruk-mm:
feat: add mesh name to the raycast result

Fix https://github.com/decentraland/godot-explorer/issues/293
Fix https://github.com/decentraland/godot-explorer/issues/63
Fix https://github.com/decentraland/godot-explorer/issues/359

### DIFF
--- a/rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
+++ b/rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
@@ -654,7 +654,7 @@ impl SceneManager {
         let scene_position = scene.godot_dcl_scene.root_node_3d.get_position();
         let raycast_data = RaycastHit::from_godot_raycast(
             scene_position,
-            raycast_from,
+            self.player_node.get_position(),
             &raycast_result,
             Some(dcl_entity_id as u32),
         )?;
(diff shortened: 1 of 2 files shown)


rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
```diff
@@ -939,80 +939,69 @@ impl INode for SceneManager {
             &current_pointer_raycast_result,
         );
 
-        let should_update_tooltip = !GodotDclRaycastResult::eq_key(
```
leanmendoza, line 942: Is there any possibility of keeping this?

rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
```diff
@@ -939,80 +939,69 @@ impl INode for SceneManager {
...
                         }
                     }
                 }
             }
-
-            self.set_pointer_tooltips(tooltips);
-            self.base.emit_signal("pointer_tooltip_changed".into(), &[]);
         }
 
+        self.set_pointer_tooltips(tooltips);
```
leanmendoza, line 1002: If this will change every frame, maybe a signal is not necessary?

leanmendoza (approved):
It looks good. Only a question about removing the same pointer event check.

rust/decentraland-godot-lib/src/scene_runner/scene_manager.rs
kuruk-mm: The distance changes on every frame and the eq_key just compare if the scene and entity are the same than the last one

PR #361, opened 2024-03-05
bunch of fixes

leanmendoza:
# What?
- Revert not generate collider for skinned meshes
- Support `AvatarTexture` for `UiBackground`
  - Implement `fetch_profile` to `ContentProvider`
  - The test does not say anything https://github.com/decentraland/scene-explorer-tests/issues/25
- Fix change camera mode inside AvatarModifierArea broken
- Don't show loading screen when preview is reloading
- Make `--skip-lobby` keeps the web3 session if exists
- Send `camera` and `player` transform only when they change
- Move RealmData inside scene-thread so it's available for APIs
  - Add SignedFetch sends proper data
- At scene-thread spawn: removed `#[allow(clippy::too_many_arguments)]` by implementing a param struct
- Update `cursor_position` when the canvas is resized (used by pointer raycast)


godot/src/main.gd
```diff
@@ -40,7 +40,7 @@ func _start():
 		)
 	elif args.has("--scene-renderer"):
 		get_tree().change_scene_to_file("res://src/tool/scene_renderer/scene.tscn")
-	elif args.has("--scene-test") or args.has("--skip-lobby"):
+	elif args.has("--scene-test") or args.has("--skip-lobby-guest"):
```
kuruk-mm, line 43:
```suggestion
	elif args.has("--scene-test") or args.has("--skip-lobby-guest"):
```

I think we can maintain just '--skip-lobby' and always try to recover the last session; if there is not, join as a guest...

rust/decentraland-godot-lib/src/avatars/avatar_type.rs
```diff
@@ -103,7 +103,7 @@ impl DclAvatarWireFormat {
     }
 
     #[func]
-    fn get_snapshots_face256(&self) -> GString {
+    pub fn get_snapshots_face256(&self) -> GString {
```
kuruk-mm, line 106: maybe im skipping something, but I didnt found where it used to make it public

rust/decentraland-godot-lib/src/content/profile.rs
```diff
@@ -0,0 +1,58 @@
...
+    let lamda_server_base_url = realm.bind().get_lambda_server_base_url().to_string();
+    let profile_base_url = realm.bind().get_profile_content_url().to_string();
+    let http_requester = global
+        .bind()
+        .get_http_requester()
+        .bind()
+        .get_http_queue_requester();
+
+    let lamda_server_base_url = if lamda_server_base_url.is_empty() {
+        "https://peer.decentraland.org/lambdas/".to_string()
```
kuruk-mm, line 26: can we add consts like:

https://github.com/decentraland/godot-explorer/blob/9cc068ec6fa9a77691a91cc3db662fce18e5161b/rust/decentraland-godot-lib/src/godot_classes/dcl_realm.rs#L39-L45

to make it easier in the future to change it?

note about the func that I referred: self is needed on the future to access the parameter problem

kuruk-mm (approved):
Great work! Minor reviews

PR #377, opened 2024-03-26
fix: avatar loading issues

### LINKED ISSUE #301: Avatar is not hidding parts of the body when equips a wearables that request it
image

This can be reproduced by equipping the black gloves with default wearable

This happens with the HEAD, LEGS

Example: Mask that hides HEAD
image

BODY that HIDES LEGS
image

leanmendoza:
Fix #301 

- feat: Add `forceRender` property
- fix: Make `name` optional, some wearables fail when loading because the deserializer couldn't find the property
- chore: move wearable vlidation to rust
- make `get_wearable_representation` body_shape_id param case insensitive
- fix: hidings not working well

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar.gd
+++ b/godot/src/decentraland_components/avatar/avatar.gd
@@ -3,6 +3,9 @@ extends DclAvatar
 
 signal avatar_loaded
 
+# Debug to store each avatar loaded in user://avatars
+const DEBUG_SAVE_AVATAR_DATA = true
+
 @export var skip_process: bool = false
 @export var hide_name: bool = false
 @export var non_3d_audio: bool = false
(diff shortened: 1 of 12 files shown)


godot/src/decentraland_components/avatar/avatar.gd
```diff
@@ -3,18 +3,18 @@ extends DclAvatar
 
 signal avatar_loaded
 
+# Debug to store each avatar loaded in user://avatars
```
kuruk-mm, line 6: Default should be FALSE, and this can be added as another script, and connecting with a signal.

kuruk-mm (approved):
Great work! Just minor reviews

godot/src/decentraland_components/avatar/avatar.gd
```diff
@@ -3,18 +3,18 @@ extends DclAvatar
...
+const DEBUG_SAVE_AVATAR_DATA = true
```
leanmendoza, line 7:
```suggestion
const DEBUG_SAVE_AVATAR_DATA = false
```

godot/src/decentraland_components/avatar/avatar.gd
leanmendoza: changed to `false` and created https://github.com/decentraland/godot-explorer/issues/378

PR #379, opened 2024-03-26
fix: correct promise base type initialization to prevent crashes on macOS and iOS

kuruk-mm:


leanmendoza (approved):

godot/export_presets.cfg
```diff
@@ -234,7 +234,7 @@ architectures/armeabi-v7a=false
 architectures/arm64-v8a=true
 architectures/x86=false
 architectures/x86_64=true
-version/code=5
+version/code=6
```
leanmendoza, line 237: incremented

PR #380, opened 2024-03-27
fix: comms sender blocks the javascript runtime

leanmendoza:
- fix: comms sender blocks the javascript runtime
- fix: ui_canvas_information is only sent when there is UI stuffs
- feat: add v8-inspector support (with feature `enable_inspector`, no default)
- feat: run scene-code without wrapping it with `require`
- fix: idle_anim is shared between avatars
- chore: remove some logs

### DIFF
--- a/rust/decentraland-godot-lib/src/dcl/js/mod.rs
+++ b/rust/decentraland-godot-lib/src/dcl/js/mod.rs
@@ -3,6 +3,8 @@ mod engine;
 mod ethereum_controller;
 mod events;
 mod fetch;
+#[cfg(feature = "enable_inspector")]
+mod inspector;
 mod players;
 mod portables;
 mod restricted_actions;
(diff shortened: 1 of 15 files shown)


rust/decentraland-godot-lib/src/dcl/js/mod.rs
```diff
@@ -206,18 +242,22 @@ pub(crate) fn scene_thread(
         .borrow_mut()
         .put(SceneStartTime(std::time::SystemTime::now()));
 
+    if inspector.is_some() {
+        // TODO: maybe send a message to announce the inspector is being waited
```
kuruk-mm, line 246: A log just not be enought?

kuruk-mm (approved):
Great

rust/decentraland-godot-lib/src/dcl/js/mod.rs
leanmendoza: Inpsector is already sending a log with the ws addres, if you have the inspector opened, it's automatically listening for new host with 9222 port, so it's not necessary, the feedback can be thought in the future for content creators debugging

PR #392, opened 2024-04-08
fix: tween doesn't use the duration so it fallbacks to 0 and finishes immediately

leanmendoza:
this effect is evident in the switchable-platform scene, when you press the button and the platform goes instantaneously to the end
also:
- feat: update last audio source spec
- fix: some mesh renderer disappear because the depth-test (material)
- fix: with scene-radius==0 when changing parcel position, it shows the loading screen and call moveTo

### DIFF
--- a/godot/src/decentraland_components/audio_source.gd
+++ b/godot/src/decentraland_components/audio_source.gd
@@ -2,6 +2,7 @@ extends DclAudioSource
 
 var last_loaded_audio_clip := ""
 var valid := false
+var _time_specified := false
 
 
 func apply_audio_props(action_on_playing: bool):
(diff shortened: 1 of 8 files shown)


godot/src/decentraland_components/audio_source.gd
```diff
@@ -10,6 +11,11 @@ func apply_audio_props(action_on_playing: bool):
 
 	self.pitch_scale = dcl_pitch
 
+	if dcl_global:
+		attenuation_model = AudioStreamPlayer3D.ATTENUATION_DISABLED
+	else:
+		attenuation_model = AudioStreamPlayer3D.ATTENUATION_INVERSE_DISTANCE
```
kuruk-mm, line 17: What about this fix?

kuruk-mm (approved):

PR #396, opened 2024-04-13
(poc) feat: experimental web

kuruk-mm:
This PR is a Proof of Concept and the ongoing effort for an experimental build for the web.

To setup the Godot Rust with `wasm` I follow this instructions: https://github.com/godot-rust/gdext/issues/438#issuecomment-1747926585

Note: All fixes are just to achieve an experimental build

# Issues found

## Issue 1
`ethers-providers` uses `jsonwebtoken` that is not working for WASM
Fix: I removed `ethers-providers` and `ethers` (that add all `ethers-*` libs) too, but I added `ethers-core` and `ethers-signers`. Also I removed the implementation of the `EthereumProvider` using `ethers-providers`.

## Issue 2
`hyper` and `tokio-tungstenite` use tokio and enable mio (I think due to the `net` feature)
Fix `hyper`: I just removed it, looks like we're using it for the inspector
Fix `tokio-tungstenite`: I removed WebSockets from the Runtime

## Issue 3 - Blocker
`tokio` support for WASM is limited, it doesn't have `fs`, `rt-multithread` and `net`. We use it a lot
Possible solution: We should isolate the behavior and have individual implementations for the Runtime using the Browser as part. For this we can use [`idb`](https://crates.io/crates/idb) and we should link some JavaScript code with `wasm-bindgen` and `wasm-bindgen-futures`.

## Issue 3 - bypassed
Temporary unblock: using conditional-compiling with `#[cfg(target_arch = "wasm32")]` and bypassing all the content-loading logic, the issue remains in the async runtime.
With this unblocked, it is possible to build and start debugging.

# Helpful commands

For trying to go one error by one error, I was building with -j1 to try to isolate the errors that I saw:
```
cargo build -j1 --no-default-features --target wasm32-unknown-emscripten
```

For tracking the dependency tree:
```
cargo tree --no-default-features --target wasm32-unknown-emscripten &> tree.txt
```


leanmendoza:
proceeded https://github.com/decentraland/godot-explorer/pull/523

PR #414, opened 2024-05-11
fix: memory issues

kuruk-mm:
Fix https://github.com/decentraland/godot-explorer/issues/415


rust/decentraland-godot-lib/src/content/content_provider.rs
```diff
@@ -69,13 +80,62 @@ impl INode for ContentProvider {
...
         self.cached.clear();
         tracing::info!("ContentProvider::exit_tree");
     }
+
+    fn process(&mut self, dt: f64) {
+        self.tick += dt;
+        if self.tick >= 1.0 {
+            self.tick = 0.0;
+
+            self.cached = self
```
leanmendoza, line 98: Wouldn't it be better to use self.cached.retain() and mutate it instead of creating a new one each time? Even with this rate 1/sec, it makes more sense

rust/decentraland-godot-lib/src/content/content_provider.rs
leanmendoza, line 1: All the `if let Some(A) = var { maybe Some } else { None }` can be replaced by `let A = var?;`

rust/decentraland-godot-lib/src/dcl/js/events.rs
```diff
@@ -37,22 +37,6 @@ struct EventBodyExpressionId {
     expression_id: String,
 }
 
-#[derive(Serialize)]
```
leanmendoza, line 40: Maybe better to add a `TODO` and `#[allow(unused)]`

rust/decentraland-godot-lib/src/godot_classes/resource_locker.rs
```diff
@@ -0,0 +1,51 @@
...
+        Self { base, reference }
+    }
+}
+
+#[godot_api]
+impl ResourceLocker {
+    #[func]
+    pub fn attach_to(mut node: Gd<Node>) {
+        if node.has_node(NodePath::from("ResourceLocker")) {
+            panic!("You cannot attach two times ResourceLocker to a Node")
```
leanmendoza, line 32: Panic seems a lot, maybe a tracing:error?

leanmendoza (approved):
Great work! It looks good.
Something I'm not really sure, using queue_free instead of remove_child it's ok in terms of memory management, but is it desirable to keep it parented to the given node until the end of the frame?

You also have commented code with remove_child

leanmendoza (approved):

PR #430, opened 2024-06-25
fix: lock resource in gltf containers and track deleted content

kuruk-mm:
lock resource in gltf containers between loading and instance collider


Gleix0:
Hola, te hable por gs zone.

PR #441, opened 2024-08-12
feat: add magic link login for Android

kuruk-mm:
Known issue: The sign has a problem that the WebView always starts refreshing. For that reason, if you open the Magic Wallet, you will see the Sign instead of the actual Wallet. Restarting the client fixes the problem. We should disable the Magic Wallet feature temporarily. Issue reported here: https://github.com/magiclabs/magic-android/issues/49

Note: When you sign in with Magic, every time that you reopen the App, it takes a bit more time, because it is validating that the session is still alive in Magic. If it's not, it will ask you to log in again.


leanmendoza (approved):
Nice work, it looks good :)

PR #451, opened 2024-09-26
fix: changerealm and reload

kuruk-mm:
- The realm was being set on NULL after you teleport to a position with /goto, and then the /changerealm gets broken
- Now the changerealms search for a valid position if there are sceneUrns in the /about


leanmendoza (approved):
EXCELLENT :D

PR #467, opened 2024-10-22
fix: blocking send and ui text tags

leanmendoza:
- fix blocking send while processing the scene messages
- fix ui text, skipping tags

### DIFF
--- a/lib/src/scene_runner/components/ui/scene_ui.rs
+++ b/lib/src/scene_runner/components/ui/scene_ui.rs
@@ -169,7 +169,7 @@ fn update_layout(
         .compute_layout_with_measure(
             root_node,
             size,
-            |size, available, node_id, node_context, _style| match node_context {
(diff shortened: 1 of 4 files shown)


lib/src/scene_runner/components/ui/scene_ui.rs
```diff
@@ -220,14 +220,14 @@ fn update_layout(
...
-                        available
-                    );
+                    // tracing::debug!(
+                    //     "text node {:?}, wrapping {:?}, size: {:?}, font_rect {:?}, available {:?}",
+                    //     node_id,
+                    //     *wrapping,
+                    //     size,
+                    //     font_rect,
+                    //     available
+                    // );
```
kuruk-mm, line 230: can we uncomment or remove this?

kuruk-mm (approved):

PR #477, opened 2024-10-31
feat: add network inspector

leanmendoza:
Add the network inspector. When it enables, the requester sends a report with the info when the request inits, then an update when finishes well or not.

Also:
- remove HttpRequester (the one is not Queue) and their tests
- adapt SceneEntityCoordinator to use HttpQueueRequester
- Remove make string header `key: value`  and use HashMap instead


lib/src/http_request/rust_http_queue_requester.rs
```diff
@@ -80,10 +85,21 @@ impl RustHttpQueueRequester {
...
+    // #[func]
+    // fn try_set_global_network_inspector(&mut self) {
+    //     let Some(sender) = DclGlobal::try_singleton()
+    //         .and_then(|v| Some(v.bind().get_network_inspector().bind().get_sender()))
+    //     else {
+    //         return;
+    //     };
+
+    //     self.http_queue_requester = set_inspector_sender(sender);
+    // }
```
kuruk-mm:
```suggestion
```

should we remove this commented code? otherwise I would add a comment to know what it does and why is commented

lib/src/tools/network_inspector.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    #[signal]
+    pub fn request_changed(&self, id: u32) {}
+
+    #[func]
+    pub fn get_request(&self, id: u32) -> Dictionary {
+        let mut dict = Dictionary::new();
+        if let Some(request) = self.requests.get(&NetworkInspectorId(id)) {
+            // dict.insert("id", request.id);
```
kuruk-mm, line 150:
```suggestion
```

godot/src/global.gd
```diff
@@ -122,13 +122,19 @@ func _ready():
...
 	GLTFDocument.register_gltf_document_extension(custom_importer)
 
 	if args.has("--raycast-debugger"):
 		set_raycast_debugger_enable(true)
 
+	if args.has("--network-debugger"):
+		self.network_inspector.set_is_active(true)
+	else:
+		self.network_inspector.set_is_active(false)
+
```
kuruk-mm, line 137:
```suggestion
	self.network_inspector.set_is_active(args.has("--network-debugger"))

```

if you think the code looks cleaner like you wrote, ignore this

godot/src/global.gd
```diff
@@ -122,13 +122,19 @@ func _ready():
 	get_tree().root.add_child.call_deferred(self.portable_experience_controller)
 	get_tree().root.add_child.call_deferred(self.testing_tools)
 	get_tree().root.add_child.call_deferred(self.metrics)
+	get_tree().root.add_child.call_deferred(self.network_inspector)
```
kuruk-mm, line 125: can we avoid adding it to the root, if it is disabled :thinking: ?


and here (dcl_global.rs):
```
pub fn get_network_inspector_sender() -> Option<NetworkInspectorSender> {
```

Do we check if the node exists?

lib/src/http_request/rust_http_queue_requester.rs
```diff
@@ -65,7 +70,7 @@ impl RustHttpQueueRequester {
         url: GString,
         method: godot::engine::http_client::Method,
         body: GString,
-        headers: VariantArray,
+        headers: Dictionary,
```
kuruk-mm, line 73: :relieved: great

lib/src/realm/scene_entity_coordinator.rs
```diff
@@ -49,10 +60,35 @@ impl SceneEntityCoordinator {
...
+            requested_entity: Default::default(),
+            cache_scene_data: Default::default(),
+            entities_active_url: Default::default(),
+            content_url: Default::default(),
+
+            version: Default::default(),
+            dirty_loadable_scenes: Default::default(),
+            loadable_scenes: Default::default(),
+            keep_alive_scenes: Default::default(),
+            empty_parcels: Default::default(),
```
kuruk-mm, line 85: we always use `..Default::default()`. I think it's a common pattern in Rust. I don´t see issues using it

lib/src/realm/scene_entity_coordinator.rs
```diff
@@ -77,6 +113,37 @@ impl SceneEntityCoordinator {
...
+                if runtime.is_err() {
+                    panic!("Failed to create runtime {:?}", runtime.err());
+                }
+                let runtime = runtime.unwrap();
+
+                runtime.block_on(async move {
+                    let result = http_requester.request(request_option, 0).await;
+                    let _ = sender.try_send(result);
+                });
+            });
```
kuruk-mm, line 143: Do you think it’s worth maintaining this fallback? 🤔 Maybe I should just log an error instead, as the fallback code is very slow if used frequently.

kuruk-mm (approved):
Great! Just a quick polish.

lib/src/realm/scene_entity_coordinator.rs
leanmendoza: if the struct doesn't derive Default or has any impl, it doesn't work :(

lib/src/realm/scene_entity_coordinator.rs
leanmendoza: only for testing

godot/src/global.gd
leanmendoza: yep, i only left the if for clarity

godot/src/global.gd
leanmendoza: the option is only if the Global doesn't exist (initial register of godot-rust)
if we don't treat it as node, we'd need to call the update anyway.

PR #479, opened 2024-11-01
fix: avatar rendering and scene crashes doesn't kill the scene anymore

leanmendoza:
cherry-pick fixes from bump-4.3


kuruk-mm (approved):
greatt

PR #480, opened 2024-11-08
feat: add network inspector UI

leanmendoza:
Also:
- feat: limit scene fetch to 2 simultaneously
- fix: avatar fetched could be null
- chore: expose if there is JS debugging
- feat: add button to debug the current scene
- refactor: set enable_inspector as default feature and disable it for mobile

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -65,7 +65,7 @@ chrono = "0.4.31"
 
 simple-easing = "1.0.1"
 
-multihash-codetable = { version = "0.1.1", features = ["digest", "sha2"] }
+multihash-codetable = { version = "0.1.1", features = ["digest", "sha2"] } 
 cid = "0.11.0"
 multipart = { version = "0.18.0", default-features = false, features = ["client", "lazy_static"] }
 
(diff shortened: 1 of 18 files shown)


lib/src/dcl/js/mod.rs
```diff
@@ -112,7 +112,7 @@ pub fn create_runtime(inspect: bool) -> (deno_core::JsRuntime, Option<InspectorS
         );
         let server = InspectorServer::new(
             SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 9222),
-            "bevy-explorer",
+            "godot-explorer",
```
kuruk-mm, line 115: Oops

lib/Cargo.toml
```diff
@@ -75,7 +75,7 @@ jni = { version = "0.21.1", features = ["invocation"] }
 paranoid-android = "0.2.1"
 
 [features]
-default = ["use_ffmpeg", "use_livekit", "use_deno"]
+default = ["use_ffmpeg", "use_livekit", "use_deno", "enable_inspector"]
```
kuruk-mm, line 78: It's ok that it's going to be able for Desktop? I think makes sense, because we don't publish Desktop.

PR #492, opened 2024-11-20
feat: loading only when you pass to another scene (default=true)

### LINKED ISSUE #425: UX improvement when loading new scenes
Sometimes, while you're walking inside a scene, the loading of other scenes introduces hiccups, and this might happen as well with the unloading.
This could be solved by fixing in advance the scenes that can be loaded while you're inside the one scene. The algorithm can maximize the scenes available or not or maybe it could be settings or set when the memory and CPU are enough.
So the hiccup is only reserved when you really change the scene

leanmendoza:
Fixes #425 

image


kuruk-mm (approved):
Code looks great!

PR #526, opened 2025-01-31
feat: asset optimization per individual assets

kuruk-mm:
Related to https://github.com/dclexplorer/consumer-processor/pull/13


kuruk-mm:
Reviewing issue with too much files open reaching the limit...

Issue #552, opened 2025-05-14
Enable glow


kuruk-mm:
Glow is currently disabled in the Environment due to inconsistency in the values.

We should enable it again.

Issue #551, opened 2025-05-13
Improve MeshRenderer performance


kuruk-mm:
Looks like MeshRenderer is creating one mesh per instance and generating a lot of DrawCalls.

Reference performance: flutterecho.dcl.eth

kuruk-mm:
Closed by https://github.com/decentraland/godot-explorer/pull/560

Issue #547, opened 2025-04-08
Dark materials on darulmawa.dcl.eth


kuruk-mm:
Investigate how to fix the dark materials that appear in darulmawa.dcl.eth

Reference images:

![Image](https://github.com/user-attachments/assets/13795207-4897-4bfc-9a3f-289219f9fb1c)

![Image](https://github.com/user-attachments/assets/0e8eac66-2db7-4043-aaee-62ff7925ea82)

![Image](https://github.com/user-attachments/assets/a2caf265-ffba-420f-a64b-229301e09b87)

![Image](https://github.com/user-attachments/assets/571713b2-bd3f-4c19-af8b-33c4b4ea6ad1)

![Image](https://github.com/user-attachments/assets/2da10c7c-d9e0-4a00-9f6b-9de5e418e29e)

manuelmaceira:
Closed by #560

Issue #527, opened 2025-02-08
Emission texture set as albedo texture on import


kuruk-mm:
Workaround:

```gdscript
func remove_emission_texture(gltf_instance: Node3D):
	# HACK: Workaround to fix an import error that sets the emisison texture as the albedo texture.
	for child in gltf_instance.get_children():
		if !(child is MeshInstance3D):
			continue
		var material = child.mesh.surface_get_material(0)
		material.emission_texture = null
```

Looks like it happens with the DAO HQ floor. We should investigate it.

manuelmaceira:
Closed by #560

PR #560, opened 2025-05-15
fix: misc material fixes

manuelmaceira:
# Miscellaneous material and rendering fixes

Some small changes to further improve material fidelity and rendering performance:

* Re-enabled bloom in the three main environments (low, medium and high) and configured it for low rendering overhead.
* Fixed a bug where materials would use vertex colors instead of its own, resulting in the whole material being black (darulmawa.dcl.eth).
* Fixed an issue where material components that had changing values would get re-created every frame instead of changing it's values (x3 fps on flutterecho.dcl.eth).


manuelmaceira:
Added a fix that converts the color format of the material components to sRGB and applies an energy multiplication in order to get the HDR values into the Mobile renderer's range (0.0 : 2.0). 
Also tweaked Bloom configurations.

manuelmaceira:
Closes #552, Closes #551, Closes #547

.github/workflows/prepare_lib.yml
kuruk-mm, line 1: can you rollback this file? this change is not required

lib/src/scene_runner/components/material.rs
```diff
@@ -132,14 +137,19 @@ pub fn update_material(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
                             .clone()
                             .multiply(pbr.emissive_intensity.0);
-                        godot_material.set_emission(emission.to_godot());
-                        godot_material.set_albedo(pbr.albedo_color.0.to_godot());
+
+                        // In the Mobile renderer, HDR will be capped at 2.0 so we'll have to reduce the energy multiplier to be able to see fluctuations in energy
+                        godot_material.set_emission_energy_multiplier(0.1);
+
+                        // In the same way, godot uses sRGB instead of linear colors.
+                        godot_material.set_emission(emission.to_godot().linear_to_srgb());
```
kuruk-mm, line 145: Great. Remember this only happens on `Materials` that the Scene Creators overrides.

The Materials from a GLTF are not affected.

kuruk-mm (changes requested):

kuruk-mm:
image

bloom is saturating in textures

kuruk-mm:
fence too shine

image

genesis reference:
image

manuelmaceira:
More consistent results.
image
image
image
image
image

Closes #527
As it was a false positive, the texture was correctly set but the emission energy was to high for the mobile renderer.

lib/src/scene_runner/scene.rs
```diff
@@ -189,7 +190,7 @@ pub struct Scene {
     pub last_tick_us: i64,
     pub next_tick_us: i64,
```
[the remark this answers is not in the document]
kuruk-mm: Disagree. It is clear, like the other HashMaps. Maybe one line comment is enought.

manuelmaceira:
> It looks like the emissive in the GLTF file is not correct. Here are two examples.
> 
> I think Genesis Plaza should be included in the testing scenes to check if everything works correctly.

Those come from a optimized scene, we are currently applying the fixes on gltf import but those are imported as tscn, maybe we should apply the fixes on the preprocessing server.
For now ive reimplemented the same fixes that happen on import and made them happen on instantiation on the godot side.

godot/src/decentraland_components/gltf_container.gd
```diff
@@ -189,18 +191,30 @@ func async_load_gltf():
...
+
+	for instance in meshes:
+		var mesh = instance.mesh
+		for idx in mesh.get_surface_count():
+			var material = mesh.surface_get_material(idx)
+			fix_material(material)
+
+
+func fix_material(mat: BaseMaterial3D):
+	mat.vertex_color_use_as_albedo = false
```
kuruk-mm, line 217: adding

```
func fix_material(mat: BaseMaterial3D):
    mat.vertex_color_use_as_albedo = false

    # In the Mobile renderer, HDR will be capped at 2.0 so we'll have to reduce the energy multiplier to be able to see fluctuations in energy
    mat.set_emission_energy_multiplier(0.2)
```

worked for me

manuelmaceira:
I've now tested on android and linux, everything looks as it should for what i can see.
There are still a lot of performance issues but I think this is progress.

Finally i've moved the processing of the materials to instantiation time, if we could move it to import time and move the same processing to the preprocessing server we could gain a bit of performance on instantiation.

manuelmaceira (commented):
Reviewed changes.

kuruk-mm:
If you want, we can schedule this for the next cycle of materials...

Robot saturated white
image

Text saturated white
image

Dialog in Genesis City saturated white
image

kuruk-mm (approved):

PR #567, opened 2025-05-27
feat: PlaceItem class

### LINKED ISSUE #562: Refactor `PlaceItem` to Reduce Code Duplication
Currently, `PlaceItem` serves as a base class for representing information about a `Place`.

Some subclasses—such as `JumpIn`—extend `PlaceItem` to display additional information using labels and other UI elements.

While attempting to implement a new card for the Satellite Map, we noticed significant code duplication from the `JumpIn` card. Much of this duplicated logic revolves around assigning values from the `Place` to UI nodes.

**Proposal:**

* Refactor `PlaceItem` to handle assigning available place information to UI nodes by default, only when those nodes exist.
* This would eliminate the need for subclasses like `JumpIn` to manually repeat this logic.
* If successful, we may even be able to remove the `JumpIn` subclass entirely or at least simplify it significantly.

This refactor would also support and streamline the work needed for #225.

sdilauro:
Resolves #562 

Create a bigger class to replace discover_item_carrousel.gd (old PlaceItem class)

Now it is more scalable according to new UI elements with different nodes.

The code uses functions instead `onready var` to get nodes safely.

```
func _get_button_close() -> Button:
	return _get_node_safe("Button_Close")
```
	
This way prevents errors in the console when components who use this class don't have a declared node in `onready`


godot/src/ui/components/place_item.gd
```diff
@@ -0,0 +1,279 @@
...
+	var like_score = item_data.get("like_score", 0.0)
+	set_likes_percent(like_score if like_score is float else 0.0)
+	set_online(item_data.get("user_count", 0))
+
+	var image_url = item_data.get("image", "")
+	if not image_url.is_empty():
+		set_image(texture_placeholder)
+		_async_download_image(image_url)
+	else:
+		set_image(texture_placeholder)
```
kuruk-mm, line 206: check if `_get_texture_image` exists... if not, don't try to _async_download_image... you can put all the code above inside the if (no need to put the placeholder also)...

PR #598, opened 2025-07-04
feat: Dev tools menu and performance benchmark

manuelmaceira:
This PR adds a unified **DCL Tools** menu in the editor for running and testing developer tools. Key additions include:

### 🧰 DCL Tools menu
- Central place in the editor for accessing development utilities.
- Added avatar test runner (existing) as a simple example tool.

### 🕵️ RenderDoc launcher
- Opens RenderDoc with the current project preloaded.
- Includes configurable default settings for smoother capture setup.

### 🌐 Typical parcel loader
- Quickly load into common/representative scenes via the DCL Tools menu.
- Useful for repro cases and visual inspection.

### 📊 Performance benchmarking system
- Counts meshes and materials, estimates deduplication potential.
- Iterates over all typical parcels, saving a CSV report and screenshots.
- Can be executed in CI with: `--headless --dcl-benchmark`.

### ⚙️ Engine config tweaks
- Reduced overhead from rendering and physics settings to better isolate content performance.

---

These additions make it easier to investigate performance, run scene-specific tests, and prepare debugging workflows without modifying the core engine or CLI.


kuruk-mm (changes requested):
When I try to run the game normal (just cargo run -- run), the client closes with the performance report after few seconds. It should not work if I don't open with the Editor. Also the plugin should be disabled on runtime. (it could be because some API is not available at the templates?)

Try to use a argument using OS.get_cmdline_args(), like --profiling

And the pipeline is failing because it is not detecting the stuff in the Loading. Can you fix those issues?

Issue #587, opened 2025-06-24
Nickname-UI enhancements


manuelmaceira:
The nickname tag should have more crisp text, also there should be a system to prevent overlapping nametags.
As a bonus we should be able to see last message sent by the user in the nickname capsule for a little while.

kuruk-mm:
Please divide this issue into multiple issues depending on how you plan to divide the work of nickname UI enhancements.

manuelmaceira:
Closed by #612

PR #612, opened 2025-08-12
feat: add message display to avatar's nickname UI

manuelmaceira:
## Summary
This PR introduces a feature that displays messages from the player in the avatar's nickname UI. The message will appear as a popup above the nickname and will disappear after a set duration.

## Changes
- **avatar.gd**: Added a connection to the `player_said` signal and a handler function `on_player_said` to display messages.
- **nickname_ui.gd**: Added functionality to show messages with animations and styling.
- **nickname_ui.tscn**: Updated the scene to include a container for displaying messages.
- **global.gd**: Added a `player_said` signal to be emitted when a player sends a message.
- **chat.gd**: Emits the `player_said` signal when a chat message is received.

Addresses part of #587


kuruk-mm (changes requested):
Can you align the text to the left easily?

Also, sometimes when the message is short, the height is too small and cuts off a part of the character like `qyj`.

And finally, increase the popup duration to 6 seconds, please

Code looks great

manuelmaceira:
Now it should correctly measure the size needed for the message, no longer truncating it nor displaying weird behavior with different line counts.
Also extended the display time to 6 seconds

kuruk-mm (approved):
Working great! Congrats

PR #728, opened 2025-10-15
refactor: softly remove ffmpeg

leanmendoza:
remove ffmpeg dependency. it's currently disabled for mobile and managed as a feature-flag in the library, but the feature-flag is coupled to the video player implementation, that's why the `softly`.
in the future the video player will be implemented with native handlers and the ffmpeg finally uncoupled


src/install_dependency.rs
```diff
@@ -394,7 +354,7 @@ fn download_prebuilt_dependencies() -> Result<(), anyhow::Error> {
     fs::create_dir_all(BIN_FOLDER)?;
 
     // Download Android dependencies
-    let android_deps_url = "https://godot-artifacts.kuruk.net/android_deps.zip";
+    let android_deps_url = "https://files.dclexplorer.com/android_deps.zip";
```
leanmendoza: @kuruk-mm : changed to stop including ffmpeg libs in the bundle

PR #754, opened 2025-10-20
chore: add `clean-cache` cargo command

leanmendoza:
if we update the deps, this command makes easier to clean the cache


kuruk-mm:
we can also remove .bin/, so it will require to execute ```cargo run -- install``` again

kuruk-mm (approved):

PR #762, opened 2025-10-22
feat: add virtual cameras

### LINKED ISSUE #558: CinematicCamera / VirtualCamera
Add Cinematic/VirtualCamera.

The Protocol Squad created a CinematicSetting in the CameraModeArea. But Foundation created a new component named "VirtualCamera".

See if we can unify the approach, create a compatibility layer, and use both.

leanmendoza:
fixes #558 

Docs:
- https://docs.decentraland.org/creator/development-guide/sdk7/camera/
- https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/main_camera.proto
- https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/virtual_camera.proto

No adrs found about it.


kuruk-mm:
When the camera is locked, we should change how the pointer behaves.

We should support two modes:

CursorLocked: Where you look is the target.

CursorUnlocked: Where you click (or touch) is the target.

We previously implemented the CursorUnlocked mode here: https://github.com/decentraland/godot-explorer/pull/256, but later removed it.

PR #790, opened 2025-10-27
feat: events metrics

### LINKED ISSUE #752: Events: Metrics Implementation
Implement Event Metrics

sdilauro:
resolves #752 
# Add Event Detail Analytics Tracking

## Overview
This PR implements analytics tracking for event detail interactions.

### Event Detail Show (ScreenViewed)
Tracks when users view event details with state, tags, and orientation information.

### Reminder set/remove (ClickedButton)
Tracks when users set or remove reminders for events.

### Event Detail Jump to (ClickedButton)
Tracks when users jump to event locations in-world.

## Implementation Details
Added extra properties to SegmentEventScreenViewed to allow sending more data

### DIFF
--- a/lib/src/analytics/data_definition.rs
+++ b/lib/src/analytics/data_definition.rs
@@ -11,27 +11,6 @@ pub struct SegmentMetricEventBody {
 }
 
 #[derive(Serialize)]
-#[serde(rename_all = "lowercase")]
-pub enum EventState {
-    Live,
-    Upcoming,
-}
-
-#[derive(Serialize)]
-#[serde(rename_all = "lowercase")]
-pub enum EventTags {
-    Trending,
-    None,
-}
-
-#[derive(Serialize)]
-#[serde(rename_all = "lowercase")]
-pub enum Orientation {
-    Portrait,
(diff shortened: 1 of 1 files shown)


lib/src/analytics/data_definition.rs
```diff
@@ -214,6 +235,29 @@ pub struct SegmentEventClickButton {
...
+    // Screen orientation.
+    pub orientation: Orientation,
+}
+
+#[derive(Serialize)]
+pub struct SegmentEventEventsGeneral {
+    // Event ID.
+    pub event_id: String,
+    // Event tags (trending or none).
+    pub event_tags: EventTags,
```
kuruk-mm: what is the purpose of this?

the only change in this file that looks needed is
```
    // JSON with extra payload, in case we need to track additional metadata.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub extra_properties: Option<String>,
```

sdilauro: yep, it was unnecessary definitions because we had other events that were discarded finally

PR #920, opened 2025-12-02
feat: livekit video streaming for `VideoPlayer` component

### LINKED ISSUE #895: Implement LiveKit Streaming
# Implement LiveKit Streams Support in Mobile Client

## Overview

The Decentraland Protocol already supports LiveKit streams. This issue tracks the implementation of this feature in the mobile client.

## Expected Behavior

When a `VideoPlayer` component has the following URL format:

```
VideoPlayer.src = 'livekit-video://current-stream'
```

This should be interpreted as a live streaming source.

If we detect a `VideoPlayer` with this `src` value **and** there is an available `VideoTrack` from a LiveKit peer, the client should automatically subscribe to the stream.

## Technical Context

- This feature relies on [LiveKit Ingress](https://docs.livekit.io/home/ingress/overview/) on the backend.
- The integration should be compatible with [DCL Live Admin Tools](https://www.youtube.com/watch?v=efjJN7Jr7Qo).

leanmendoza:
fix #895 

chores:
- for now it holds #919 
- add filter for android logging
- add a way to open the js source of current scene from debug panel
- remove use_ffmpeg feature flag, get rid of ffmpeg at all


lib/src/godot_classes/dcl_global.rs
```diff
@@ -34,16 +34,36 @@ use super::{
...
+
+        let filter = EnvFilter::new(
+            // TODO: Modify this line to change logging levels
+            // "warn"  // Only warnings and errors
+            // "libwebrtc=error,debug" // Filter out libwebrtc noise, show debug for everything else
+            // "debug"  // Debug, info, warnings and errors (shows all debug logs)
+            // "dclgodot::scene_runner=trace,warn"  // Trace for scene_runner, warn for everything else
+            // "dclgodot::scene_runner=debug,dclgodot::comms=info,warn"  // Debug for scene_runner, info for comms, warn for everything else
+            "info", // Info, warnings and errors
+        );
```
kuruk-mm: This should not be from the `RUST_LOG` env variable?

PR #934, opened 2025-12-10
feat: implement auth / signin deep link flow

### LINKED ISSUE #909: Modify Client to implement Deeplink Session Auth

leanmendoza:
fix #909 #1070 

it works locally ✅ 

TODO:
- once https://github.com/decentraland/auth/pull/218 is merged to main
  - [x] test flow in Android with using .zone
    - only work if there is session active ⚠️ 
      - [x] fixed
  - [x] test flow in iOS with using .zone
    - only work if there is session active ⚠️ 
      - [x] fixed
  - [x] test old-flow in Desktop with using .zone
    - not working :x:
      - [x] fixed
- once a release that include the PR merged to main is rolled out to .org
  - [x] test flow in Android with using .org
  - [x] test flow in iOS with using .org
  - [x] test old-flow in Desktop with using .org

### DIFF
--- a/lib/src/godot_classes/dcl_global.rs
+++ b/lib/src/godot_classes/dcl_global.rs
@@ -57,7 +57,7 @@ mod android {
             // "debug"  // Debug, info, warnings and errors (shows all debug logs)
             // "dclgodot::scene_runner=trace,warn"  // Trace for scene_runner, warn for everything else
             // "dclgodot::scene_runner=debug,dclgodot::comms=info,warn"  // Debug for scene_runner, info for comms, warn for everything else
-            "dclgodot::auth=debug", // Info, warnings and errors
+            "info",
         );
 
         let android_layer = paranoid_android::layer(env!("CARGO_PKG_NAME"))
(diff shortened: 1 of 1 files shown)


lib/src/godot_classes/dcl_global.rs
```diff
@@ -57,8 +57,7 @@ mod android {
             // "debug"  // Debug, info, warnings and errors (shows all debug logs)
             // "dclgodot::scene_runner=trace,warn"  // Trace for scene_runner, warn for everything else
             // "dclgodot::scene_runner=debug,dclgodot::comms=info,warn"  // Debug for scene_runner, info for comms, warn for everything else
-            "dclgodot::scene_runner::components::video_player=debug", // Debug for scene_runner, info for comms, warn for everything else
-                                                                      // "info", // Info, warnings and errors
+            "dclgodot::auth=debug", // Info, warnings and errors
```
kuruk-mm: are we sure that we want this? i think `info` or `warn` should be the default... a lot of logs are going to be generated with this...

PR #986, opened 2025-12-19
fix: hdpi textures

### LINKED ISSUE #864: High DPI for UiBackground
UiBackground has low resolution, we should use high dpi for that

leanmendoza:
fixes #864 

disadvantage (maybe to keep tracking)
- if the texture is used both in world and ui, the asset will be optimized by resizing in world and the original for UI, so we have redundancy here:
  - option 1: take a smarter approach that overwrite the optimized asset with original, the cons here is maybe we're adding overhead for in-world representation


kuruk-mm:
Even if it is not optimized. I would compress the texture using ETC2.

PR #1015, opened 2025-12-23
fix: ExoPlayer keeps video playing in background 

leanmendoza:
now it pauses the ExoPlayer when the app is paused and resumes when the app is resumed


kuruk-mm (approved):
Great!

PR #1099, opened 2026-01-05
feat: add gltf node modifiers

leanmendoza:
## Summary

  This PR implements the GltfNodeModifiers SDK component, enabling scenes to dynamically modify materials and shadow casting settings on specific nodes within loaded GLTF models at runtime.

## Key capabilities:
  - Apply custom PBR or unlit materials to specific GLTF nodes by path
  - Control shadow casting per-node
  - Support for both Unity-style (**/NodeName) and Bevy-style (EntityName) path matching
  - Per-surface material overrides using mesh_name/surface_index syntax
  - Automatic texture loading including video textures
  - Material reuse to minimize Vulkan descriptor pool pressure

## Architecture

  New module structure under lib/src/scene_runner/components/gltf_node_modifiers/:

  | File             | Purpose                                               |
  |------------------|-------------------------------------------------------|
  | mod.rs           | Main update loop and dirty entity processing          |
  | state.rs         | State tracking (original materials, pending textures) |
  | path_matching.rs | Node path matching with wildcard support              |
  | mesh_utils.rs    | GLTF hierarchy traversal utilities                    |
  | material.rs      | Material creation, application, and texture loading   |

## Path Matching Examples

  | Pattern      | Matches                                              |
  |--------------|------------------------------------------------------|
  | NodeName     | Any node named "NodeName" (Bevy-style)               |
  | **/NodeName  | Any node named "NodeName" at any depth (Unity-style) |
  | Parent/Child | Node "Child" under "Parent"                          |
  | MeshName/0   | Surface 0 of mesh "MeshName"                         |

DISCLAIMER: this path matching is hacking to match the same beahviour as foundation has, and also the same as bevy has

## Test Plan
- Functionality in general should be the same (except for those scenes which uses GltfNodeModifier, mostly Foundation scenes)
- Test in world `mannakia`-> decentraland://open?realm=mannakia.dcl.eth cubes indicates the effect on the placed gltfs

### DIFF
--- a/godot/src/ui/components/backpack/backpack.gd
+++ b/godot/src/ui/components/backpack/backpack.gd
@@ -53,6 +53,8 @@ var is_loading_profile: bool = false
 @onready var emote_editor = %EmoteEditor
 
 @onready var container_navbar = %PanelContainer_Navbar
+@onready var button_emotes = %Button_Emotes
+@onready var button_wearables = %Button_Wearables
 
 
 # gdlint:ignore = async-function-name
@@ -498,9 +500,18 @@ func _on_button_wearables_pressed():
 
 
 func _on_button_emotes_pressed():
+	show_emotes()
+
+
+func show_emotes() -> void:
 	avatar_preview.focus_camera_on(Wearables.Categories.BODY_SHAPE)
 	wearable_editor.hide()
 	emote_editor.show()
+
+
+func press_button_emotes() -> void:
+	button_emotes.set_pressed_no_signal(true)
+	button_wearables.set_pressed_no_signal(false)
 
 
 func _on_check_box_only_collectibles_toggled(toggled_on):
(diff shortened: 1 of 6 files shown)


lib/src/godot_classes/dcl_global.rs
leanmendoza: revert

lib/src/scene_runner/components/material.rs
leanmendoza: Fixes:
- [x] material only get the first texture set, it should be able to change
- [x] inherit from above: material doesnt change the texture if you clean it

leanmendoza (commented):
some auto review

lib/src/scene_runner/components/gltf_node_modifiers/material.rs
```diff
@@ -0,0 +1,499 @@
...
+    let mut content_provider = DclGlobal::singleton().bind().get_content_provider();
+    for tex in dcl_material.get_textures().into_iter().flatten() {
+        if let DclSourceTex::Texture(hash) = &tex.source {
+            content_provider.call_deferred(
+                "fetch_texture_by_hash",
+                &[
+                    hash.to_godot().to_variant(),
+                    DclContentMappingAndUrl::from_ref(content_mapping.clone()).to_variant(),
+                ],
+            );
```
kuruk-mm: new way on GDExt to call_deferred

```
// Old way (string-based, error-prone):
node.call_deferred("set_position", &[pos.to_variant()]);

// New way (type-safe):
node.run_deferred(|obj| obj.set_position(pos));
```

😲

if you want to use it, otherwise we can change all calls in the project in one PR

PR #1119, opened 2026-01-07
feat: implement clean orientation script and unify avatar preview on the profile

### LINKED ISSUE #1089: Split Menu.tscn into MenuLandscape.tscn and MenuPortrait.tscn
## Context

From the node analysis spike in #937, having both portrait and landscape menu layouts loaded simultaneously is inefficient.

### Problem

- `Menu.tscn` contains both landscape and portrait layouts
- Both exist in memory even though only one is used at a time
- This unnecessarily increases the node count

## Proposed Solution

1. Split `Menu.tscn` into two separate scenes:
   - `MenuLandscape.tscn` - for landscape/desktop orientation
   - `MenuPortrait.tscn` - for portrait/mobile orientation
2. Load only the appropriate scene based on current orientation
3. Handle orientation changes by swapping scenes if needed

## Expected Impact

- Reduce node count by eliminating duplicate layout nodes
- Lower memory footprint
- Cleaner separation of concerns between orientations

## References

- Spike analysis: #937
- Related: #1075 (lazy-load menu panels)

sdilauro:
Closes #1089 

## Overview

This PR optimizes memory usage by implementing a unified approach to handle UI elements across different screen orientations. Instead of creating separate menu instances for portrait and landscape orientations, we've introduced a utility script (`clean_orientation.gd`) that dynamically removes unused nodes from memory based on the current screen orientation. Additionally, we've unified the avatar preview instances to use a single instance that gets reparented between containers.

### 1. Introduction of `clean_orientation.gd` Script

Instead of maintaining separate menu scenes for portrait and landscape orientations, we've created a utility script that removes orientation-specific nodes from memory when they're not needed.

**How it works:**

The `clean_orientation.gd` script extends `Control` and provides two exported properties:
- `clean_on_landscape`: When `true`, the node and all its children are completely removed from memory when the screen is in landscape orientation
- `clean_on_portrait`: When `true`, the node and all its children are completely removed from memory when the screen is in portrait orientation

**Key features:**
- Automatically listens to window size changes via the `size_changed` signal
- Properly disconnects signal handlers before freeing nodes to prevent memory leaks
- Completely removes nodes from memory (not just hides them) when they're not needed for the current orientation

### 2. Application in Menu (`menu.tscn`)

The script has been applied to orientation-specific elements in the main menu. For example:
- **Bottom navigation bar**: Applied with `clean_on_landscape = true` - this bar is only used in portrait mode, so it's completely removed from memory when switching to landscape orientation
- Other portrait-specific UI elements that are not needed in landscape mode

This approach ensures that when the user switches to landscape orientation, portrait-specific UI elements (like the bottom navigation bar) are not just hidden but completely freed from memory, reducing the overall memory footprint.

### 3. Limited Usage in Child Components

The `clean_orientation` script has been applied to some child components of the menu where it makes sense to remove orientation-specific nodes. However, the number of use cases is limited because:

- **Most UI elements are shared**: Many components work well in both orientations with responsive layouts, so they don't need orientation-specific cleanup
- **Visibility vs. removal**: For many cases, simply hiding elements (using `visible` property) is sufficient and more efficient than removing them
- **Reparenting strategy**: Some components (like the avatar preview) use a different optimization strategy (see below)

The script is most effective when applied to large, orientation-specific UI sections that contain multiple child nodes and resources that won't be used in the other orientation.

### 4. Unified Avatar Preview with Reparenting

To further optimize memory usage, we've unified the avatar preview instances in the profile component. Instead of maintaining separate `AvatarPreview` instances for portrait and landscape orientations, we now use a single instance that gets reparented between containers based on the current screen orientation.

**Implementation:**

```gdscript
@onready var avatar_preview: AvatarPreview = %AvatarPreview
@onready var control_landscape_avatar: Control = %Control_landscape_avatar
@onready var margin_container_portrait_avatar: MarginContainer = %MarginContainer_PortraitAvatar

func _ready() -> void:
	get_window().size_changed.connect(self._relocate_avatar_preview)
	_relocate_avatar_preview()
	# ... rest of initialization

func _relocate_avatar_preview():
	var window_size: Vector2i = DisplayServer.window_get_size()
	var is_landscape: bool = window_size.x > window_size.y
	if is_landscape:
		avatar_preview.reparent(control_landscape_avatar)
	else:
		avatar_preview.reparent(margin_container_portrait_avatar)
```

**Benefits:**
- Only one `AvatarPreview` instance exists in memory at any time, regardless of orientation
- All avatar operations (emotes, rotations, updates) work seamlessly on the single instance
- The reparenting is non-destructive and maintains all node properties and state
- No synchronization issues between multiple avatar instances

### Basic Usage

1. **Attach the script** to any `Control` node that should be removed based on orientation
2. **Set the appropriate property** in the editor:
   - Set `clean_on_landscape = true` if the node should be removed when in landscape mode
   - Set `clean_on_portrait = true` if the node should be removed when in portrait mode
3. **The script handles the rest**: It will automatically detect orientation changes and remove the node when appropriate

### Important Notes

- The script automatically connects to the window's `size_changed` signal in `_ready()`
- Before freeing the node, it properly disconnects the signal handler to prevent memory leaks
- Once a node is freed, it cannot be restored - this is intentional for memory optimization
- Use this script for nodes that are truly orientation-specific and won't be needed in the other orientation

### DIFF
--- a/godot/src/ui/components/menu/menu.gd
+++ b/godot/src/ui/components/menu/menu.gd
@@ -104,7 +104,8 @@ func close():
 
 func show_discover():
 	select_discover_screen()
-	hud_button_discover.toggled.emit(true)
+	if is_instance_valid(hud_button_discover):
+		hud_button_discover.toggled.emit(true)
 	_open()
 
 
--- a/godot/src/ui/components/menu/menu.tscn
+++ b/godot/src/ui/components/menu/menu.tscn
@@ -94,7 +94,6 @@ grow_vertical = 2
 
 [node name="Control_Discover" parent="VBoxContainer/ColorRect_Background/Menus" instance=ExtResource("4_5qyiw")]
 unique_name_in_owner = true
-visible = false
 layout_mode = 1
 
 [node name="Control_Settings" parent="VBoxContainer/ColorRect_Background/Menus" instance=ExtResource("3_1stdy")]
@@ -134,6 +133,7 @@ stretch_mode = 6
 
 [node name="Control_ProfileSettings" parent="VBoxContainer/ColorRect_Background/Menus" node_paths=PackedStringArray("backpack") instance=ExtResource("18_x0kg3")]
 unique_name_in_owner = true
+visible = false
 layout_mode = 1
 backpack = NodePath("../Control_Backpack")
 


godot/src/ui/components/utils/clean_orientation.gd
```diff
@@ -0,0 +1,27 @@
+extends Control
+
+@export var clean_on_landscape: bool = false
+@export var clean_on_portrait: bool = false
+
+
+func _ready() -> void:
+	get_window().size_changed.connect(self._on_size_changed)
```
kuruk-mm: this should not be connected to size_changes... should just execute the function on the _ready()...

kuruk-mm (changes requested):

kuruk-mm (approved):
Great job!

PR #1155, opened 2026-01-09
chore: add rust errors in sentry

leanmendoza:
Enable logging tracing::error and tracing::warn in Sentry


kuruk-mm (approved):
great! can we add the tracing::info!() logs if you're in `staging` (DclGlobal.is_staging())?

PR #1182, opened 2026-01-12
feat: add alpha_texture support to UnlitMaterial with custom shader

### LINKED ISSUE #856: Move alpha_texture from PBRMaterial to UnlitMaterial
Implement: https://github.com/decentraland/protocol/pull/223

manuelmaceira:
## Summary
Adds `alpha_texture` support to `UnlitMaterial` using a GPU shader instead of CPU pixel merge.

### Changes
- New `dcl_unlit_alpha.gdshader` that combines albedo + alpha in fragment shader
- `ShaderMaterial` only when `alpha_texture` is present (transparency pipeline)
- `StandardMaterial3D` for materials without `alpha_texture` (opaque pipeline)

### Two rendering paths
To avoid putting all unlit materials into the transparency pipeline:

| Case | Material | Pipeline |
|------|----------|----------|
| Has `alpha_texture` | `ShaderMaterial` | Transparency |
| No `alpha_texture` | `StandardMaterial3D` | Opaque |

This ensures only materials that actually need transparency pay the cost of the transparency pipeline.

### Benefits over CPU approach
- No texture decompression on CPU
- No pixel-by-pixel iteration
- Compressed textures work directly on GPU

### Unity compatibility fixes
- Remove alpha_test/scissor (Unity uses alpha blend, not scissor for unlit)
- Ignore `diffuse_color` alpha (Unity ignores it for unlit materials)
- Add `repeat_disable` to textures to avoid border artifacts with clamp mode

### Plane mesh UV fix
- Fixed plane UVs to use positive values (0-1) instead of negative values
- Negative UVs broke texture clamping (`repeat_disable`)
- Inverted UV Y to match Godot's coordinate system while maintaining Unity compatibility

### Test scene
[test-scene-material-switching.zip](https://github.com/user-attachments/files/24646230/test-scene-material-switching.zip)

Closes #856

### DIFF
--- a/lib/src/scene_runner/components/material.rs
+++ b/lib/src/scene_runner/components/material.rs
@@ -145,14 +145,20 @@ pub fn update_material(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
                         mat.upcast::<Material>()
                     }
                 } else if let Some(material) = existing_material {
-                    let mut mat = material.to::<Gd<StandardMaterial3D>>();
-
-                    // Clear textures that are no longer present in the new material
-                    // This is needed when changing from a textured material to a non-textured one
-                    clear_removed_textures(&mut mat, &dcl_material);
-                    apply_dcl_material_properties(&mut mat, &dcl_material);
-
-                    mat.upcast::<Material>()
+                    // Try to reuse existing StandardMaterial3D, but if coming from ShaderMaterial
+                    // (e.g., switching from unlit+alpha_texture to pbr), create a new one
+                    if let Ok(mut mat) = material.try_to::<Gd<StandardMaterial3D>>() {
+                        // Clear textures that are no longer present in the new material
+                        // This is needed when changing from a textured material to a non-textured one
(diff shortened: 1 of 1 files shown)


godot/src/decentraland_components/mesh_renderer.gd
```diff
@@ -104,14 +104,15 @@ static func build_plane_arrays() -> Array:
...
-	uvs.append(Vector2(1, -1))
 	uvs.append(Vector2(1, 0))
-	uvs.append(Vector2(0, 0))
-	uvs.append(Vector2(0, -1))
-	uvs.append(Vector2(1, -1))
+	uvs.append(Vector2(1, 1))
+	uvs.append(Vector2(1, 1))
 	uvs.append(Vector2(1, 0))
+	uvs.append(Vector2(0, 0))
+	uvs.append(Vector2(0, 1))
```
leanmendoza: could you please add some reference to this? maybe prev/after screenshots and foundation client as well (for same scene)

PR #1215, opened 2026-01-15
fix: connect profile copy address and copy nick buttons

sdilauro:
This PR connects missing signals from the copy nickname and copy address buttons. They were also enlarged.


kuruk-mm (approved):
Thankss!

PR #1423, opened 2026-02-09
fix: search-bar on focus breaks the layout

sdilauro:
Remove unnecessary safe margin container. In this case, the design includes space for a notch or a camera.


godot/src/ui/components/discover/discover.tscn
```diff
@@ -208,7 +201,7 @@ theme_override_styles/separator = SubResource("StyleBoxEmpty_242e4")
 
 [node name="Label" type="Label" parent="TextureRect/VBoxContainer/MarginContainer/ScrollContainer_Content/VBoxContainer/MessageError"]
 layout_mode = 2
-text = "Something
+text = "Somethin
```
kuruk-mm:
```suggestion
text = "Something
```

Issue #1286, opened 2026-01-21
[Bug]: Plaza: Genesis Plaza stops responding


ileana-pr:
### 🕹️Platform

Android

### 🔢App Version

v. 45 - f6bfcca 

### 📱Device Information

SAMSUNG Galaxy S10 + 

### 📄Issue description

Gift drone at entrance to genesis plaza does not open when user hovers over it and clicks button. 

### ✅Expected behavior

Gift drone should open and show UI 

### 🖼️Screenshots / Media

**PC** 

https://github.com/user-attachments/assets/6e3ce223-7531-43fa-8400-8115b327733e

**Android** 

https://github.com/user-attachments/assets/d5c2a334-caee-43a8-9b5f-c51295fd85c7 

### ▶️Steps to reproduce

- boot app 
- jump in to Genesis Plaza 
- click on the gift drone by the entrance portal 

### 👁️Occurrence

100% of the time

ileana-pr:
recheck v. 48 - e1d32f7 

drones are responsive at first encounter but if you ignore them and come back to click on them after a few minutes they become unresponsive. 

https://github.com/user-attachments/assets/b62e5479-3ec2-4c0d-b210-bcd39b0ca31b

PR #1444, opened 2026-02-13
fix: prevent scene freezing from deferred RendererResponse

kuruk-mm:
Fix https://github.com/decentraland/godot-explorer/issues/1286
## Summary

- Fix scene thread getting permanently blocked when the main thread's time-budgeted state machine defers processing across frames without ever reaching `SendToThread`
- Stop sending `DeleteEntity` and component deletion CRDT messages for dead entities back to JS, matching bevy-explorer behavior and preventing `syncEntity failed because the id provided is already in use` errors
- Add stuck-scene detector that forces completion after 10 consecutive deferred frames (~170ms), matching bevy's approach of always processing each scene to completion

### Scene thread freezing (root cause)
The Godot explorer uses a time-budgeted state machine in `_process_scene` that can return early before reaching `SendToThread`. Unlike bevy-explorer (which always processes each scene atomically), this could leave the scene thread blocked at `recv_from_renderer` indefinitely.

**Fix:** Added a `stuck_frames` counter that tracks consecutive frames where `_process_scene` returns `false` while `waiting_process` is `true`. After 10 frames, `force_complete` mode activates: the time budget becomes infinite and `should_break` is ignored, ensuring the state machine processes through to `SendToThread`.

### syncEntity errors
Dead entities' LWW/GOS component updates were being serialized and sent back to JS, corrupting the SDK's `syncEntity` state. Bevy-explorer never sends entity deletions or their component updates back to JS.

**Fix:** Filter out entities in `dirty_crdt_state.entities.died` from both LWW and GOS component iterations in `op_crdt_recv_from_renderer`.

### CRDT lock contention
The CRDT mutex was held during `blocking_send`, creating contention with the scene thread trying to lock it after receiving.

**Fix:** Restructured `_process_scene` into two phases — CRDT-locked states and unlocked states (ProcessRpcs, SendToThread) — releasing the lock before the channel send.

### Error handling
Matched bevy-explorer's error handling: increment before logging, `<= 10` threshold, `== 10` shutdown check, no decrement on success.

## Test plan

- [x] `cargo fmt --all && cargo clippy -- -D warnings` passes
- [ ] Navigate to Genesis Plaza (0,0) — no `syncEntity` errors in logs
- [ ] Send many chat messages — scene remains responsive
- [ ] If a scene stalls, `warn!` log appears: "Scene stuck for N frames, forcing completion"
- [ ] `cargo run -- run --itest` passes

### DIFF
--- a/lib/src/dcl/js/engine.rs
+++ b/lib/src/dcl/js/engine.rs
@@ -114,7 +114,7 @@ async fn op_crdt_recv_from_renderer(
                     dirty_crdt_state.gos.len(),
                 );
                 for entity_id in dirty_crdt_state.entities.died.iter() {
-                    tracing::info!("  died entity: {:?}", entity_id);
+                    tracing::debug!("  died entity: {:?}", entity_id);
                 }
             }
 
@@ -162,7 +162,7 @@ async fn op_crdt_recv_from_renderer(
             }
 
             if skipped_lww > 0 || skipped_gos > 0 {
-                tracing::info!(
+                tracing::debug!(
                     "recv_from_renderer: skipped {} lww + {} gos updates for dead entities",
                     skipped_lww,
                     skipped_gos,
--- a/lib/src/scene_runner/update_scene.rs
+++ b/lib/src/scene_runner/update_scene.rs
@@ -82,11 +82,13 @@ pub fn _process_scene(
 ) -> bool {
     let crdt = scene.dcl_scene.scene_crdt.clone();
 
-    // When force_complete is set, use an effectively infinite time budget so the state machine
+    // When force_complete is set, use a generous 2-second time budget so the state machine
     // processes to completion in a single call. This prevents the scene thread from being
-    // blocked indefinitely when the normal time budget would cause deferral across frames.
+    // blocked indefinitely when the normal time budget would cause deferral across frames,
+    // while still capping execution to avoid freezing the client on buggy scenes.
+    const FORCE_COMPLETE_BUDGET_US: i64 = 2_000_000; // 2 seconds
     let effective_end_time_us = if force_complete {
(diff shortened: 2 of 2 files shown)


lib/src/dcl/js/engine.rs
```diff
@@ -107,8 +105,33 @@ async fn op_crdt_recv_from_renderer(
...
+                tracing::info!(
+                    "recv_from_renderer: {} entities died, {} born, {} lww dirty components, {} gos dirty components",
+                    dirty_crdt_state.entities.died.len(),
+                    dirty_crdt_state.entities.born.len(),
+                    dirty_crdt_state.lww.len(),
+                    dirty_crdt_state.gos.len(),
+                );
+                for entity_id in dirty_crdt_state.entities.died.iter() {
+                    tracing::info!("  died entity: {:?}", entity_id);
+                }
```
leanmendoza: ::debug! instead?

lib/src/scene_runner/update_scene.rs
```diff
@@ -78,390 +78,35 @@ pub fn _process_scene(
...
 ) -> bool {
     let crdt = scene.dcl_scene.scene_crdt.clone();
-    let Ok(mut crdt_state) = crdt.try_lock() else {
-        return false;
+
+    // When force_complete is set, use an effectively infinite time budget so the state machine
+    // processes to completion in a single call. This prevents the scene thread from being
+    // blocked indefinitely when the normal time budget would cause deferral across frames.
+    let effective_end_time_us = if force_complete {
+        i64::MAX
```
leanmendoza: can we put an end_time_us for two seconds instead ? i64:MAX could bring undesired user experience

PR #1659, opened 2026-03-09
feat: use SFSafariViewController for Apple Sign In and Skip FTUE

leanmendoza:
## Summary
- Adds `open_safari_auth_url` method using `SFSafariViewController` specifically for Apple Sign In, which requires shared cookies/localStorage with Safari
- Keeps `open_auth_url` using `ASWebAuthenticationSession` for all other auth providers (Google, Discord, WalletConnect, etc.)
- Routes `provider=apple` URLs through the Safari VC path in `global.gd`
- Skips FTUE after avatar naming, going directly to discover

## Test plan
- [ ] Test Apple Sign In on iOS device — should open SFSafariViewController
- [ ] Test Google/Discord sign in on iOS — should use ASWebAuthenticationSession as before
- [ ] Test WalletConnect sign in on iOS
- [ ] Verify deeplink callback dismisses SFSafariViewController automatically
- [ ] Verify auth flow completes and returns to app correctly for all providers


kuruk-mm (approved):
tested and working!

PR #1757, opened 2026-03-25
fix: apple review issues

### LINKED ISSUE #1756: Issues encountered by Apple reviewer
The apple reviewer encountered a breaking issue that prevented scenes from loading.

To reproduce:
- Open app, login
- Enter a scene
- Logout
- Login again
- Enter a scene (the scene will be broken)

https://github.com/user-attachments/assets/4277f6a4-6f9c-4169-990e-34f5e6d7a3ce

EibrielInv:
## Summary                                                                                                                                                                                                      
                                                                                                                                                                                                                  
  - Fix crash on scene load after logout/login: `SceneManager.base_ui` becomes a dangling reference when the Explorer scene is torn down; added `recreate_base_ui()` called at the start of `explorer._ready()`   
  - Fix `set_preview_url` being called unconditionally — now only called when `deep_link_obj.preview` is not empty (prevents connecting preview WebSocket to non-preview realms)                                  
  - Fix crash in `SafeMarginContainer._on_size_changed()` when node is not yet in the tree or has no viewport                                                                                                     
                                                                                                                                                                                                                  
Closes  #1756 

## Test plan                                                                                                                                                                                                    
                                                                                                                                                                                                                  
  - [ ] Login, logout, login again — verify no crash on scene load                                                                                                                                                
  - [ ] Launch with a normal realm deeplink (no `?preview=`) — verify no WebSocket connection attempt
  - [ ] Launch with `?preview=<url>` — verify hot-reload WebSocket connects correctly                                                                                                                             
  - [ ] Test on a device with safe area insets — verify no crash during orientation changes or scene transitions

### DIFF
--- a/lib/src/scene_runner/scene_manager.rs
+++ b/lib/src/scene_runner/scene_manager.rs
@@ -299,6 +299,9 @@ impl SceneManager {
     /// leaving SceneManager with a dangling Gd<DclUiControl> reference.
     #[func]
     fn recreate_base_ui(&mut self) {
+        if self.base_ui.is_instance_valid() {
+            self.base_ui.clone().free();
+        }
         let mut base_ui = DclUiControl::new_alloc();
         base_ui.set_anchors_preset(LayoutPreset::FULL_RECT);
         base_ui.set_mouse_filter(MouseFilter::IGNORE);


lib/src/scene_runner/scene_manager.rs
```diff
@@ -293,6 +293,22 @@ impl SceneManager {
...
+    /// leaving SceneManager with a dangling Gd<DclUiControl> reference.
+    #[func]
+    fn recreate_base_ui(&mut self) {
+        let mut base_ui = DclUiControl::new_alloc();
+        base_ui.set_anchors_preset(LayoutPreset::FULL_RECT);
+        base_ui.set_mouse_filter(MouseFilter::IGNORE);
+        base_ui.set_name("scenes_ui");
+        let callable_on_ui_resize = self.base().callable("_on_ui_resize");
+        base_ui.connect("resized", &callable_on_ui_resize);
+        self.base_ui = base_ui;
```
regenesis-claw: **Potential node leak on first call:** `SceneManager::init()` already allocates a `base_ui` via `DclUiControl::new_alloc()`. On the very first `explorer._ready()`, `recreate_base_ui()` overwrites `self.base_ui` without freeing the original node. Since `DclUiControl` extends `Control` (a `Node`), it won't be automatically freed when the `Gd` reference is dropped — it becomes an orphan node leak.

On subsequent calls this is fine because the scene teardown frees the old node. But the first call always leaks one.

Suggestion: free the old instance if it's still valid before replacing:
```rust
fn recreate_base_ui(&mut self) {
    if self.base_ui.is_instance_valid() {
        self.base_ui.clone().free();
    }
    // ... rest of the method
}
```
This also makes the method safe to call in any scenario, not just after teardown.

PR #1772, opened 2026-03-31
feat: poor connection notification and modal

### LINKED ISSUE #255: Bad Network Connection HUD Awareness + Modal
Problem: Currently the client starts failing and shows an empty Avatar Preview when connectivity is bad

As a user, I want to be aware of network performance issues and disconnections so I can react accordingly:

- [ ] -Design Implement a bad connection HUD indicator when network performance is degrading
- [ ] -Design Implement a Disconnect / Reconnect Error Handling Modal when Connection is Lost

Desktop Sample for Connection Error 
Image

sdilauro:
Closes #255 

## Summary                                                                                                      
                                                            
  - Add ConnectionQualityMonitor autoload that periodically pings the realm health endpoint to detect poor/lost
   connection states
  - Show an alert toast when connection quality degrades (POOR state) and a blocking modal with retry/exit when
   connection is lost (LOST state)                          
  - Add blur background shader and blocker property to modals, replacing the old dismissable flag
  - Wrap modals in a CanvasLayer (layer 100) to ensure they render above all overlays
  - Mitigate race condition by discarding health check results when the realm URL changes mid-request

  Test plan

  - Verify poor connection toast appears after ~10s of degraded connectivity (2 consecutive failures × 5s
  interval)
  - Verify connection lost modal appears after ~20s without connectivity (4 consecutive failures)
  - Verify RETRY button resets state and re-polls at 2s intervals, escalating to modal after 2 more failures
  - Verify EXIT button quits the app without closing the modal first
  - Verify existing modals (external link, teleport, change realm) remain dismissable by tapping outside
  - Verify scene crash modal still blocks input correctly
  - Verify blur background renders correctly on mobile
  
  
  
  
  
  # Connection Quality Monitor                                                                                      
                                                            
  ## Polling intervals                                                                                               
   
  - Fast (0.5s): Default interval, used at startup and whenever connection is degraded.                           
  - Slow (2.0s): Used after a successful ping.              

  ## Flow

  1. The monitor starts pinging the realm health endpoint every 0.5s.
  2. On a successful ping, the interval switches to 2.0s.
  3. On a failed ping (error or response > 5000ms), the interval switches back to 0.5s.
  4. After 2 consecutive failures:
    - With explorer UI: A "poor connection" toast notification is shown. If failures continue to 4, the
  "connection lost" modal is shown.
    - Without explorer UI: The "connection lost" modal is shown directly.
  5. The monitor keeps polling at 0.5s while the connection is degraded.
  6. If a ping succeeds while the modal is open, the modal auto-closes and the interval goes back to 2.0s.

  ## Retry button behavior

  - Closes the modal and resets the error counter.
  - Polling continues at 0.5s.
  - After retry, a single failure is enough to reopen the modal directly (skipping the toast notification).
  - If the connection recovers, the retry flag is cleared and the interval goes back to 2.0s.

  ## iOS specifics

  - The EXIT button is always hidden (Apple doesn't allow programmatic quit).
  - The first time connection is lost, the modal shows a RETRY button.
  - If the retry fails and the modal appears a second time, it shows no buttons at all with the message "Try
  restarting the app." — it can only be dismissed by the connection restoring automatically.

### DIFF
--- a/godot/src/ui/components/modal/modal.gd
+++ b/godot/src/ui/components/modal/modal.gd
@@ -16,8 +16,9 @@ var blocker: bool = false
 @onready var icon: TextureRect = %Icon
 @onready var button_secondary: Button = %Button_Secondary
 @onready var button_primary: Button = %Button_Primary
-@onready var buttons_container: HBoxContainer = %HBoxContainer
 @onready var panel_container: PanelContainer = $PanelContainer
+@onready var buttons_separator: HSeparator = %HSeparator_ButtonsSeparator
+@onready var buttons_container: HBoxContainer = %HBoxContainer_Buttons
 
 
 func _ready() -> void:
--- a/godot/src/ui/components/modal/modal.tscn
+++ b/godot/src/ui/components/modal/modal.tscn
@@ -155,17 +155,18 @@ max_lines_visible = 2
 visible_characters = 106
 visible_ratio = 1.0707071
 
-[node name="HSeparator" type="HSeparator" parent="PanelContainer/MarginContainer_Content/VBoxContainer"]
+[node name="HSeparator_ButtonsSeparator" type="HSeparator" parent="PanelContainer/MarginContainer_Content/VBoxContainer"]
+unique_name_in_owner = true
 layout_mode = 2
 theme_override_constants/separation = 53
 theme_override_styles/separator = SubResource("StyleBoxEmpty_figov")
 
-[node name="HBoxContainer" type="HBoxContainer" parent="PanelContainer/MarginContainer_Content/VBoxContainer"]
+[node name="HBoxContainer_Buttons" type="HBoxContainer" parent="PanelContainer/MarginContainer_Content/VBoxContainer"]
 unique_name_in_owner = true
 layout_mode = 2
 theme_override_constants/separation = 21
 
-[node name="Button_Secondary" type="Button" parent="PanelContainer/MarginContainer_Content/VBoxContainer/HBoxContainer"]
+[node name="Button_Secondary" type="Button" parent="PanelContainer/MarginContainer_Content/VBoxContainer/HBoxContainer_Buttons"]
 unique_name_in_owner = true
 custom_minimum_size = Vector2(0, 74)
 layout_mode = 2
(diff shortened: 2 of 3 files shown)


godot/src/connection_quality_monitor.gd
```diff
@@ -0,0 +1,197 @@
...
+
+func _get_health_url() -> String:
+	var realm_url: String = Global.realm.realm_url
+	if realm_url.is_empty():
+		return DclUrls.peer_base() + "/about"
+	return realm_url + "about"
+
+
+func _on_poor_connection() -> void:
+	if not Global.get_explorer():
```
regenesis-claw: This blanket disconnect of all `pressed` connections is fragile — it removes whatever the modal_manager itself wired up. If modal_manager's button handling changes later, this will silently break. Consider storing and disconnecting only the specific callable you want to replace.

godot/src/connection_quality_monitor.gd
```diff
@@ -0,0 +1,212 @@
...
+## ConnectionQualityMonitor
+##
+## Autoload that periodically pings a lightweight endpoint to assess
+## connection quality. Emits signals consumed by the toast and modal systems.
+##
+## Polling:
+##   - Fast (0.5s): default, and while connection is degraded
+##   - Slow (2.0s): after a successful ping
+##
+## After 2 consecutive failures:
```
regenesis-claw: The fast poll (0.5s) also doubles as the HTTP timeout via `_create_timeout_promise(_poll_timer.wait_time)`. On mobile networks, 500ms is a very tight timeout — healthy connections with normal latency (300–800ms) would register as failures. Consider a separate `REQUEST_TIMEOUT_SECONDS` constant (e.g. 3–5s) for the timeout, independent of the poll interval.

godot/src/connection_quality_monitor.gd
```diff
@@ -0,0 +1,212 @@
...
+	):
+		_state = State.POOR
+		poor_connection_detected.emit()
+	elif (
+		_consecutive_errors
+		>= (CONSECUTIVE_ERRORS_FOR_LOST if has_explorer else CONSECUTIVE_ERRORS_FOR_DEGRADED)
+	):
+		_state = State.LOST
+		connection_lost_detected.emit()
+
```
regenesis-claw: The `SceneTreeTimer` created here will still fire even if the HTTP promise wins the race and resolves first. When it does, it calls `p.reject("timeout")` on a potentially already-resolved promise. If `Promise` silently ignores double-resolution, this is fine. Otherwise, consider adding a guard (`if not p.is_resolved()`) or disconnecting the timer callback after the race completes.

PR #1822, opened 2026-04-09
feat: add mentions autocomplete in chat

### LINKED ISSUE #1002: Text Chat - Tagging Name Autocomplete Feature
### 🕹️Platform

Android

### 🔢App Version

5afae02

### 📱Device Information

SAMSUNG Galaxy S10 +

### 📄Issue description

user names are not auto populating when entering @ character in chat 

### ✅Expected behavior

names of users nearby should populate in chat when the user enters @ char and starts typing a user name. 

### 🖼️Screenshots / Media

**Media Folder** 
https://drive.google.com/drive/folders/1wRFyZR0nVzvvyyzo2WVSdeEkNjKPtsxd?usp=drive_link

**PC Client Video** 
https://drive.google.com/file/d/1GMqD2WNpgpIyLsJFZ66jIQp8UHVPTSN5/view?usp=drive_link

**Android Video** 
https://drive.google.com/file/d/1JWVqbcttbv6jb4qgK5hxO-GzalgPB30S/view?usp=drive_link

### ▶️Steps to reproduce

- log in
- jump to any scene
- click chat icon 
- start typing a user name beginning with @ char

### 👁️Occurrence

100% of the time

sdilauro:
Close #1002 
## Summary

- Adds **@ mention autocomplete** to the chat: typing `@` shows a scrollable list of nearby players, filtered as you type
- `MentionItem` changed from `Button` to `Control` with manual `_gui_input` to prevent focus stealing from the `LineEdit`
- Autocomplete panel and scroll set to `focus_mode=0` / `mouse_filter=IGNORE` so clicking a suggestion doesn't defocus the input
- After inserting a mention programmatically, syncs the OS virtual keyboard buffer via `DisplayServer.virtual_keyboard_show()` so backspace works correctly on the inserted text
- Autocomplete updates are debounced (one deferred call per frame) and child nodes are properly freed on hide

## Test plan

- [ ] Open chat, type `@` → autocomplete panel appears with nearby players
- [ ] Type partial name → list filters correctly (case-insensitive)
- [ ] Click a suggestion → mention is inserted as `@Name ` with no duplicate `@`
- [ ] After inserting, press backspace repeatedly → all characters delete normally (no "locked" text)
- [ ] After inserting, type more text, then backspace through everything → works seamlessly
- [ ] Click outside the chat / press Escape → autocomplete hides, chat closes normally
- [ ] Verify on mobile (Android/iOS) that the virtual keyboard stays in sync after mention insertion
- [ ] Send a message with a mention → message sends correctly with `@Name` in it


godot/src/ui/components/chat/chat.gd
```diff
@@ -212,3 +228,146 @@ func _on_line_edit_command_focus_exited() -> void:
...
+
+func _update_autocomplete() -> void:
+	_autocomplete_queued = false
+	var query = _get_mention_query()
+	if query == null:
+		_hide_autocomplete()
+		return
+
+	var avatars: Array = _get_matching_avatars(query)
+	if avatars.is_empty():
```
regenesis-claw, line 300: Nit: you could slice `avatars` before passing to `_show_autocomplete` (e.g. `avatars.slice(0, MAX_AUTOCOMPLETE_RESULTS)` or a slightly higher cap like 10) to avoid instantiating nodes that will never be scrolled into view in practice. Not blocking — just a minor optimization.

regenesis-claw (approved):
Clean and well-structured implementation — nice work! 🎉

The focus-management approach (using `Control` + `_gui_input` instead of `Button`) is the right call to prevent the `LineEdit` from losing focus, and the virtual keyboard buffer sync for mobile is a great detail.

One minor suggestion:

**Consider capping the number of instantiated `MentionItem` nodes.**
`_get_matching_avatars()` returns all matches but `_resize_autocomplete_scroll` only shows `MAX_AUTOCOMPLETE_RESULTS` (3) at a time. If the user types just `@` with many nearby players, all of them get instantiated as scene nodes even though only 3 are visible. Not a real problem for typical player counts, but a `.resize(mini(results.size(), some_cap))` or slicing the array before instantiation would keep it bounded.

Everything else looks solid — query parsing, debounce, cleanup on hide, edge cases. LGTM ✅

godot/src/ui/components/chat/chat.gd
sdilauro: The instances are lightweight nodes; I don't think we need to overcomplicate things—I think we can easily handle the scene's nearby-players limit.

rodrigo685 (approved):

PR #1826, opened 2026-04-10
fix: connection health checker bug + monitor cleanup

sdilauro:
## Summary

Fixes a bug where loading a non-existent world (e.g. `/world world-that-does-not-exist`) wrongly triggered the *Connection Lost* modal, and reworks the `ConnectionQualityMonitor` to be lighter, simpler and decoupled from `ModalManager` internals.

## Bug fix: `/world non-existent` no longer shows "Connection Lost"

**Root cause** — `Realm.async_set_realm` mutated `realm_url` / `realm_string` *before* validating the new `/about` endpoint. When the endpoint returned 404, the explorer was left in a half-state pointing to a broken URL, and `ConnectionQualityMonitor` started counting those 404s as connection failures, eventually showing the *Connection Lost* modal.

**Fix**
- `Realm.async_set_realm` now validates `/about` against a *candidate* URL and only commits the new realm state on success. On failure (network error, non-2xx, invalid JSON) the previous realm state is left intact.
- New signals: `realm_changing` (emitted at the start of every realm change) and `realm_change_failed(new_realm_string, reason)` (emitted on validation failure).
- `async_set_realm` now returns `bool`. All existing call-sites are unchanged because they ignored the (previously `void`) return value.
- A "World unavailable" toast is shown on `realm_change_failed`, wired only to `Global.realm` so transient `Realm.new()` instances (e.g. portable experiences) don't generate spurious user-facing toasts.

### Pause polling during realm changes
The monitor subscribes to the new `realm_changing` / `realm_changed` / `realm_change_failed` signals and stops the poll timer while a realm change is in flight. Any in-flight check is discarded via a generation counter. This prevents 404s from a broken realm switch from being counted as connection errors.

### Lower polling cost (~25× less traffic)
- `SLOW_POLL_SECONDS`: `2.0 → 10.0`. The monitor now polls every 10s on the happy path instead of every 2s. `FAST_POLL_SECONDS` stays at `0.5s` so detection remains fast once a failure is observed.
- `GET` → `HEAD` against `/about`. Same status-code signal, no JSON payload downloaded.
- Combined: roughly **5× fewer requests × 10–50× less bytes per request**.

Worst-case detection latency: ~12s for *poor connection*, ~16s for *connection lost* — acceptable for a UX-level monitor that previously cost 2 req/s of multi-KB responses against the catalyst.

### Decoupled from ModalManager internals
- Removed a 12-line hack in `_async_on_connection_lost` that reached into `modal_manager.current_modal.button_secondary`, disconnected a private handler and connected its own. The hack existed so the *Connection Lost* modal would stay visible while `get_tree().quit()` runs (instead of flashing the broken UI underneath for a frame).
- `ModalManager._on_connection_lost_secondary` no longer calls `close_current_modal()`; the listener is now responsible for both quitting and modal lifecycle. The signal-based wiring (`connection_lost_exit → CQM._on_exit`) handles everything cleanly.

### Modal type guard
`_on_connection_restored` no longer dismisses *any* current modal — it only closes the *Connection Lost* modal if it was the one **we** opened (tracked via `_showing_our_modal`). This prevents accidentally dismissing unrelated modals that may have been opened in the meantime.

### `_retrying` removed from the state machine
The "post-retry single failure goes straight to LOST" rule was previously a flag inspected inside `_update_state`, mixing retry handling with the GOOD/POOR/LOST transitions. It now lives in `_async_check_connection` as an explicit branch on the failure path. `_update_state` is back to being a pure function of `_consecutive_errors`.

### Other cleanups
- Removed three internal signals (`poor_connection_detected`, `connection_lost_detected`, `connection_restored`) that were only being bounced through to private methods on the same class. Replaced with direct calls. No external listeners existed.
- `print` → `printerr` for error/timeout cases so they show up in the right log channel.
- The poll timer now starts in `_ready()` again (after briefly being gated on the first realm). It uses the existing `peer_base()` fallback so connection monitoring still works on the lobby / backpack / discover screens before any realm is loaded.

## Test plan

- [ ] Type `/world some-world-that-does-not-exist` while in a valid realm → "World unavailable" toast appears, no Connection Lost modal, the previous realm stays loaded.
- [ ] Disconnect the network in-game → Connection Lost modal appears within ~12s.
- [ ] Press *Retry* on the Connection Lost modal with the network still down → modal reappears immediately on the next failed ping (no waiting 4 fails).
- [ ] Press *Retry* on the Connection Lost modal after restoring the network → modal closes and gameplay resumes.
- [ ] Press *Exit* on the Connection Lost modal → modal stays visible until the app fully closes (no flash of the broken UI).
- [ ] Disconnect the network on the lobby / backpack / discover screens (no realm loaded) → Connection Lost modal still appears (uses `peer_base()` fallback).
- [ ] `/changerealm` to a valid realm → realm loads normally, monitor resumes against the new realm.
- [ ] Switch between several valid worlds in quick succession → no false Connection Lost notifications.

### DIFF
--- a/godot/src/connection_quality_monitor.gd
+++ b/godot/src/connection_quality_monitor.gd
@@ -40,12 +40,20 @@ func _ready() -> void:
 	_poll_timer.timeout.connect(_on_poll_timeout)
 	add_child(_poll_timer)
 	_poll_timer.start()
+	# Before a realm is set (lobby / backpack / discover), _get_health_url() falls
+	# back to peer_base() so we still detect real connection loss on those screens.
 
 	poor_connection_detected.connect(_on_poor_connection)
 	connection_lost_detected.connect(_async_on_connection_lost)
 	connection_restored.connect(_on_connection_restored)
 	Global.modal_manager.connection_lost_retry.connect(_on_retry)
 	Global.modal_manager.connection_lost_exit.connect(_on_exit)
+
+	# Pause polling while a realm change is in flight so 404s / slow /about calls
+	# on the new realm don't get counted as connection failures.
+	Global.realm.realm_changing.connect(_on_realm_changing)
+	Global.realm.realm_changed.connect(_on_realm_changed)
+	Global.realm.realm_change_failed.connect(_on_realm_change_failed)
 
 
 func _on_poll_timeout() -> void:
@@ -63,7 +71,9 @@ func _async_check_connection() -> void:
 		_is_checking = false
 		return
 
-	var promise: Promise = Global.http_requester.request_json(url, HTTPClient.METHOD_GET, "", {})
+	# HEAD instead of GET: we only care about reachability + status code, not the
+	# /about payload (which is several KB of realm metadata per poll).
+	var promise: Promise = Global.http_requester.request_json(url, HTTPClient.METHOD_HEAD, "", {})
 	var timeout_promise := _create_timeout_promise(REQUEST_TIMEOUT_SECONDS)
 
 	var start_ms := Time.get_ticks_msec()
@@ -211,3 +221,30 @@ func _on_retry() -> void:
 	_retrying = true
 	if OS.get_name() == "iOS":
 		_ios_retry_used = true
+
+
+func _on_realm_changing() -> void:
+	# Discard any in-flight check and stop pinging until the realm change settles.
+	_poll_timer.stop()
+	_check_generation += 1
+	_is_checking = false
+	_consecutive_errors = 0
+
+
+func _on_realm_changed() -> void:
+	# New realm is validated: (re)start polling from a clean slate.
+	_resume_polling()
+
+
+func _on_realm_change_failed(_new_realm_string: String, _reason: String) -> void:
+	# Realm change was rejected. Resume polling either against the previous realm
+	# (if any) or against the peer_base() fallback used in the pre-realm screens.
+	_resume_polling()
+
+
+func _resume_polling() -> void:
+	_state = State.GOOD
+	_consecutive_errors = 0
+	_retrying = false
+	_set_poll_interval(FAST_POLL_SECONDS)
+	_poll_timer.start()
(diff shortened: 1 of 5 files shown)


regenesis-claw (approved):
## Review — Approved ✅

Excellent PR. The root cause analysis and fix are solid: validating the candidate realm URL *before* committing state is the right approach, and the new signal trio (`realm_changing` / `realm_changed` / `realm_change_failed`) gives the monitor clean lifecycle hooks to pause/resume polling.

### Highlights

**Realm validation (realm.gd)**
- Candidate URL pattern is correct — state only commits after `/about` returns valid JSON.
- All failure paths emit `realm_change_failed` and return `false`, so callers and the monitor stay in sync.
- Return type change (`void` → `bool`) is backward-compatible since all existing call-sites ignore the return.

**ConnectionQualityMonitor**
- Pausing the poll timer during realm transitions prevents false positives from in-flight 404s. Generation counter correctly discards stale checks.
- `_showing_our_modal` guard is a nice fix — previously `_on_connection_restored` could dismiss any unrelated modal.
- `HEAD` + 10s slow poll is a meaningful traffic reduction with acceptable latency tradeoff.
- Moving the retry-failure shortcut out of `_update_state` into `_async_check_connection` makes the state machine easier to reason about.
- Removing the button-rewiring hack in `_async_on_connection_lost` and relying on signal-based wiring is much cleaner.

**ModalManager**
- Not closing the modal in `_on_connection_lost_secondary` so the CQM `_on_exit` handler keeps it visible during quit — clean separation of concerns.

### Minor suggestion (non-blocking)

In `realm.gd`, the validation checks that the response is a `Dictionary`, but doesn't verify the HTTP status code. If a server returns a 2xx with unexpected JSON (or a 404 with a valid JSON error body), the realm would be committed with potentially broken data. Consider adding a status code check (`response.status_code == 200`) before parsing the body. Not blocking since this is pre-existing behavior and unlikely in practice.

Requested by Seba via Slack

PR #1830, opened 2026-04-13
feat: add Scene Inspector wire protocol

leanmendoza:
# Scene Inspector

Instrumentation system that captures CRDT traffic, JS op-calls, lifecycle events, and performance snapshots from running Decentraland scenes, and streams them to an external consumer (Creator Hub, Godot editor addon, or JSONL file).

## Activation

- `--scene-inspector=true` / `?scene-inspector=true` — reuse the preview WebSocket channel
- `--scene-inspector=ws://host:port` — dedicated WebSocket target (auto-reconnect w/ exponential backoff)
- `--scene-inspector-file` — write JSONL on disk (rotates at 100 MB, caps total at 1 GB)
- `--scene-debug` — alias for `--scene-inspector=true --scene-inspector-file`

## What it captures

- **CRDT messages** in both directions (`s2r` / `r2s`), payload decoded to JSON + optional hex; `build.rs` auto-generates the proto decoder table so it stays in sync with the `.proto` sources
- **Deno op-calls** with args, result, duration, async flag, and error — wraps `Deno.core.ops` only for debugged scenes (zero overhead otherwise)
- **Scene lifecycle**: `SceneInit → MainCrdtLoaded → ScriptLoaded → OnStart{,End} → OnUpdate{,End} → SceneShutdown`
- **Console**: `op_log` / `op_error` output
- **Performance snapshot** every 2s: FPS, draw calls, primitives, objects, memory (static / GPU / Rust heap), V8 heap (used/total/limit/external), scene count, assets loading/loaded, download speed
- **Session markers** with version, platform, device name (iOS/Android native plugins)

## Wire protocol

WebSocket JSON: `{ type: "SCENE_INSPECTOR", payload: { sessionId, entries } }`.
Incoming commands via `SCENE_INSPECTOR_CMD` / `SCENE_INSPECTOR_CMD_ACK`:
`pause`, `resume`, `reload_scene`, `get_status`, `set_file_logging`, `set_perf_interval`, `set_lifecycle_verbose`.

## Output channels

- WebSocket (preview or dedicated)
- JSONL file with rotation and cleanup
- **Hot-connect**: newly attached consumers receive a CRDT state snapshot (LWW + GOS) so they can reconstruct the entity tree without replaying history

## Future work

- Per-scene inspector activation (currently all-or-nothing via `scene_inspector_active`)
- Multi-consumer fan-out from a single dispatcher
- Authentication / token on the WebSocket for untrusted networks
- Richer command set (step, pause-on-component, breakpoints)
- Inspector editor addon panel consuming the `EngineDebugger` channel (wire is bridged, UI pending)
- Scene-scoped filters on the dispatcher to keep payloads small on multi-scene deployments


godot/src/global.gd
```diff
@@ -56,6 +56,7 @@ const FORCE_TEST_LOCATION = Vector2i(54, -55)
 # const FORCE_TEST_REALM = "http://localhost:8000"
 
 const FORCE_DEEPLINK = ""
+#const FORCE_DEEPLINK = "decentraland://open?scene-inspector=true"
```
leanmendoza:
```suggestion
```

godot/project.godot
```diff
@@ -35,8 +35,8 @@ boot_splash/minimum_display_time=300
 
 [autoload]
 
-UiSounds="*res://src/helpers_components/ui_sounds.gd"
 Global="*res://src/global.gd"
+UiSounds="*res://src/helpers_components/ui_sounds.gd"
 NotificationsManager="*res://src/notifications_manager.gd"
```
leanmendoza, line 40: revert this

godot/src/tool/scene_inspector_bridge.gd
leanmendoza, line 1: do not support preview ws at the moment,only dedicated ws

lib/src/dcl/crdt/message.rs
```diff
@@ -204,12 +340,6 @@ pub fn append_gos_component(
     Ok(())
 }
 
-pub fn delete_entity(entity_id: &SceneEntityId, writer: &mut DclWriter) {
-    writer.write_u32(CRDT_DELETE_ENTITY_HEADER_SIZE as u32);
-    writer.write(&CrdtMessageType::DeleteEntity);
-    writer.write(entity_id);
-}
-
```
leanmendoza, line 212: maybe this should be tagged as dead code but not removed

lib/src/dcl/js/js_modules/fetch.js
leanmendoza, line 1: remove modified spaces

regenesis-claw (approved):
## Scene Inspector Wire Protocol — Review

Well-architected instrumentation system. The zero-overhead design for non-debug scenes is solid: `#[cold] #[inline(never)]` on all logging paths, a single predictable branch on `should_debug`, and the JS `setupOpLogging` wrapper bails immediately when the flag is off. The channel design (bounded mpsc with non-blocking `try_send` + `DROPPED_COUNT`) is the right trade-off — never block scene threads, report loss instead.

### Architecture highlights
- Clean module split: `config`, `dispatcher`, `logger`, `storage` under `tools/scene_inspector/`
- `build.rs` auto-generates both the proto decoder table and the `component_id_to_name` mapping — stays in sync with `.proto` sources automatically
- Hot-connect CRDT snapshot (LWW + GOS) lets late-joining consumers reconstruct entity state
- File rotation (100 MB per file, 1 GB total cap) with graceful per-file error handling in cleanup

1. **`--scene-inspector=true` target routing**: When the target is `"true"`, `SceneInspectorBridge._connect_to_target()` rejects it (`begins_with("ws://")` check fails), so the batch signal handler is connected but `_dedicated_ws` stays null — entries go nowhere via WebSocket. The Rust instrumentation and file logging still work, but the "reuse preview WS channel" path mentioned in the PR description doesn't appear to be wired up yet. Might be intentional (future work listed in the PR body), but worth a note so consumers of `--scene-inspector=true` know they need `--scene-inspector-file` to get output.

2. **Paused scene skip in `scene_manager.rs`** (line ~931): The new `if scene.paused { continue; }` inside the `SceneState::Alive` block skips the thread-finished check for paused scenes. If a scene thread crashes while paused, the "scene closed without kill signal" detection won't fire until it's resumed. Low risk since pausing is only triggered via inspector commands.

3. **GOS snapshot cap**: The per-key cap of 100 entries in `update_crdt_snapshot` is per-(scene, entity, component). For scenes with many entities using append-based components, total snapshot memory could grow. The comment acknowledges this — just confirming it's on the radar.

4. **JS `safeSerialize` depth limit**: Max depth of 3 with a 20-key cap per object — good defensive defaults for the hot path. Arrays > 10 items get sampled to 5, which is reasonable.

LGTM — clean design, zero overhead for production scenes, well-documented code with thorough comments explaining trade-offs.

Reviewed by Claw 🤖 — Requested by Lean via Slack

manuelmaceira (approved):
lgtm

PR #1837, opened 2026-04-14
fix: preserve emoji in iOS notification DB queries

### LINKED ISSUE #1832: [Bug] Wave emoji not showing in iOS Comeback notification
### Platform
iOS

### App Version
Not specified

### Device Information
Not specified

### Issue Description
The wave emoji is not displaying correctly in the iOS Comeback notification. User reported that this issue is still occurring ("sigue mal lo de la notificacion, no se ve el emoji").

### Expected Behavior
The wave emoji should be visible in the notification text.

### Screenshots / Media
User provided screenshot (IMG_6605.PNG) showing the notification without the emoji.

### Steps to Reproduce
1. Leave the app and wait for Comeback notification to trigger
2. View the notification on iOS device
3. Observe that the wave emoji is missing

### Occurrence
Not specified

---
Requested by Rodri via Slack

manuelmaceira:
## Summary
- Use `String::utf8()` instead of `String()` when converting NSString values back to Godot Strings in `db_query_notifications()` and `db_get_notification()`
- Godot 4's `String(const char*)` interprets bytes as Latin-1, corrupting multi-byte UTF-8 characters (like emoji 👋) when reading notification data from SQLite
- This was the root cause of the missing wave emoji in the Day 1 "Comeback" notification on iOS

## Test plan
- [x] Tested on iOS device: uninstall app, deploy fresh, accept notification permissions, wait 30s (with DEBUG_DAY1_SHORT_DELAY=true) — emoji 👋 now displays correctly in the notification

Fixes #1832


regenesis-claw (approved):
Clean fix 👍

The change from `String(const char*)` → `String::utf8(const char*)` is the correct approach — Godot 4's `String(const char*)` constructor interprets bytes as Latin-1, which silently corrupts any multi-byte UTF-8 sequences (emoji, CJK, accented chars, etc.). `String::utf8()` properly decodes the UTF-8 bytes that `[NSString UTF8String]` returns.

Both call sites (`db_query_notifications` and `db_get_notification`) are patched consistently.

Minor note: the dictionary *keys* on lines 987 and 1065 still use `String([key UTF8String])` — not an issue today since keys are ASCII identifiers, but worth keeping in mind if that ever changes.

kuruk-mm (approved):

PR #1848, opened 2026-04-15
feat: landscape settings panel - disable skybox handler when scene force it

### LINKED ISSUE #1690: Disable custom skybox setting when scene controls skybox time
### Description

When a scene forces the skybox time, the mobile app should detect this and handle the conflict between scene-controlled skybox and user settings:

1. **Display a warning** in the settings UI indicating that the scene is controlling the skybox
2. **Disable the custom skybox option** since the scene has taken control

This prevents user confusion when their custom skybox settings have no effect because the scene is overriding them.

### Expected Behavior

- When a scene uses the skybox time control, the settings menu should clearly indicate this state
- The custom skybox controls should be disabled/grayed out while the scene is in control
- When the user leaves the scene or the scene releases skybox control, the custom settings should be re-enabled

### Related Issues

- #700 — Skybox Review

---
Requested by Mateo via Slack

### LINKED ISSUE #1770: New Settings Design (Landscape) - Implementation
Implement new settings on landscape 

[FIGMA](https://www.figma.com/design/l6t9Qs4ZueQdFhE6Gcp6hZ/Settings?node-id=1002-2897&t=BPr2yNH0xaxOSU4f-1)

[ATLAS](https://www.figma.com/design/l6t9Qs4ZueQdFhE6Gcp6hZ/Settings?node-id=1086-6085&t=BPr2yNH0xaxOSU4f-1)

**Changes:**
SETTINGS title 44pt

_Apply all the sections:_
- [ ] Title font size, change to 24
- [ ] In section, if have more than 1 subcategory, spacing change to 56px. Example:
- [ ] Content width adapt to width panel

Image

_Dropdown_
Dropdown open, shows 3 option max and scroll.

_Image preview skybox_
change height 

_Changes buttons:_
Usa medium button on sign out and clean cache.

REFE
Image

sdilauro:
Closes #1770
Closes #1690

## Summary                                                                                                                                                                
                                                                                                                                                                            
  - Added landscape mode support for the settings panel, allowing it to be used as a side panel inside the explorer                                                         
  - Added a Rust signal `sdk_skybox_time_active_changed(bool)` on `DclGlobal` that fires when a scene starts or stops controlling the skybox via the SkyboxTime SDK component
  - Settings now subscribes to this signal to disable the dynamic skybox toggle and custom skybox dropdown when the current scene is controlling the skybox, showing a warning label to the user


  ## Test plan

  - [ ] Open settings in landscape mode and verify it displays correctly as a side panel
  - [ ] Enter a scene that sets a fixed skybox time — verify the dynamic skybox toggle and dropdown are disabled and the warning label is visible
  - [ ] Teleport to a scene without fixed skybox — verify the controls are re-enabled and the warning label is hidden
  - [ ] Toggle dynamic skybox on/off in a scene without fixed skybox — verify the dropdown enables/disables accordingly


regenesis-claw (approved):
## Review: Landscape settings panel + skybox handler disable

Nice work — clean refactor with solid Rust-side signal plumbing. A few notes:

### ✅ What looks good
- **Rust signal design**: `sdk_skybox_time_active_changed` fires correctly on state transitions (became active / was active), with proper `drop(global_bind)` before `emit_signal` to avoid borrow issues.
- **Optimization improvement** in `skybox_time.rs`: the early-return now also compares `fixed_time` and `transition_forward`, which is more correct than the previous version that skipped updates when already active regardless of value changes.
- **Scene manager cleanup**: properly emits the signal when resetting skybox on scene change.
- **Panel mode pattern**: the `panel_mode` export with landscape-aware layout is well structured. The `_apply_layout` function cleanly adjusts sizing/fonts for both orientations.
- **Skybox warning UX**: disabling the toggle + dropdown + showing a warning label when the scene controls the skybox is the right approach. Guard in `_on_check_button_dynamic_skybox_toggled` prevents user from overriding SDK control.
- **Cleanup**: removed stale `print()` debug line, unused radio selectors, dead HSeparator nodes, and the old back button — good housekeeping.
- **Mouse filter fixes**: adding `mouse_filter = 2` (IGNORE) on separators and `mouse_filter = 1` (PASS) on interactive elements prevents touch/click interception issues.
- **`_on_container_storage_visibility_changed`** guard (`if not is_node_ready()`) prevents crashes during scene tree setup.

### 💡 Minor observations (non-blocking)
- In `_apply_layout`, `find_children("*", "PanelContainer", true, false)` + script check iterates the whole tree on every resize. Shouldn't be a perf concern for settings UI, but caching the list once could be a future optimization if needed.
- The `panel_closed` signal is declared and emitted in `hide_panel()`, but I don't see it connected anywhere in `explorer.gd` (the close is handled via `_on_settings_panel_closed` which calls `settings_panel.hide()` directly). Not a bug, but the signal is currently unused — may be useful for future consumers.
- The `SettingsPanel` in `explorer.tscn` doesn't start with `visible = false`, though `_close_all_panels()` is called during init flow, so it should be fine in practice.

LGTM 👍

Requested by Seba via Slack

rodrigo685 (approved):

PR #1849, opened 2026-04-16
feat: minimal support for spring bones (ADR-316)

### LINKED ISSUE #1789: Spring Bones Support: Verify Avatar Assembly Handles Extra Bones
## Context

The Spring Bones (jiggle bones) project is progressing and aims to be released by the end of this cycle. This means some wearables will include extra bones for dynamic movement/physics simulation.

**Technical Specification**: [ADR-316: Wearable Spring Bones](https://github.com/decentraland/adr/blob/feat/adr-316/content/ADR-316-wearable-spring-bones.md)  
**Notion doc**: https://www.notion.so/decentraland/Jiggle-Bones-2c25f41146a5800b8db1d653f0566c21

Spring bones are extra bones (not part of the base avatar armature) whose transforms are driven by physics simulation. They enable hair, earrings, capes, belts, and similar wearable elements to move dynamically in response to avatar locomotion and gravity.

## Requirements

While implementing spring bone physics is not immediately required, we need to verify that:

1. **Avatar assembly code** can handle wearables with extra bones without breaking
2. **Skinning system** correctly processes the additional bone data
3. No crashes or visual artifacts occur when equipping items with spring bones

## Test Resources

A collection of wearables with spring bones is available for testing in .zone:

- **Curator collection editor**: https://decentraland.zone/builder/item-editor?collection=357b5d06-0ffa-40ae-b399-e8d2e918a257&reviewing=true
- **Marketplace**: https://decentraland.zone/marketplace/collections/0x81004ea82f4af8337e357bef49cc746fce881dee

## Tasks

- [ ] Test equipping wearables from the spring bones collection
- [ ] Verify no crashes or errors in avatar assembly
- [ ] Verify skinning renders correctly (even without physics simulation)
- [ ] Document any issues found
- [ ] (Future) Implement spring bone physics simulation if desired

## Contact

Ping @RocioCorralMena (Rocío Corral Mena) with any questions or issues about the test collection.

---
Requested by Gon via Slack

leanmendoza:
Fixes #1789

Adds spring bone support at the avatar assembly step: extra bones from wearables are copied into the base skeleton and mesh skins are rebound by name. No physics simulation — bones render in rest pose.

Also includes dev-only deep link params used to test unowned wearables:
- `dclenv=zone` routes the catalyst to `peer.decentraland.zone`
- `fake-owned-wearables=urn1,urn2` injects URNs into the backpack
- `disable-profile-deploy=true` keeps profile updates local (auto-enabled when `fake-owned-wearables` is set, otherwise the deploy would fail)


regenesis-claw (commented):
Nice first-pass for ADR-316 — the merge-by-depth algorithm and rebind-by-name approach are sound and the Rust tests for the new deep link params are solid. A few observations below; nothing that I think has to block merging for the "minimal / no physics" scope, but the bone-accumulation point is worth tracking.

## Findings

### 1. Bone accumulation across wearable / body-shape changes (medium)
`body_shape_skeleton_3d` is the scene-bound `$Armature/Skeleton3D`. `async_try_to_set_body_shape` only removes mesh children — it never calls `clear_bones()` / rebuilds the skeleton. That means bones merged by `_merge_extra_wearable_bones_into_base` persist forever after de-equip or body-shape change. Consequences:
- Memory growth over many outfit changes.
- On a subsequent load, `body_shape_skeleton_3d.find_bone(bone_name) != -1` will hit a stale bone, so a new wearable whose spring bone shares a name silently binds to the previous wearable's bone (potentially wrong parent / rest pose).
- Two different wearables that happen to use the same spring-bone name (e.g. `hair_tip_1`) will collide in unpredictable ways.

Not critical for "verify avatar assembly handles extra bones" (#1789), but worth either fixing now (reset skeleton bones when the body shape is (re)assembled) or filing a follow-up.

### 2. Doc comment references a non-existent constant (minor)
`lib/src/profile/profile_service.rs` `set_deploy_disabled` doc says:
> Used together with `FORCE_FAKE_OWNED_WEARABLES` to test unowned wearables

There's no `FORCE_FAKE_OWNED_WEARABLES` — the actual param is the `fake-owned-wearables` deep link key. Please update the comment so future readers can grep it.

### 3. Dev-only deep link params are live in release builds (minor)
`disable-profile-deploy` and `fake-owned-wearables` are parsed unconditionally. A crafted deep link could desync a user's local profile from the catalyst. Self-griefing only, but consider gating behind `#[cfg(debug_assertions)]` or an internal-build flag — or at minimum log a very loud warning when either is active in a release build.

### 4. Accidental whitespace change (nit)
`godot/src/global.gd` removes the blank line between `profile_changed.connect(...)` and `self.testing_tools = TestingTools.new()` — looks unintentional.

### 5. Skin duplication per mesh (nit / question)
`_rebind_skin_by_name` duplicates the Skin on every `MeshInstance3D`. Correct, but if a wearable has multiple mesh instances sharing one Skin resource, each ends up with its own copy. Caching per-skin isn't required, just flagging.

## Positives
- Topological ordering by depth before adding bones is the right call and correctly guarantees parents exist when children are inserted.
- `Skin.set_bind_name` is the idiomatic Godot way to resurvive a reparent — good choice over remapping indices.
- `mesh.skin.duplicate()` before mutating — good, avoids clobbering the shared resource.
- The auto-implication `fake-owned-wearables` ⇒ `disable-profile-deploy` is a nice safety net; covered by a dedicated test. 👍
- Zone → `peer.decentraland.zone` cleanup in `urls/mod.rs` is a welcome side-fix.

Overall: ship-able with the doc fix; would like a follow-up issue for the bone-accumulation / skeleton-reset concern.

regenesis-claw (approved):
## Re-review on 17bc6fc (delta from 5c0e1e0)

Thanks for the follow-up — three of the four items are addressed; the remaining one was flagged as minor.

### Status of previous findings

| # | Item | Status |
|---|------|--------|
| 1 | Bone accumulation | ✅ Fixed via slot-recycle pool |
| 2 | `FORCE_FAKE_OWNED_WEARABLES` doc typo | ✅ Now references `fake-owned-wearables` deep link param |
| 3 | Dev flags unconditional in release builds | ⚠️ Not addressed (was flagged minor — fine to defer) |
| 4 | Accidental blank-line removal in `global.gd` | ✅ Restored |

### On the bone-recycle mechanism

Nice workaround for the Godot 4.6 `Skeleton3D` API missing `remove_bone()`. The design is sound:

- `_active_extra_bone_indices` tracks slots in use for the current assembly.
- `_recycle_extra_wearable_bones` renames each active slot to a unique `__stale_bone_N`, disables it, detaches (parent=-1), resets rest to identity, and pushes the index into `_free_bone_pool`.
- `_merge_extra_wearable_bones_into_base` pops from the free pool first before growing the skeleton, so `bone_count` is bounded by the max number of extra bones ever needed concurrently.
- The depth-sorted merge still works with recycled slots because `find_bone(parent_name)` sees the *renamed* slots: stale sentinels don't collide with real parent names, and just-reused parents are renamed before their children look them up. 👍

1. *Early returns in `async_try_to_set_body_shape` skip recycling.* If body-shape load fails (`body_shape == null` / `new_skeleton == null`), `_recycle_extra_wearable_bones()` isn't called. In practice the next successful load recycles, but worth a note if you ever see lingering stale-but-not-recycled bones after a failed load.
2. *`_stale_bone_counter` is monotonic.* Fine in practice (u64-ish ceiling), just flagging.
3. *Commit messages `fix` / `fix`.* Consider a squash-merge or amended message so the bone-recycle design is discoverable in `git log`.
4. Item #3 from the original review (gating `disable-profile-deploy` / `fake-owned-wearables` behind `#[cfg(debug_assertions)]` or a release-build guard) is still open — only self-griefing, so punting is reasonable.

### CI

At review time: `📊 Static checks` ✅; Linux / Clippy / Android builds still pending on `17bc6fc`. Please confirm Validate GDScript + Rust checks are green before merging.

Approving — nice fix.

PR #1850, opened 2026-04-16
fix: one-way colliders for plane meshes in GLTF scenes (#1529)

### LINKED ISSUE #1529: [Bug]: One-way colliders act as two-way colliders (SPIKE)
### 🕹️Platform

Android

### 🔢App Version

v242

### 📱Device Information

PIXEL 6A

### 📄Issue description

One-way colliders (should allow passage only in one direction) currently behave as two-way colliders, meaning you can move through them from both sides.

### ✅Expected behavior

One-way colliders should permit passage only in the intended direction and block movement from the other side.

Test Scene: https://mobile.dclexplorer.com/open?position=-68%2C-65

### 🖼️Screenshots / Media

Desktop

https://github.com/user-attachments/assets/b88e2fc7-74ff-426d-801e-1334e141271b

Mobile

https://github.com/user-attachments/assets/baabb2e3-be8a-43cc-bc23-bcb937bd2d03

### ▶️Steps to reproduce

- Place a one-way collider in your scene
- Attempt to move through the collider from both directions
- Observe the collider allows passage from both sides

### 👁️Occurrence

100% of the time

manuelmaceira:
## Summary

Closes #1529.

Plane meshes (e.g. \`barrier._collider\`) were behaving as two-way colliders in Godot because backface collision was unconditionally enabled on all \`ConcavePolygonShape3D\` shapes created from GLTF meshes (\`scene.rs:128-146\`). This broke one-way doors and similar gameplay.

## Why Godot differs from Unity

Unity's PhysX \`MeshCollider\` is **always double-sided** for physics, so volumetric meshes (streets, ramps, buildings) work correctly without any extra setup. Godot's \`ConcavePolygonShape3D\` with \`backface_collision_enabled = false\` is genuinely single-sided for all physics interactions, including character movement — pisar un triángulo cuyo normal apunta hacia abajo te hace caer a través.

We can't simply remove backface collision (breaks ramps/streets) or always enable it (breaks one-way colliders). We need to discriminate.

## Fix

Heuristic based on the mesh AABB:

- **Planar mesh** (any axis < 1cm) → backface OFF → one-way collider
- **Volumetric mesh** (3 axes with thickness) → backface ON → double-sided like PhysX

This matches Unity behavior across all collider types:

| Case | Unity | Godot (with fix) |
|------|-------|-----------------|
| GLB plane \`_collider\` (1 face) | One-way | One-way |
| GLB plane \`_collider\` (2 opposite faces) | Double | Double |
| GLB thin box \`_collider\` | Double | Double |
| \`MeshCollider.setPlane()\` (SDK) | Double | Double |
| \`MeshCollider.setBox()\` (SDK) | Double | Double |

Creators that want a double-sided thin wall can model the plane with two opposite faces, extrude with thickness, or use \`MeshCollider.setBox()\` — same workflow as Unity.

## Test plan

- [x] Test scene with 5 cases (SDK plane / SDK plane rotated / GLB plane \`_collider\` with visible mesh / GLB bare plane \`_collider\` / SDK box) — all behave as expected
- [x] Goerli Plaza ramp at 78,-7 — \`Street_collider\` (volumetric, 14% down-facing triangles) keeps blocking correctly after the fix
- [x] \`cargo fmt\`, \`cargo clippy -D warnings\`, full pre-commit suite passes


lib/src/content/gltf/scene.rs
```diff
@@ -153,3 +158,20 @@ fn create_scene_colliders(node_to_inspect: Gd<Node>, root_node: Gd<Node3D>) {
...
+/// Check if a mesh is essentially planar (very thin in at least one axis).
+/// Planar meshes are used as one-way colliders and should NOT have backface collision.
+fn is_mesh_planar(mesh_instance: &Gd<MeshInstance3D>) -> bool {
+    if let Some(mesh) = mesh_instance.get_mesh() {
+        let aabb = mesh.get_aabb();
+        let size = aabb.size;
+        size.x < PLANAR_THICKNESS_THRESHOLD
+            || size.y < PLANAR_THICKNESS_THRESHOLD
+            || size.z < PLANAR_THICKNESS_THRESHOLD
+    } else {
```
regenesis-claw, line 174: Minor: the threshold is compared against the *mesh-local* AABB, so a MeshInstance3D (or ancestor) with a non-unit scale can skew the classification. A mesh that's 2 cm thick in geometry but placed under a `scale=0.1` node is effectively 2 mm in world space (planar) but would be flagged volumetric here; conversely a 5 mm geometry scaled 10× would be flagged planar but is really 5 cm thick in world space.

Not blocking — GLB collider meshes authored in Blender are almost always unit-scale, and the Goerli Plaza / SDK test cases cover the common shapes. Worth a comment on the function noting the assumption, so someone hitting a false negative later has a breadcrumb.

lib/src/content/gltf/scene.rs
```diff
@@ -153,3 +158,20 @@ fn create_scene_colliders(node_to_inspect: Gd<Node>, root_node: Gd<Node3D>) {
...
+        false
+    }
```
regenesis-claw, line 176: Nit: when `get_mesh()` is `None`, this returns `false` (→ volumetric → backface ON). In practice `create_trimesh_collision()` above wouldn't have produced a `ConcavePolygonShape3D` either, so the outer loop finds nothing and the fallback is unreachable. Still, a one-line comment like `// No mesh → no ConcavePolygonShape3D will exist, value is unused` would save the next reader a minute.

regenesis-claw (approved):
LGTM — heuristic is sound and the Unity-parity rationale is well explained. Single-file change, logic is clean, and the matrix in the PR description matches what I'd expect for volumetric vs planar meshes.

A couple of non-blocking notes (inline).

*Verified:*
• `MeshCollider.setPlane()` goes through `create_or_update_mesh` with a `BoxShape3D` of size `(1,1,0)` — separate code path, not affected by this change, and remains double-sided as claimed.
• `ConcavePolygonShape3D.backface_collision_enabled` defaults to `false` in Godot, so skipping the setter for planar meshes correctly leaves them single-sided.
• `mesh.get_aabb()` returns the mesh's local AABB, which matches the collider geometry produced by `create_trimesh_collision()` (same source mesh). Good choice over `VisualInstance3D::get_aabb()`.

CI: static checks + clippy green, Android/Linux still in progress at review time — assuming they land green.

kuruk-mm (approved):

lib/src/content/gltf/scene.rs
manuelmaceira: I think having the heuristic as the unscaled model aligns better with artist intent. I dont think we want to flag scaled objects as one way colliders

PR #1854, opened 2026-04-16
fix: avatar Y-axis sway when orbiting camera

kuruk-mm:
## Summary
- Avatar was a child of Player, so Player's Y rotation (driven by camera yaw) propagated into the avatar. The three input handlers tried to cancel this with a counter-rotation (`_player.avatar.rotate_y(+Δ)` against `_player.rotate_y(-Δ)`), but the cancellation was fragile — floating-point drift from repeated basis multiplications and conflicts with `avatar.look_at` in `_physics_process` caused a long-standing subtle sway on the avatar's Y axis while the camera orbited.
- Fix decouples the avatar from Player's transform entirely by making it `top_level = true`, manually syncing its world position each physics frame, and updating all places that composed avatar world rotation as `player.rot + avatar.rot` to operate on world-space avatar rotation directly.

## Changes
- `player.tscn`: `top_level = true` on Avatar node.
- `player.gd`:
  - Sync `avatar.global_position = global_position` after `move_and_slide()`.
  - `avatar_look_at()` sets avatar's world Y to `y_rot + PI` (not zero).
  - `set_avatar_rotation_independent()` writes world Y directly (no `- rotation.y`).
  - `get_broadcast_rotation_y()` in third-person returns `avatar.rotation.y` (no sum).
  - Third-person switch aligns avatar world Y to player body.
- `player_desktop_input.gd`, `player_mobile_input.gd`, `player_gamepad_input.gd`: removed the `_player.avatar.rotate_y(...)` counter-rotations.

## Test plan
- [ ] Orbit camera while standing still → avatar stays perfectly frozen on its Y axis (no sway).
- [ ] Orbit camera while walking/jogging/running → avatar smoothly re-orients to face movement direction.
- [ ] First-person ↔ third-person toggle behaves correctly; avatar faces player body after switch.
- [ ] Scene-driven RPCs: `avatar_look_at`, `camera_look_at`, `avatar_look_at_independent` still point avatar/camera at the right world target.
- [ ] Broadcast rotation to remote peers looks correct (avatar facing matches what the local player sees).
- [ ] Teleport still lands avatar at player position (position sync runs each physics frame).
- [ ] Desktop mouse, mobile touch, and gamepad right-stick camera control all feel normal.
- [ ] iOS build (triggered via `build-ios` label).

### DIFF
--- a/godot/src/logic/player/player.gd
+++ b/godot/src/logic/player/player.gd
@@ -152,6 +152,10 @@ func _ready():
 	Global.scene_runner.locomotion_settings_changed.connect(_on_locomotion_settings_changed)
 	_on_scene_changed(Global.scene_runner.get_current_parcel_scene_id())
 
+	# Avatar is top-level: initialize its world transform to match the player
+	avatar.global_position = global_position
+	avatar.rotation = Vector3(0, rotation.y, 0)
+
 
 func _on_player_profile_changed(new_profile: DclUserProfile):
 	var new_version = new_profile.get_profile_version()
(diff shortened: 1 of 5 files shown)


manuelmaceira (approved):
LGTM

PR #1870, opened 2026-04-20
feat: apply the figma style to the warning: skybox configured by the scene creator

sdilauro:
Apply the Figma style to the warning: Skybox configured by the scene creator.


ElzathRusso (approved):
great work!

PR #1872, opened 2026-04-20
fix: blocked users list not persisting across restarts

### LINKED ISSUE #1809: [Bug] Blocked users get unblocked after app restart
### Platform
Mobile (iOS and Android)

### App Version
Not specified

### Device Information
Motorola Edge 60 Fusion, Android 16 (also confirmed on iOS devices)

### Issue Description
When a user blocks another user, the block is applied successfully during the current session. However, blocked users are automatically unblocked every time the app is restarted. The block list does not persist between app sessions.

### Expected Behavior
Blocked users should remain blocked across app restarts. The block list should be persisted locally and/or synchronized with the user's profile, so that blocks are maintained permanently until the user manually unblocks them.

### Steps to Reproduce
1. Open the mobile app and log in
2. Block a user through the social/user interface
3. Verify the user is blocked (they should no longer appear in chat, etc.)
4. Fully close and restart the app
5. Check the previously blocked user
6. Observe that the user is no longer blocked

### Occurrence
100% - happens every app restart

---

sdilauro:
## Summary
- **Root cause**: `_on_profile_changed` called `init_from_profile()` which overwrote the Social Service's authoritative blocked list with stale profile data
- Removed `init_from_profile()` for blocked users; muted users still load from profile via `append_muted`
- Validate `BlockUserResponse`/`UnblockUserResponse` oneof response instead of silently ignoring server errors
- Show error toast when block/unblock fails
- Fetch blocked profiles in parallel and defer `remove_items()` until all data is ready to prevent partial list rendering from race conditions

## Test plan
- [ ] Block a user, restart the app, verify the user is still blocked
- [ ] Open blocked list tab, verify all blocked users appear
- [ ] Block/unblock from profile panel and social panel, verify toast appears on failure
- [ ] Mute a user, restart, verify mute persists
- [ ] Test on a second device, verify blocked list replicates

Closes #1809

### DIFF
--- a/godot/src/logic/player/player_identity.gd
+++ b/godot/src/logic/player/player_identity.gd
@@ -116,9 +116,11 @@ func _on_profile_changed(new_profile: DclUserProfile):
 	_current_profile = new_profile.duplicated()
 	_mutable_avatar = _mutable_profile.get_avatar()
 
-	# Note: blocked/muted lists are managed by the Social Service (get_blocking_status),
-	# not by the profile. Do NOT call init_from_profile here as it would overwrite
-	# the server's authoritative blocked list with stale profile data.
+	# Blocked users are managed by the Social Service (get_blocking_status), not the profile.
+	# Only load muted users from profile (no Social Service RPC for muted).
+	var muted_list = new_profile.get_muted()
+	if muted_list.size() > 0:
+		Global.social_blacklist.append_muted(muted_list)
 
 
 func get_mutable_avatar() -> DclAvatarWireFormat:
--- a/godot/src/logic/player/player_identity.gd
+++ b/godot/src/logic/player/player_identity.gd
@@ -116,8 +116,9 @@ func _on_profile_changed(new_profile: DclUserProfile):
 	_current_profile = new_profile.duplicated()
 	_mutable_avatar = _mutable_profile.get_avatar()
 
-	# Update social blacklist from the profile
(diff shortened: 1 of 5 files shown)


regenesis-claw (approved):
## Review: Approved ✅

Clean fix with a well-identified root cause. The changes are coherent and well-structured:

**Root cause fix** — Removing `init_from_profile()` for blocked users makes sense: the Social Service is the source of truth, and the profile data was stale by definition. Good that muted users still load from profile since there's no Social Service RPC for those.

**Response validation (Rust)** — Properly matching on the `BlockUserResponse`/`UnblockUserResponse` oneof variants instead of silently discarding server errors is a solid improvement. The error messages are descriptive and cover all variants.

**Parallel fetch + deferred `remove_items()`** — Starting all profile fetches before awaiting any of them is the right call — prevents the sequential waterfall. Moving `remove_items()` after all data is collected eliminates the partial-render flash. One minor note: the per-fetch `request_id` check from the old code is gone, so if a newer request fires mid-fetch, we'll still await all in-flight promises before discarding. This is fine in practice since the promises are already running and can't be cancelled anyway, but worth being aware of.

**Error toasts** — Good UX touch surfacing block/unblock failures to the user instead of only logging to stderr.

LGTM 👍

Requested by Seba via Slack

kuruk-mm (approved):
Greatt!!

PR #1873, opened 2026-04-21
feat: add reload scene button in HUD for preview mode

### LINKED ISSUE #1524: Preview Mode - Add reload button in HUD
Add a reload button to the HUD in Preview Mode (for Creator Hub).

When debugging or developing in Preview Mode, a reload button would assist developers/creators by quickly reloading the current scene without having to exit/re-enter. 

- This would streamline workflow and allow faster iteration during development
- Button should only be visible in debug mode

Reload button example in HUD

Acceptance Criteria:
- [ ] Reload button is present in HUD when in debug mode (Preview Mode)
- [ ] Clicking reload button reloads the current scene
- [ ] Button is not visible in normal mode

sdilauro:
## Summary

- Adds a **Reload Scene** button to the HUD, visible only in preview mode (CLI `--preview` or `?preview=` deeplink)
- Both the reload button and the auto-enabled debug panel hide when navigating away from the preview realm (e.g. via Discover) and reappear when returning
- The debug panel opened explicitly via `--debug-panel` is unaffected

Closes #1524


regenesis-claw (approved):
Code looks good — clean implementation. A few notes from the review:

**Logic ✅**
- `_debug_panel_from_preview` correctly distinguishes explicit `--debug-panel` from auto-enabled preview debug — so the explicit one stays visible regardless of realm changes.
- `_is_in_preview_realm()` properly handles both deeplink (`preview` URL match) and CLI (`preview_mode` flag) paths.
- `_update_preview_ui()` toggles the parent `Panel_ReloadScene` visibility, keeping the button and its container in sync.
- Reload button starts hidden in the scene and only appears on `_on_change_scene_id`, which makes sense — can't reload before a scene loads.
- Null check in `_on_button_reload_scene_pressed()` is a good guard.

**Scene file (.tscn)**
- The `unique_id` additions throughout are Godot editor noise — harmless.
- `SafeMarginContainerDebug` margin_right changed from 0 → 92, presumably to avoid overlap with the reload button area.

No issues found. 👍

Requested by Seba via Slack

leanmendoza (commented):
LGTM, just my two cents: if decoupling the reload button from debug panel is not a strong requirement, I'd simplify the PR by always showing the button when debug panel is on. This makes sense since the reload button is a debug tool

leanmendoza (changes requested):
Please revert the unique_id changes since it's going to be a conflict later

sdilauro (commented):
Done

kuruk-mm (approved):

PR #1887, opened 2026-04-23
feat: implement version update screen 

### LINKED ISSUE #1702: Version Update Screen - Implementation
### 🛍️ Storefront Version Update Screen

**Description**  
Add a non-blocking nudge Screen in the Entry Flow to inform users when a new app version is available and encourage them to update.

🎨 [Design](https://www.figma.com/design/H1Alvbe7sh0e0IyLckewMI/Mobile-Entry-Flow?node-id=771-8488&t=gIZvhxsZ85zFiYQD-1)

**Goals**
- Increase update adoption rate  
- Ensure users are aware of new features, fixes, and improvements  

**Scope**
- Detect when a newer app version is available 
- Display a nudge Screen that redirect to appstore/platystore

**Acceptance Criteria**
- Screen is shown only when a newer version is available to users that don't have the latest version of the app
- Needs to check latest version against an endpoint (where we set latest)
- Users can dismiss the  Screen and continue using the app  
- CTA correctly redirects to the appropriate store (deeplink)
- Screen does not appear again within the same session (or respects cooldown rules)
- Screen only triggers ONCE every 3 App Launches

Metrics:
-`UPGRADE_SCREEN_SHOWN`: Screen_Show_Event
-`UPGRADE_SCREEN_UPDATE`: Click Event
-`UPGRADE_SCREEN_NOTNOW`: Click Event

**QA / Validation**
- Verify screen appears on outdated versions only  
- Test update redirection on iOS and Android  
- Validate dismissal and reappearance logic  
- Ensure no impact on Storefront performance or navigation

Image

**Title**: Update available
**Copy**: A new version of Decentraland is available. Update now for the latest fixes and improvements.
Buttons: UPDATE / LATER

leanmendoza:
Fixes #1702

Shows a full-screen update prompt on mobile when the app-versions BFF reports the installed build is below the configured minimum (hard block) or recommended (soft, dismissible) version.

- Boot-time, blocking check during the lobby loading screen with a 3s timeout; times out / errors silently proceed (no retries).
- Version encoded as `major*100000 + minor*100 + patch` for monotonic comparison against the BFF values.
- "Later" snoozes the soft prompt for 24h; the hard block ignores the snooze.
- Overlay lives on its own CanvasLayer (99), one notch below the modal manager (100).


godot/src/ui/components/auth/lobby.gd
```diff
@@ -734,3 +740,14 @@ func _on_ftue_jump_in(parcel_position: Vector2i, realm_str: String) -> void:
 
 func _on_ftue_jump_in_world(realm_str: String) -> void:
 	Global.async_join_world(realm_str)
+
+
+func _async_run_version_gate() -> String:
+	var gate: Node = preload("res://src/version_gate.gd").new()
+	add_child(gate)
+	var result: String = await gate.async_check()
+	if result == "hard":
```
regenesis-claw, line 749: Minor: the `gate` Node stays in the tree after `async_check()` finishes (for `proceed`/`soft` paths). Since the overlay is parented to `get_tree().root`, not to `gate`, you could safely `gate.queue_free()` here for all three paths. Not urgent — it's a lightweight Node with no `_process`.

regenesis-claw (approved):
Clean, well-scoped feature — version gate logic is sound. A couple of minor observations:

### ✅ What looks good
- Version encoding (`major*100000 + minor*100 + patch`) is monotonic and handles pre-release suffixes correctly via the `split("-")` strip.
- Fail-open design: timeout / HTTP error / malformed JSON all resolve to `RESULT_PROCEED` — no risk of bricking the app if BFF is down.
- Snooze persistence round-trips correctly through `config_data` load/save.
- `CanvasLayer(99)` sits correctly below the modal manager.
- `async_race` + `PromiseError` guard is the right pattern here.

### 💬 Minor observations (non-blocking)

1. **Gate `Node` leak** — In `_async_run_version_gate()`, the gate node is added as a child of the lobby but never freed after `async_check()` returns. For `"proceed"` and `"soft"` paths, it stays in the tree doing nothing. Trivial to fix with `gate.queue_free()` before the return (except on `"hard"` where the overlay's parent reference matters — though the overlay is actually parented to root, not the gate, so freeing should be safe in all cases).

2. **Hard-block vs lobby init** — When `gate_decision == "hard"`, `_ready()` returns early, skipping analytics tracking, FTUE setup, and the rest of lobby initialization. That's probably intentional (the user is stuck on the update screen), but worth confirming there are no timers/signals from earlier in `_ready()` that expect the skipped code to have run.

Neither blocks approval — the feature works correctly and fails gracefully. 👍

Requested by Lean via Slack

kuruk-mm (commented):
code looks great!

rodrigo685 (approved):

PR #1893, opened 2026-04-23
fix: show loading screen immediately on jump in, defer ban check to post-loading

### LINKED ISSUE #1892: [Bug] Delay between tapping Jump In and loading screen
### Platform
Mobile (Android/iOS)

### App Version
Current development version

### Device Information
Not specified

### Issue Description
There is a noticeable 1-2 second delay after tapping the "Jump In" button in the Discover screen before the loading screen appears and the orientation changes from portrait to landscape. The dialog closes but the user sees a brief wait before anything happens.

### Expected Behavior
The transition should be immediate (or at least feel immediate with loading feedback), as it was previously. The flow should be: tap Jump In → instant loading screen → orientation change.

### Root Cause
The delay appears to be caused by the ban verification check happening before the loading screen is shown. This pre-loading check is introducing latency that affects the perceived responsiveness of the UI.

### Discussion Context
From team discussion:
- **Rodri's suggestion**: If the ban-checker is taking 2-3 seconds, consider removing it from the pre-loading phase and handle ban verification in-world instead. The priority should be on the experience of well-behaved users, not optimizing for banned users.

### Occurrence
100% - occurs on every Jump In interaction

Requested by Seba via Slack

sdilauro:
Fixes #1892

## Problem

When clicking **Jump In** (from Discover or in-game modals), the dialog closed and there was a noticeable ~2 second delay before the loading screen appeared. During this gap the user saw either the previous scene or a blank screen with no feedback.

The root cause was the **ban check HTTP request** (gatekeeper call with signed auth chain) running **before** the loading screen was shown and **before** navigation started. This blocking request added 1–2 seconds of dead time on every scene transition.

## Solution

Move the ban check from a **blocking pre-navigation gate** to a **reactive post-loading check**:

1. **Loading screen shows immediately** — `enable_loading_screen()` is called before the orientation change and before any navigation begins, so the user gets instant visual feedback on Jump In.

2. **Ban check runs after loading completes** — `explorer._on_loading_finished()` triggers `_async_run_ban_check()`, which resolves the current scene ID and checks gatekeeper access. If the user is banned, the modal appears immediately after loading.

3. **Discover back button intercept** — If `ban_pre_check_active` is set (user was banned), pressing back in Discover re-shows the ban modal instead of returning to the banned scene.

4. **Generation counter prevents stale results** — Each ban check captures a `_ban_check_generation` value. If a new loading starts or a new check is triggered, stale results are discarded, preventing race conditions from showing incorrect ban modals.

## Test plan

- [ ] Jump In from Discover to a non-banned scene → loading screen appears instantly, scene loads normally
- [ ] Jump In from Discover to a banned scene → loading screen appears instantly, ban modal shows after loading
- [ ] Ban modal → Discover → Jump In to non-banned scene → loads normally, no stale ban modal
- [ ] Ban modal → Discover → Jump In to banned scene → loading, ban modal appears again
- [ ] Ban modal → Discover → press Back → ban modal re-appears (does not return to scene)
- [ ] Deep link to a banned scene → loading, ban modal shows after loading
- [ ] In-game `/goto` to banned scene → loading, ban modal shows


kuruk-mm (approved):
if you tested, code looks great

regenesis-claw (approved):
## ✅ Approved

Clean and well-structured refactor that eliminates the loading delay on Jump In by deferring the ban check to post-loading.

### What I like

- **Generation counter** — solid pattern for preventing stale ban check results from race conditions (e.g., user navigates to scene A, then quickly to scene B before A's check resolves). Both the loading start and the check itself increment the counter, and results are discarded if the generation doesn't match.
- **Minimal surface area** — the ban check logic is centralized in `_async_run_ban_check()` in explorer.gd rather than scattered across navigation paths.
- **Simplified API** — removing `scene_id` and `skip_ban_check` parameters from `async_teleport_to`/`async_join_world` makes the callsites much cleaner.
- **Back button intercept** in Discover correctly re-shows the ban modal instead of returning to the banned scene.

### Minor observation

In `async_join_world`, the chat message and `LOADING_START` metrics tracking moved inside the `if is_instance_valid(explorer)` branch. This means the initial app launch path (else branch, no explorer yet) no longer tracks those. Probably fine since there's no chat to show and the first load has its own tracking, but worth confirming this doesn't affect analytics.

LGTM — nice UX improvement. 🚀

Requested by Seba via Slack

PR #1901, opened 2026-04-27
fix: clear avatar entity components when a remote player disconnects

leanmendoza:
Fix #1818

When a remote player disconnects, `remove_avatar` only killed the entity in the scene CRDT. `engine.rs` skips component updates for died entities, so the JS scene kept stale `PlayerIdentityData` / `AvatarBase` / `AvatarEquippedData` and `onLeaveScene` / `onPlayerDisconnected` never fired.

Now every component for that entity is set to `None` with a bumped timestamp and marked dirty, so `DELETE_COMPONENT` propagates to the JS runtime and the player tracking events fire.


leanmendoza:
Already addressed by Eibriel

PR #1916, opened 2026-04-28
feat: friends online carousel in discover

### LINKED ISSUE #1550: Discover - Add Friends Status/Join
### DISCOVER

[Atlas](https://www.figma.com/design/Tz8ekjVNmv2OpkmfZb8KwU/Discover?node-id=1653-32260&t=bIJFPP1dlTmfAP4b-1) 

**Online Friends**
In the Discover section, add a top carousel showing **online friends**  to jump into the place/world they’re currently in.

https://github.com/user-attachments/assets/ac0344ef-5bec-4c18-9e4c-073547735d1b

**CAROUSEL:**
Image

Carousel Title: Online Friends
Next to the title: Number of friends (5)
Horizontal scroll

--

**FRIEND:**

Image

Profile picture
Name Friend
Name Place/World (2 lines, elipsis)

--

**PLACE/WORLD PANEL**
On tap open panel with the info of the place:
Image

PFP and Name friend
Name Place/World
Creator Name
Image Place/World
JUMP IN button

---
Corner cases:
- If there are no online friends?
It follows the logic of the other carousels; if there's no content, it's hidden corousel.

- If I only have 1 friend, is the rest of the row empty?
Yes.

- If the friend system fails, do we just hide it or does it have an error state?
If friedn system fails, hide carousel.

- IOS, blocked scenes.
For now, we show on iOS even if the scene is blocked.

sdilauro:
Closes #1550 

## Summary

Adds a "Friends Online" carousel to the Discover screen showing friends currently in Genesis City with their profile picture, name, and the experience they're visiting. Tapping a friend card opens a detail panel with Jump In.

- New `friends_generator.gd` fetches friends list, crosses it with Archipelago peers to find who's in Genesis City, resolves place names, and renders cards
- Diff-based card updates: only adds/removes cards that changed, doesn't rebuild the whole list
- Reactive refresh via `friend_connectivity_updated` signal (debounced at 1s to collapse bursts) + 15s polling timer as fallback
- New `ProfilePicture` sizes (`DISCOVER`, `DISCOVER_HEADER`) for the card and detail panel
- Friend detail panel reuses `place_item.gd` with `allow_full_drag = false` to prevent full-screen expansion
- `is_instance_valid` guards after every `await` to handle the user navigating away mid-load
- Reverted an accidental `quick_blur.tres` float precision artifact

## Test plan

- [ ] Log in with an account that has friends online in Genesis City
- [ ] Open Discover — "Friends" carousel should appear at the top with online friends, showing profile pic + name + place name
- [ ] Tap a friend card — detail panel opens with friend info and Jump In button
- [ ] Tap Jump In — teleports to the friend's location
- [ ] Have a friend go online/offline while Discover is open — carousel should update within a few seconds
- [ ] Log in as guest — Friends carousel should not appear
- [ ] Navigate away from Discover while it's loading — no crashes


regenesis-claw (approved):
Solid work! The architecture is well thought out — diff-based card sync, debounced reactive updates, LRU place cache, concurrent HTTP throttling, and `is_instance_valid` guards after every `await`. A few observations (none blocking):

1. **Rust idempotency guard** (`dcl_social_service.rs`): the `try_read()` on the manager RwLock silently falls through on contention, which would re-initialize. Extremely unlikely in practice since this is called from the main thread, but worth noting.

2. **Streaming world fetch** (`_async_fetch_connected_worlds_streaming`): unlike the batch version, this path has no overall timeout — it relies on Godot's default HTTP timeout per request. Fine for now, but if you ever see hangs on this path, that's where to look.

3. **`_update_title` count label**: dynamically creates and injects a Label into the carousel's HBox. Slightly fragile if the carousel layout changes, but pragmatic.

4. **`locations.gd` fix**: emitting `in_genesis_city_changed` on error paths is a nice catch — prevents the `await` in `friends_generator` from hanging forever.

Overall the code is clean, handles edge cases well, and the signal flow (connectivity updates → debounce → diff sync) is robust. 👍

Requested by Seba via Slack

ElzathRusso (approved):

kuruk-mm (approved):

PR #1924, opened 2026-04-29
fix(ios): restore Universal Link & custom-scheme deeplink delivery on Godot 4.6

### LINKED ISSUE #1919: [Bug] Deeplink not triggering scene change if app is already opened
### Platform
iOS

### App Version
v 422 (release candidate)

### Device Information
iPhone 16

### Issue Description
When the app is already running, tapping a deeplink does not trigger the expected scene change. The deeplink appears to be ignored when the app is in the foreground.

### Expected Behavior
Deeplinks should trigger scene navigation regardless of whether the app is freshly launched or already running in the foreground.

### Screenshots / Media
Video demonstration attached (to be uploaded manually)
||

https://github.com/user-attachments/assets/d497bdf6-0778-4d14-9752-f7dc387eee60

### Steps to Reproduce
1. Open the Decentraland mobile app
2. Navigate to any scene
3. Tap a deeplink that should navigate to a different scene
4. Observe that the scene change does not occur

### Occurrence
Not specified

---

**Related Issues:**
- #1792 - Feature: Show scene highlight modal when entering via deeplink from ads/posts
- #1776 - Support Event deeplinks to open Discover and show event details
- #1895 - [Bug] Direct world URLs in chat open browser instead of triggering world/scene travel via deeplink

Requested by Rodri via Slack

kuruk-mm:
## Summary
Closes #1919.

Restores both **custom URL scheme** (`decentraland://…`) and **HTTPS Universal Link** (`https://mobile.dclexplorer.com/…`) deeplink delivery on iOS, broken since the 4.5.1 → 4.6.2 Godot upgrade.

The engine-side fix is on `decentraland/godotengine` `4.6.2` via [PR #12](https://github.com/decentraland/godotengine/pull/12) → [`7e5bfa4282`](https://github.com/decentraland/godotengine/commit/7e5bfa4282) (squash-merge of the full investigation). This PR bumps the godot submodule to that commit and keeps a defensive runtime-injection fallback in the iOS plugin so users on stock 4.6.2 templates also get the fix.

## Symptoms

| Path | Cold launch (app killed) | Warm / foreground |
|---|---|---|
| `decentraland://…` | ✅ worked | ❌ broken |
| `https://…` Universal Link | ✅ worked (URL extracted from `launchOptions[UIApplicationLaunchOptionsUserActivityDictionaryKey]`) | ❌ broken — app foregrounded but no event, no teleport |

### 1. UIScene migration dropped URL forwarding ([godotengine/godot#116395](https://github.com/godotengine/godot/pull/116395))

Godot 4.6.2 cherry-picked "[iOS] Add UIScene lifecycle events". The PR:
- Adds `UIApplicationSceneManifest` to the iOS Info.plist template.
- Makes `GDTApplicationDelegate` conform to `UIWindowSceneDelegate` and registers itself as the scene delegate via `application:configurationForConnectingSceneSession:options:`.
- Forwards the six scene **lifecycle** callbacks (`sceneDidBecomeActive:`, `sceneWillResignActive:`, …) to plugin services.

It does **not** forward the three scene **URL/userActivity** callbacks: `scene:openURLContexts:`, `scene:willConnectToSession:options:`, `scene:continueUserActivity:`. In scene-based iOS apps those are the only methods iOS dispatches URLs through — the legacy `application:openURL:options:` and `application:continueUserActivity:restorationHandler:` on the AppDelegate are no longer called. Result: every Godot iOS plugin that implemented those legacy selectors silently stops receiving URLs, including `DeeplinkService`.

Confirmed upstream issue [godotengine/godot#118600](https://github.com/godotengine/godot/issues/118600). Same regression broke `godot-sdk-integrations/godot-deeplink` ([#14](https://github.com/godot-sdk-integrations/godot-deeplink/issues/14)) — `csueiras` posted the same forwarding fix.

### 2. SwiftUI WindowGroup intercepts URL events on iOS 17+

Godot 4.6 also moved the iOS app shell to SwiftUI:

```swift
@main struct SwiftUIApp: App {
    @UIApplicationDelegateAdaptor(GDTApplicationDelegate.self) var appDelegate
    var body: some Scene {
        WindowGroup { GodotSwiftUIViewController() }
    }
}
```

In a SwiftUI `WindowGroup` app the runtime wraps the configured scene delegate with a SwiftUI-internal proxy. **Even after fixing #1**, scene URL events for HTTPS Universal Links never reach `GDTApplicationDelegate.scene:continueUserActivity:`. On **iOS 17+** the routing changed further: HTTPS Universal Links arrive at SwiftUI's `.onOpenURL` modifier as a plain `URL`, not at `scene:continueUserActivity:` as an `NSUserActivityTypeBrowsingWeb` activity. Without an explicit handler SwiftUI silently consumes them.

Diagnostic capture from a real device confirmed (`scene:continueUserActivity:` never fires, `.onOpenURL` does):

```
[DEEPLINK] GDTApplicationDelegate.scene:willContinueUserActivityWithType: fired type=NSUserActivityTypeBrowsingWeb
[DEEPLINK] SwiftUI .onOpenURL fired, url=https://mobile.dclexplorer.com/open?position=150,150
```

Custom URL schemes (`decentraland://`) are unaffected by #2 because they don't go through `.onOpenURL`'s pipeline — they reach `application:openURL:options:` once #1 is fixed.

### Engine-side ([decentraland/godotengine#12](https://github.com/decentraland/godotengine/pull/12) → [`7e5bfa4282`](https://github.com/decentraland/godotengine/commit/7e5bfa4282) on `4.6.2`)

**Layer A — forward UIScene URL/userActivity to AppDelegate services.** Adds `scene:openURLContexts:` / `scene:willConnectToSession:options:` / `scene:continueUserActivity:` on `GDTApplicationDelegate`. Each forwards to the existing `application:openURL:options:` / `application:continueUserActivity:restorationHandler:` service selectors, so plugins keep using the legacy AppDelegate selectors they already implement. Fixes regression #1.

**Layer B — bridge SwiftUI WindowGroup to the AppDelegate.** Adds two SwiftUI modifiers to the WindowGroup body that re-dispatch through the real `GDTApplicationDelegate` (`@UIApplicationDelegateAdaptor` instance):

```swift
.onContinueUserActivity(NSUserActivityTypeBrowsingWeb) { userActivity in
    _ = appDelegate.application(UIApplication.shared, continue: userActivity, restorationHandler: { _ in })
}
.onOpenURL { url in
    _ = appDelegate.application(UIApplication.shared, open: url, options: [:])
}
```

`.onOpenURL` is what actually fires for HTTPS Universal Links on iOS 17+; `.onContinueUserActivity` covers older iOS / non-WindowGroup configurations. Both bypass SwiftUI's scene-delegate wrapping. Fixes regression #2.

This PR bumps `plugins/dcl-godot-ios/godot` to the post-merge `4.6.2` head.

### Plugin-side — runtime injection fallback (defensive)

`plugins/dcl-godot-ios/plugins/dcl_godot_ios/deeplink_service.mm` adds the same three scene methods on `GDTApplicationDelegate` via `class_addMethod` from a `+ (void)load` on a dedicated `DeeplinkServiceLoader` class. This runs at image-load time — before `UIApplicationMain` instantiates the SceneDelegate. The class is referenced from `force_deeplink_service_initialization` so the linker keeps it under `-dead_strip` (the iOS app does not use `-ObjC`).

Only fires on templates that lack the engine-side fix. Once templates are rebuilt from `4.6.2` (post-merge) `class_respondsToSelector` returns YES and injection is skipped — the engine's own forwarders take over. Kept as a safety net so the plugin keeps working against any 4.6.2 template that hasn't been rebuilt yet (and Layer B in `app.swift` is the only one that fixes warm-start Universal Links on iOS 17+ — that one always needs the rebuilt template).

## Verifying the fix on device

`DeeplinkService` logs are mirrored to Godot's stdout via `print_line`, so a warm-tap of `https://mobile.dclexplorer.com/open?position=150,150` produces (in order):

```
[DEEPLINK] DeeplinkService.application:openURL: fired, URL=https://mobile.dclexplorer.com/open?position=150,150
[DEEPLINK]   applicationState=…
[DEEPLINK] emit_deeplink_received called with URL: https://mobile.dclexplorer.com/open?position=150,150
[DEEPLINK] DclGodotiOS::deeplink_received signal emitted to GDScript
[DEEPLINK] process_deep_link params: { "position": "150,150" }
```

Custom URL schemes (`decentraland://…`) follow the same path.

## Test plan
- [ ] **Warm-start custom scheme**: in-game, tap `decentraland://open?position=10,20` → teleport.
- [ ] **Warm-start Universal Link**: in-game, tap `https://mobile.dclexplorer.com/open?position=10,20` → teleport.
- [ ] **Cold-start custom scheme** (regression check): force-quit, tap `decentraland://jump?position=5,5` → launch → auth → teleport.
- [ ] **Cold-start Universal Link** (regression check): force-quit, tap `https://mobile.dclexplorer.com/open?position=5,5` → launch → auth → teleport.
- [ ] **Background → warm**: app backgrounded, tap UDL → app foregrounds + teleports.
- [ ] **Path coverage**: `/jump`, `/open`, `/events?id=…`, `/places?id=…` route correctly.
- [ ] **WalletConnect callback**: `decentraland://walletconnect?…` → no teleport, wallet flow proceeds.
- [ ] **Signin deeplink**: `decentraland://signin?identityId=…` mid mobile-auth → auth completes, no teleport.

## Out of scope / follow-ups
1. **Rebuild stable iOS templates from `4.6.2`** — once the templates served at `godot-engine-releases.dclexplorer.com/4.6.2.stable/compressed-templates/ios.zip` are rebuilt from the post-merge `4.6.2` head, the runtime-injection fallback in this PR becomes dead code and can be removed in a follow-up PR.
2. **Submit upstream** — open a Godot PR closing [godotengine/godot#118600](https://github.com/godotengine/godot/issues/118600). Both layers (UIScene forwarding + SwiftUI bridge) benefit any Godot iOS app on the SwiftUI shell.


godot/export_presets.cfg
```diff
@@ -254,14 +259,18 @@ xr_features/passthrough=0
 name="ios"
 platform="iOS"
 runnable=true
-advanced_options=true
 dedicated_server=false
 custom_features=""
 export_filter="all_resources"
 include_filter="*, *.*"
```
regenesis-claw, line 265: The iOS export path changed from `../exports/...` to `../../ws/exports/...`. This looks like a local workspace path — should it be committed, or was this accidentally picked up from your local Godot project settings?

godot/export_presets.cfg
```diff
@@ -530,6 +539,10 @@ storyboard/custom_image@2x=""
 storyboard/custom_image@3x=""
 storyboard/use_custom_bg_color=false
 storyboard/custom_bg_color=Color(0, 0, 0, 1)
+privacy/collected_data/search_hhistory/collected=false
```
regenesis-claw: These four lines re-introduce the `search_hhistory` typo that was corrected a few lines above (→ `search_history`). Looks like the Godot editor auto-generated a duplicate block with the old misspelling. Harmless, but worth cleaning up while you're here.

regenesis-claw (commented):
Thorough fix — the root-cause analysis is clear and the two-layer defense (Godot fork patch + runtime `+load` injection) is exactly the right approach for an upstream regression with no fix yet.

The Obj-C runtime injection code is solid: `+load` timing is correct for pre-SceneDelegate setup, the `(void)[DeeplinkServiceLoader class]` anti-strip reference is the canonical pattern, `scene:continueUserActivity:` covers the Universal Link warm-start gap the original code missed, and the `DEEPLINK_LOG` macro with the OS singleton guard is clean.

Two observations on `export_presets.cfg` (possibly just Godot editor artifacts — flagged inline).

godot/export_presets.cfg
kuruk-mm: Thanks — confirmed it was a local workspace path picked up from my Godot editor. Reverted to `../exports/decentraland-godot-client.xcodeproj` in [`a9adece7`](https://github.com/decentraland/godot-explorer/commit/a9adece7).

godot/export_presets.cfg
kuruk-mm: Good catch — duplicate auto-generated block dropped in [`ec3ab4d2`](https://github.com/decentraland/godot-explorer/commit/ec3ab4d2). Only the corrected `search_history` keys remain.

PR #1930, opened 2026-04-30
chore: bump sentry godot 1.6.0

leanmendoza:
## Why

We were stuck on `sentry-godot` 1.0.0 (Oct 2025), six months behind upstream. The bump bridges 11 releases worth of mobile-relevant fixes and unlocks the proper `attach_log` API (see follow-up commit) that was the real driver behind our recent attachment-quota blow-ups.

### Crash reporting becomes more reliable
- **iOS**: fix for a crash that occurred when the SDK tried to send a crash report from a previous session on next app launch (1.3.1). Until now our iOS users could enter a loop where the very act of reporting a crash triggered another crash.
- **Android**: fix for occasional crash-on-exit (1.1.1). We see plenty of `godot-rust function call failed: SceneManager::physics_process` events at shutdown — at least some of those should disappear.
- **Both**: guard against accessing the view hierarchy before the scene tree is ready (1.6.0). Removes another startup crash path.

### Better signal in mobile crashes
- **Android**: native crashes now carry our custom tags (1.2.0). Previously, when the app crashed at the NDK level, all our context (`dcl_session_id`, `branch_name`, `commit_hash`, etc.) was stripped — forcing us to guess which build/session the crash came from.
- **Android**: tombstone support enabled (1.5.0). On Android 11+, the OS keeps a structured tombstone on `ANR`/`SIGSEGV`/native abort. The SDK now ingests these, which gives us native stack traces from devices that didn't crash inside our process.
- **iOS**: scene tree captured in events (1.5.0). The Cocoa side previously sent crashes without scene context.
- **iOS**: app-hang timeout lowered to 5s on Apple platforms and disabled by default (1.1.0). Stops a class of false-positive hang reports while still letting us opt back in.

### iOS App Store / TestFlight footguns closed
- iOS XCFramework plist now correctly reports min version 15.0 instead of 12.0 (1.6.0). Prevents potential validation failures on App Store submission.
- Export-time warning when the project's iOS minimum version is below what the SDK requires (1.6.0). Catches misconfigurations before we ship.

### Attachment-quota architecture fixed
- `options` set inside the `SentrySDK.init` configuration callback now actually affect the event processor registration (1.4.2). This was the underlying reason our 1.0.0 attach-log sampling had to call `SentrySDK.add_attachment(...)` manually after init. With the fix, we can set `options.attach_log = sampled_bool` directly and the SDK does the right thing — both Android and iOS.
- Attachment handling hardened across all platforms; blank attachment names on Apple platforms fixed (1.4.1).

### Quality-of-life
- Per-build environment via project setting (1.3.0).
- `SentrySDK.set_attribute()` / `remove_attribute()` for global attributes that flow into structured logs and metrics (1.5.0).
- `device_type` populated in device context for better device-class filtering in the Sentry UI (1.5.0).
- `enable_logs` is now off by default to align with other Sentry SDKs (1.3.2). We don't ship structured logs today; this keeps it that way unless we explicitly opt in.
- Configurable `shutdown_timeout_ms` (1.4.0) — lets us bound how long the SDK blocks app exit while flushing.

### Underlying native SDK bumps

| platform | from | to |
|---|---|---|
| Android (sentry-java) | 8.23.0 | 8.39.1 |
| iOS / macOS (sentry-cocoa) | 8.56.2 | 9.10.0 |
| Native (Crashpad) | 0.11.2 | 0.13.7 |

Six months of upstream stability and performance fixes for free, including better ANR detection on Android and Swift 6 readiness on iOS.

## What's NOT in this PR

- No changes to which events we send or how we sample.
- No changes to user-facing behavior. The SDK loads the same way; events look the same in the Sentry UI (apart from the new context fields we now get for free).
- The follow-up commit (`refactor(sentry): use options.attach_log instead of manual add_attachment`) leverages the 1.4.2 fix to clean up our 1.0.0 workaround.

## Compatibility

- iOS minimum jumped to 15.0 in `sentry-godot` 1.5.0. Our `application/min_ios_version` is **18.0**, so no impact.
- Android minimum unchanged.
- The custom `godot/sentry.gdextension` was updated to match the new layout: `entry_symbol` renamed (`gdextension_init` → `sentry_gdextension_init`), macOS switched from `.framework` to `.dylib`, Linux/arm64 + Windows/arm64 now use real binaries instead of noop, Android `x86_32` added.

## Verification

- `cargo run -- doctor` passes after fresh install.
- Headless editor smoke test on macOS: `Verifying GDExtensions... DONE`, exit 0.
- Live event in Sentry confirms native SDK versions match 1.6.0 on both Android (`sentry.java.android.godot/8.39.1`) and iOS (`sentry.cocoa.godot/9.10.0`).
- Mobile build pipelines on CI cover the platforms not validated locally (iOS, Android, Windows, Linux).

### DIFF
--- a/godot/project.godot
+++ b/godot/project.godot
@@ -245,7 +245,6 @@ options/auto_init=false
 options/dsn="https://03559fa545b3fa2bc9e876a41d6aab2f@o4510187684298752.ingest.us.sentry.io/4510187688361984"
 options/debug_printing=0
 options/diagnostic_level=4
-options/attach_log=false
 
 [shader_globals]
 
--- a/godot/sentry.gdextension
+++ b/godot/sentry.gdextension
@@ -1,22 +1,26 @@
 [configuration]
 
-entry_symbol = "gdextension_init"
-compatibility_minimum = "4.6"
+entry_symbol = "sentry_gdextension_init"
+compatibility_minimum = "4.5"
 
 [libraries]
 
-macos.debug = "res://addons/sentry/bin/macos/libsentry.macos.debug.framework"
-macos.release = "res://addons/sentry/bin/macos/libsentry.macos.release.framework"
+macos.debug = "res://addons/sentry/bin/macos/libsentry.macos.debug.dylib"
+macos.release = "res://addons/sentry/bin/macos/libsentry.macos.release.dylib"
 
 windows.debug.x86_64 = "res://addons/sentry/bin/windows/x86_64/libsentry.windows.debug.x86_64.dll"
 windows.release.x86_64 = "res://addons/sentry/bin/windows/x86_64/libsentry.windows.release.x86_64.dll"
 windows.debug.x86_32 = "res://addons/sentry/bin/windows/x86_32/libsentry.windows.debug.x86_32.dll"
 windows.release.x86_32 = "res://addons/sentry/bin/windows/x86_32/libsentry.windows.release.x86_32.dll"
+windows.debug.arm64 = "res://addons/sentry/bin/windows/arm64/libsentry.windows.debug.arm64.dll"
+windows.release.arm64 = "res://addons/sentry/bin/windows/arm64/libsentry.windows.release.arm64.dll"
 
 linux.debug.x86_64 = "res://addons/sentry/bin/linux/x86_64/libsentry.linux.debug.x86_64.so"
 linux.release.x86_64 = "res://addons/sentry/bin/linux/x86_64/libsentry.linux.release.x86_64.so"
 linux.debug.x86_32 = "res://addons/sentry/bin/linux/x86_32/libsentry.linux.debug.x86_32.so"
 linux.release.x86_32 = "res://addons/sentry/bin/linux/x86_32/libsentry.linux.release.x86_32.so"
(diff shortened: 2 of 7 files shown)


regenesis-claw (approved):
Clean bump — the diff is small and well-scoped.

**What I checked:**

- `sentry.gdextension` layout matches 1.6.0 conventions: renamed entry symbol, macOS `.framework` → `.dylib`, noop libs pruned, new real binaries for linux/arm64 + windows/arm64 + android/x86_32 + web. ✅
- `options.attach_log = self.attach_log_sampled` inside the init callback now works (1.4.2 fix) — the old manual `add_attachment()` workaround is correctly removed along with its comment. ✅
- `options/attach_log=false` removed from `project.godot` — makes sense since the SDK default changed to `true` in 1.3.2 and the sampling now happens in code. ✅
- Download URL and cache key updated (`sentry-1.6.0.zip`) — cache busted properly. ✅
- `build.rs`: `cargo:rerun-if-changed` for `.git/HEAD`, `.git/refs/heads`, `.git/index`, and the env vars — good addition, keeps the embedded commit hash in sync. ✅
- Dead `@onready` refs removed in `login.gd` — no references left dangling.

CI all green. LGTM.

Requested by Lean via Slack

kuruk-mm (approved):

PR #1932, opened 2026-04-30
fix: show skeleton placeholder while profile loads

### LINKED ISSUE #1691: [Bug]: Profile placeholder data flashes briefly on first load
### 🕹️Platform

Android, iOS, Web Client

### 🔢App Version

Not specified

### 📱Device Information

Not specified

### 📄Issue description

When loading a profile for the first time, placeholder data (e.g., "Namemaxcharatce_1234") appears briefly before the selected profile is displayed. This creates a visual flash/flicker that impacts the user experience.

### ✅Expected behavior

The profile loading should either:
- Display a loading indicator instead of placeholder data
- Wait to render the profile until actual data is available
- Hide the profile display until real data is loaded

### 🖼️Screenshots / Media

https://github.com/user-attachments/assets/e162912b-3603-464c-b407-cecddec9a9a8

### ▶️Steps to reproduce

1. Open the app
2. Load a profile for the first time
3. Observe the brief display of placeholder data before actual profile appears

### 👁️Occurrence

100% of the time

---

sdilauro:
## Summary

Fixes #1691

- **Replaced the old loading overlay** (`PanelContainer_GettingData` with spinner) with a skeleton placeholder approach: `HBoxContainer_PlaceHolder` shows skeleton panels that mirror the real profile layout while data loads, then swaps to `HBoxContainer_Data` when ready
- **Fixed async timing**: `_set_avatar_loading()` was called *after* the avatar `await`, leaving placeholder text visible during load. Now the skeleton shows immediately and the profile panel stays hidden until all data (avatar, wearables, name, about, links) is resolved
- **Awaited `profile_equipped.async_refresh()`** — previously called without `await`, causing the loading state to end before wearables finished fetching
- **Added blur background** to `profile_container.tscn` for polish while the profile loads

## Test plan

- [ ] Open another player's profile for the first time — should see skeleton placeholder, then smooth swap to real data
- [ ] Open own profile — should show skeleton, then real data with edit button
- [ ] Verify profile size stays consistent between placeholder and data states (no resize jump)
- [ ] Test on portrait orientation — placeholder text should not flash
- [ ] Deploy a profile change and verify the deploy-waiting state works (shows placeholder until server confirms)


godot/src/ui/components/profile/profile_portrait.gd
```diff
@@ -91,17 +91,17 @@ func _refresh_content() -> void:
 	if profile == null:
 		return
 	var content = draggable_bottom_sheet.get_content_instance()
-	if content and content.has_method("refresh"):
-		content.refresh(profile)
+	if content and content.has_method("async_refresh"):
```
regenesis-claw, line 94: This calls `async_refresh` (now a coroutine) without `await`. It will still run, but the caller won't wait for `profile_equipped.async_refresh` inside it to complete. If that's intentional (fire-and-forget), consider adding a comment so future readers know it's deliberate.

regenesis-claw (commented):
Nice improvement — the skeleton placeholder approach is much better UX than the old spinner overlay, and fixing the async timing so the skeleton shows immediately is the right call.

One thing worth considering:

**Unawaited async calls after the rename**

In `profile_portrait.gd`, both `_refresh_content()` and `_refresh_content_from_mutable()` call `content.async_refresh(profile)` without `await`. Since `async_refresh` now contains `await profile_equipped.async_refresh(...)`, the equipped section will load asynchronously without the caller knowing when it finishes. This is probably fine for the signal-handler context, but if there's ever downstream code that depends on the refresh being complete, it could be a subtle issue. Same pattern in `explorer.gd` where `_open_profile` calls `async_open` without `await` — `release_mouse()` runs immediately, which seems intentional.

Just flagging so it's a conscious decision rather than an oversight. If you're happy with fire-and-forget semantics there, no change needed.

Everything else looks solid — the `_show_placeholder()`/`_show_data()` swap is clean, the deploy-waiting logic with the boolean flag is simpler than the counter approach, and the blur background is a nice polish touch. 👍

ElzathRusso (approved):

regenesis-claw (approved):
Re-reviewed after latest commit (bc6fa18). The fire-and-forget comments on the unawaited `async_refresh` calls in `profile_portrait.gd` address the previous feedback nicely — intent is now explicit.

Note: `_open_profile` in `explorer.gd` still calls `async_open` without await, but that's fine since it's a signal callback (can't await there anyway).

LGTM ✅

Requested by Seba via Slack

kuruk-mm (approved):

rodrigo685 (approved):

PR #1933, opened 2026-04-30
chore: sentry digest: filter production env, show only crash-free rate

leanmendoza:
Filters Sentry sessions queries to `environment:production`. Strips sessions/users/errors counts from the Slack message, leaving only the crash-free rate (and error-free rate for `auth-mobile`).

Background: previous digest aggregated prod + staging + dev together, inflating user count ~21% and reporting an optimistic crash-free rate.


kuruk-mm (approved):
looks great!

regenesis-claw (approved):
Clean refactor. Filtering to production-only is the right fix for the inflated numbers, and stripping the raw counts in favor of just the rate makes the digest much more scannable.

One minor note: `day_rate()` returns `0.0` when `sessions == 0` — this means a day with zero sessions shows `0.00%` which could be misread as "0% healthy" rather than "no data". Probably fine for a daily digest but worth knowing.

Thresholds bump from 90/75 → 99/95 makes sense for crash-free rates on a production mobile app. LGTM.

PR #1934, opened 2026-04-30
feat: travel modal for teleport and world links

### LINKED ISSUE #1156: Modal: Teleport to Scene & World Modals
Teleport to Scene and World modals look outdated and have different implementation, they should be the same from the user POV.
Desktop 

**[Figma](https://www.figma.com/design/IG9xWnAveE63CrLNSIIJ7V/Modals?node-id=50-47&t=wKd2T09SUHfsOxeo-1)**

[ATLAS - Specs](https://www.figma.com/design/IG9xWnAveE63CrLNSIIJ7V/Modals?node-id=233-190&t=wKd2T09SUHfsOxeo-1)

----
Modal with an image using a horizontal layout to better fit the screen orientation.

**Content**:
Place / World / Event name
Creator
JUMP IN button
Close (X) button (not cancel button)

## New
Image

## Old
Image

sdilauro:
Closes #1156

## Summary

- New `TravelModal` scene (`travel_modal.tscn` / `travel_modal.gd`) with horizontal layout: place image on the left, purple content panel on the right with title, creator, close button, and JUMP IN button
- Skeleton loading states for both the image and text content — skeletons show until real data arrives, JUMP IN stays disabled until data is loaded
- Replaces the old generic `Modal` for teleport actions (`async_show_teleport_modal`)
- Chat messages containing `.dcl.eth` world names (e.g. `myworld.dcl.eth`) are now detected as clickable links that open the travel modal
- World names are validated against the Places API before showing the modal; if the world doesn't exist, an alert toast is shown instead
- Responsive layout: landscape uses minimum sizes (600x460 image, 460x460 content), portrait resets both to zero

## How to integrate / Implementation details

The travel modal is managed through `ModalManager` (accessible via `Global.modal_manager`):

### Teleport to coordinates
```gdscript
# Opens the travel modal, loads place data async (name, creator, image)
Global.modal_manager.async_show_teleport_modal(Vector2i(10, 20))

# With a specific realm
Global.modal_manager.async_show_teleport_modal(Vector2i(10, 20), "my-realm.dcl.eth")
```

### Teleport to a world
```gdscript
# Validates the world exists first, then opens the modal with world data
Global.modal_manager.async_show_world_modal("myworld.dcl.eth")
```

### Flow
1. `ModalManager` creates a `TravelModal` instance inside a `CanvasLayer` (layer 100)
2. Modal opens immediately with skeleton loading state
3. Places API is queried for place/world data
4. On success: title, creator, and image are populated, skeletons hide, JUMP IN enables
5. On JUMP IN: `Global.async_teleport_to()` is called, modal closes
6. On close (X button or background tap): modal is dismissed

### Chat integration
- Coordinates like `10,20` → clickable → opens teleport modal (existing behavior)
- World names like `something.dcl.eth` → clickable → validates via API → opens world modal or shows alert toast

## Test plan

- [ ] In chat, type coordinates (e.g. `52,-10`) and click them — travel modal should open with skeleton, then load place data
- [ ] Click JUMP IN — should teleport to those coordinates
- [ ] Click X or tap background — modal should close
- [ ] In chat, type a valid `.dcl.eth` world name and click it — travel modal should open with world data
- [ ] Type an invalid `.dcl.eth` world name — alert toast should appear, no modal
- [ ] Test in both landscape and portrait orientations
- [ ] Verify the old teleport modal no longer appears (SDK-triggered teleports from scenes)


godot/src/ui/components/modal/travel_modal.gd
```diff
@@ -0,0 +1,100 @@
...
+func set_image(texture: Texture2D) -> void:
+	if texture:
+		texture_rect_image.texture = texture
+		panel_skeleton.hide()
+		texture_rect_image.show()
+
+
+func _async_update_modal_size() -> void:
+	if not is_inside_tree():
+		return
```
regenesis-claw: `_async_update_modal_size()` is defined but never called anywhere — looks like dead code. Was this intended for future use, or leftover from development? If it's not needed, consider removing it to keep things clean.

godot/src/ui/components/modal/travel_modal.gd
```diff
@@ -0,0 +1,100 @@
...
+	if panel_container and panel_container.has_method("_request_update"):
+		await get_tree().process_frame
+		await get_tree().process_frame
+		panel_container._request_update()
+
+
+func _on_gui_input(event: InputEvent) -> void:
```
regenesis-claw, line 85: This only handles `InputEventScreenTouch` — on desktop/Linux builds, clicking the background with a mouse won't close the modal. If desktop support matters, you'd want to also check for `InputEventMouseButton`:
```gdscript
func _on_gui_input(event: InputEvent) -> void:
	if (event is InputEventScreenTouch and event.pressed) or (event is InputEventMouseButton and event.pressed and event.button_index == MOUSE_BUTTON_LEFT):
		closed.emit()
```
If this is intentionally mobile-only, ignore this.

godot/src/ui/components/chat/chat_message.gd
```diff
@@ -195,6 +195,23 @@ func make_urls_clickable(text: String) -> String:
...
+		var start_pos = world_match.get_start()
+		var end_pos = world_match.get_end()
+
+		var clickable_world = (
+			"[url=world:%s][color=#66B3FF]%s[/color][/url]" % [world_name, world_name]
+		)
+		processed_text = (
+			processed_text.substr(0, start_pos) + clickable_world + processed_text.substr(end_pos)
+		)
+
```
regenesis-claw, line 216: The world regex runs before the URL regex, which means `.dcl.eth` names get wrapped in bbcode first. The URL regex (`[a-zA-Z0-9][-a-zA-Z0-9]*\.[a-zA-Z]{2,}[^\s]*`) could then re-match parts of the bbcode-wrapped text. In practice `_is_safe_url` rejects these because of `[` / `]` characters, so it won't double-wrap — but it's fragile. A more robust approach: after processing world names, also skip URL matches whose range overlaps with an already-processed world region (same pattern used for coordinates already).

godot/src/ui/components/modal/modal_manager.gd
```diff
@@ -138,41 +145,70 @@ func async_show_connection_lost_modal(hide_buttons: bool = false) -> void:
...
+		)
+		return
+
+	var world_data: Dictionary = json.data[0]
+
+	if not await _async_create_travel_modal():
+		return
+
+	current_travel_modal.closed.connect(close_travel_modal)
+	current_travel_modal.jump_in_pressed.connect(_on_world_jump_in.bind(world_name))
```
regenesis-claw, line 194: Minor: `_async_load_travel_modal_image` is called without `await` here. Since it's an async function (does `await` internally), this is fire-and-forget, which means any error in the image download won't propagate. That's probably fine for a non-critical image load, but worth being explicit — a comment like `# Fire-and-forget: image loading failure is non-blocking` would clarify intent.

regenesis-claw (commented):
Nice work on the new TravelModal! The skeleton loading states and the world link detection in chat are solid additions. A few things I noticed below — nothing blocking, but worth a look.

regenesis-claw (approved):
All review comments addressed — dead code removed, desktop dismiss added, clarifying comments in place. LGTM :rocket:

regenesis-claw (changes requested):
Requesting one addition before merge: the new TravelModal should also handle realm changes (SDK `changeRealm()` and `/changerealm` command), as this will be used for plaza portals.

Currently `async_show_change_realm_modal()` still uses the old generic Modal. It should be updated to use TravelModal similar to how `async_show_teleport_modal()` was updated.

The realm change modal would need:
- Load place data for the destination realm (if available via Places API)
- Show the skeleton → place image → title/creator flow
- Use the same JUMP IN button UX
- Include the optional scene message if provided

This keeps the modal UX consistent across all travel scenarios (teleport, world links, and realm changes).

regenesis-claw (approved):
Realm change modal now uses TravelModal — looks good! Clean conversion with proper fallback to realm name when Places API fails.

One minor note: the `message` parameter from the scene is now silently ignored (`_message`). The old modal displayed it as "Scene message: ...". This is probably fine since the new TravelModal doesn't have a message field, but worth being aware that scene-provided context is dropped.

LGTM :rocket:

regenesis-claw:
**Testing feedback (iOS build 440):**

The new TravelModal works great for scene teleports (coordinate-based portals), but portal doors in Plaza that use `changeRealm()` (i.e. `goToWorld` destination type) are still showing the old modal UI.

The Plaza `portalsManager.ts` calls `changeRealm({ realm: config.destination, message: 'Jump to ' + config.destination + '?' })` for world portals — this code path doesn't appear to be triggering the new TravelModal on build 440. Either the change didn't make it into this build, or there's a missing case for realm changes initiated via portal interactions.

Could you verify whether the `async_show_change_realm_modal` → TravelModal conversion is included in build 440, or if a new build is needed?

Requested by Rodri via Slack

rodrigo685:
Reviewed and Approved, /changerealm command now supported

image

rodrigo685 (approved):

kuruk-mm (approved):
great job!

PR #1942, opened 2026-05-01
fix: use OLE Automation Date for chat timestamps

### LINKED ISSUE #1928: [Bug] Chat messages from mobile showing 1904 timestamps in Explorer (Android and iOS)
### Platform
Explorer 

### App Version
v422

### Device Information
iPhone 16E

### Issue Description
Chat messages sent from mobile clients are displaying with timestamps from the year 1901 when viewed in the Explorer (desktop) client.

Chat messages on iOS mobile are displaying timestamps from 1904 instead of the current date. The user's message timestamp shows "Wed, 6 Jul, 1904" at "09:10 AM" (see screenshot).

This appears to be a timestamp initialization/conversion bug related to the Mac/iOS epoch date (January 1, 1904), which often shows up when there's an uninitialized Unix timestamp or incorrect date format conversion on Apple platforms.

Image

### Expected Behavior
Messages should display with the correct timestamp matching when they were actually sent.

### Steps to Reproduce
1. Send a chat message from the mobile app
2. View the same chat message in the Explorer desktop client
3. Observe the timestamp shows a date from 1901

### Occurrence
Not specified

---

**Related Issues:**
- #1925 — Chat messages showing timestamps from 1904 on iOS (similar timestamp issue but different symptom)

Requested by Rodri via Slack

Dogebusters:
Chat messages currently use elapsed time since connection as timestamp.

The Unity-based client uses OLE Automation Date
(days since 1899-12-30)

This mismatch can cause issues with chat ordering and interoperability.

This PR switches Chat.timestamp to OLE format while keeping movement/emote timestamps unchanged.

See Unity implementation:
https://github.com/decentraland/unity-explorer/blob/3ef5c5eb0f8331b7ca5294de52f1e8d6c187ab13/Explorer/Assets/DCL/Chat/History/ChatMessage.cs#L24

### DIFF
--- a/lib/src/comms/communication_manager.rs
+++ b/lib/src/comms/communication_manager.rs
@@ -3,7 +3,7 @@ use godot::prelude::*;
 use http::Uri;
 #[cfg(feature = "use_livekit")]
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 #[cfg(feature = "use_livekit")]
 use crate::{
@@ -1093,7 +1093,7 @@ impl CommunicationManager {
         let packet = rfc4::Packet {
             message: Some(rfc4::packet::Message::Chat(rfc4::Chat {
                 message: text.to_string(),
-                timestamp: self.start_time.elapsed().as_secs_f64(),
+                timestamp: ole_timestamp_now(),
             })),
             protocol_version: DEFAULT_PROTOCOL_VERSION,
         };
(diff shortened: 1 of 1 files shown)


Dogebusters:
Convert Unix time to OLE Date:

Unix = seconds since 1970
OLE = days since 1899

unix_seconds / 86400.0 + 25569.0

= seconds → days + epoch offset (1970 → 1899)

regenesis-claw (approved):
LGTM — small, focused fix. The OLE conversion math is correct (`unix_seconds / 86400.0 + 25569.0`), and the `unwrap_or_default()` fallback is safe (returns epoch 0 → OLE 25569, still a valid monotonic value).

Verified that:
- Duplicate-detection in `message_processor.rs` works fine since OLE timestamps are monotonically increasing just like the old elapsed-time values
- GDScript side doesn't interpret the timestamp format (the parameter is prefixed with `_` in `set_chat`), so no downstream breakage
- Movement and emote timestamps are correctly left unchanged

Requested by Mateo via Slack

kuruk-mm (approved):
looking great! thanks for the contribution!

PR #1947, opened 2026-05-03
fix: scene interactable area + hide camera button when blocked

### LINKED ISSUE #1898: [Bug] Full-screen UI overlays don't cover safe area on mobile devices
## Summary
Full-screen UI overlays rendered via React-ECS (`width: '100%', height: '100%'`) do not extend to the physical screen edges on mobile devices. A strip of the 3D scene remains visible behind the overlay, breaking the intended effect of black screens, death screens, fade transitions, and modal backgrounds.

## Steps to Reproduce
1. Open any Decentraland scene on mobile that uses a full-screen UI overlay (e.g. black fade, death screen, title splash)
2. Hold device in landscape orientation
3. Observe that the overlay does not cover the safe area inset region — a thin strip of the 3D scene is visible on one or both edges

## Expected Behavior
Full-screen UI elements with `width: '100%', height: '100%'` and `positionType: 'absolute', position: { top: 0, left: 0 }` should cover the entire physical screen, including the area behind the notch/safe area insets.

## Actual Behavior
The UI canvas is inset from the screen edges to respect the device safe area. Scene code has no mechanism to draw outside this canvas, so full-screen overlays (black screens, fade transitions, modal backdrops) leave a visible gap where the 3D scene shows through.

Image

Image

## Suggested Fix
The mobile app renderer should either:
1. **Extend the UI canvas to the full physical screen** — let `width: '100%'` truly mean 100% of the display, and let scene developers handle safe area padding themselves
2. **Fill the safe area inset region with the nearest UI background color** — similar to iOS `edgesIgnoringSafeArea(.all)` or Android `WindowCompat.setDecorFitsSystemWindows(window, false)`
3. **Provide a scene-accessible API** (e.g. a component or canvas property) that lets developers opt in to drawing behind the safe area for specific UI elements

Option 1 or 2 would fix this transparently for all existing scenes without requiring scene code changes.

## Environment
• Decentraland Mobile App: v0.58.0-1c2b5f1-prod - Opt
• SDK: @dcl/sdk ^7.22.5
• Scene: Flag Tag (flagtag.dcl.eth) — 1024-parcel world

---
*Requested by Rodri via Slack*

kuruk-mm:
Fix https://github.com/decentraland/godot-explorer/issues/1898

## Summary
- Fix `UiCanvasInformation.interactable_area`: emit per-edge indents (BorderRect semantics) instead of raw rect positions, and push updates from Godot whenever the safe area or window size changes (including `--emulate-ios` / `--emulate-android` and orientation flips). The redundant `SafeMarginContainer` on `SceneUIContainer` is dropped in favor of a single explicit push from `explorer.gd`.
- Hide the virtual-joystick camera button while a `CameraModeArea` forces a camera mode, so users can't try to switch into a mode the scene won't allow. Adds `Global.camera_mode_blocked` + `camera_mode_block_changed` signal, with a press-handler short-circuit as a safety net.

## Test plan
- [ ] Desktop: resize window and verify scenes receive correct `interactable_area` / `device_pixel_ratio`.
- [ ] Mobile (or `--emulate-ios` / `--emulate-android`): rotate / change safe area and verify the scene UI rect tracks it.
- [ ] Walk into a scene with a `CameraModeArea` on mobile: camera toggle button should disappear, and reappear on exit.
- [ ] Open/close the navbar while inside a blocked area: camera button stays hidden.
- [ ] Cinematic camera mode still hides the button as before.


kuruk-mm:
@rodrigo685 we need to test with the existing scenes that try to implement Mobile UI and check if this PR doesn't break them...

regenesis-claw (approved):
Clean fix that correctly addresses both issues. The old `interactable_area` was swapping x/y into the wrong BorderRect fields — the new indent-from-edge computation is correct and well-clamped. The camera-block signal + centralized visibility refresh is a solid pattern.

One minor observation:

**`_refresh_camera_button_visibility` doesn't account for navbar-open state** — if `camera_mode_block_changed` fires while the navbar is open (e.g., player leaves a CameraModeArea with navbar open), the button could momentarily become visible under the navbar. Very unlikely in practice since movement is paused with navbar open, but tracking an `_is_navbar_open` bool in the refresh would make it airtight. Not blocking.

Also noting the pre-existing `prints(camera_mode, ...)` on line ~107 of `virtual_joystick.gd` is a debug leftover from before this PR — worth cleaning up in a follow-up.

Issue #1986, opened 2026-05-05
[Bug][Sentry] social_item.gd: call_deferred to nonexistent _update_buttons method (1SX)


kuruk-mm:
## Sentry
[GODOT-EXPLORER-1SX](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1SX) — `ERROR: Error calling deferred method: 'Button(social_item.gd)::_update_buttons': Method not found.`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | 1 | 2 |
| 0.64.0 | 2 | 3 |
| 0.58.0 | 1 | 2 |

Low-volume but persistent. Confirmed bug in current code.

## Root cause (confirmed)
`godot/src/ui/components/social/social_item/social_item.gd:264`:
```gdscript
func _sync_blacklist_ui(changed_avatar_id: String) -> void:
    if social_data and social_data.address == changed_avatar_id:
        call_deferred("_update_buttons")
```
No `_update_buttons` function exists in `social_item.gd`. (Only `profile.gd` defines `_update_buttons`.)

## Suggested fix
Either:
- Add a `_update_buttons` method to `social_item.gd` that calls the right refresh logic (likely `_update_elements_visibility` based on context), OR
- Replace `call_deferred("_update_buttons")` with the correct method name.

Issue #1984, opened 2026-05-05
[Bug][Sentry] ia_any InputMap action looked up on player_gamepad_input.gd (1Z9)


kuruk-mm:
## Sentry
[GODOT-EXPLORER-1Z9](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1Z9) — `ERROR: The InputMap action 'ia_any' doesn't exist. Did you mean 'ia_action_3'?`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | 2 | 14 |
| 0.64.0-b0532c7 | 1 | 6 |

## Root cause (confirmed)
`godot/src/ui/components/pointer_tooltip/tooltip_label.gd` special-cases `ia_any` as a virtual action (line 105, 171). But `godot/src/logic/player/player_gamepad_input.gd:18` blindly does:
```gdscript
for event in InputMap.action_get_events(action):
```
When `action` is `ia_any` (the virtual any-action), the InputMap lookup fails and emits the engine error.

## Suggested fix
- Add the same `if action == "ia_any": continue` guard in `player_gamepad_input.gd` before calling `InputMap.action_get_events`.
- Or define `ia_any` as a no-op action in `project.godot`.

Issue #1964, opened 2026-05-05
[Bug][Sentry] Discover category_tag.gd: load() fails for missing places_categories SVGs (8 users new in v0.65)


kuruk-mm:
## Sentry
[GODOT-EXPLORER-231](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-231) — `No loader found for resource: res://assets/ui/places_categories/featured.svg (expected type: unknown)`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | **8** | 20 |
| 0.64.0 | 0 | 0 |

**NEW in v0.65** — first seen 12h ago.

## Root cause (confirmed)
`godot/src/ui/components/discover/categories/category_tag.gd:18` does:
```gdscript
texture_rect.texture = load(CATEGORIES_ICONS_PATH + category + ".svg")
```
The `category` value comes from the catalyst/server. The asset directory `godot/assets/ui/places_categories/` does not contain `featured.svg` (only art, business, casino, crypto, education, entertainment, fashion, fav, game, gaming, live, market, music, poi, social, sports). When the server returns `featured` as a category, `load()` fails.

## Suggested fix
1. Add `featured.svg` (and audit any other categories the server may return that lack an asset).
2. Or fall back to a default icon if `load()` returns null. Wrap in `ResourceLoader.exists()` check first.
3. Better long-term: ship a manifest of supported categories or a default category icon.

kuruk-mm:
Superseded by #1988 — that issue tracks the underlying missing assets (`featured.svg`, `favorites.svg`, `shop.svg`) and is assigned to @sdilauro. Closing this one to avoid duplication; will reopen if the missing-asset fix doesn't drop the Sentry signature.

PR #1987, opened 2026-05-05
fix: avatar_attach duplicate parent + dup tree_changed connect; social_item dead method call

### LINKED ISSUE #1961: [Bug][Sentry] avatar_attach: duplicate parent + duplicate tree_changed connect — regression in v0.65 (12 users)
## Sentry
- [GODOT-EXPLORER-1Z0](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1Z0) — `Can't add child 'AvatarAttach' to 'eXXX', already has a parent`
- [GODOT-EXPLORER-1YZ](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1YZ) — `Signal 'tree_changed' is already connected to given callable 'Node(avatar_attach.gd)::_on_tree_changed'`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | **12** | 300 each |
| 0.64.0 | 3 | 17–22 |

**4× regression in v0.65** — likely tied to avatar impostor rollout (#1914, b0321d3503).

## Root cause (suspected)
`godot/src/decentraland_components/avatar_attach.gd` extends `PlayerColliderFilter`. When `init_player_collider_filter()` runs more than once on the same node (re-parent / re-init):
- It calls `scene.tree_changed.connect(self._on_tree_changed)` again → duplicate connect (1YZ)
- Whatever path adds `AvatarAttach` to the entity is firing twice → already has parent (1Z0)

`godot/src/decentraland_components/player_collider_filter.gd:18` — no guard against double-connect.

## Suggested fix
- Wrap connect in `if not scene.tree_changed.is_connected(self._on_tree_changed): scene.tree_changed.connect(...)`.
- For 1Z0, find the AvatarAttach add-child callsite and check parent is null first or use `reparent` / `call_deferred`.

kuruk-mm:
## Summary

Two real fixes to engine errors surfaced by Sentry triage in v0.64/v0.65. Both target root causes — no error muting.

| Sentry | Issue | File | Fix |
|---|---|---|---|
| 1YZ | #1961 | `lib/src/scene_runner/components/avatar_attach.rs` | When the AvatarAttach node already exists for an entity, only update its properties — don't re-`add_child` (was creating "duplicate parent") and don't re-`init()` (was creating a duplicate `tree_changed` connection). |
| 1SX | #1986 | `godot/src/ui/components/social/social_item/social_item.gd` | `_sync_blacklist_ui` was `call_deferred`-ing the nonexistent `_update_buttons`; route to the existing `_update_blocked_visibility_for_type` instead, matching the `_on_blacklist_changed` pathway. |

Closes #1961, #1986.

## Why the Rust fix (not a GDScript guard)

In `update_avatar_attach`, the code reuses the existing `AvatarAttach` Godot node when the component changes (e.g. anchor_point flips). It then unconditionally re-adds it as a child and re-calls `init()` — which itself connects `scene.tree_changed` and walks the collider tree. That's the underlying cause of:

- "Signal already connected" warnings on `scene.tree_changed`
- Duplicate parents (the same node appearing twice in `node_3d`'s children)
- Per-frame collider-tree work being doubled (or worse) per attach update

A GDScript `is_connected()` guard in `player_collider_filter.gd` would silence the signal warning but leave the duplicate parenting + duplicate work in place. The Rust change fixes the source.

## Out of scope

- **#1984 (gamepad `ia_any`)** — closed-out as misdiagnosed during triage. The actual `ia_any` lookup happens in `tooltip_label.gd`, which already special-cases it correctly. The `player_gamepad_input.gd` actions all exist in `project.godot` so no guard is warranted there. Leaving #1984 open with the corrected note for separate investigation.
- **#1964 (places category SVGs)** — separate; needs the missing assets added (`featured.svg`, `favorites.svg`, `shop.svg`) rather than a load() guard.
- Other Sentry items from the triage need design decisions, span Rust+GDScript, or were already addressed on `main`.

## Test plan

- [ ] Spawn an SDK scene with a `AvatarAttach` component that flips `anchor_point` (e.g. left hand → right hand) → no "Signal already connected" warnings, no duplicate `AvatarAttach` children under the entity node, the attach still updates its transform correctly.
- [ ] Open Nearby Players panel, block/unblock another user → row hides/shows correctly, no "Cannot find method `_update_buttons`" engine error in the console.
- [ ] `gdformat` / `gdlint` clean (verified locally), `cargo fmt` clean, `cargo check` clean.

### DIFF
--- a/godot/src/ui/components/social/social_item/social_item.gd
+++ b/godot/src/ui/components/social/social_item/social_item.gd
@@ -261,7 +261,7 @@ func _notify_other_components_of_change() -> void:
 
 func _sync_blacklist_ui(changed_avatar_id: String) -> void:
 	if social_data and social_data.address == changed_avatar_id:
-		call_deferred("_update_buttons")
+		call_deferred("_update_blocked_visibility_for_type")
 
 
 func _update_elements_visibility() -> void:
--- a/lib/src/scene_runner/components/avatar_attach.rs
+++ b/lib/src/scene_runner/components/avatar_attach.rs
@@ -35,14 +35,15 @@ pub fn update_avatar_attach(scene: &mut Scene, crdt_state: &mut SceneCrdtState) 
                     // TODO: resolve the current transform
                 }
             } else if let Some(new_value) = new_value {
-                let mut avatar_attach_node = if let Some(avatar_attach_node) = existing {
-                    avatar_attach_node
+                let (mut avatar_attach_node, is_new) = if let Some(avatar_attach_node) = existing {
+                    (avatar_attach_node, false)
                 } else {
-                    godot::tools::load::<PackedScene>(
+                    let node = godot::tools::load::<PackedScene>(
                         "res://src/decentraland_components/avatar_attach.tscn",
                     )
                     .instantiate()
-                    .unwrap()
+                    .unwrap();
+                    (node, true)
                 };
 
                 avatar_attach_node.set(
@@ -52,10 +53,11 @@ pub fn update_avatar_attach(scene: &mut Scene, crdt_state: &mut SceneCrdtState) 
 
                 avatar_attach_node.set("attach_point", &Variant::from(new_value.anchor_point_id));
 
-                avatar_attach_node.set_name("AvatarAttach");
-
-                node_3d.add_child(&avatar_attach_node.clone());
-                avatar_attach_node.call("init", &[]);
+                if is_new {
+                    avatar_attach_node.set_name("AvatarAttach");
+                    node_3d.add_child(&avatar_attach_node.clone());
+                    avatar_attach_node.call("init", &[]);
+                }
             }
         }
     }


regenesis-claw (approved):
Clean, targeted Sentry fixes — both address root causes rather than muting symptoms.

**Rust (avatar_attach.rs):** The `is_new` guard correctly skips `add_child` + `init()` on reuse, eliminating the duplicate parent and duplicate `tree_changed` connection. Straightforward and safe.

**GDScript (social_item.gd):** Routing to `_update_blocked_visibility_for_type` instead of the nonexistent `_update_buttons` — matches the existing `_on_blacklist_changed` pathway as described. One-liner, obvious fix.

CI all green. No iOS-sensitive changes. LGTM.

Requested by Mateo via Slack

PR #1990, opened 2026-05-05
feat: implement z_index for SDK7 UI components

### LINKED ISSUE #556: z-index on UI
Add z-index parameter for the SDK7 UI, so we can draw UI layers accordingly in the client for the player

EibrielInv:
## Summary                                                                                                                                                                                                      
  - Reads the `z_index` field from `UiTransform` (proto field 77, already declared but previously unused)                                                                                                         
  - Implements DCL stacking context semantics: z_index orders siblings within the same parent; a child's z_index cannot escape its parent's context                                                               
  - Purely visual — does not affect pointer/input handling                                                                                                                                                        
                                                                                                                                                                                                                  
  ## How z_index values are assigned in Godot                                                                                                                                                                     
                                                                                                                                                                                                                  
  DCL z_index is `int32` (range ±2 billion), while Godot's `CanvasItem.z_index` is clamped to **-4096 to 4096**. The DCL value is never used directly as a Godot z_index — it is only used as a **sort key**.
                                                                                                                                                                                                                  
  After the layout pass, a recursive DFS traversal visits each entity's children sorted by DCL z_index (ascending) and assigns a monotonically increasing counter as the absolute Godot z_index (`z_as_relative = 
  false`). The counter only increments between siblings, never when descending into children — so all nodes in a subtree share the same z_index as their parent, with Godot's tree order handling within-group    
  rendering. This naturally enforces stacking context containment.     
                                                                                                                                                                                                                  
  SDK UI starts at `RenderingServer.CANVAS_ITEM_Z_MIN + 1000` (-3096) so it always renders behind the app UI, which occupies the default z_index range (0+). The 1000-unit margin reserves space at the bottom for
   scenes using negative z_index values.                                                                                                                                                                                                                                                  
   
## Image

<img width="851" height="823" alt="Captura de pantalla 2026-05-05 a la(s) 18 49 05" src="https://github.com/user-attachments/assets/3e1494c1-80ee-40ec-a7d6-df5c70510a3d" />


  Closes #556

### DIFF
--- a/lib/src/scene_runner/components/ui/scene_ui.rs
+++ b/lib/src/scene_runner/components/ui/scene_ui.rs
@@ -8,7 +8,7 @@ use godot::{
     builtin::math::FloatExt,
     classes::{
         text_server::{JustificationFlag, LineBreakFlag},
-        Node,
+        Node, RenderingServer,
     },
     obj::Gd,
 };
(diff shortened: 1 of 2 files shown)


kuruk-mm (approved):
Great job!

regenesis-claw (commented):
Clean implementation. The DFS traversal with per-parent sorting + monotonically increasing absolute z_index is the right way to enforce stacking context containment.

One observation regarding the thread discussion (SDK UI rendering over game UI): `set_z_as_relative(false)` makes these values absolute across the entire CanvasItem hierarchy. If the game HUD (chat, minimap, notifications) uses lower or similar z_index values, SDK scenes will render on top. This may be intentional (SDK scenes own their UI layer), but if it's not, a simple fix would be to ensure game UI lives on a higher `CanvasLayer` or to start `z_counter` at a negative offset so SDK UI sits below the game HUD by default.

Not blocking — the z_index logic itself is correct and matches the Unity stacking context behavior.

Requested by Mateo via Slack

PR #1991, opened 2026-05-05
chore: use toon shader in compatibility via tonemap_exposure

### LINKED ISSUE #2061: Avatar snapshot renderer flat/dark in Compatibility (GLES3) mode
## Problem

The headless avatar snapshot renderer rendered flat/dark in **Compatibility (GLES3) mode** vs the prod-style look in **Mobile/Vulkan**.

## Root cause

`avatar_preview.tscn` used `Environment.adjustment_brightness = 1.4` to compensate for the toon shader's `EMISSION = ALBEDO * 0.4` floor. That works in Vulkan but fails silently in Compatibility — adjustments are only supported in Forward+ and Mobile rendering methods, not Compatibility ([godotengine/godot#92853](https://github.com/godotengine/godot/issues/92853)). Mobile got the intended +40% boost; Compatibility got 0%. The earlier S3-published reference snapshot was generated by a Vulkan/Mobile build, hence the mismatch.

Second-order issue: `tonemap_exposure` is supported in both renderers, but Vulkan renders to an HDR linear buffer while GLES3 renders to LDR sRGB (`RGBA8`). The same exposure value lands ~1.7× dimmer in Compatibility after the gamma round-trip.

## Out of scope (separate issue)

The backpack / lobby / profile UI in Compatibility mode is also affected by the `adjustment_brightness` not-implemented bug — those screens render dimmer for Compat-using players. PR #1991 scopes the fix to the snapshot tool only; the in-game UI fix needs a separate visual review.

kuruk-mm:
Closes #2061

## Summary

The headless avatar snapshot renderer rendered flat/dark in **Compatibility (GLES3) mode** vs the prod-style look in **Mobile/Vulkan**. After investigation, the toon shader itself is fine in both renderers — the problem was a **renderer-feature gap in `Environment.adjustment_*`** plus an **HDR-vs-LDR framebuffer precision difference** between Vulkan and GLES3. This PR keeps the toon shader and fixes the snapshot renderer to work consistently in both backends, **scoped to the snapshot tool only**.

## Root cause

`avatar_preview.tscn` was using `Environment.adjustment_brightness = 1.4` to compensate for the toon shader's `EMISSION = ALBEDO * 0.4` floor. That works in Vulkan but **fails silently in Compatibility**:

> Adjustments are only supported in the Forward+ and Mobile rendering methods, not Compatibility.

Sources:
- [Adjustments don't work in Compatibility unless glow or 3D scaling is used — godotengine/godot#92853](https://github.com/godotengine/godot/issues/92853)
- [4.x OpenGL Compatibility renderer issues tracker — godotengine/godot#66458](https://github.com/godotengine/godot/issues/66458)
- ["ambient_light_disabled" shader render mode caveats — godotengine/godot#66726](https://github.com/godotengine/godot/issues/66726)
- [GLES3 scene shader exposure path — `drivers/gles3/shaders/scene.glsl#L2345`](https://github.com/godotengine/godot/blob/master/drivers/gles3/shaders/scene.glsl)

So Mobile got the intended +40% boost; Compatibility got 0%. The earlier S3-published reference snapshot was generated by a Vulkan/Mobile build, hence the mismatch.

There is also a second-order issue once `adjustment_brightness` is replaced: **`tonemap_exposure` is supported in both renderers**, but Vulkan renders to an HDR linear buffer while GLES3 renders to an LDR sRGB buffer (`RGBA8`). The same exposure value lands ~1.7× dimmer in Compatibility after the gamma round-trip. Without compensation, Compat ends up dimmer than Mobile by the gamma factor.

## Fix (scoped to the snapshot tool)

`avatar_preview.tscn` is **shared** with the backpack, lobby, profile, `snapshot.gd`, `impostor_capturer.gd`, and several test/dev tools. To avoid changing all of those, the env override is applied **from code inside `avatar_renderer_standalone.gd`'s `start()`**, on a `duplicate()`d Environment so the shared resource is never mutated.

1. **`avatar_preview.tscn` — unchanged.** Still ships with `adjustment_enabled = true` / `adjustment_brightness = 1.4` for in-game use.
2. **`avatar_renderer_standalone.gd`** — at startup, duplicate the Environment, then on the duplicate:
   - `env.adjustment_enabled = false` (kill the no-op for Compat / equivalent for Mobile)
   - `env.tonemap_exposure = 1.4` (renderer-agnostic equivalent — applied per-fragment in GLES3 `scene.glsl:2345`, in the Vulkan tonemap post pass `tonemap.glsl:849`)
   - In Compatibility only (detected via `RenderingServer.get_rendering_device() == null`, which only the GLES3 renderer satisfies): multiply `tonemap_exposure` by **2.8** to compensate for the LDR sRGB framebuffer gamma round-trip. The 2.8× factor was empirically tuned for parity with the Mobile output.
3. **Remove the previous `_apply_pbr_overrides` / `_toon_to_pbr` scaffolding** from the snapshot tool. The toon shader is now used as-is in both renderers, so this hack is no longer needed.

## Validation (wallet `0x481b…07d6`, 256×512 body PNG)

| Variant | mean RGB (R,G,B) | max RGB |
|---|---|---|
| **prod** (S3 reference) | 58.7, 50.2, 65.3 | 255, 255, 255 |
| **mobile-toon** (this PR) | 52.3, 47.4, 60.9 | 198, 255, 253 |
| **compat-toon** (this PR) | 50.3, 45.8, 59.3 | 208, 255, 255 |
| ~~compat-toon, old PBR-override approach~~ | 84.7, 78.1, 91.0 | 255, 255, 255 (over-bright) |
| ~~compat-toon, no fix at all~~ | ~21 | ~131 (too dark) |

Compat is **within ~4%** of Mobile in mean brightness; max RGB matches. Visual side-by-side confirms compat ≈ mobile, both close to prod.

## Why this is better than the previous (PBR-override) approach

The previous commit on this branch (`9aec6606`) walked the avatar tree and replaced each toon `ShaderMaterial` with a reconstructed `StandardMaterial3D`, scoped to the snapshot tool only. That worked but:

- duplicated material logic that the toon shader already handles (alpha clip / blend, double-sided, etc.)
- diverged snapshots from the in-game look (snapshots were PBR; backpack/world preview used toon)
- hid the underlying environmental misconfiguration (`adjustment_brightness` on a Compat path)

The new approach keeps a single rendering path (toon everywhere), fixes the actual environment misconfiguration, and is a few lines instead of a 50-line override walker.

## Out-of-scope (intentional)

- The backpack / lobby / profile UI in **Compatibility mode** is also affected by the `adjustment_brightness` not-implemented bug — those screens render dimmer for Compat-using players. **This PR does not fix that** by design (different visual review needed). The fix here is scoped to the snapshot tool only; the shared `avatar_preview.tscn` is untouched.

## Test plan

- [ ] CI builds Linux + exports + runs avatar snapshot test
- [ ] Compare images with snapshots: reference snapshots in `tests/snapshots/avatar-image-generation/` were last refreshed in #1438 with the toon look in **Mobile**; the new Compat output should match closely. If the diff fails, regenerate.
- [ ] Manually run `scripts/local_profile_snapshot.sh --compatibility 0x481b...07d6` and `--mobile 0x481b...07d6` and compare against `avatars-output*`.
- [ ] Sanity-check backpack / lobby / profile look unchanged in both renderers (no shared-scene drift — `avatar_preview.tscn` is untouched).
- [ ] `publish-docker` label is set, so on merge/CI the Docker image will be tagged `<sha>` + `next` and pushed to `quay.io/decentraland/godot-explorer`. On merge to `main`, it gets the `staging` tag.

## Triggering the Docker image build → quay.io

This PR has the `publish-docker` label. The flow:

1. `runner.yml` → `linux-build` → `docker_build_test.yml` (workflow_call)
2. `docker_build_test.yml` runs `buildah-build` with tags `${github.sha}` and `next`
3. `redhat-actions/push-to-registry@v2` pushes to `quay.io/decentraland/godot-explorer` because `inputs.publish_docker` is true (or `github.ref == 'refs/heads/release'`)
4. After this PR merges, the image picks up the `staging` tag automatically


regenesis-claw (commented):
## Review: Snapshot renderer fix + tooling

Nice work — the root-cause analysis of the `adjustment_brightness` vs `tonemap_exposure` Compat gap is thorough, and the scoped fix (duplicate env, snapshot-only) is the right call. The new scripts and `PROFILE_IMAGE.md` doc are a strong addition for reproducibility.

Followed `REVIEW.md` conventions. All CI checks pass ✅. No iOS build needed (no platform-sensitive iOS changes).

### ⚠️ Prod code *does* change — but safely

The PR title says "no prod code changes" — that's not strictly true. Here's what touches prod paths:

| File | What changed | Prod impact |
|---|---|---|
| `global.gd` | `--dclenv` CLI precedence + `disable_telemetry` gating | Telemetry gate is a no-op unless `disable_telemetry` feature is compiled in (mobile/release builds don't have it). `--dclenv` precedence is new code that doesn't activate without the flag — safe. |
| `project_main_loop.gd` | Early return on `is_telemetry_disabled()` | Same — only fires in CI Linux builds with the feature enabled. |
| `avatar_preview.gd` | New `ssaa` param on `async_get_viewport_image()` | Defaults to `1` — all existing callers (backpack, lobby, profile) are unaffected. Backward compatible ✅ |
| `dcl_cli.rs` | `--key=value` parsing + `--dclenv` arg | The `=` form parsing is additive — previously `--key=value` was silently treated as a flag with no value. No regression risk for existing args. |
| `dcl_global.rs` | New `is_telemetry_disabled()` func | Static, no side effects |
| `linux_builds.yml` | `disable_telemetry` feature added to CI Linux build | CI-only; Android/iOS/release unaffected |

**Verdict: all prod changes are either gated behind a compile-time feature not present in mobile/release builds, or backward-compatible with safe defaults.** The claim in the title is *mostly* right in spirit, but worth being precise about it.

### Notes

- The `SSAA_FACTOR = 2` combined with `scaling_3d_scale = 2.0` means the snapshot renders at 4× internally before Lanczos downsample — that's a quality-over-performance tradeoff that's fine for an offline snapshot tool but would be brutal in realtime. The scoping to `avatar_renderer_standalone.gd` keeps it safe.
- The 2.8× Compat gamma compensation is empirically tuned — might be worth a comment noting what prod/reference images it was validated against (the PR body has it, but a one-liner in-code helps future maintainers).
- `scripts/data/issue-1880-entities.txt` (830 CIDs) is a large data file — reasonable for issue reproduction but worth noting it'll live in the repo permanently.

No blockers — looks good to merge once the title/description accurately reflects the small prod-touching changes.

Requested by Mateo via Slack

Issue #1985, opened 2026-05-05
[Bug][Sentry] Wearables dict: invalid access for unknown URN (e.g. 'schoolshoes') (1TR)


kuruk-mm:
## Sentry
[GODOT-EXPLORER-1TR](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1TR) — `SCRIPT ERROR: Invalid access to property or key 'urn:decentraland:off-chain:base-avatars:schoolshoes' on a base object of type 'Dictionary'.`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | 1 | 52 |
| 0.58.0-1c2b5f1 | 1 | 15 |

## Root cause (suspected)
`schoolshoes` isn't a valid base avatar URN. Some user profile contains this malformed URN, and our wearable dictionary lookup uses direct key access (`dict[urn]`) instead of `.get(urn)`. The script error fires on every avatar update for that user.

## Suggested fix
1. Audit wearable dictionary access in `godot/src/decentraland_components/avatar/` and replace `dict[urn]` with `dict.get(urn)` or a bounds check.
2. Filter invalid URNs at profile load time.

Issue #1979, opened 2026-05-05
[Bug][Sentry] StreamProtocol generator yield/dropped errors (35/26 users v0.65)


kuruk-mm:
## Sentry
- [GODOT-EXPLORER-1B3](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1B3) — `StreamProtocol > to_generator > Generator > error r#yield, probably it's closed`
- [GODOT-EXPLORER-1B2](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1B2) — `StreamProtocol > process_messages > Error on sending through the Generator, seems to be dropped`

## Impact (last 30d, prod)
| Release | Users (1B3 / 1B2) | Events |
|---|---|---|
| 0.65.0-e5ada5c | 35 / 26 | 108 / 90 |
| 0.64.0-e26741f | 39 / 31 | 161 / 138 |

Persistent. Likely RPC stream cleanup noise.

## Root cause (suspected)
When an RPC stream consumer is dropped (e.g. user closes UI or signs out), the StreamProtocol generator side logs an error trying to yield to a closed receiver. This is part of the normal teardown path.

## Suggested fix
- Demote both messages to `tracing::debug!` since they're expected on stream close.
- Add explicit close-detection to bail early without erroring.

Issue #1975, opened 2026-05-05
[Bug][Sentry] WebSocket disconnect noise: ClientMessagesHandler Closed/Reset (115 users)


kuruk-mm:
## Sentry
- [GODOT-EXPLORER-1AC](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1AC) — `ClientMessagesHandler > process > Error on receive Closed`
- [GODOT-EXPLORER-1AD](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-1AD) — `ClientMessagesHandler > process > Error on receive Internal(Protocol(ResetWithoutClosingHandshake))`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | 115 / 96 | 319 / 251 |
| 0.64.0-e26741f | 101 / 89 | 248 / 211 |

Long-running websocket noise — almost certainly normal disconnect path being logged at ERROR.

## Root cause (suspected)
Normal websocket close sequence is being logged as ERROR. `Closed` and `ResetWithoutClosingHandshake` are expected when the comms server closes a connection (idle timeout, client backgrounded, server-side reload).

## Suggested fix
- Demote to `tracing::debug!` or `tracing::info!` for clean closes.
- Only log ERROR for unexpected protocol errors.

Issue #1972, opened 2026-05-05
[Bug][Sentry] Mouse is not supported by this display server (122 users v0.65)


kuruk-mm:
## Sentry
[GODOT-EXPLORER-6E](https://dcl-regenesis-labs.sentry.io/issues/GODOT-EXPLORER-6E) — `ERROR: Mouse is not supported by this display server.`

## Impact (last 30d, prod)
| Release | Users | Events |
|---|---|---|
| 0.65.0-e5ada5c | 122 | 474 |
| 0.64.0-e26741f | 252 | 970 |
| 0.64.0-b0532c7 | 79 | 343 |

Persistent across versions.

## Root cause (suspected)
GDScript or Rust code is calling `Input.warp_mouse()`, `Input.set_mouse_mode()`, or similar on mobile where the display server doesn't support mouse. Engine emits an error but the code path is desktop-only.

## Suggested fix
1. Wrap mouse-related calls in `if DisplayServer.has_feature(DisplayServer.FEATURE_MOUSE):` or the equivalent check.
2. Probably 1-2 callsites doing `Input.set_mouse_mode(...)` or warp without platform guard.

PR #1999, opened 2026-05-07
fix: easy Sentry triage cluster (5 fixes)

kuruk-mm:
## Summary

Sentry-label triage: fixing the 8 issues classified as **easy (0.5d)** in the DAO Explorers project. 5 land as code changes here; 3 are already resolved on `main` and just need closing.

## Issues fixed in this PR

| # | Issue | Fix |
|---|---|---|
| #1958 | placeholder_image Dictionary warning (188u) | Pre-seed `placeholder_image` with nil on the Rust-side `GLTFState` in `lib/src/content/gltf/common.rs` so `custom_gltf_importer.gd`'s preflight read never hits a missing key. Earlier `has_additional_data()` guard was reverted because it regressed the Docker avatar-image-generation snapshot path; this root-cause fix avoids the guard entirely while leaving the placeholder branch usable by external tooling that overwrites the key with a truthy value |
| #1972 | Mouse not supported on display server (122u) | Wrap every `Input.set_mouse_mode(...)` callsite with `DisplayServer.has_feature(FEATURE_MOUSE)` — `global.gd`, `explorer.gd`, `navbar.gd`, `camera.gd` |
| #1979 | StreamProtocol yield/dropped (35u) | Add `"StreamProtocol"` to `NOISE_PATTERNS` in `project_main_loop.gd` — these come from `dcl-rpc` on stream-consumer drop and are expected teardown |
| #1984 | `ia_any` InputMap action doesn't exist | `tooltip_label.gd`: skip `Input.action_press/release` for the virtual `ia_any` action in the mobile touch path (`mobile_on_panel_container_gui_input`) |
| #1985 | Wearables dict invalid URN (e.g. 'schoolshoes') | `backpack.gd`: switch `wearable_data[urn]` to `.get(urn)` + null guard in `_on_wearable_equip`, `_on_wearable_unequip`, and the unequip-by-category loop. Malformed URNs from user profiles no longer raise per-frame script errors |

## Issues already fixed on main — close as resolved

| # | Issue | Where it was fixed |
|---|---|---|
| #1971 | Chat panel `%PanelContainer_Navbar` / `%TextureRect_Logo` not found (195u) | Chat rework #1894 (`e4a4f2d2`) — `chat.gd` no longer has those `@onready` lookups. v0.65 still hits it; resolves on next release |
| #1975 | WebSocket disconnect noise: ClientMessagesHandler Closed/Reset (115u) | Already in `NOISE_PATTERNS` from #1931 (`ad62c3e7`) |
| #1986 | `social_item.gd` dead `_update_buttons` call_deferred (1SX) | Already fixed in #1987 (`adfbafcf`) — `_sync_blacklist_ui` now routes to `_update_blocked_visibility_for_type` |

## Notes on the #1958 fix

The `has_additional_data()` guard approach (commits `72471c577`, `8f7bec2b7`) was reverted in `f474afe0c` because it broke the actual system. Instead, this PR seeds `placeholder_image` to `Variant::nil()` from the Rust GLTF loader. Behavior is identical because `if placeholder_image:` stays falsy on nil and the else branch (base_path/mappings remap) runs as before. The avatar-image-generation tooling that wants the placeholder branch still works because it overwrites the seeded nil with a truthy value upstream of `append_from_file`.

## Test plan

- [x] gdformat / gdlint pass (verified locally)
- [x] `cargo check` passes
- [ ] Open backpack with a profile containing an unknown wearable URN — no per-frame "Invalid access to property or key" script error
- [ ] Hover/tap a tooltip whose action is `ia_any` on mobile — no `InputMap action 'ia_any' doesn't exist` log
- [ ] Run on mobile (Android/iOS) — no `Mouse is not supported by this display server` log when capturing/releasing mouse
- [ ] Load a scene with a GLTF that doesn't set `placeholder_image` — no Dictionary `placeholder_image` engine warning
- [ ] Verify Docker avatar-image-generation snapshots still render wearables correctly (mappings path unchanged + placeholder branch still reachable via external truthy overwrite)
- [ ] Sign out / disconnect comms — no `StreamProtocol > to_generator` ERROR-level logs reach Sentry


godot/src/logic/custom_gltf_importer.gd
```diff
@@ -2,7 +2,11 @@ extends GLTFDocumentExtension
 
 
 func _import_preflight(state: GLTFState, _extensions: PackedStringArray) -> Error:
-	var placeholder_image = state.get_additional_data("placeholder_image")
+	# Guard with has_additional_data() — bare get_additional_data() emits a
+	# Dictionary engine warning when the key is missing, captured by Sentry.
+	var placeholder_image: Variant = null
+	if state.has_additional_data("placeholder_image"):
```
regenesis-claw: Minor: the PR description mentions guarding `base_path` and `mappings` too, but only `placeholder_image` got the `has_additional_data()` treatment. `base_path` has a `!= null` check on line 29 which covers it, but `mappings` on line 31 could still emit the Dictionary warning if `base_path` is set without `mappings`. Probably never happens in practice — just flagging the description/code mismatch. Non-blocking.

regenesis-claw (approved):
Nice batch — five targeted Sentry fixes, all defensive and minimal. Each one is the obvious correct fix for its issue. ✅

**One pre-existing bug spotted while reviewing** (`global.gd:release_mouse()`):

The fallback branch (when `explorer` is not valid) sets `MOUSE_MODE_CAPTURED` instead of `MOUSE_MODE_VISIBLE`:
```gdscript
func release_mouse():
    ...
    elif DisplayServer.has_feature(DisplayServer.FEATURE_MOUSE):
        Input.set_mouse_mode(Input.MOUSE_MODE_CAPTURED)  # should be VISIBLE?
```
This was already there before this PR (your diff only added the `DisplayServer` guard), so not blocking — but worth a follow-up fix.

**CI note:** The 🐳 Docker snapshot failure is pre-existing — the log shows `ResourceLoader failed to load!` which is exactly the bug #2000 fixes. Not related to this PR's changes.

**iOS build:** ✅ Green.

LGTM — approve.

Requested by Mateo via Slack

PR #2010, opened 2026-05-11
fix: audio->global is not being set

### LINKED ISSUE #1761: [Bug]: AudioSource with global:true + loop:true broken on mobile
### Platform
Mobile (iOS and Android)

### App Version
Not specified

### Device Information
Not specified

### Issue Description
Audio sources with both `global: true` and `loop: true` do not work correctly on the mobile client. The audio either fails to play entirely, or plays but is incorrectly spatialized at the entity's Transform position (e.g., only audible near 0,0,0) instead of playing globally as expected.

**Additional findings:**
- `global: true` + `loop: false` works correctly and plays globally on mobile
- Even with `loop: false`, `global: true` audio appears to be spatialized at the entity's Transform position rather than truly non-spatialized
- Multiple global audio sources may conflict - only one global audio source can play at a time (needs confirmation)
- This is problematic for background music use cases

**Scene:** flagtag.dcl.eth

### Expected Behavior
Audio with `global: true` should:
1. Play continuously when `loop: true` is set
2. Be non-spatialized (audible equally from any position in the scene)
3. Work identically to desktop client behavior
4. Allow multiple global audio sources to play simultaneously

### Steps to Reproduce
1. Create an entity with `AudioSource { global: true, loop: true, playing: true }`
2. Open the scene on mobile client
3. Walk away from the entity's position (e.g., away from 0,0,0)

**Observed on mobile:**
- Audio either doesn't play at all, OR
- Audio plays but is spatialized at the entity's Transform position instead of globally

### Occurrence
100% - consistently reproducible

**Workaround:** Use `loop: false` with manual re-trigger on a timer, and move the audio entity to the player's position every frame.

---

leanmendoza:
fix #1761 

also it fixes loading wav files!

### DIFF
--- a/lib/src/content/audio.rs
+++ b/lib/src/content/audio.rs
@@ -44,9 +44,12 @@ pub async fn load_audio(
     let bytes = PackedByteArray::from_vec(&bytes_vec);
     let audio_stream: Option<Gd<AudioStream>> = match extension.as_str() {
         ".wav" => {
-            let mut audio_stream = AudioStreamWav::new_gd();
-            audio_stream.set_data(&bytes);
-            Some(audio_stream.upcast())
+            // AudioStreamWav::set_data expects RAW PCM samples with the format
+            // properties (mix_rate / stereo / format) configured manually. To
+            // play a full RIFF/WAVE container we must parse it via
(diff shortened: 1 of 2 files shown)


kuruk-mm (approved):
greattt

PR #2018, opened 2026-05-11
chore: setup iOS Swift library

leanmendoza:
Initial setup for having the Swift library


kuruk-mm (approved):
Looking great for initial setup.

Can we add step by step about how to build the lib? Can we also add the build in plugins/build_ios.sh?

For future reference. Try to use the wrapper when you implement the calls using the DclGodotiOS in Rust

regenesis-claw (commented):
Nice initial setup — clean separation between the real Swift framework and the desktop stub, and the GDExtension wiring looks correct. A few observations:

### 🟡 Suggestions

1. **SwiftGodot pinned to a branch, not a tag** (`Package.swift:19`)
   `branch: "barebone-split"` is a moving target — if the upstream branch is force-pushed or deleted, builds will break without any change on your side. Consider pinning to a specific commit hash (`.package(url: ..., revision: "<sha>")`) once you've confirmed which commit works. A tag would be even better if one becomes available.

2. **Committed binary artifacts** — `DclSwiftLib.framework/DclSwiftLib` (binary) and `libdcl_swift_lib_stub.macos.arm64.dylib` are checked into git. This is fine for bootstrapping, but these will grow the repo history permanently. Worth considering whether CI should build and cache these instead, especially once the framework gains real StoreKit functionality and the binary size increases.

3. **Smoke test runs on every launch** (`global.gd:410-423`)
   The `_dcl_swift_lib_smoke_test()` call runs unconditionally in `_ready()`. The TODO comment says to remove it once StoreKit lands — just flagging so it doesn't get forgotten. For now it's fine since it early-returns on non-iOS and the overhead is negligible.

4. **Missing Linux/Windows stub entries** in `dcl_swift_lib.gdextension` — only `ios` and `macos` are listed. On Linux/Windows, Godot may log a warning about the extension having no matching library for the platform. If you see those warnings in CI logs or dev builds, a similar no-op stub (or just an empty shared lib) would silence them.

### ℹ️ Notes

- Mateo's comment about adding build steps to `plugins/build_ios.sh` and documenting the step-by-step is worth doing before this merges — the `build_swift_lib.sh` and `build_ios_swift.sh` scripts exist but aren't referenced from the main iOS build pipeline yet.
- Android and iOS CI builds are still in progress. The sticky build report shows Android as failed, but that may be from an earlier run — worth confirming once the current run finishes.
- iOS build was triggered (the `build-ios` label is present) — good, since this PR directly touches iOS native plugins.

Overall this is a solid scaffold. The SwiftGodot integration pattern (custom entry symbol to avoid clashes, RefCounted class, stub for non-iOS) is well thought out.

Requested by Mateo via Slack

regenesis-claw (commented):
Excellent work on the feedback! All the key items have been addressed:

## ✅ Changes Applied

### From my review:
- **SwiftGodot pinned to commit SHA** (`Package.swift`) — now uses `revision: "f60a71fd22f932f3eed2626e2282386f9ce7d14a"` with a clear comment explaining reproducibility. Perfect fix.
- **Smoke test cleanup** (`global.gd`) — now uses `DclSwiftLibPlugin.is_available()` instead of raw `ClassDB.class_exists()`, much cleaner and consistent with the new wrapper pattern.

### From Mateo's review:
- **Build documentation** — comprehensive `README.md` added with step-by-step instructions, layout overview, requirements, usage examples, and even SwiftGodot update procedure. Well done.
- **Integration with iOS build** (`build_ios.sh`) — now calls `build_swift_lib.sh` automatically at the end of the iOS build pipeline.
- **Rust wrapper** (`lib/src/godot_classes/dcl_swift_lib.rs`) — new `DclSwiftLibPlugin` mirrors the `DclIosPlugin` pattern perfectly: typed wrapper, `is_available()` check, clean fallback for non-iOS. Exactly the right architecture.

## 📊 CI Status

All checks passing (Android still pending, but iOS build ✅ passed in 15m24s, which is the critical one for this PR).

## 🟡 Minor notes (non-blocking)

The two unaddressed items from my original review are both non-functional suggestions:
- **Binary artifacts in git** — acknowledged as fine for bootstrapping; can move to CI later if framework size grows
- **Linux/Windows stub entries** in `.gdextension` — would only silence console warnings; desktop stub already works for macOS

Neither blocks merge. This is ready to ship.

Requested by Lean via Slack

PR #2019, opened 2026-05-11
fix: lock camera-parented scene entities to render rate

### LINKED ISSUE #2020: Camera-parented scene entities visibly trail the camera on high-refresh displays
## Description

Decentraland scenes can attach entities (e.g. a `GltfContainer`) to `engine.CameraEntity` with a static local offset to create first-person viewmodels — held tools, weapons, HUD-like 3D overlays, etc.

In the current explorer build, these entities **do not stay locked to the camera**. When the player rotates or moves the camera quickly, the attached entity visibly trails the camera and "swims" behind it before snapping into place. The effect is most noticeable on displays running above 60 Hz and during fast mouse-look.

## Expected behaviour

An entity parented to `engine.CameraEntity` with a static local `Transform` should appear rigidly fixed relative to the camera — the same way a viewmodel is locked to the view in a typical first-person game. There should be no visible lag, sway, or snapping at any camera rotation speed or display refresh rate.

## Actual behaviour

The entity trails the camera by a perceptible amount and visibly "catches up" in steps. The faster the camera moves and the higher the display refresh rate, the more pronounced the trailing.

## Reproduction steps

1. Run the explorer on a display with refresh rate > 60 Hz (e.g. 120 / 144 / 240 Hz).
2. Load a scene that attaches a visible entity to `engine.CameraEntity` with a static local offset. For example:
   ```ts
   const item = engine.addEntity()
   Transform.create(item, {
     parent: engine.CameraEntity,
     position: Vector3.create(0.35, -0.2, 0.6),
     rotation: Quaternion.fromEulerDegrees(195, 240, 0),
     scale: Vector3.create(0.25, 0.25, 0.25)
   })
   GltfContainer.create(item, { src: 'assets/scene/items/hook.glb' })
   ```
   (The hackathon-scene's `src/factories/heldItem.ts` is one real example.)
3. Enter first-person camera mode.
4. Move the mouse rapidly back and forth to rotate the camera.
5. Observe the held item visibly trailing the camera rotation and snapping back.

For a clearer signal, lock the physics tick to a low value (e.g. 30 Hz via `Engine.physics_ticks_per_second`) while keeping the display at the native refresh rate — the trailing becomes very obvious.

## Affected use cases

- First-person held items / viewmodels.
- Any HUD-style 3D overlay parented to `engine.CameraEntity`.
- Camera-attached gizmos used in scene tooling.

## Not affected

- Entities that read `Transform.get(engine.CameraEntity)` from scene JS and recompute their own position each scene tick — those have a separate (data-side) lag profile tied to scene tick rate, not display rate.

kuruk-mm:
Closes #2020

## Summary

Camera-parented DCL entities (held viewmodels, first-person overlays — anything parented to `engine.CameraEntity`) were visibly trailing the rendering camera by up to one physics tick on high-refresh displays.

### Root cause

`engine.CameraEntity` is not the real `Camera3D` — it's a proxy `Node3D` inside each scene's subtree (created in `godot_dcl_scene.rs`). The real camera updates every render frame (mouse-look is applied inside `_input`), but the proxy was only being re-synced from inside `SceneManager::physics_process` at the physics tick rate (60 Hz). On a 120/144/240 Hz display, the proxy stayed at its last position for several render frames, and any GltfContainer parented to it visibly "swam" behind the camera.

### Fix

Move the `set_global_transform` block for `SceneEntityId::PLAYER` and `SceneEntityId::CAMERA` (current parcel scene + all global scenes) out of `physics_process` and into a new `process` callback on `SceneManager`. That makes the proxy track the real camera at display rate, so camera-parented entities stay glued instead of trailing.

The CRDT-side Transform write in `update_scene.rs:441-446` (what scene JS reads via `Transform.get(engine.CameraEntity)`) is intentionally untouched — that path stays at physics rate where the scene tick samples it.

## Test plan

- [ ] Run the client on a high-refresh display (>60 Hz) and move the mouse around quickly in first-person — held viewmodel should stay perfectly locked to the camera, no visible trailing.
- [ ] Verify the same on a 60 Hz display (no regression).
- [ ] Verify portable/global scenes that attach to `CameraEntity` still follow the camera.
- [ ] Verify scenes that read `Transform.get(engine.CameraEntity)` in JS (raycasts, projectile spawn, etc.) still get fresh values — physics-rate CRDT write is unchanged.


regenesis-claw (approved):
Clean, well-scoped fix. Moving the proxy camera/player transform sync from `physics_process` to `process` is the textbook correct approach — camera-parented entities need to track at display rate, not physics tick rate. The CRDT-side write staying in physics_process is the right call since scene JS samples it on the scene tick anyway.

The logic is identical before and after the move — just lifted into its own `process` callback. No new allocations, no new failure modes.

All CI checks green, iOS build present and green. LGTM.

Reviewed following repo REVIEW.md guidelines.

PR #2022, opened 2026-05-11
fix: prevent travel modal from becoming unreachable after repeated use

### LINKED ISSUE #2014: [Bug] Portal Manager Breaking in Plaza After Entering/Exiting Portal Several Times
### Platform
iOS

### App Version
466

### Device Information
iOS - iPhone 16E

### Issue Description
The Portal Manager breaks in Genesis Plaza after entering and exiting portals multiple times. The exact manifestation of the breakage is not yet documented.

### Expected Behavior
The Portal Manager should continue functioning correctly regardless of how many times a user enters and exits portals.

### Steps to Reproduce
1. Navigate to Genesis Plaza
2. Enter and exit a portal multiple times
3. Observe Portal Manager functionality breaking

### Occurrence
Not specified

---
Requested by Rodri via Slack

sdilauro:
## Summary
- Fixes a race condition in `ModalManager` where the old travel modal's deferred `tree_exited` signal would nullify the reference to a newly created modal, making it unreachable
- Disconnects `tree_exited`, `closed`, and `jump_in_pressed` signals before `queue_free` to prevent stale callbacks
- Removes old canvas layers from the tree immediately (`remove_child`) before `queue_free` to avoid two layers coexisting

Closes #2014

## Test plan
- [ ] Navigate to Genesis Plaza, enter a portal zone so the travel modal appears
- [ ] Close the modal (X button or tap background), re-enter the zone — modal should appear again
- [ ] Repeat enter/exit multiple times — modal should always reappear
- [ ] Click "Jump In", teleport, return, and trigger the modal again — should work

### DIFF
--- a/godot/src/ui/components/modal/modal_manager.gd
+++ b/godot/src/ui/components/modal/modal_manager.gd
@@ -362,6 +362,7 @@ func _async_create_modal() -> Modal:
 
 	# Wrap in a CanvasLayer with high layer to ensure modals render above all overlays
 	if _canvas_layer and is_instance_valid(_canvas_layer):
+		_canvas_layer.get_parent().remove_child(_canvas_layer)
 		_canvas_layer.queue_free()
 
 	_canvas_layer = CanvasLayer.new()
@@ -405,6 +406,7 @@ func _async_create_travel_modal() -> TravelModal:
 		return null
 
 	if _travel_canvas_layer and is_instance_valid(_travel_canvas_layer):
+		_travel_canvas_layer.get_parent().remove_child(_travel_canvas_layer)
 		_travel_canvas_layer.queue_free()
 
 	_travel_canvas_layer = CanvasLayer.new()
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
Clean, focused fix for the race condition. The approach is solid:

- `remove_child` before `queue_free` ensures no two canvas layers coexist during the deferred-free window
- Disconnecting `tree_exited`, `closed`, and `jump_in_pressed` before `queue_free` prevents the stale callback from nullifying the new modal reference

One observation: `_disconnect_travel_signals()` uses the `get_connections()` loop to disconnect *all* listeners, while `_remove_modal` does a targeted `is_connected` check for `tree_exited`. Both are fine in practice since only the manager connects to these signals — just noting the slight asymmetry.

LGTM — approving. CI still running (Linux, Android, iOS pending); no platform-sensitive native code here, just GDScript UI logic.

Requested by Seba via Slack

kuruk-mm (approved):

PR #1992, opened 2026-05-06
spike: GP profiling benchmark (#1862) — bench-infra base PR for the perf series

manuelmaceira:
## What

Base PR for the perf-optimization stack. Holds **every modification made to the benchmark machinery** — runner, deeplink param routing, profile forcing, DG/HW-bench disable during warmup, visibility-grid as opt-in tool, profiling helpers — with minimal client-behavior changes (see "Behavior changes" section below).

Downstream PR #2025 stacks on this branch and cites its bench A/B against the baseline numbers below.

## Why

A perf series needs a stable measurement environment:
- Without `force-graphic-profile`, every fresh install runs HW-Benchmark and picks a profile based on a one-time hardware probe — A/B noise dominates signal across installs.
- Without DG/HW-bench disable during warmup, dynamic FPS caps (`thermal_fps_cap`, `param_changed` re-application) clamp the bench output.
- Without deeplink param routing, every A/B requires a full rebuild/install cycle.

This PR centralizes all of that.

## Baseline (A54, Mali-G68, GP preview ea9cc738 — current catalyst-deployed state)

True pre-perf baseline: this branch + revert of #2036 (shadow tweaks) + revert of #2038/#2039 (subviewports), running against the **currently-deployed Genesis Plaza (ea9cc738)**.

| metric                | value      | notes |
|---|---:|---|
| fps mean              | **13.88**  | HIGH profile, GP preview, position 0,0, 90s warmup |
| render_gpu_ms mean    | **67.27 ms** | Mali-G68 fragment-bound |
| render_cpu_ms mean    | ~7 ms      | small |
| primitives mean       | **2,967,296** | LOD0 only, no LOD chain |
| draw_calls mean       | 1,970      | |
| render_objects_in_frame | 2,104    | |
| texture_mem_mb        | 611        | Godot-tracked |
| video_mem_mb          | 762        | Godot-tracked |
| resource_count        | 5,861      | |

JSON: `bench-results/newgp-truebaseline-no2036-no2038.json` (attached).

**Branch used for the baseline measurement**: `baseline/bench-infra-pre-perf` (= this PR + reverts of #2036 / #2038 / #2039). Build APK from that branch to reproduce.

## Reproducing

```bash
# 1. Spin up GP preview (pinned via godot/bench/genesis_plaza.config.json → ea9cc738)
scripts/bench/launch_devices.sh --android --gp-benchmark --preview http://<HOST>:8000 --position 0,0

# 2. Pull results from the device
[...]

PR #2025, opened 2026-05-11
perf(rendering): mobile GPU optimization stack

### LINKED ISSUE #2064: Stop blocking the render thread on avatar impostor saves and tween updates
## Context

On Samsung A54 (Mali-G68, Genesis Plaza) the render pipeline is bound by two CPU hotspots that punch through the frame budget independently of GPU work:

1. **Avatar impostor PNG encode runs on the render thread.** `Image::save_png` shows 13.8% inclusive on VkThread, with `longest_match` (zlib hot loop) at 6.4% self-time. Each impostor capture (256×512 RGBA) hits the synchronous encode path directly, pinning A54 to ~14 FPS regardless of GPU work.

2. **`TweenState` is emitted to JS every frame.** The `TsActive` component does a `put` every tick as a keep-alive. On the GP benchmark trace that means **15162 dirty entries/frame** just for Tween updates, flooding the V8↔Rust CRDT pipeline.

## Experiment

- **Async PNG**: snapshot the RGBA8 mip0 on the main thread before `generate_mipmaps`, hand bytes off to a tokio blocking worker that does encode + atomic write (`.tmp` → rename) so a partial crash never leaves a half-written PNG on disk.
- **Emit-on-transition**: track `last_emitted_state` per Tween; only `put()` when the state transitions. For active tweens this collapses to 1 put on start + 1 on completion.

## Hypothesis

VkThread reclaims time previously stolen by zlib encode; V8 stops processing useless dirty entries → +2-3 FPS on A54 HIGH.

## Success criteria

- `frame_proc_ms` drops by ≥10% with no visual regressions.
- Tween dirty entries/frame down by ≥90%.
- No half-written PNGs on crash.

Tracked by #2054.

### LINKED ISSUE #2065: Reduce V8 garbage collection caused by CRDT messages from the renderer
## Context

Profile on Samsung A54 (Genesis Plaza) shows Thread-14 (V8 isolate for the GP scene) at **31.9% of total CPU**, with:

- **24.1%** in `Builtin_ArrayBufferConstructor`
- **20.6%** in GC scavenge

The bottleneck isn't compute — V8 is thrashing the BackingStore allocator on every CRDT recv tick. Each inner `Vec<u8>` in the payload returned by `op_crdt_recv_from_renderer` causes a fresh BackingStore in V8, all of which become garbage on the next frame.

## Experiment

Replace `op_crdt_recv_from_renderer` (serde → `Vec<Vec<u8>>`) with two ops:

- `op_crdt_recv_wait` (async): waits on the renderer channel, builds one framed `Vec<u8>` payload `[u32 LE main_crdt_len][main_crdt bytes][data bytes]`, stashes it in op_state.
- `op_crdt_recv_drain` (sync `#[fast]`): copies the stashed bytes into the JS-owned `#[buffer] &mut [u8]`. Returns 0 if the buffer is too small (caller grows + retries).

JS side keeps a persistent 64 KB buffer that grows on demand. `decodeRecv` reads the length prefix to split off `main_crdt` and `data` without allocating new `Uint8Array` backings.

## Hypothesis

0 new BackingStores per tick after warmup → drastically less GC pressure → V8 thread frees up CPU → higher FPS in scene-runner-bound scenes.

## Success criteria

- ArrayBufferConstructor % on the V8 thread drops near zero.
- GC scavenge % on the V8 thread drops near zero.
- FPS on GP HIGH up by ≥15%.

Tracked by #2053.

### LINKED ISSUE #2067: Replace per-container load-timeout Timers with a single global tracker
## Context

Genesis Plaza loads ~1419 `GltfContainer` instances. Each one spawns its own Godot `Timer` node for the 120s load-timeout. That produces:

- 1419 `_process` callbacks per frame against the scene tree's idle process list.
- Each `Timer` is a full `Node` with internal time-accumulation logic (which runs every frame, even when stopped).
- Zero useful work in 99.99% of those frames.

## Experiment / Requirement

Replace the per-container Timers with a single `GltfLoadTimeoutCoalescer` node holding one `Dictionary` keyed by `container_instance_id → deadline_ms`. Per frame: walk the dict and call `_on_load_timeout()` on any container whose deadline passed.

## Constraint discovered during bisect

Adding `_process` directly to the `DclGlobal` autoload reproducibly triggers a VkThread NULL deref inside `libgodot_android.so` on A54. The coalescer **must** live as a child Node of Global, not on the autoload itself.

## Success criteria

- One Dictionary walk/frame instead of 1419 callbacks/frame.
- Identical load-timeout behavior to the existing implementation.
- No new crashes on A54.

Tracked by #2050.

### LINKED ISSUE #2070: Skip rendering scene geometry hidden behind buildings
## Context

DCL scenes are open-world and have no built-in occlusion structure (no BSP, no portals, no author-supplied occluders). Godot handles frustum culling but **does nothing about "geometry behind that building"**. On Genesis Plaza ~30–40% of geometry per frame is occluded by larger buildings yet still rasterized.

We need an occlusion mechanism that:
1. Works without authored metadata.
2. Has near-zero per-MI runtime cost.
3. Is precomputable offline in a future phase.

## Experiment

Cell-based PVS culler:

**Build (one-time, post scene load, GDScript)**
- Walk the scene tree, bucket every static `MeshInstance3D` into a uniform 16×16m grid by its world AABB.
- Collect `CollisionShape3D` AABBs as blockers (>8m diagonal).
- Bake an O(n²) PVS bitmap by raycasting cell-pair midpoints against blockers.

**Per-frame (Rust)**
- For each cell: distance test → screen-space-size threshold → frustum AABB test → PVS bit lookup against the camera's cell.
- On a 0↔1 crossing: `RenderingServer.instance_set_visible(rid, bool)` (bypasses the scene-tree cascade of `mi.visible = bool`).
- Temporal hysteresis (60-frame hide-streak) prevents popping when crossing boundaries.

## Requirements

- The hot path **must** be in Rust — an earlier GDScript prototype cost 5-7 ms/frame.
- Default ON; opt-out via deeplink `visibility-grid=false`.
- Must not alter the final image (beyond correctly hiding occluded geometry).

## Follow-up (not in this PR)

Move work offline into an asset preprocessor:
1. PVS bake per GLTF → sidecar `.vgrid-blockers.bin`.
2. Per-GLTF MI AABBs → sidecar.
3. Cell layout snapped to the DCL 16×16m parcel grid.

What **can't** move offline: world-space transforms (DCL composes dynamically via CRDT), per-MI RID assignment, the per-frame visibility loop itself.

## Success criteria (A54 HIGH, GP preview, warm cache)

- `fps.mean` ≥ +50%
- `render_gpu_ms` ≥ −35%
- `primitives` ≥ −35%
- Accept tradeoff: increased VRAM and `load_seconds` until the bake moves offline.

Tracked by #2037.

### LINKED ISSUE #2073: Cheap-PBR mode: switch materials to per-vertex shading at import time
## Context

On Forward Mobile + Mali-G68 (A54) the bottleneck on Genesis Plaza is **fragment-bound** — per-pixel shading is what pins the GPU. Tile-based mobile GPUs pay expensive ALU per fragment and cheap ALU per vertex.

## Experiment

Hook in `load_gltf_pipeline`, between `append_from_file_ex` and `generate_scene`, that overrides `shading_mode` to `PER_VERTEX` (Gouraud) on every `BaseMaterial3D`. Gated by the `--cheap-pbr` CLI flag (deeplink `cheap-pbr=true`).

`ShaderMaterial`-backed surfaces (custom shaders, avatar facial-feature mask materials) are left untouched.

## Accepted tradeoffs

- PBR specular highlights flatten slightly.
- At typical mobile camera distances the visual delta is acceptable.
- `load_seconds` jumps +151% from the import-time per-material flip (see follow-up).

## Requirements

- Opt-in via flag (no default-on until load time is cached to disk).
- `ShaderMaterial` must remain untouched.
- Tests: per-vertex set on every `BaseMaterial3D`, textures/colors preserved, `ShaderMaterial` skipped.

## Follow-ups identified

- `GltfState` iteration is single-threaded → consider Rayon for high material counts.
- Cache `DclGlobal::try_singleton` in a context struct instead of per-call fetch.
- The flag also gates the LOD chain (#2033) today — split the knobs for independent tuning.

Tracked by #2034.

### LINKED ISSUE #2074: Auto-generate LODs for imported meshes to lighten far-distance rendering
## Context

Vertex throughput on Mali-G68 in Genesis Plaza is dominated by far meshes (buildings, signage, skyboxes) that end up sub-pixel on the A54 screen but still rasterize at full resolution. Godot supports LODs natively but DCL assets don't ship a baked LOD chain.

## Experiment

Generate LOD1 / LOD2 / LOD3 per triangle surface at GLTF import time via `meshopt::simplify` with `SimplifyOptions::Permissive | Sparse`.

- LOD0 stays untouched.
- LODs land in the surface's `lods` dict, keyed by screen-space-error threshold (0.5 / 1.5 / 3.0 px).
- Godot's C++ renderer picks the highest-detail LOD whose error ≤ `viewport.mesh_lod_threshold` for the instance's projected size at draw time.

Gated by the same flag as cheap-pbr (deeplink `cheap-pbr=true`). The two features share the pre-generate pipeline.

## Requirements

- LOD0 must remain untouched (visible pass full quality).
- Skip surfaces with blend shapes and skinned surfaces (simplify breaks those).
- Tests: LODs present when index count is sufficient, skip when indices < min, LOD0 preserved, blend-shape skip, skinned skip.

## Success criteria

A54 HIGH, GP preview, warm cache, `cheap-pbr=true`:
- `render_gpu_ms` ≥ −35%
- `fps` ≥ +8%
- No visual regression at mobile distances.

## Follow-ups identified

- `meshopt::simplify` is single-threaded per surface → multi-surface meshes process sequentially. Rayon across surfaces.
- Thresholds `0.5 / 1.5 / 3.0` are hardcoded → per-profile tuning.
- Persistent disk cache (`.lod.bin` sidecar) to eliminate the ~2-5ms/GLTF rebake on each load.

Tracked by #2033.

manuelmaceira:
## What

Mobile GPU optimization stack for the Godot client, targeting Genesis Plaza on Mali-class hardware (Samsung A54). Stacks on top of #1992 (bench infra).

Genesis Plaza, post-stack (A54, HIGH, new GP ea9cc738)

## Stack contents

**Shipping (default ON when `--asset-server` bakes the `.scn`):**
- **cheap-pbr** — flip every `BaseMaterial3D` to `SHADING_MODE_PER_VERTEX` post-generate. Trades per-fragment ALU for per-vertex; major win on Mali-G68 fragment-bound scenes.
- **Godot LOD chain bake** — `ImporterMesh::generate_lods(60°, 25°)` per MI post-generate, **skipping MIs whose any surface uses a transparent/alpha-tested material** (foliage / billboards). Without that skip, generate_lods collapsed bushes into degenerate triangles; with it, opaque buildings get LODs and foliage stays intact.
- **VBO/material dedup across MIs that share a source mesh** — `apply_post_generate_godot_lods` memoizes the baked `ArrayMesh` by source `instance_id`. Without it, `ImporterMesh::get_mesh()` returned a fresh `ArrayMesh` per MI; N MIs that originally shared 1 mesh ended up with N unique meshes → VBOs / index buffers / shader variants duplicated N×. GP has ~5800 surfaces but only ~200 unique source meshes — the dedup recovers that sharing.
- **BoxOccluder bake** — `OccluderInstance3D` siblings on big opaque MIs (`MIN_AABB_DIAG_M=2.0`, mesh-shaped, AABB-derived). Per-axis inset: 50 % on thin axes (< 0.5 m) so wall occluders don't overshoot decals/posters; 85 % on fat axes. Gated by `--asset-server` so the cliente uses pre-baked occluders from the optimized `.scn` instead of re-running the spawn on device.
- **mesh_lod_threshold per-profile** — viewport pixel threshold for LOD swap tuned per graphic profile (VeryLow=8, Low=6, Medium=3, High=2).
- **Shadow tweaks (already merged via #2036, kept in stack)** — directional shadow `size=512`, `mode=PARALLEL_2_SPLITS`, `split_1=0.3`, `max_distance=30 m`.

**Bench infra** lives in the base PR #1992.

**Architectural fix (no flag, default ON):**
- **shadow_proxy disabled in `create_scene_colliders`** — the old `cast_shadow=OFF` on visible + `SHADOWS_ONLY` + `cull_front` on collider sibling broke the GP zeppelin's AnimationPlayer drive. Visible meshes now cast shadows normally; LOD3 covers the shadow-pass cost at distance.

## What's NOT in the stack (dropped / abandoned)

- ~~paired shadow_mesh~~ — visual regressions on Mali; LOD chain already covers shadow pass.
- ~~impostor / octahedral_impostor~~ — framework worked end-to-end on NVIDIA Vulkan, bench A/B inconclusive vs LOD-chain baseline, removed.
- ~~decimator (post-load aggressive LOD2-as-LOD0)~~ — broke close-range visuals (replaced LOD0 with the 25%-triangle LOD2). LOD chain via `generate_lods` already handles distance-based detail.
- ~~vertex_strip (UV2 / tangents / colors)~~ — measured no perf win on its own; removed to simplify.
- ~~runtime asset_preprocessor (`--asset-preproc`)~~ — moved entirely to import-time inside `--asset-server`; no runtime per-entity loop.
- ~~visibility_grid~~ — default-OFF opt-in tool, lives in the bench-infra base PR #1992, not this stack.
- ~~material_atlas / textureless_merger / mesh_lod (runtime) / occluder_gen / auto_shadow_cull / auto_distance_cull / cheap_pbr_materials (runtime)~~ — all gated off + not validated; removed.

## Bench — A/B/C/D progression on **same hardware + same content** (Samsung A54, HIGH profile, GP ea9cc738)

Each row is a different APK against the same Genesis Plaza preview (`ea9cc738`, the catalyst-deployed version).

| state | source | fps | render_gpu_ms | primitives | video_mem |
|---|---|---:|---:|---:|---:|
| **A** — true pre-perf baseline | `baseline/bench-infra-pre-perf` (bench infra + revert of #2036/#2038/#2039) | 13.88 | 67.27 | 2,967,296 | 762 MB |
| **B** — + #2036 shadow tweaks | `baseline/bench-infra-no-perf-stack` | 17.06 | 55.36 | 2,278,718 | 710 MB |
| **C** — + #2038/#2039/#2077/#2078/#2095 (current main) | `baseline/bench-infra-no-perf-stack` rebased on main | 17.60 | 54.26 | 2,181,687 | 735 MB |
| **D** — + this stack | `feat/performance-maximizing` + `optimized-content-base-url` | **18.51** | **51.66** | **1,920,326** | **690 MB** |

Per-step delta on fps (the headline):

| step | Δ fps | Δ render_gpu_ms | what it cost |
|---|---:|---:|---|
| A → B (#2036 shadow tweaks) | **+3.18 (+23 %)** | **−11.9 ms** | shadow res 2048 → 512, range 50 m → 30 m, 4 splits → 2 |
| B → C (subviewports + texture API + DirectionalLight) | +0.54 | −1.1 ms | misc render-side fixes |
| C → D (this stack) | **+0.91 (+5 %)** | **−2.6 ms** | LOD chain + cheap-pbr + dedup + occluders, all baked at import-time |
| **A → D total** | **+4.62 (+33 %)** | **−15.6 ms (−23 %)** | |

**Honest summary**: the largest single win is #2036 (shadow tweaks, already merged). This stack adds another **+5 % fps / −5 % GPU** on top of current main. The headline gain of the stack is **memory**, not perf:

### Device memory (`dumpsys meminfo`, sampled at end-of-bench)

Measured against the same APK + same content scenes, only differing by whether VBO/material dedup was enabled in the asset-server bake:

| metric         | pre-dedup | post-dedup (this PR) | Δ |
|---|---:|---:|---:|
| **GL mtrack**  | 1,882 MB  | **715 MB**  | **−1,167 MB (−62 %)** |
| EGL mtrack     | 71 MB     | 23 MB       | −48 MB |
| Native Heap    | 1,107 MB  | 476 MB      | −631 MB |
| **TOTAL Pss**  | 3,556 MB  | **1,328 MB**| **−2,228 MB (−63 %)** |

Pre-dedup, `ImporterMesh::get_mesh()` returned a fresh `ArrayMesh` per MI: 5,800 surfaces × 200 unique source meshes = each source loaded ~29 times into VRAM. The dedup recovers VBO sharing → cascading dedup of shader variants → ~2.2 GB Pss reclaimed. **App no longer OOMs on mid-range Android.**

### Asset-server output size

| | before dedup | after dedup |
|---|---:|---:|
| `.scn` total disk | 447 MB | **160 MB** (−64 %) |

Smaller `optimized-content` downloads + smaller on-device cache.

## Methodology caveats

- The PR's original baseline table compared against PR #1992's published numbers (fps 7.14, gpu 95.33), which were measured against an older GP commit (`30cdaffd`, 2026-04-29). The deployed catalyst has since moved to `ea9cc738` (2026-05-19, +24 GLBs, +14 % triangles, Sakura theme additions). Bench results above are re-measured against current catalyst-deployed GP for an honest apples-to-apples comparison.
- The 90-second warmup is needed because the floor mesh sometimes doesn't finish streaming within the default 30 s warmup + 60 s settling window. Without it the baseline screenshot shows incomplete loading.

## How to reproduce

- Branch `baseline/bench-infra-pre-perf` reproduces row A.
- Branch `baseline/bench-infra-no-perf-stack` reproduces row B.
- Current `main` (with PR #1992 merged) reproduces row C with raw content.
- `feat/performance-maximizing` + `optimized-content-base-url` pointed at the asset-server output reproduces row D.

## Test plan

- [x] `cargo check --release` clean.
- [x] `cargo clippy --release -- -D warnings` clean.
- [x] APK install on A54, GP loads end-to-end via deeplink.
- [x] Bench result JSONs match the table.
- [x] `dumpsys meminfo` confirms memory table.
- [x] Foliage / alpha-tested materials visually intact post-bake.
- [ ] No regression in non-GP realms (needs review).

## Closes (auto-close on merge to main)

Work from these issues is folded into this stack:

- Closes #2073 — Cheap-PBR per-vertex shading at import (`cheap_pbr_enabled`).
- Closes #2074 — Auto-generate LODs for imported meshes (`apply_post_generate_godot_lods`).
- Closes #2070 — Skip rendering geometry behind buildings (BoxOccluder bake, `mesh_occluder::try_spawn_for`).
- Closes #2065 — Reduce V8 GC from CRDT recv (`op_crdt_recv_wait` + `op_crdt_recv_drain`, framed buffer).
- Closes #2053 — Split CRDT recv into wait + drain to reuse the JS recv buffer (same work as #2065).
- Closes #2064 — Stop blocking the render thread on avatar impostor saves (`save_cached_texture_async`) + TweenState emit-on-transition. (The full native-Godot-tween migration is a separate draft, #2170.)
- Closes #2067 — Replace per-container load-timeout Timers with a single global tracker (`GltfLoadTimeoutCoalescer`, coalesces ~1419 per-container Timers into one Global node).

Shipping config also carries #2036 (shadow tweaks, already merged) and #2071 (lower shadow cost, already closed).

Superseded / folded PRs: #2054 (async impostor PNG + TweenState), #2050 (GltfContainer Timer coalescer) — both closed, content here; and the earlier per-feature PRs (#2033/#2034/#2035/#2037).


docs/bench/screenshots/baseline-pre-perf.png
kuruk-mm, line 1: can we remove these screenshots?

kuruk-mm (changes requested):
## Review: Request changes 🔄

Thanks for the very thorough write-up and the honest assessment — the benchmark rigor here is excellent, and several changes are clear wins I'd happily merge (VBO/material dedup memory recovery, the GLTF load-timeout coalescer replacing ~1,419 `Timer` nodes, the CRDT recv-buffer reuse, the `TweenState` emit-on-transition gate, and moving impostor PNG encode/write off the render thread).

I'm requesting changes for a few items where **benchmark scaffolding appears to ship to production users**, plus some **dead/mislabeled code** that should be removed or wired before merge. Details below, grouped by severity.

---

### 🔴 Blocking

**1. Uncapped FPS + VSync disabled unconditionally on every build (`godot/src/global.gd:229-231` and `:322-324`)**
`_ready()` runs `Engine.max_fps = 0`, `window_set_vsync_mode(VSYNC_DISABLED)`, `low_processor_usage_mode = false`, and then after `config.load_from_settings_file()` force-sets `config.limit_fps = NO_LIMIT` again — **overriding the user's saved FPS cap**. The comments label these "Bench-only knob," but there is no bench/CLI gate, so they execute for real users. On the Mali minspec target this means uncapped GPU + no vsync → more heat, thermal throttling, tearing, and battery drain — the opposite of the PR's goal. Please gate these behind `cli.bench_mode`/`is_gp_benchmark()`.

**2. `DynamicGraphicsManager` initialization commented out for all builds (`godot/src/global.gd:508`)**
`# _init_dynamic_graphics_manager.call_deferred()` is hard-commented, so the thermal-adaptive FPS cap and adaptive profile-downgrade never run on any build. The comment says "Re-enable for production builds," but as written it's disabled for everyone. Note `_init_dynamic_graphics_manager` already self-guards with `if cli.bench_mode: return`, so commenting the call out is unnecessary to skip it during bench — please uncomment it so production keeps thermal protection.

**3. `cheap_pbr_enabled` is parsed but never consumed; CLI arg is mislabeled (`lib/src/godot_classes/dcl_cli.rs:735`, `:500-501`)**
The flag is parsed (`!args_map.contains_key("--no-cheap-pbr")`, default ON) and written by the bench runner (`gp_benchmark_runner.gd:813`), but **nothing reads it** — there is no `diffuse_mode`/specular material code in `lib/src/content/gltf/` keyed on it. Separately, the registered `ArgDefinition` is `--cheap-pbr` ("Default OFF") while the parse reads the unregistered `--no-cheap-pbr` (default ON). The PR description lists runtime "Cheap-PBR (Shipping, Default ON)," but the only per-vertex material conversion that actually runs is `apply_post_material_overrides` (`common.rs:227`), which is **asset-server-bake-only**. Please either wire the runtime flag to a real consumer or drop it, and fix the arg name/help to match reality.

**4. PVS runtime hook loads a script that doesn't exist → latent crash (`godot/src/decentraland_components/gltf_container.gd:165`)**
When the `pvs-runtime` deeplink param is set, the code does `load("res://src/tools/dcl_pvs_runtime.gd").new()`, but that file is not in the repo (`git ls-files | grep -i pvs` → nothing), and `init_with_buffer` is defined nowhere. The producer side is also disconnected: `packer.rs:192` reads a `.pvs.bin` whose generator (`apply_pvs_bake`) exists only in a comment, so no sidecar is ever produced. Please remove the entire PVS chain (runtime hook + `packer.rs` sidecar block) or land the missing pieces — as-is it's a dead, crash-if-enabled path.

---

### 🟠 Should address before merge

**5. Large amount of gated-off / dead code carried in the diff**
The whole `DclGltfRenderManager` + MultiMesh path (`dcl_gltf_render_manager.rs`, `mesh_pool.rs`, `promotion_tracker.rs`, ~1,200 lines) is behind `--rs-gltf-direct` (default OFF) and isn't listed as shipping — that's fine as an experiment, but the siblings `gltf_render/slot_allocator.rs` and `gltf_render/load_timeout.rs` have **zero callers** (superseded by the inline freelist in `mesh_pool` and the GDScript coalescer). In `common.rs`, `apply_pre_generate_material_overrides`, `apply_pre_generate_mesh_simplification`, `apply_post_generate_lod_chain`, and `build_lods_for_surface` are all `#[allow(dead_code)]` with no callers, and **the `meshopt` dependency (`common.rs:17`) is imported only by this dead code**. `apply_post_generate_mesh_split` is a commented call to a function that no longer exists. `vertex_strip::strip_unused` + `metrics::record_stripped`/`drain_global_stats` are dead. Please strip the dead code (and the `meshopt` dep if nothing live uses it), or split the experimental render-manager into a follow-up PR so this one is reviewable as "what actually ships."

**6. MultiMesh RID leak under `--rs-gltf-direct` (`lib/src/godot_classes/gltf_render/mesh_pool.rs:109`)**
`MeshPoolManager::clear()` (which frees the `multimesh_rid`/`instance_rid`) has no caller, and there's no `Drop`/exit hook on the manager that calls it. Per-key RIDs are only ever released slot-by-slot via `release()`. When the flag is on and the manager tears down, those RIDs leak until scenario teardown. Add a `clear()` on exit.

**7. Occlusion culling is enabled project-wide but has nothing to cull on the runtime path (`godot/project.godot:238`)**
`use_occlusion_culling=true` ships, but no `OccluderInstance3D` exists in any committed scene, and occluder generation (`mesh_occluder::try_spawn_for`) is gated by `cli.asset_server`. On a default device loading raw GLBs this is added CPU occlusion-pass cost with no benefit. It only pays off if the device pulls asset-server-baked `.scn` from the optimized-assets CDN — please confirm the CDN is/will be re-baked with this branch, or note the dependency.

**8. `mesh_lod_threshold` is inert without a baked LOD chain (`godot/src/config/graphic_settings.gd:243`)**
The per-profile thresholds (8/6/3/2) only do work when the loaded `.scn` actually carries a generate_lods chain — which is asset-server-bake-only here. They also auto-apply only on a mobile first-launch benchmark (the dynamic-graphics re-apply path is the one commented out in #2). Worth a comment clarifying that this knob is paired with CDN-baked assets, not raw-GLB runtime loads.

---

### 🟡 Non-blocking / nits

- **`TweenState.current_time` is now frozen between state transitions** (`lib/src/scene_runner/components/tween.rs:292`). Great for cutting dirty entries, but please confirm no SDK7 scene reads mid-flight tween progress — this is the one change with SDK-visible semantics.
- **Stale `excludedOps` entry** for the removed `op_crdt_recv_from_renderer` (`lib/src/dcl/js/js_modules/main.js:39`) — cosmetic, but clean it up alongside the recv-op rename.
- **Duplicate `optimized-content-base-url` deeplink handling** in both `deep_link_router.gd` and `global.gd:280` — harmless but redundant.
- **Stale doc comment**: `mesh_occluder.rs` advertises `ArrayOccluder3D`/decimated silhouette, but the code spawns a coarse `BoxOccluder3D` — please fix the comment so reviewers don't expect silhouette-accurate culling.

---

### Framing suggestion

The description's "Shipping, Default ON" bucket (LOD bake, VBO dedup, BoxOccluder, cheap-PBR) is accurate for the **asset-server bake**, but on the **device-runtime first-load path** these are all gated off — the device only gets them via pre-baked optimized `.scn`. Splitting the description into "baked at the asset server" vs "runs on the device at runtime" would make the actual shipping surface much clearer, and would make it obvious that the genuine default-on device wins are the CPU-side ones (timer coalescer, CRDT buffer reuse, TweenState gate, impostor async I/O) plus the graphics-profile retune.

Happy to re-review once the blocking items are gated/removed. 🙏

docs/bench/screenshots/baseline-pre-perf.png
manuelmaceira: yes

kuruk-mm:
Already merged in main with another PR

PR #2037, opened 2026-05-12
perf(rendering): port visibility-grid hot path to Rust + enable by default

manuelmaceira:
## What

Adds a cell-based PVS visibility culler (`DclVisibilityGrid` GDScript + `DclVisibilityGridRust` for the per-frame hot path):

1. **Build** (one-time, post scene load, GDScript): walk the scene tree, bucket every static `MeshInstance3D` into a uniform 16×16m grid by its world AABB, collect `CollisionShape3D` AABBs as blockers (>8m diagonal), bake an O(n²) PVS bitmap by raycasting cell-pair midpoints against blockers.

2. **Per-frame** (Rust): for each cell, distance test → screen-space size threshold → frustum AABB test → PVS bit lookup against the camera cell. If a cell flips visible/hidden, bump per-MI ref-counts; on 0↔1 crossings call `RenderingServer.instance_set_visible(rid, bool)` directly (bypasses `mi.visible = bool`'s scene-tree cascade).

Temporal hysteresis (60-frame hide-streak) prevents popping when the camera moves across cell boundaries.

ENABLED BY DEFAULT (opt out with deeplink `visibility-grid=false`).

## Why

DCL scenes are open-world and have no built-in occlusion structure (no BSP, no portals). Godot's frustum culling handles "behind the camera" but does nothing about "behind that big building". On Genesis Plaza, ~30-40% of rendered geometry per frame is occluded by larger buildings but Godot rasterizes it anyway. The PVS bitmap captures cell-to-cell visibility through blockers so the renderer never sees the occluded MIs.

The per-frame work runs in Rust to avoid the ~5-7ms/frame GDScript overhead that an earlier prototype incurred — same culling decisions, native dispatch, native bit ops, native RenderingServer calls.

| metric | baseline | after | Δ |
|---|---:|---:|---:|
| fps.mean | 7.14 | **10.99** | **+54.0%** |
| render_gpu_ms.mean | 95.33 | 56.99 | **−40.2%** |
| render_cpu_ms.mean | 7.35 | 3.95 | **−46.2%** |
| frame_time_process_ms | 169.27 | 101.75 | −39.9% |
| primitives | 2,610,926 | 1,602,132 | −38.6% |
| video_mem_mb | 778.67 | 884.91 | +13.6% |
| load_seconds | 26.28 | 37.99 | +44.5% |

Largest individual fps win of the perf series. The +12s load_seconds and +14% VRAM are the cost of the runtime PVS bake + the per-MI ref tables.

### Visual diff

| baseline | this PR |
|---|---|
| baseline | after |

## Follow-ups: move work offline into an asset-preprocessor

The +12s build cost is the main remaining wart. Good candidates to move OFFLINE:

1. **PVS bake** — `_build_pvs` (~400ms GDScript) is 100% deterministic given (cell layout, blockers). Per static GLTF, precompute its CollisionShape3D AABBs into a sidecar `.vgrid-blockers.bin`. Runtime loads sidecars, transforms local→world via GLTF instance transform, skips the slow recursive walk + segment-vs-blocker O(n²) raycasts.

2. **Per-GLTF MI AABBs** — Each MeshInstance3D's LOCAL AABB is static. Precompute the list per GLTF into the sidecar; runtime bucketing becomes: apply GLTF transform once, do AABB-vs-cell-grid math.

3. **Cell layout snapping to parcel grid** — DCL parcels are 16×16m, free at preprocessor time.

What CAN'T move offline:
- World-space transforms (DCL scenes compose dynamically per CRDT).
- Final per-MI RID assignment.
- Per-frame visibility loop itself.

Estimated combined effect: runtime build drops from ~2.2s to ~200ms.

### DIFF
--- a/godot/src/decentraland_components/visibility_culling.gd
+++ b/godot/src/decentraland_components/visibility_culling.gd
@@ -1,0 +1,40 @@
+## Visibility-grid culling driver.
+##
+## Rebuilds the cell grid + PVS on every scene-set load (the same boundary
+## the floating-islands feature uses — N parcels load/unload as a batch via
+## `Global.scene_runner.loading_complete`). Per-frame update delegates to
+## the Rust hot path (DclVisibilityGridRust) to keep CPU < 1ms/frame on GP.
+##
+## Off when `Global.cli.visibility_grid_enabled == false` (--no-visibility-grid).
+class_name VisibilityCulling
+extends Node
+
+const GRID_SCRIPT := preload("res://src/tools/visibility_grid.gd")
+
+var _grid: Node = null
+
+
+func _ready() -> void:
+	if not Global.cli.visibility_grid_enabled:
+		set_process(false)
+		return
+	Global.scene_runner.loading_complete.connect(_on_loading_complete)
+
+
+func _on_loading_complete(_session_id: int) -> void:
+	if _grid != null:
+		_grid = null
+	var scene_root: Node = Global.scene_runner
+	if scene_root == null:
+		return
+	_grid = GRID_SCRIPT.new()
+	_grid.build_from_scene_tree(scene_root)
+
+
+func _process(_delta: float) -> void:
+	if _grid == null:
+		return
+	var camera: Camera3D = Global.player_camera_node
+	if camera == null:
+		return
+	_grid.update_visibility(camera)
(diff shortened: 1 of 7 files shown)


manuelmaceira:
Superseded by #2025 (full mobile perf stack, draft). Closing as not planned in this split form — the rebased stack consolidates cheap-pbr, LOD chain pre-gen, paired shadow_mesh (gated OFF), visibility-grid (gated OFF), BoxOccluder bake, and shadow_proxy fix into one reviewable branch.

PR #2038, opened 2026-05-12
perf(rendering): disable always-on SubViewports when not in use

manuelmaceira:
perf(rendering): disable always-on SubViewports when not in use

Visual Profiler revealed two SubViewports rendering every frame even
when there was no UI to display / no avatar to outline:

* `avatar_preview.tscn` — 256x500 SubViewport with player avatar + sky
  for the backpack/profile/lobby portrait.
  `render_target_update_mode`: UPDATE_ALWAYS (4) -> UPDATE_WHEN_VISIBLE (2).

* `outline_system` — 1920x1080 SubViewport (depth pre-pass + outline
  quad). Now gated: starts at UPDATE_DISABLED, flips to UPDATE_ALWAYS
  only when `set_outlined_avatar(non_null)` fires. Avoids a full-screen
  depth pass + outline quad pass every frame when there's nobody to
  outline.

`settings.tscn` preview SubViewport had a similar bug but already lands
at UPDATE_DISABLED in main (probably from a parallel cleanup).

Combined effect on Samsung A54, Genesis Plaza floor (skip-gltf+kill-sky
baseline): fps mean 31 -> 68, p50 29 -> 75, process_ms 42 -> 16. These
viewports were eating a substantial slice of every frame.

Extracted from ebd3b68b on feat/performance-maximizing.


kuruk-mm (approved):
looks great!

PR #2040, opened 2026-05-12
fix: use project setting base resolution for UI zoom + add safemargindebug deep link param

### LINKED ISSUE #2059: UI zoom uses hardcoded 720x720 base resolution instead of project setting
## Problem

`GraphicSettings.get_max_ui_zoom` and `apply_ui_zoom` had a hardcoded `Vector2(720, 720)` base resolution, ignoring the actual project setting (`display/window/size/viewport_width`/`viewport_height` = `1600x720`). The resulting `content_scale_factor` is wrong everywhere except square windows — most visibly on mobile portrait, where the scale collapses to ~0.45×.

UI zoom is also not recomputed on orientation change. On mobile the OS rotation is async, so `Window.size_changed` alone is not enough.

## Expected

UI zoom should use the project's configured base resolution and recompute on orientation change (mobile rotations + desktop emulation toggles).

kuruk-mm:
Closes #2059

### UI zoom base resolution (original fix)
- `GraphicSettings.get_max_ui_zoom` and `apply_ui_zoom` had a hardcoded `Vector2(720, 720)` base resolution, ignoring the actual project setting (`display/window/size/viewport_width` / `viewport_height` = `1600x720`). The resulting `content_scale_factor` was wrong everywhere except square windows.
- Added a `_get_base_resolution(screen_size)` helper that reads the base from `ProjectSettings` and swaps the axes when the screen orientation differs from the project base (so a portrait window scales against a portrait design space instead of collapsing to ~0.45×).
- Connected `apply_ui_zoom` to `Global.orientation_changed` (deferred), so the scale is recomputed when the orientation flips. `Window.size_changed` already fires on actual resize, but on mobile the OS rotation is async — the deferred call guarantees the recompute lands after the resize.
- Added an itest that fails if `display/window/size/viewport_*` in `project.godot` drifts from the values `graphic_settings.gd` reads via `GraphicSettings.get_base_resolution`, so accidental edits surface in CI instead of silently distorting the HUD scale.

### Mobile-preview plugin no longer clobbers the design base
While verifying the fix on device, the `safemargindebug` overlay revealed the design base at runtime was `720x720`, not `1600x720`. Root cause: the `dcl_mobile_preview` editor plugin's "Default" device was hardcoded as `720x720` for both portrait and landscape, and `_apply_settings` writes the selected device's dimensions straight into `display/window/size/viewport_width`/`viewport_height` — the same key our new `get_base_resolution()` reads. Any project save while "Default" was selected persisted `720x720` as the project's design base.

- Updated the `DEVICES[0]` ("Default") entry to `[720, 1600, 1600, 720]` so it matches the real design base in both orientations, and selecting it can no longer corrupt the project's viewport size.
- Removed the now-redundant `"Figma Base"` entry (same dimensions as the new Default).
- Added a defensive clamp on the persisted `last_device` index so users whose editor remembered `"Figma Base"` (index 4) don't end up with an out-of-range selection on next open.

### `safemargindebug` deep link param (new diagnostic tool)
While iterating on the base-resolution math it was hard to inspect safe-area insets and the effective scale on real devices without attaching a debugger. To make that diagnosable in-app, this PR adds a way to **enable a debug safe-margins overlay via a parameter** — which is exactly how the plugin bug above was discovered.

- New `safemargindebug=true` (or `=1`) query param parsed in `lib/src/deep_link.rs` and exposed on `DclParseDeepLink`.
- Example: `decentraland://open?safemargindebug=true`
- When the flag is set, `Global` instantiates a transparent, top-level, click-through `CanvasLayer` overlay (`godot/src/tool/safe_margin_debug_overlay.gd`) that:
  - tints the four unsafe margin strips so the safe area is visually obvious,
  - shows window / scaled / base resolution and the per-side safe-margin pixel values in a center HUD,
  - persists across scene changes,
  - is **only instantiated when the flag is set** — zero memory/CPU cost when off.
- Three-finger touch toggles show/hide on device, so the overlay can be dismissed without restarting the deep link.
- Includes Rust unit tests for the new param (`true` / `1` / default-off cases) and explicit `maxf` / `maxi` types in the overlay to keep it loadable under the strict-mode GDScript parser.

## Test plan
- [ ] Desktop: launch the client, verify the UI scales correctly at the default landscape window (`1600x720`) and after manually resizing.
- [ ] Desktop emulation: toggle `Global.set_orientation_portrait()` / `set_orientation_landscape()` (`1280x720` ↔ `720x1280`) and confirm the UI re-scales each time without manual resize.
- [ ] Mobile (Android): rotate the device landscape ↔ portrait ↔ landscape and confirm the UI re-scales on each rotation (no stuck scale factor).
- [ ] Confirm the zoom dropdown in graphics settings (driven by `get_ui_zoom_available` / `get_max_ui_zoom`) shows sensible max-zoom values at both orientations.
- [ ] Launch with `decentraland://open?safemargindebug=true` and verify the overlay appears, the four unsafe strips are tinted, and the HUD shows window/scaled/base resolution + safe-margin pixel values — and that **Base** reads `1600x720` (landscape) or `720x1600` (portrait), never `720x720`.
- [ ] In the editor, cycle the mobile-preview "Default" device on a scene with no orientation meta and confirm `project.godot` is **not** rewritten to `viewport_width=720` / `viewport_height=720`.
- [ ] Open the editor with `dcl_mobile_preview/last_device` previously set to `4` (Figma Base) in editor settings and confirm the device dropdown falls back to "Default" without errors.
- [ ] Launch without the param and confirm no overlay is created (no extra `CanvasLayer` in the tree).
- [ ] On a touch device with the overlay enabled, use a three-finger tap to toggle the overlay show/hide.
- [ ] CI: the new `project_settings_test` itest passes and would fail if `display/window/size/viewport_*` is edited away from `1600x720`.


godot/src/config/graphic_settings.gd
```diff
@@ -56,13 +56,29 @@ const SKYBOX_TIME_NAMES: Array[Dictionary] = [
...
+		func(_is_portrait): GraphicSettings.apply_ui_zoom.bind(root).call_deferred()
+	)
+
+
+## Returns the design base resolution from project settings, with axes swapped
+## when the current window is portrait so that the design space follows the
+## screen orientation (e.g. 1600x720 landscape -> 720x1600 portrait).
+static func _get_base_resolution(screen_size: Vector2) -> Vector2:
+	var base_width: float = ProjectSettings.get_setting("display/window/size/viewport_width", 1600)
+	var base_height: float = ProjectSettings.get_setting("display/window/size/viewport_height", 720)
```
leanmendoza, line 71: this is fixed settings? or it's changed in function of mobile preview?

kuruk-mm: it should not be changed by the mobile preview tool...

i will add a test to double check no resolutions change

godot/src/tool/safe_margin_debug_overlay.gd
```diff
@@ -0,0 +1,165 @@
...
+		" base=",
+		base_resolution
+	)
+
+	# Safe-area margins in window pixels (what notch/cutout costs us).
+	var top_px: int = maxi(0, safe_area.position.y)
+	var left_px: int = maxi(0, safe_area.position.x)
+	var bottom_px: int = maxi(0, window_size.y - safe_area.end.y)
+	var right_px: int = maxi(0, window_size.x - safe_area.end.x)
+
```
regenesis-claw, line 110: This `print()` fires on every resize/orientation change. The HUD label already displays all this info visually — consider removing this print or gating it behind a one-time flag to keep logs clean when the overlay is active on-device.

regenesis-claw (approved):
Clean, well-structured fix. The root cause analysis (hardcoded `720x720` vs the actual `1600x720` project base) is solid, the `get_base_resolution()` helper is the right abstraction, and the itest that locks the base resolution to catch future drift is a nice touch.

The safe-margin debug overlay is a useful diagnostic tool — the 3-finger toggle and deep-link activation are both practical for on-device debugging.

### Minor notes (non-blocking)

🟡 **Verbose prints in `_refresh()` (`safe_margin_debug_overlay.gd:98-110`)** — This fires on every window resize and orientation change. Since the overlay is debug-only and deep-link-gated, it's not a production concern per se, but if someone enables this on a device and rotates a few times the log gets noisy. Consider gating behind a flag or removing the per-refresh print (the HUD label already shows all the same data visually).

🟡 **Deep link processed in two places** — `global.gd:_ready()` and `deep_link_router.gd:process_deep_link()` both call `set_safe_margin_debug_enable(true)` for the same param. The idempotency guard in `set_safe_margin_debug_enable` makes this safe, and both paths are needed (initial launch vs runtime deep links), so no action needed — just noting it for clarity.

✅ CI all green (static checks, clippy, Linux, Android, Docker)
✅ iOS build present and successful
✅ Rust unit tests for the new deep-link param cover true/1/default-off
✅ itest locks the viewport base to 1600x720

Approving — this fixes the real root cause of the HUD scaling issues reported by Elizabeth and Rodri.

Requested by Mateo via Slack

PR #2044, opened 2026-05-12
fix: HIDE_AVATARS modifier area now disables profile click collider

### LINKED ISSUE #2016: Avatar modifier areas with HIDE_AVATARS not hiding profile interaction collider
## Issue
Avatar modifier areas configured with HIDE_AVATARS successfully hide the visual representation of avatars, but the collider for interacting with player profiles remains active. Users can still click on the invisible avatar to open the profile.

## Expected Behavior
When an avatar is hidden via HIDE_AVATARS modifier area, the profile interaction collider should also be disabled.

## Current Behavior
- Avatar visual: hidden ✓
- Profile interaction collider: still active ✗

## Impact
Creates confusing UX where users can interact with invisible avatars.

---
Requested by Mateo via Slack

kuruk-mm:
## Summary
- Avatar modifier areas with `HIDE_AVATARS` were hiding visuals but leaving the `ClickArea` collider active, so users could click on invisible avatars and open their profile.
- The fix mirrors `set_hidden(true)`: when the modifier hides the avatar, also call `_set_click_area_enabled(false)`. When the modifier is unset, re-enable the click area (guarded by `if not hidden` so blocked avatars remain non-interactable).

Fixes #2016

## Test plan
- [ ] Enter a scene with an `AvatarModifierArea` configured with `HIDE_AVATARS`; confirm avatars inside the area are invisible and clicking where they stand does not open a passport.
- [ ] Leave the area; confirm avatars re-appear and become clickable again.
- [ ] Verify a separately-blocked/muted avatar (set via `set_hidden(true)`) remains non-clickable after entering/leaving a modifier area.
- [ ] Verify `DISABLE_PASSPORT` modifier behavior is unchanged.

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar.gd
+++ b/godot/src/decentraland_components/avatar/avatar.gd
@@ -263,6 +263,7 @@ func _on_set_avatar_modifier_area(area: DclAvatarModifierArea3D):
 		if modifier == 0:  # hide avatar
 			hide()
 			_hide_impostor_render()
+			_set_click_area_enabled(false)
 		elif modifier == 1:  # disable passport
 			passport_disabled = true
 
@@ -322,6 +323,7 @@ func _set_click_area_enabled(enabled: bool) -> void:
 func _unset_avatar_modifier_area():
 	if not hidden:
 		show()
+		_set_click_area_enabled(true)
 	passport_disabled = false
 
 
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
Clean fix that correctly mirrors the `set_hidden()` pattern. The `_set_click_area_enabled(false)` + `passport_disabled = true` in the hide-avatar modifier path, and the guarded re-enable in `_unset_avatar_modifier_area()` (respecting the `if not hidden` check so blocked avatars stay non-interactable) are both correct.

No platform-specific code touched — iOS build not required.

LGTM ✅

Requested by Mateo via Slack

leanmendoza (approved):

PR #2048, opened 2026-05-13
feat: analytics improvements

### LINKED ISSUE #2023: Entry Flow: Increase EULA Checkbox Size
## Description
Increase size of checkbox and font.

### Text
- Increase font size by 2 points

### Checkbox
- Enlarge checkbox to 64x64
- Add outline: 2 points, Color #FF2D55

## Design Reference

Figma: https://www.figma.com/design/H1Alvbe7sh0e0IyLckewMI/Mobile-Entry-Flow?node-id=1276-10439&t=aGVTqXZ9NhzOfF6E-1

## Refe
Image

## Context

Design iteration by Elizabeth to improve readability and visual prominence of the Terms of Service acceptance UI.

Requested by Rodri via Slack

### LINKED ISSUE #2027: Add Firebase Identifiers to INSTALL_ATTRIBUTION
Part of #2026 (LMT1 Tracking v2 + Fixes)

## Objective

Modify INSTALL_ATTRIBUTION table to include firebase data essential for cross-funnel tracking and enhanced analytics as specified in the [following sample table](https://docs.google.com/spreadsheets/d/1hAklFjxuR4Zz2Fve1sWObZo8rqvVGhlz9tt8-dvm9rY/edit?gid=370022304#gid=370022304).

## Fields to Add:
```
- FIREBASE_USER_PSEUDO_ID, 
```

Image

## Context
Need to ensure consistent user/device identification across all tracking and analytics platforms for accurate attribution and debugging.

### LINKED ISSUE #2028: Pre-EULA Events Firing logic
Part of #2026 (LMT1 Tracking v2 + Fixes)

## Objective

Review and optimize the firing logic for events that occur before the EULA is accepted, based on #1996 

## Tasks
- [ ] Check flush moment / time cadence and when we trigger the metrics to Metabase server

## Context
Events fired before EULA acceptance require special handling to ensure they are properly captured and sent at the right time. Need to verify the current flush timing and cadence.

### LINKED ISSUE #2029: Add New Firebase custom events
Part of #2026 (LMT1 Tracking v2 + Fixes)

## Objective

Add Firebase custom analytics events for key user actions to improve tracking granularity and enable cross-checking with segment.

## Tasks
- [ ] Add Firebase custom event for `eula_accept`, appending mobile user_id.
- [ ] Add Firebase custom event for `login_success`
- [ ] Add Firebase custom event for `move_in_world`

## Context

These events will provide better visibility into user behavior within Firebase Analytics, complementing existing Metabase tracking.

Requested by Rodri via Slack

### LINKED ISSUE #2041: [Bug] AUTH_BROWSER_OPEN: Missing 'email' in screen_view extra_properties
Part of #2026 (LMT1 Tracking v2 + Fixes)

## Description

The `AUTH_BROWSER_OPEN` event is missing the `otp` field in `screen_view` `extra_properties` — the value is coming through as `null` instead of the expected OTP value.

## Expected Behavior

The `otp` field should contain the actual OTP value in the `extra_properties` of the `screen_view` event when `AUTH_BROWSER_OPEN` is triggered.

## Actual Behavior

The `otp` field is `null` in `extra_properties`.

## Impact

This prevents proper tracking and debugging of authentication flows involving OTP codes.

Requested by Rodri via Slack

leanmendoza:
Fixes #2023
Fixes #2027
Fixes #2028
Fixes #2029
Fixes #2041

# Analytics improvements

Wires Firebase Analytics (Android) on top of the existing Segment pipeline, adds an EULA consent gate, and reorganizes auth-related event placement in the lobby.

## Cross-system correlation: Segment ↔ Firebase

- New `Firebase Init` Segment event (`SegmentEventFirebaseInit`) carrying the Firebase Analytics app instance id (`ga_pseudo_user_id`) + the existing `session_id`. This is the anchor analysts use to pivot Segment → Firebase.
- For the inverse direction (Firebase → Segment), `Metrics::ready` seeds `user_id` and `session_id` as Firebase user properties so every subsequent Firebase event (including SDK auto-events) carries them.
- The signal `firebase_app_instance_id_ready` is fired once by the Android plugin when the SDK resolves the id. We close the cache-hit race by reading back the cached value after `connect`: if non-empty, deliver it manually (idempotent via `firebase_init_queued`).
- If the id arrives empty (SDK failure), we `tracing::error!` and skip queueing — an event without an id has no pivot value.

## EULA consent gate

- New `Metrics::eula_accepted` flag. While false, events keep queueing in `self.events` but `flush()` / the periodic timer never serialize or ship anything (Segment or Firebase).
- `set_eula_accepted(true)` flips the gate and auto-flushes queued pre-consent events (ACCEPT_EULA screen/click, START, etc.).
- Returning users open the gate at startup if their stored `terms_and_conditions_version` matches the current. Fresh users open it on Accept-EULA click.

## Firebase Key Events (Android-only)

Three custom Firebase events fired through the plugin's `logFirebaseEvent`:

- `eula_accepted` — fired right after the gate opens on fresh accept.
- `login_success` — fired on `wallet_connected`. **Suppressed for session recoveries** via a one-shot flag the lobby sets before calling `try_recover_account`. The flag is cleared on the next `process_frame` (not `call_deferred`, which would FIFO-clear before Rust's deferred `emit_signal` runs).
- `first_move_in_world` — fired once per install the first time the player crosses 0.5 m/s horizontal velocity after `loading_finished`. Persisted via the new `first_move_in_world_sent` config flag. The flag is only persisted when the Firebase event actually dispatched (Android + EULA accepted); on other platforms or pre-EULA we stop polling for the current session but the flag stays false so a future session retries.

## AnalyticsController

New RefCounted (`godot/src/logic/metrics/analytics_controller.gd`) instantiated in `Global._ready` after `metrics`. Owns:

- The EULA gate setup (returning user path).
- The `wallet_connected` handler that fires `login_success` (with session-recovery suppression).
- The transient Timer (10 Hz) that polls `player_body.actual_velocity_xz` for the one-shot `first_move_in_world` detection. The Timer is created lazily on `loading_finished`, freed on either event firing or `loading_started`.

The controller connects its `wallet_connected` handler in `Global._ready`, before the lobby scene loads, so it runs **before** lobby's own handler — important because `track_login` reads the wallet address from the signal arg (lobby's handler is what later calls `update_identity` to refresh `common.dcl_eth_address`). Documented inline.

## Flush cadence

- New `Metrics::set_flush_interval(seconds)` — restarts the periodic timer with the new wait time (clamped to ≥ 0.5 s).
- Lobby drops to 2 s in `_ready` so onboarding/auth events ship snappily.
- `explorer._ready` and `menu._ready` restore the 10 s default so feed/discover events batch.

## Auth event placement (lobby)

- Moved `update_identity` and `track_screen_viewed("AUTH_SUCCESS")` from `async_close_sign_in` to `_on_wallet_connected` so they also fire on session recoveries (not only on fresh sign-ins). Added `flush.call_deferred()` after.
- Added `AUTH_DEPLOY_SUCCESS` screen-view event after the new-account profile deploy succeeds, plus a `flush.call_deferred()`.

## Android plugin (Kotlin)

New methods in `GodotAndroidPlugin.kt` (all currently accessed via reflection so the plugin doesn't depend on the Firebase SDK; see in-file TODO for a future `compileOnly` refactor):

- `getFirebaseAppInstanceId()` — returns cached id or empty.
- `setFirebaseUserId(String)` / `setFirebaseUserProperty(String, String)` — pass-through to FirebaseAnalytics.
- `logFirebaseEvent(eventName, paramsJson)` — parses primitive JSON into a Bundle and logs.
- One-shot signal `firebase_app_instance_id_ready` emitted from the `getAppInstanceId` `OnCompleteListener`.

Firebase initialization is kicked off in `onGodotSetupCompleted` so the id has a head start by the time `Metrics::ready` connects. Reflection failures (Firebase missing on classpath, e.g. dev builds without `google-services.json`) are caught and the plugin no-ops cleanly.

## Rust ↔ Kotlin bindings (`dcl_android_plugin.rs`)

New `#[func]` wrappers: `get_firebase_app_instance_id`, `connect_firebase_app_instance_id_ready`, `set_firebase_user_id`, `set_firebase_user_property`, `log_firebase_event`.

## Lobby EULA checkbox restyle

New `checkbox_2_on.svg` / `checkbox_2_off.svg` icons (red stroke, 64×64) replacing the previous checkbox visuals on the EULA accept control. `lobby.tscn` updated to use them with `icon_max_width = 64`, `icon_alignment = 1`.

## Test plan

- [ ] Fresh install (Android, with `google-services.json` present): verify nothing ships until Accept-EULA; then `Firebase Init` Segment event + `eula_accepted` Firebase event ship.
- [ ] Fresh install: walk into the world → `first_move_in_world` Firebase event fires once; `first_move_in_world_sent` persisted.
- [ ] Returning user (EULA already accepted): events ship immediately on startup; session recovery does NOT fire `login_success`; deliberate sign-in flow does.
- [ ] Lobby flush cadence is 2 s; menu/explorer is 10 s.
- [ ] Build without `google-services.json` (dev clone): plugin logs `[Firebase] FirebaseAnalytics class not present` and all Firebase events no-op; Segment events still flow after EULA accept.
- [ ] Non-Android build: Firebase events are no-ops; `first_move_in_world_sent` stays false across sessions (we still stop polling within a session).


regenesis-claw (approved):
## Review — Analytics improvements (Firebase + EULA gate + flush cadence)

Followed `REVIEW.md`. 698 additions across Kotlin, Rust, and GDScript — full structured review below.

### Pre-flight

- **Branch**: 1 commit behind main — not a concern for this diff.
- **CI**: All 5 checks green (static checks, clippy, Linux, Android, Docker).
- **iOS build**: Not required — no iOS-specific code paths touched. Firebase integration is Android-only; GDScript/Rust changes are platform-gated via `MobilePlatform::Android` / plugin availability checks.

### Architecture read

The layering is clean and well-separated:

| Layer | Responsibility |
|---|---|
| `GodotAndroidPlugin.kt` | Firebase SDK access via reflection, signal emission |
| `dcl_android_plugin.rs` | Rust ↔ Kotlin JNI bridge (typed wrappers) |
| `metrics.rs` | Consent gate, event queueing, flush cadence, Firebase key events |
| `analytics_controller.gd` | GDScript orchestration (EULA setup, login suppression, first-move poller) |

The `AnalyticsController` as a `RefCounted` kept alive by `Global.analytics_controller` is a good pattern — no scene-tree noise except the transient first-move Timer. The docstring correctly notes that signal connections alone don't prevent RefCounted collection.

### Key observations

1. **Race-condition handling for Firebase app instance id** — the `firebase_init_queued` guard + manual cache-read after `connect` closes the cache-hit race cleanly. The early-set of the flag before the empty-check is correct given the one-shot nature of the signal: the manual path and signal path are mutually exclusive in practice.

2. **Session-recovery suppression** — the `process_frame` one-shot for clearing `_wallet_connected_is_session_recovery` is well-reasoned (the inline comment explains why `call_deferred` wouldn't work here due to FIFO ordering vs Rust's deferred `emit_signal`). Signal connection order is documented as load-bearing (controller connects before lobby loads).

3. **EULA consent gate** — clean design. Events queue freely; `flush()` and `timer_timeout` both respect the gate; `set_eula_accepted(true)` auto-flushes. The ordering in `_on_button_accept_eula_pressed` (controller opens gate → config persists version) has a tiny crash window but the worst case is re-showing EULA — acceptable.

4. **`first_move_in_world` lifecycle** — poller created on `loading_finished`, destroyed on event fire or `loading_started`. The `track_first_move_in_world() → bool` contract (persist only when actually dispatched) correctly handles non-Android and pre-EULA platforms without burning the one-shot.

5. **Flush cadence** — lobby 2s, menu/explorer 10s. `set_flush_interval` clamps to ≥0.5s and uses epsilon comparison to skip no-op restarts. Both `menu.gd` and `explorer.gd` restore the default — covers the two exit paths from lobby.

6. **Kotlin reflection** — all Firebase calls go through reflection with comprehensive try/catch. The TODO for `compileOnly` dep is fair. `ensureFirebaseInitialized` intentionally does NOT set `firebaseInitTried` when `applicationContext` is null (allows retry) — correct for early-startup transient state.

7. **Auth event placement** — moving `update_identity` + `AUTH_SUCCESS` from `async_close_sign_in` to `_on_wallet_connected` means session recoveries now also fire these. This is the right call since the wallet address comes from the signal arg directly rather than stale `common.dcl_eth_address`.

### Minor notes (non-blocking)

- The `"actual_velocity_xz" in player_body` duck-type check in the poller is defensive but would fail silently if the property is ever renamed. A typed reference would be safer, but given this is a one-shot polling window it's fine for now.

### Positives

- PR description is excellent — the cross-system correlation explanation, the race-condition analysis, and the test plan are all thorough.
- Inline comments throughout explain *why* decisions were made (connection ordering, frame-deferred vs call_deferred, when to persist the one-shot flag).
- No debug prints, no `.claude/` files, no stray config changes.
- The checkbox restyle SVGs come with proper `.import` files.

Approving — solid work.

Requested by Mateo via Slack

PR #2050, opened 2026-05-13
perf(gltf): coalesce 1419 GltfContainer Timers into one Global node

manuelmaceira:
## What

Genesis Plaza loads ~1419 GltfContainer instances. Each one used to spawn a Godot Timer node for its 120s load-timeout. That's 1419 \`_process\` callbacks per frame doing nothing useful 99.99% of the time.

Replaces them with a single \`GltfLoadTimeoutCoalescer\` node on Global, holding one Dictionary keyed by container instance id → deadline_ms. Per frame the coalescer walks the dict and fires \`_on_load_timeout()\` on any container whose deadline passed.

## Why

Per-instance Timer nodes are expensive at scale on mobile:
- Each Timer is a full \`Node\` with its own \`_process\` callback, registered with the scene tree's idle process list
- 1419 idle-process traversals/frame on Genesis Plaza load
- Each Timer node also has internal time accumulation logic that runs every frame even when stopped

A single coalescer node does ~1 \`Dictionary\` lookup + 1 comparison per active container per frame, no scene-tree per-node callbacks.

## Why the coalescer is a separate child Node (not on the autoload)

Adding \`_process\` directly to the \`DclGlobal\` autoload reproducibly triggers a VkThread NULL-deref crash inside \`libgodot_android.so\` on Android (Samsung A54). Bisected. The coalescer lives as a child of Global to keep \`_process\` off the autoload itself.

## Changes

| file | change |
|---|---|
| \`gltf_load_timeout_coalescer.gd\` | NEW: 33-line coalescer node |
| \`gltf_container.gd\` | replace \`timer.start/stop\` with \`coalescer.schedule/cancel\`; rename \`_on_timer_timeout\` → \`_on_load_timeout\` |
| \`gltf_container.tscn\` | remove Timer child node |
| \`global.gd\` | add \`get_gltf_load_timeout_coalescer()\` lazy getter |

## Extracted from

Originally bundled inside the spike PR (#1992). Pulled out as a standalone perf fix.


manuelmaceira:
Superseded by #2025 (perf stack). The GltfLoadTimeoutCoalescer is already folded into `feat/performance-maximizing`:
- `Global.get_gltf_load_timeout_coalescer().schedule/cancel(self, ...)` in `gltf_container.gd` (replaces the per-container Timer).
- One Global `GltfLoadTimeoutCoalescer` node coalesces the ~1419 per-container load-timeout Timers.

Closing as not-planned; the work ships via #2025. The issue it addresses (#2067) is referenced in #2025's Closes block so it auto-closes when the stack lands on main.

PR #2051, opened 2026-05-13
perf(gltf): RenderingServer-direct + MultiMesh batching for static GLBs (opt-in)

manuelmaceira:
## What

New opt-in rendering pipeline that bypasses Godot's scene-tree per-mesh nodes for static GLBs. Static MeshInstance3Ds are extracted from the loaded GLTF tree, registered with the RenderingServer directly, and batched into MultiMesh slots per (mesh, material) bucket.

**Default OFF** — gated by \`--rs-gltf-direct\` CLI flag (or \`rs-gltf-direct=true\` deeplink param). Animated GLBs are detected at registration time and skip the migration, falling through to the legacy scene-tree path.

## Architecture

| file | role |
|---|---|
| \`dcl_gltf_render_manager.rs\` (430) | manager GodotClass — owns the scenario, slot allocator, promotion tracker, per-frame \`_process\` for transform sync |
| \`gltf_render/mesh_pool.rs\` (198) | ref-counted mesh resource pool keyed by \`(mesh_rid, material_rid)\` |
| \`gltf_render/slot_allocator.rs\` (162) | MultiMesh slot allocation with free-list reuse |
| \`gltf_render/promotion_tracker.rs\` (133) | tracks containers eligible for promotion from instance-rendering to MultiMesh once a bucket reaches its threshold |
| \`gltf_render/load_timeout.rs\` (125) | per-container retry/timeout state |

## Why it's still opt-in

This path is functionally complete but trades some fidelity for throughput:
- Material parameter changes per-instance (e.g. tint overrides via SDK7) don't yet propagate through the MultiMesh batch
- AnimationPlayer-driven GLBs are detected and skipped — but the heuristic for \"static\" is conservative
- The render manager's \`_process\` does transform sync every frame; for a fully static scene with N batched slots this is N matrix copies per frame. Future: dirty-flag transforms.

Once those caveats are addressed, this can flip to default-on (similar to the cheap-pbr / shadow-mesh path).

## Why a single PR

The 4 helper modules (mesh_pool, slot_allocator, promotion_tracker, load_timeout) are tightly coupled and only used by the render manager — splitting them across PRs would just create a chain of stubs.

## Extracted from

Originally bundled inside the spike PR (#1992). Pulled out so the spike stays bench-only and this stack rendering pipeline can be reviewed on its own merit.


manuelmaceira:
Closing as not planned — same conclusion we reached for #2066.

We re-tested this path in depth to see if it was worth it for **RAM** even if it didn't move fps. Findings:

- **MultiMesh batching loses per-instance culling.** A MultiMesh is a single cull unit, so the whole batch is drawn whenever any part of its AABB is on-screen. On GP that *increased* visible primitives vs. plain MeshInstance3D, which the renderer culls per-instance (frustum + occlusion). Validated on A54 with the `rs-direct-test` sweep: raw RS instances kept culling (primitives stayed ~2.22M), MultiMesh did not.
- **RenderingServer-direct instances (the non-batched tier) do keep Godot's frustum + occlusion culling** — that part works. But the only thing it buys over MeshInstance3D nodes is a bit of node overhead, and the RAM win didn't justify the complexity of the slot allocator / promotion tracker / mesh pool machinery.
- The real RAM lever turned out to be **VBO/material dedup at import time** (memoize the baked `ArrayMesh` by source `instance_id`), which lands in the #2025 stack: −63% Pss on A54 without touching the render path. That gives us the memory win this PR was chasing, without giving up culling.

So: no fps gain (GP is GPU/primitive-bound, not node-overhead-bound) and the RAM goal is met more simply elsewhere. Keeping the branch for reference. Issue #2066 already closed with the matching rationale.

PR #1841, opened 2026-04-14
feat: non-pointer (proximity) interactions

EibrielInv:
## Summary                                                                              
                                                                                                                                                                                                              
  - Filters proximity interactions from cursor/raycast pointer events system so they don't trigger tooltips, hover, or click events via the raycast path                                                      
  - Implements proximity event firing: detects the active proximity entity each frame based on player distance, visibility cone (1/3 shortest screen side), and priority; fires \`PetHoverEnter\`,            
  \`PetHoverLeave\`, \`PetDown\`, \`PetUp\` accordingly                                                                                                                                                       
  - Shows proximity tooltip for the closest/highest-priority entity in range; suppressed when player is pointing at a cursor-interaction entity                                                               
  - Refactors pointer events into \`pointer_events_system_without_proximity\` and \`proximity_events_system\`, sharing \`find_active_proximity_entity\` to avoid duplication                                  
  - Updates DCL protocol to commit \`5e66d77\` and adds new proto fields for movement compression and social settings                                                                                         
                                                                                                                                                                                                              
  Closes #1795                                                                                                                                                                                                

  ## Video

https://github.com/user-attachments/assets/82085738-7a4c-4a89-b97a-4c0ea740fe78


  ## Test plan                                                                                                                                                                                                
                                                                                                                                                                                                              
  - [ ] Entity with \`onProximityDown\` only: no tooltip or interaction when pointing at it from outside range; tooltip and events fire when within \`maxPlayerDistance\` and looking at it                   
  - [ ] Entity with \`onPointerDown\` only: tooltip and events fire normally via raycast                                                                                                                      
  - [ ] Both entity types in scene: pointing at cursor entity suppresses proximity tooltip; walking away restores it                                                                                          
  - [ ] Multiple proximity entities: only highest-priority one shows tooltip; distance as tiebreaker on equal priority                                                                                        
  - [ ] Hover enter/leave events reach JS scene correctly for both interaction types

PR #2052, opened 2026-05-13
fix: PointerEvents HOVER on 3D entities populates hit and respects range (#2017)

### LINKED ISSUE #2017: [Bug] PointerEvents HOVER/LEAVE events not triggered when focusing 3D objects
## Issue Description

PointerEvents + HOVER/LEAVE events are not being triggered when focusing something in the 3D scene.

## Expected Behavior

When focusing on a 3D object, the PointerEvents HOVER and LEAVE events should be properly triggered.

## Current Behavior

These events fail to trigger when focusing 3D objects.

---

Requested by Mateo via Slack

kuruk-mm:
## Summary

Fixes [#2017](https://github.com/decentraland/godot-explorer/issues/2017) — PointerEvents `HOVER_ENTER` / `HOVER_LEAVE` appeared not to trigger on 3D entities.

The emission path existed but was incomplete:

- `hit` was always `None` for cursor hover (DOWN/UP populate it). SDK7 scripts that read `event.hit.position` / `event.hit.normalHit` saw null and treated the event as not delivered.
- `max_distance` / `max_player_distance` from `pointer_events.proto` were not enforced on hover, diverging from DOWN/UP and from the SDK spec.

## Changes

- `PbPointerEvents::has_active_cursor_event(pet, camera_distance, player_distance)` — checks if any non-proximity entry of the given type passes its distance gate. Implements the proto rule: only `max_distance` → camera check; only `max_player_distance` → player check; both → OR; neither → default `max_distance = 10`.
- `entity_player_distance()` helper — looks up an entity's `Node3D` global position, returns `f32::MAX` if the entity is gone (fail-closed).
- Cursor `HOVER_ENTER` / `HOVER_LEAVE`:
  - `hit` now populated with `RaycastHit` from the relevant frame.
  - Emission gated by the distance helper above.
- Proximity hover blocks: comment clarifying `hit: None` is intentional (no raycast involved).
- `player_position` threaded into `pointer_events_system_without_proximity`.

Single-file change: `lib/src/scene_runner/components/pointer_events.rs` (+114/-40).

## Out of scope

- `PET_PROXIMITY_ENTER` / `PET_PROXIMITY_LEAVE` event-type emission (proximity flow currently emits `PetHoverEnter/Leave` — separate gap from #1841).
- UI PointerEvents (separate `base_ui` code path).
- Mobile tap-to-focus.

## QA — Hover Range

Head to the QA scene at parcel **`-31,19`** on mobile, open the test switcher, and pick **Hover Range**. You'll see six numbered cubes in a row; while you hover a cube, a yellow sphere marks the hit point and a green arrow shows the surface normal.

For the first three, just walk toward and away from each cube and confirm `HOVER_ENTER` / `HOVER_LEAVE` fire at the expected boundary, always with a real `hit` (never `hit: none`):
- [x] **Cube 1 (`maxDistance 3`)** should fire only when the camera is within ~3 m.
- [x] **Cube 2 (`maxPlayerDistance 4`)** should fire only when your avatar is within ~4 m.
- [x] **Cube 3 (default range)** should fire when the camera is within ~10 m.

The last three prove *which* distance each gate reads. Since mobile has no camera zoom, the **Far Camera** buttons park the camera ~12 m from the cube while your avatar stays close (a custom joystick appears so you can still move):
- [x] For **Cube 4 (`maxPlayerDistance 4`)**, tap **Far → 4**: hover should still fire while you're within 4 m, confirming the gate uses avatar distance.
- [x] For **Cube 5 (`maxDistance 3`)**, tap **Far → 5**: hover should stay silent even standing next to it, confirming the gate uses camera distance.
- [ ] For **Cube 6 (click + `maxPlayerDistance 4`)**, tap **Far → 6**: tapping the cube from within 4 m should fire `PET_DOWN` and flash the cube, even with the camera far away.
- [x] Finally, tap **Player Cam** to return, and confirm the **Pointer** tab still behaves normally with working DOWN/UP clicks.


lib/src/scene_runner/components/pointer_events.rs
EibrielInv: Distance is reevaluated with a different criterion. Making it show the overlay even when the conditions are not met.

lib/src/scene_runner/components/pointer_events.rs
```diff
@@ -361,56 +412,79 @@ fn pointer_events_system_without_proximity(
     changed_inputs: &HashSet<(InputAction, bool)>,
     previous_raycast: &Option<GodotDclRaycastResult>,
     current_raycast: &Option<GodotDclRaycastResult>,
+    player_position: godot::prelude::Vector3,
```
EibrielInv, line 431: eq_key only compares the raycast target (scene + entity). So the enter/leave logic — and therefore the new has_active_cursor_event distance check — runs only on the frame the cursor moves onto or off a
different entity. Distance is sampled once, at that instant, and never re-evaluated while you keep pointing at the same entity. Resulting on events not firing if only the distance changed.

(comment intended for line 418)

EibrielInv (commented):
I found a couple of issues when testing, take a look.
- Events not firing under some conditions
- Outline being shown when there is no event

regenesis-claw (approved):
## ✅ Approved

Followed `REVIEW.md`. Pre-flight clean: branch merged with `main` today, CI all green (Clippy, Linux, Android, iOS, Docker), no submodule drift. iOS build not required (Rust scene_runner change, no iOS-specific paths).

### What this does

Fixes two long-standing gaps in cursor `HOVER_ENTER`/`HOVER_LEAVE` for 3D entities:
1. **`hit` was always `None`** — SDK scripts reading `event.hit.position` saw null.
2. **`max_distance` / `max_player_distance` were ignored** on hover, diverging from DOWN/UP and the proto spec.

Eibriel's "distance fix" commit (`8036a63`) is the key improvement over the original approach: `current_hover` is now re-evaluated **every tick** based on range, so ENTER/LEAVE respond to distance changes while pointing at the same entity — not only to raycast target switching. This directly addresses the two issues Eibriel flagged in his earlier review.

### Review highlights

| Area | Verdict |
|---|---|
| **Correctness** | ✅ `current_hover` re-evaluation per tick makes `eq_key` fire on range boundary crossings, not just entity changes. LEAVE correctly skips re-checking range (being out of range *is* the reason we leave). |
| **Safety** | ✅ `.unwrap()` on `scenes.get_mut()` replaced with `if let Some(scene)` — no panic if scene is removed mid-frame. `entity_player_distance` returns `f32::MAX` for missing nodes (fail-closed). |
| **Consistency** | ✅ `event_info_in_range()` extracted and shared across hover events, click gating, and feedback overlay — all three now respect `max_player_distance` identically. Previously, clicks only checked `max_distance`. |
| **Hit on LEAVE** | ✅ `prev.hit` comes from `last_hover_entity`, which is updated each tick — so the hit is from the last in-range frame, not stale from initial enter. |
| **Performance** | ✅ One extra `entity_player_distance()` per tick (HashMap lookup + `get_global_position()`) — negligible in the pointer events path. `raycast.clone()` is a few floats + IDs. |
| **Sentry / logs** | ✅ No new error/warning logging added. |
| **Debug artifacts** | ✅ No debug prints, no `.claude/` files, no non-English comments. |

### Non-blocking edge case (for awareness, not blocking)

`has_active_hover_in_range` unions all hover entries' ranges into a single "active hover" state, but the per-transition emission checks (`emit_enter`/`emit_leave`) only verify event-type *presence*, not per-entry range. If an entity configured `HoverEnter(max_distance=3)` alongside `HoverLeave(max_distance=10)`, ENTER would fire at distance 5 (outside its own 3m range but inside the union). In practice this configuration is unlikely — and the previous code didn't respect range at all — so the current union approach is a pragmatic and significant improvement.

Solid fix. Well-commented code, clean design.

Ludmilafantaniella (commented):
Tested at parcel -31,19, Hover Range tab, all 3 probes.

### ✅ Validated

- [x] **maxDistance 3** - HOVER_ENTER/LEAVE fire correctly on distance-boundary crossing (hit dist ~2.92–2.98m)
- [x] **maxPlayerDistance 4** — HOVER_ENTER/LEAVE fire correctly on distance-boundary crossing (hit dist ~3.34–3.46m)
- [x] **default range** - HOVER_ENTER/LEAVE fire correctly on distance-boundary crossing (hit dist ~4.03–6.57m)
- [x] **Core fix confirmed**: enter/leave now track the distance boundary, not just raycast-target change - verified by walking backward while still aiming at the cube (gizmos disappear even with reticle on target)
- [x] **`hit` populated** - every ENTER/LEAVE shows real hit distance/position/normal, never `none`
- [x] **Regression - Pointer tab**: standard cube + proximity cubes behave unchanged, DOWN/UP clicks work end-to-end

### ⚠️ Could not fully verify

- **maxPlayerDistance gate isolation**: mobile has no independent camera zoom (only 1st/3rd person toggle), so camera-distance and player-distance are always coupled. Couldn't confirm the gate is specifically reading avatar distance vs. camera distance - behaves correctly from a user-facing standpoint either way.

### ❓ Open question (posted separately, see comment above)
Step 6 ("Clicks respect maxPlayerDistance") - Hover Range cubes don't show a click tooltip or respond to click input on mobile; only the Pointer tab cubes are clickable, and those don't have a maxPlayerDistance gate. Waiting on @Eibriel to confirm if this step applies to a different cube or isn't testable in the mobile Hover Range scene.

--

📎 Video + screenshots of each probe (enter/leave with hit data) attached below.

https://github.com/user-attachments/assets/a3f42c20-2a7d-4585-99c0-b6f6dfda0207

Image
Image
Image
Image
Image
Image

Ludmilafantaniella:
## 🔍 Request Changes / Findings

Tested Hover Range QA scene at -31,19 on mobile.

**Cubes 4, 5, 6 — OK**
- [x] All behave as expected with Far Camera checks

**Cube 3 — OK**
- [x] ~9.98m activation in both first and third person, consistent with default 10m range

⚠️ **Cube 1 (`maxDistance 3`) — needs confirmation**
- Only activates in first person. In third person, hover never activates even standing right next to the cube.
- This might be expected if `maxDistance` measures from the camera (third-person camera sits a few meters behind the avatar, potentially pushing it past the 3m threshold even when the avatar is adjacent to the cube) — but flagging to confirm this is the intended behavior, not a silent failure. @EibrielInv 

https://github.com/user-attachments/assets/91c61298-bb95-4c90-b005-9a2dbae3ad06

⚠️ **Cube 2 (`maxPlayerDistance 4`) — bug**
- Activation zone (where hover turns on/off) is consistent between first and third person, which is correct since this gate should read avatar distance.
- However, the reported hit distance value is inconsistent/unreliable between the two perspectives — doesn't match the expected ~4m boundary. The gate itself seems to work, but the distance value exposed in the hit looks wrong.

https://github.com/user-attachments/assets/5cb82bae-2b53-4300-bca8-07b4828232d2

Ludmilafantaniella (approved):
## ✅ Approve
Tested on mobile using the updated QA scene (with the new E-tooltip cube and camera position menu).

**Hover Range**
- [x] Cubes 3-6 behave as expected per the updated QA steps (distance gates, Far Camera checks, click respecting maxPlayerDistance)
- [x] Hover fires on distance boundary crossings, always with a populated hit
- [x] No regressions on the Pointer tab - DOWN/UP clicks still work normally

**Known issues (tracked separately, not blocking)**
Two findings from Cube 1 and Cube 2 (Hover Range) were confirmed as real bugs by @Eibriel:
- Cube 1 (`maxDistance 3`): only activates hover in first person - never activates in third person, even standing right next to the cube.
- Cube 2 (`maxPlayerDistance 4`): activation zone is consistent between first/third person (correct), but the reported hit distance value is inconsistent/unreliable between the two perspectives.

A new issue will be created to track these separately - confirmed out of scope for this PR, not blocking approval.

No blockers on this PR.

https://github.com/user-attachments/assets/f5b53fdd-fc52-4b1a-9089-030232c53a52

Image
Image

https://github.com/user-attachments/assets/87e7752f-e25a-4095-b629-f11ca42b460d

https://github.com/user-attachments/assets/baccba1b-f510-42d4-967f-309f12f3a672

https://github.com/user-attachments/assets/b33db886-5c3c-42b1-9f78-1fa5a9122124

EibrielInv (approved):
GW!

PR #2053, opened 2026-05-13
perf(dcl): split CRDT recv into wait + drain to reuse JS recv buffer

manuelmaceira:
## What

Replaces \`op_crdt_recv_from_renderer\` (returned \`Vec<Vec<u8>>\` via \`#[serde]\`, which made V8 allocate a fresh BackingStore per inner Vec on every recv) with two ops:

- **\`op_crdt_recv_wait\`** (async): waits on the renderer channel, builds a single framed \`Vec<u8>\` payload \`[u32 LE main_crdt_len][main_crdt bytes][data bytes]\`, stashes it in op_state, returns total length.
- **\`op_crdt_recv_drain\`** (sync \`#[fast]\`): copies the stashed bytes into the JS-owned \`#[buffer] &mut [u8]\` \`out\`. Returns 0 if \`out\` was too small (caller grows buffer + retries).

JS side keeps a 64 KB persistent buffer that grows on demand. \`decodeRecv\` reads the 4-byte length prefix to split off main_crdt + data without allocating new Uint8Arrays for the backing buffers.

## Why

Profile on Samsung A54 (Genesis Plaza) showed Thread-14 (V8 isolate of the GP scene) was at 31.9 % of CPU, with **24.1 % spent in \`Builtin_ArrayBufferConstructor\`** and **20.6 % in GC scavenge**. The bottleneck wasn't compute — V8 was thrashing the BackingStore allocator on every CRDT recv tick (one fresh allocation per Vec<u8> in the serde-returned payload).

Reusing the JS buffer means: 0 new BackingStores per tick after the first, drastically less GC pressure, V8 thread frees up CPU for the actual work.

| run | fps mean | frame_proc_ms |
|---|---:|---:|
| baseline (HEAD revert) | 14.48 | 74.7 |
| + async-png + TweenState | 17.09 | 63.9 |
| **+ CRDT recv split (this)** | **18.03** | **69.3** |

**+3 FPS / +20%** on the V8-thread-bound case. The frame_proc bumped 64 → 69 ms because the win is *upstream* of frame_proc — V8 catches up faster, so frame_proc starts seeing the work that was previously held back.

## Files

- \`lib/src/dcl/js/engine.rs\`: new op pair + \`PendingRecvBuffer\` op_state stash.
- \`lib/src/dcl/js/js_modules/EngineApi.js\`: 64 KB persistent buffer + grow-on-demand + frame-prefix decoder.

Extracted from \`feat/rendering-server-animations-1948\` commit \`6262c41e\` for isolated A/B benchmarking.

### DIFF
--- a/lib/src/dcl/js/engine.rs
+++ b/lib/src/dcl/js/engine.rs
@@ -42,8 +42,23 @@ use super::{
 
 // list of op declarations
 pub fn ops() -> Vec<OpDecl> {
-    vec![op_crdt_send_to_renderer(), op_crdt_recv_from_renderer()]
+    vec![
+        op_crdt_send_to_renderer(),
+        op_crdt_recv_wait(),
+        op_crdt_recv_drain(),
+    ]
 }
+
+/// Framed recv buffer produced by `op_crdt_recv_wait`, drained into a
+/// JS-owned `Uint8Array` by `op_crdt_recv_drain`.
+///
+/// Layout: `[u32 LE main_crdt_len][main_crdt bytes][data bytes]`.
+///
+/// Two ops because async ops can't take `#[buffer] &mut [u8]` (deno_core:
+/// the borrow can't span an `await`). The split lets JS reuse the same
+/// `Uint8Array` across every recv — no per-call V8 BackingStore allocation
+/// in steady state.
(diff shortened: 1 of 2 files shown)


kuruk-mm:
Already merged in main with another PR

PR #2054, opened 2026-05-13
perf: async impostor PNG save + TweenState emit-on-transition

manuelmaceira:
## What

Two CPU wins identified by post-async-png profiling on Samsung A54 (Genesis Plaza).

### 1. Async impostor PNG save (\`avatar_scene.rs\`)

Profile showed \`Image::save_png\` 13.8 % inclusive on the Android VkThread, with \`longest_match\` (zlib hot loop) at 6.4 % self-time — the synchronous encode was running on the render thread. Each impostor capture (256×512 RGBA) hit that path directly, pinning A54 at ~14 FPS regardless of GPU work.

Snapshot the RGBA8 mip0 on the main thread before \`generate_mipmaps\` (a 0.5 MB memcpy), then hand the bytes off to a tokio blocking worker that does the encode + atomic write (\`<path>.tmp\` → rename so a partial crash never leaves a half-written PNG on disk).

### 2. TweenState emit-on-transition (\`tween.rs\`)

\`TweenState\` was emitted to JS every frame (\`TsActive\` per-tick keep-alive). On the GP benchmark trace that meant **15162 dirty entries/frame** just for Tween updates — flooding the V8↔Rust CRDT pipeline.

Track \`last_emitted_state\` per Tween; only \`put()\` when the state transitions. For active tweens this drops to one initial put + one on completion — measured **15162 → 331 dirty entries/frame (-98 %)** on the benchmark trace.

| run | fps mean | frame_proc_ms |
|---|---:|---:|
| baseline (HEAD revert) | 14.48 | 74.7 |
| **+ async-png + TweenState (this)** | **17.09** | **63.9** |

**+2.6 FPS / +18%**. Pure CPU win — VkThread reclaims time previously stolen by zlib encode + CRDT thrash.

## Why bundled

Both fixes target the **CPU side of the same frame budget** (VkThread + V8 isolate respectively) and the bench A/B above was measured together. The original commit grouped them; splitting would just create two PRs that have to be benched together anyway.

## Files

- \`lib/Cargo.toml\` / \`Cargo.lock\`: adds the tokio blocking worker dep
- \`lib/src/avatars/avatar_scene.rs\`: async PNG path (+137)
- \`lib/src/scene_runner/components/tween.rs\`: transition-only emit (+32)

Extracted from \`feat/rendering-server-animations-1948\` commit \`4b35522f\` for isolated A/B benchmarking.

### DIFF
--- a/lib/Cargo.lock
+++ b/lib/Cargo.lock
@@ -899,6 +899,7 @@ dependencies = [
  "once_cell",
  "oslog",
  "paranoid-android",
+ "png",
  "poll-promise",
  "prost 0.11.9",
  "prost-build 0.11.9",
@@ -1341,6 +1342,15 @@ dependencies = [
  "simdutf8",
  "tokio",
  "utf-8",
+]
+
+[[package]]
+name = "fdeflate"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e6853b52649d4ac5c0bd02320cddc5ba956bdb407c4b75a2c6b75bf51500f8c"
+dependencies = [
+ "simd-adler32",
 ]
 
 [[package]]
@@ -2618,6 +2628,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2d80299ef12ff69b16a84bb182e3b9df68b5a91574d3d4fa6e41b65deec4df1"
 dependencies = [
  "adler2",
+ "simd-adler32",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


manuelmaceira:
Superseded by #2025 (perf stack). Both features of this PR are already folded into `feat/performance-maximizing`:

- **Async impostor PNG save** → `avatar_scene.rs::save_cached_texture_async()` (moves zlib encode off the render thread).
- **TweenState emit-on-transition** → `tween.rs` `last_emitted_state` gating (skips per-frame TweenState puts when state hasn't transitioned).

Closing as not-planned; the work ships via #2025.

Issue #2076, opened 2026-05-14
Asset pipeline: bake textures at multiple resolutions for runtime quality selection


kuruk-mm:
## Context
The offline asset pipeline currently bakes a single `.res` per texture at one fixed
quality — whichever `texture_quality` was set in the processor context at import time
(`lib/src/asset_server/processor.rs:480`, where `ctx.texture_quality.to_max_size()` caps
the baked texture). At runtime, `ContentProvider::fetch_texture_by_hash` has a fast-path
that loads this pre-baked `.res` from disk
(`lib/src/content/content_provider.rs`, optimized-asset branch in `fetch_texture_by_hash`),
bypassing the network and the decode step.

## Problem
Because there's only one baked variant, the optimized fast-path doesn't honor the runtime
`TextureQuality` setting:

- If a user has runtime quality `High` but the `.res` was baked at `Medium`, they
  silently get `Medium`. No warning, no re-decode, no way to opt out.
- The new `fetch_texture_by_hash_with_quality` / `fetch_texture_by_url_with_quality` APIs
  (`lib/src/content/content_provider.rs`) — which let callers ask for an explicit quality
  (`DclConfig.TEXTURE_QUALITY_LOW/MEDIUM/HIGH/SOURCE`) — cannot use the optimized
  fast-path. They always fall back to a fresh network download + decode, because the
  baked `.res` has unknown quality metadata and can't be trusted to match the request.
- This means any UI or scene component that wants Source / non-default quality pays the
  full bandwidth and decode cost on every load, even when an appropriately-sized variant
  could have been pre-baked.

## Desired behaviour
The asset pipeline should bake multiple resolutions per texture (e.g. a Low/Medium/High
variant set, plus the source bytes), record which quality each baked artifact represents,
and expose enough metadata for the runtime to pick the closest variant for a requested
`TextureQuality`. The runtime fast-path can then serve any quality from disk without
hitting the network.

## Touch points
- `lib/src/asset_server/processor.rs` — bake more than one variant per texture and persist
  per-variant quality metadata.
- `lib/src/content/content_provider.rs` — extend the optimized-asset lookup to choose a
  variant by requested quality; teach `fetch_texture_by_hash_with_quality` to use the
  fast-path when a matching baked variant exists.
- `lib/src/godot_classes/dcl_config.rs` — `TextureQuality` enum is the natural key for
  variant selection.

## Not in scope
- Choosing the exact set of resolutions to bake.
[...]

PR #2077, opened 2026-05-14
refactor(content): quality-parameterized texture fetch API + AsyncImage fallback detection

### LINKED ISSUE #2063: AsyncImage downloads textures at source resolution by default
## Summary
The reusable `AsyncImage` component (`godot/src/ui/components/async_image/async_image.gd`)
defaults its `quality` export to `Quality.ORIGINAL`. This makes every instance bypass
the global texture quality setting and decode the image at its source resolution.

## Where
- `godot/src/ui/components/async_image/async_image.gd` — `@export var quality: Quality = Quality.ORIGINAL`
- The `ORIGINAL` branch calls `Global.content_provider.fetch_texture_by_url_original(...)`,
  which forces `TextureQuality::Source` in `lib/src/content/content_provider.rs`
  (`fetch_texture_by_url_original`), skipping the resize step in `load_image_texture`.

## Impact
- **GPU memory.** A 2048×2048 thumbnail uses ~5.5 MB ETC2 on mobile / ~16 MB uncompressed
  on desktop, vs. ~170 KB at the default 512 px cap. The cost scales with the number of
  `AsyncImage` instances on screen (e.g. carousels).
- **Cache fragmentation.** `fetch_texture_by_url_original` appends `_original` to the cache
  key, so the same URL fetched at `ORIGINAL` and at `DEFAULT` lives in two separate cached
  promises — two downloads, two decodes, two GPU textures.

## Current call sites
- `snap_carousel_card` (`godot/src/ui/components/snap_carousel/snap_carousel_card.tscn`)
  instantiates `AsyncImage` without overriding `quality`, so every snap-carousel card is
  loading full-resolution images for a thumbnail-sized rect.

## Repro
1. Load any UI that hosts a snap carousel with remote images.
2. Inspect texture sizes via the Godot debugger or `image_not_supported`-style probes —
   textures arrive at the source URL's native resolution instead of the global
   texture-quality cap.

kuruk-mm:
## Summary
- Adds `TextureEntry.failed: bool` so callers can distinguish a real texture from the `image_not_supported.png` placeholder that `load_image_texture` returns for unsupported formats / decode errors. `AsyncImage` now checks `result.failed` and routes those cases to its no-image state instead of silently displaying the placeholder.
- Replaces `fetch_texture_by_{url,hash}_original` with `fetch_texture_by_{url,hash}_with_quality(file_hash, url|content_mapping, quality)`. `fetch_texture_by_url` becomes a thin wrapper that reads the global default from `DclConfig::static_get_texture_quality()` — cache keys converge on `{hash}_q{N}`, so the same URL fetched at the same quality is no longer downloaded twice.
- Exposes `DclConfig.TEXTURE_QUALITY_LOW/MEDIUM/HIGH/SOURCE` `#[constant]` items so GDScript callers don't pass magic numbers.
- Updates `dcl_ui_background.rs` (the only Rust caller of the former `_original` hash variant) to use `fetch_texture_by_hash_with_quality(..., TextureQuality::Source)` and the matching `{hash}_q3` `waiting_hash`.

## Why
- Reported: `AsyncImage` silently rendered the magenta/placeholder texture when the URL pointed at an unsupported image format. The fallback was hidden inside the content provider behind an `Ok(...)` result, so the GDScript layer had no way to detect failure. (See #2063.)
- The two existing entry points (`fetch_texture_by_url` and `fetch_texture_by_url_original`) used separate cache namespaces (bare hash vs. `_original` suffix), which meant the same texture could be downloaded and decoded twice when one site asked for the global default and another asked for source quality. Unifying on `{hash}_q{N}` keys avoids that.

## Out of scope
- The optimized-asset (`.res`) fast-path inside `fetch_texture_by_hash` is unchanged. It only honors a single baked quality variant; supporting per-quality variants requires asset-pipeline changes tracked in #2076.

Closes #2063

## Test plan
- [ ] Load a snap carousel with valid images — textures appear as before, no regression.
- [ ] Point an `AsyncImage` at a URL serving an unsupported format (e.g. AVIF) — verify the no-image state renders instead of the magenta placeholder.
- [ ] Confirm a UI background using `DclSourceTex::Texture` still loads (covers `fetch_texture_by_hash_with_quality` + `waiting_hash`).
- [ ] `cargo fmt --all`, `cargo clippy`, `cargo run -- check-gdscript` clean.

### DIFF
--- a/godot/src/ui/components/async_image/async_image.gd
+++ b/godot/src/ui/components/async_image/async_image.gd
@@ -124,7 +124,9 @@ func _async_download(url: String) -> void:
 	var url_hash := _get_hash_from_url(url)
 	var content_mapping
 	if quality == Quality.ORIGINAL:
-		content_mapping = Global.content_provider.fetch_texture_by_url_original(url_hash, url)
+		content_mapping = Global.content_provider.fetch_texture_by_url_with_quality(
+			url_hash, url, DclConfig.TEXTURE_QUALITY_HIGH
+		)
 	else:
 		content_mapping = Global.content_provider.fetch_texture_by_url(url_hash, url)
 	var result = await PromiseUtils.async_awaiter(content_mapping)
@@ -133,5 +135,8 @@ func _async_download(url: String) -> void:
 		_finish_with_error()
 		return
 	if not is_instance_valid(self):
+		return
+	if result.failed:
+		_finish_with_error()
 		return
 	set_texture(result.texture)
--- a/godot/src/ui/components/async_image/async_image.gd
+++ b/godot/src/ui/components/async_image/async_image.gd
@@ -6,11 +6,13 @@ extends Control
 
 signal image_loaded
 
-## ORIGINAL fetches at full resolution (bypasses global texture quality).
-## DEFAULT uses the global texture quality setting (typically Medium = 512px max).
-enum Quality { ORIGINAL, DEFAULT }
+## Optional per-instance texture quality override.
+## NONE → uses the global texture quality setting (DclConfig).
+## LOW/MEDIUM/HIGH/SOURCE values mirror DclConfig.TEXTURE_QUALITY_* — keep in sync with
+## TextureQuality in lib/src/godot_classes/dcl_config.rs.
+enum ForcedQuality { NONE = -1, LOW = 0, MEDIUM = 1, HIGH = 2, SOURCE = 3 }
 
-@export var quality: Quality = Quality.ORIGINAL
+@export var forced_quality: ForcedQuality = ForcedQuality.NONE
 @export var border_radius: int = 12
 @export var border_color: Color = Color("E8B9FF")
 @export var background_color: Color = Color(0.20784314, 0.03137255, 0.32941177, 0.5)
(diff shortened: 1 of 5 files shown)


regenesis-claw (approved):
Clean refactor — the API simplification and cache-key unification read well, and the `TextureEntry.failed` flag is a minimal, correct fix for #2063. A few observations (none blocking):

- **Cache-key migration**: existing entries cached under `{hash}_original` will miss after this lands, causing a one-time re-download for affected textures. Not a bug, just something to be aware of.
- **GDScript ↔ Rust enum sync**: `ForcedQuality` values in `async_image.gd` are kept in sync with `TextureQuality` via comments + the `DclConfig.TEXTURE_QUALITY_*` constants. If the Rust enum ever gains a variant, the GDScript side would silently fall behind — but the current approach is pragmatic and the comment makes the dependency clear.

The deleted `fetch_texture_by_url` duplication is a nice cleanup. `fetch_texture_by_url` as a thin wrapper over `fetch_texture_by_url_with_quality` converges the cache paths properly.

LGTM ✅

Requested by Mateo via Slack

leanmendoza (approved):

PR #2083, opened 2026-05-14
refactor(mobile): unified joystick input via gui_input

### LINKED ISSUE #1998: [Bug] pointer-filter: 'block' not working on mobile - click events pass through UI to 3D entities
## Device Information

- **Device:** iPhone 15
- **App Version:** 458

## Description

`pointer-filter: 'block'` is not working properly on mobile UIs. The pointer interaction goes through the UI, allowing users to interact with 3D entities behind it.

## Expected Behavior

Users should not be able to click 3D entities located behind a UI with `pointer-filter: 'block'`.

## Actual Behavior

When a UI element with `pointer-filter: 'block'` is open, pointer events still pass through to 3D entities behind it. The 3D entities remain interactive (hover text appears, click handlers trigger).

## Reproduction Steps

1. Jump into the "Dead Surge" scene deployed at `thecodingcave.dcl.eth`
2. Open the shop by clicking the NPC
3. While the shop UI is open, notice that the NPC behind the UI still receives pointer events
4. The NPC hover text ("Open Store") and pointer down interaction can still be triggered through the UI

## Screenshots

Screenshots/videos are available but need to be uploaded manually to this issue (GitHub API limitation).

---

Requested by Agustin Carriso via Slack

EibrielInv:
## Summary

Unifies mobile input handling on Godot's native `gui_input` + `MOUSE_FILTER_STOP` arbitration, and closes a gap where `pointerFilter: 'block'` UI didn't actually block 3D entity clicks.

**Joystick (original scope)**

- Migrate the virtual joystick from a global `_input` listener to `gui_input` on its `ActiveArea` Control with `MOUSE_FILTER_STOP`. Godot now handles touch routing natively, removing the custom HUD/scene-UI hit-tests (`_is_hud_ui_at_position`, `_is_scene_ui_at_position`, `_is_point_inside_joystick_area`, `_check_children_for_pointer_control`).
- Fix camera-lock in `PlayerMobileInput`: a second finger landing on the joystick capture area is no longer mis-classified as a two-finger gesture. `gui_input` runs after `_input`, so we probe `ActiveArea.get_global_rect()` directly. Also tracks chat touch indices for clean release and resets `_two_fingers` when either touch lifts.
- `ActiveArea` now owns its lifecycle: `MOUSE_FILTER_STOP` normally, `MOUSE_FILTER_IGNORE` while a Godot-side menu is open.
- Include chat notifications in `ChatPanel.is_interactive_area_at` so taps on visible notifications don't bleed through as camera drags.

**Camera-drag catcher (symmetric to the joystick)**

- Add `MobileCameraInput` — a full-screen `Control` at child index 0 of `%UI`, `mouse_filter = STOP`. Owns mobile camera drag, two-finger detection, and cinematic-mode `ia_pointer` emission via `gui_input`. Godot's native input arbitration now answers "is UI on top?" for camera-drag the same way the joystick already does.
- **Reorder `%UI` children** so `SceneUIContainer` sits *after* `SafeMarginContainerHUD`. Godot's reverse-tree-order walk now reaches Scene UI's STOP Controls before the joystick's `ActiveArea`, so interactive Scene UI (`UiBackground` with `PointerEvents`, or `pointerFilter='block'`) blocks the joystick instead of losing to it. Visual order is unchanged — Scene UI's `z_index = CANVAS_ITEM_Z_MIN + 1000` keeps it rendered behind HUD.
- Set `focus_mode = None` on the joystick's `Button_Camera` so tapping it no longer captures keyboard focus and silently swallows subsequent `Input.action_press` calls (which broke the joystick + camera rotation after any camera-mode toggle until another control re-grabbed focus).

**`pointerFilter: 'block'` actually blocks the raycast**

- The 3D physics raycast in `SceneManager::get_current_mouse_entity` doesn't consult the Godot UI tree, so before this change a block UI failed to stop clicks reaching world entities whenever `ia_pointer` fired from a path that bypasses `gui_input` (mobile joypad interact button, desktop key, gamepad button).
- Add a thread-local `UiBlockerRegistry` maintained by `DclUiControl::update_mouse_filter`: every transition into `MOUSE_FILTER_STOP` inserts the Control, every transition back removes it. The set is consulted in `get_current_mouse_entity` right after computing `screen_point`; if any registered, visible blocker contains the point, we return `None` so no `PetDown` fires. O(blockers) per frame instead of O(all Scene-UI controls).
- No `exit_tree` handler — `scene_ui` reparents Controls during initial tree reorganization, which fires `exit_tree` without the Control being destroyed. Lazy `is_instance_valid()` pruning in `blocks_point` handles the real free path correctly.

**Cleanup**

- Delete the now-unreachable HUD hit-test chain: `chat_panel.is_interactive_area_at`, `chat.is_interactive_area_at`, `chatbar.is_point_inside`, `notifications.is_point_over_notification`.
- Inline the camera-mode-set listener into `Player`. `PlayerMobileInput`'s only remaining job after the catcher was forwarding `Global.camera_mode_set`, identical to `PlayerDesktopInput`'s. Delete `player_mobile_input.gd` entirely and drop the duplicate handler from `player_desktop_input.gd`.
- Remove `_on_ui_root_gui_input` from `explorer.gd`; the catcher subsumes its cinematic-pointer flow.

### Joystick
- [ ] On mobile, joystick movement works in FIXED and DYNAMIC modes.
- [ ] Camera rotation works when touching the non-joystick area.
- [ ] Tapping/dragging on chat or chat notifications no longer rotates the camera.
- [ ] **Regression**: start rotating the camera with finger A, then drop finger B on the joystick — camera should keep rotating, no lock-up.
- [ ] Opening/closing a Godot-side menu correctly toggles `ActiveArea`'s `mouse_filter`.

### Camera-drag catcher
- [ ] Drag in empty world space rotates the camera as before; clamps in first/third person match prior behavior.
- [ ] Cinematic / virtual-camera mode: tapping a 3D entity still fires `ia_pointer` and the scene receives the pointer event.
- [ ] After tapping `Button_Camera` to switch between first/third person, the joystick and camera-drag both keep working (focus_mode regression).
- [ ] Tapping a Scene UI element (any `UiBackground` with `PointerEvents`) no longer also rotates the camera.

### `pointerFilter: 'block'` and Scene UI input priority
- [ ] Place a `UiBackground` with `pointerFilter='block'` over the crosshair on mobile non-cinematic, then press the joypad interact button. The 3D entity behind the UI must **not** receive `onMouseDown`.
- [ ] Negative control: remove the block (or set `pointerFilter='none'`) — entity should fire as before.
- [ ] Desktop: clicking a 3D entity through a block UI must not fire it.
- [ ] Scene UI with `onMouseDown` continues to fire on tap; joystick does not also activate.
- [ ] Scene UI scroll/drag inside the joystick's bottom-left ActiveArea now works (Scene UI wins arbitration over the joystick).

Closes #1998


kuruk-mm (approved):
looks great!

Issue #1328, opened 2026-01-26
Modal: Cross Platform Disconnection Modal


rodrigo685:
Adjust modals to new design provided by Eli

<img width="1270" height="702" alt="Image" src="https://github.com/user-attachments/assets/14c94436-42fc-4b93-a56c-a8615dc3e76f" />

PR #2085, opened 2026-05-15
fix: route DuplicateIdentity disconnect to modal + sign-out flow

### LINKED ISSUE #1742: [Bug]: Multi-device login with same account not disconnecting previous session
### Platform
Mobile (Android) and Unity client interaction

### App Version
v318

### Device Information
iOS // Android

### Issue Description
When logging in with the same account on both mobile and Unity client simultaneously, the system should disconnect the previous session but instead both sessions remain active, causing synchronization issues between clients.

### Expected Behavior
When logging in with the same account from a different device/client, the previous session should be automatically disconnected to prevent conflicts.

Image

### Actual Behavior
Both sessions remain active with the following issues:
- **Mobile client**: Appears to behave normally (needs further testing for live streams, games, etc.)
- **Unity client**: Other users flash in and out of existence
- **Other users' perspective**: The avatar appears and disappears repeatedly, though communication still works
- **Chat functionality**: 
  - Messages typed on mobile do not display
  - Messages typed in Unity client do show up in chat

### Steps to Reproduce
1. Log in to Decentraland using the same account on mobile client
2. Log in to Decentraland using the same account on Unity client (without logging out from mobile)
3. Observe that both sessions remain active
4. Notice avatar synchronization issues and chat message inconsistencies

### Occurrence
Not specified

---
Requested by Rodri via Slack

kuruk-mm:
Fixes #1742, #1328

## Summary

- **`DuplicateIdentity` silently dropped.** LiveKit sends `Disconnected` with `H160::zero()`, so `MessageProcessor` filtered it as a non-peer event — no UI, scene room stuck in a 5s reconnect loop. Now routed as a room-level event, parallel to `RoomMetadataChanged`.
- **Disconnect overlay → `ModalManager`.** Replaced the hand-rolled overlay with `async_show_session_ended_modal` (SIGN IN/EXIT) and room-closed/disconnected variants (RECONNECT/EXIT). SIGN IN reuses `Global.sign_out()`.
- **Guard `SceneManager`** against a freed `player_avatar_node` after `change_scene_to_file`.

## Test plan

- [ ] Dual sign-in → modal + SIGN IN routes to lobby, no panic
- [ ] Transient disconnect → RECONNECT/EXIT works
- [ ] Ban-kicked modal unchanged


regenesis-claw (approved):
Clean fix — good root-cause analysis on the `H160::zero()` filtering, and the modal migration + SceneManager guards are solid.

**What I checked (following REVIEW.md):**

### Pre-flight
- ✅ **iOS build**: present and green (PR touches `lib/src/comms/`).
- ⚠️ **Branch is 5 commits behind main** (protocol bump, UI scroll, cleanup). None overlap with comms/disconnect/modal code, so the diff is still trustworthy — but worth rebasing before merge.

### Findings

All clear — no blockers.

1. **Rust – early `Disconnected` routing** (`message_processor.rs`): correctly moves the disconnect-reason capture above the `is_player_address` gate so `H160::zero()` room-level events are no longer silently dropped. The old match arm is now documented as unreachable and kept as a no-op — fine for safety, though it could be replaced with a `tracing::warn!` if it ever fires unexpectedly.

2. **Rust – `player_avatar_node` validity guards** (`scene_manager.rs`): two `is_instance_valid()` checks prevent dangling-ref crashes after `change_scene_to_file` during sign-out. Both are in the right spots (tick entry and the pointer/raycast block in `INode`).

3. **GDScript – disconnect overlay → ModalManager**: the ~80-line hand-rolled `ColorRect` overlay is replaced with three clean delegates to `modal_manager`. Signal wiring in `_ready()` is safe here — `DisconnectHandler` is a scene node, not an autoload, so `Global.modal_manager` is already initialized.

4. **GDScript – `modal_manager.gd` additions**: the private `_async_show_disconnect_modal` helper + `_disconnect_button_signals()` call before reconnecting is the correct pattern to avoid stacking handlers. `blocker = true` is appropriate for session-ending modals.

5. **REASON_KICKED**: not handled in `_show_disconnect_modal` — verified this is intentional since kicked/banned users are routed through the ban modal path earlier in `_on_disconnected`.

Approving. Consider rebasing onto current main before merge.

Requested by Mateo via Slack

PR #2120, opened 2026-05-19
feat(preview): scene stats / limits overlay for creators

### LINKED ISSUE #2410: Creator Hub: Scene Performance Limit Plugin
## Problem: 
In Mobile, content creators don't have clear scene performance indicators or caps to optimize against as they develop.

## Scope
Display a plugin in creator hub showing current scene measures and recommended limits (plus colours to indicate status), so creators can be aware of their scene limitations and optimize for mobile

## Acceptance Criteria
- [ ] In Creator Hub Mobile Preview Mode, a plugin is available to see the scene limits defined in #2368 as the creator develops.

### LINKED ISSUE #2451: [Bug] Mobile Preview: Debug Tools and Reload Buttons Missing
### Platform
Mobile Preview

### App Version
v1138

### Device Information
IPHONE 16E

### Issue description
In Mobile Preview, the Debug Tools and Reload buttons are missing from the UI.

### Expected behavior
Debug Tools and Reload buttons should be visible and accessible in Mobile Preview, as before.

Image

### Screenshots/Media
N/A

### Steps to reproduce
1. Open Mobile Preview
2. Look for Debug Tools and Reload buttons

### Occurrence
Always

---
Requested by Rodri via Slack

kuruk-mm:
## What

An **opt-in** on-screen **Scene Stats / Limits overlay** that gives content creators live feedback on how the previewed scene measures against resource budgets. It grew out of the Genesis Plaza performance audit (#2368), where scenes were shown to massively overshoot reasonable budgets (≈3 GB textures, 1.3k GltfContainers, 3.5k colliders) with no signal to the author.

**Enabled in preview** (`preview=` deeplink / preview realm), or in any realm via the **`scene-stats=true` deep link** (e.g. `decentraland://open?position=0,0&scene-stats=true`) — in **every build flavor, production included**, so creators can measure scenes on store builds. Without one of those triggers nothing is instantiated, so the zero-cost guarantee holds.

Toggle it with the **red monitor button** placed to the left of the camera (1st/3rd-person) button. Each metric is a progress bar:

- **green** — under the soft (recommended) limit
- **yellow** — between soft and hard
- **red + overpass %** (e.g. `200%`) — over the hard limit

## Metrics

**Per scene** (collected by walking the scene's `DclSceneNode` subtree, dedup by instance id with a per-mesh triangle cache; entities from `debug_list_entities`):
triangles · entities · bodies · geometries · materials · textures · colliders · **content size** · **external content** (url-textures + JS `fetch()` bytes, from the ResourceProvider's in-memory cache metadata).

**Whole app** (labeled as *not per-scene* — Godot shares one viewport):
memory (real process RSS) · draw calls · FPS.

## Limits

Defaults in `scene_limits.gd` (`FIXED_ABSOLUTE` model, seeded from the #2368 audit — **tunable, flagged for a review pass**). Bars go **green** under soft, **yellow** between soft and hard, **red + overpass %** above hard.

| Metric | Scope | Soft (warn) | Hard (max) |
|---|---|---:|---:|
| Triangles | per-scene | 1,000,000 | 1,200,000 |
| Entities | per-scene | 4,800 | 6,000 |
| Meshes (bodies) | per-scene | 2,400 | 3,000 |
| Geometries | per-scene | 1,000 | 2,000 |
| Materials | per-scene | 400 | 500 |
| Textures | per-scene | 400 | 500 |
| Colliders | per-scene | 1,200 | 1,500 |
| Content size | per-scene | 120 MB | 150 MB |
| External content | per-scene | 40 MB | 50 MB |
| Memory (process RSS) | whole-app | 1,638 MB | 2,048 MB |
| Draw calls | whole-app | 1,000 | 2,000 |
| PERFORMANCE *(inverse: higher is better)* | whole-app | 50 | 30 |

- **FPS** is an *inverse* metric: green at ≥ 50, yellow between 30–50, red below 30; its bar full-scale is computed at runtime (`dynamic_max`), and no overpass % is shown.
- **Memory** is a single unified row (phones have unified memory — GPU allocations live in the same pool), read as real process RSS via a passive Rust getter (`SceneManager.get_process_memory_mb()`); `Performance.MEMORY_STATIC` is compiled out of release/mobile templates and reported 0.
- A `PER_PARCEL` model (the canonical Decentraland per-parcel formula: 10k triangles / 200 entities / 300 bodies per parcel, log2-scaled materials/textures/geometries) is stubbed in `scene_limits.gd` but not active.

## Content size = real on-disk footprint
`content_size` measures what actually lives in `user://content` for the scene's content-mapping hashes, not the manifest size. A scene asset persists as raw `{hash}` (textures/audio/crdt/js), a runtime-processed `{hash}.scn` (the raw glb is *deleted* after processing — so the earlier raw-only sum under-reported to near-zero), or an optimized `{hash}-mobile.zip` pack (mobile: assets are mounted from the pack, no separate `.scn` on disk). The scan adds every file whose leading hash is in the mapping — the on-disk forms are mutually exclusive per hash, so nothing double-counts. Recomputed each tick, so it converges as assets download.

## Zero cost when not enabled (hard constraint)

The entire system — metric collection **and** UI — is instantiated **only when opted in**, behind `_is_in_preview_realm() or Global.deep_link_obj.scene_stats` in `explorer.gd:_update_scene_stats_ui()`. It works in **every build flavor including production** (so creators can measure scenes on store builds), but a normal run without the preview realm or the deep link creates nothing: no nodes, no timer, no polling, no allocation. The only Rust additions are passive: the `scene-stats` deep-link field and the `get_process_memory_mb()` getter — no default code path touched. The debug-tools row is hosted in a `SafeMarginContainer` (`use_top`/`use_right` with 110/12 px floors), so the overlay stays inside device safe-area insets (notch/punch-hole).

## Deep-link enable (`scene-stats=true`)

Plumbed the standard way (mirrors `safe_margin_debug`): pure-Rust `DeepLinkResult.scene_stats` field + `"scene-stats"` parse arm + unit tests in `lib/src/deep_link.rs`; `#[var] scene_stats: bool` in `lib/src/godot_classes/dcl_parse_deep_link.rs`, read as `Global.deep_link_obj.scene_stats`. Works for both `--fake-deeplink` (CLI) and real mobile deeplinks; retargets on scene change via the existing `_on_change_scene_id` path.

## Files

- **New** `godot/src/ui/components/utils/scene_limits.gd` — tunable fixed-absolute limit table (seeded from #2368) + a per-parcel model stub.
- **New** `godot/src/ui/components/utils/scene_stats_collector.gd` — per-scene subtree walk + on-disk content size + global Performance/RSS stats.
- **New** `godot/src/ui/components/organisms/scene_stats_panel/{scene_stats_panel.gd,.tscn}` — the overlay organism (bars, in-bar value text, red monitor toggle).
- **New** `godot/assets/ui/scene_stats_monitor.svg` (+ `.import`) — toggle icon.
- **Modify** `godot/src/ui/components/organisms/debug_panel/debug_panel.{gd,tscn}` — renamed "Console"; its Console/Reload-Scene buttons dock into the navbar next to the camera button; shares the right-aligned debug-tools row with the stats panel.
- **Modify** `godot/src/ui/explorer.gd` + `explorer.tscn` — preview/deep-link-gated instantiate/free + `_preview_scene_id()` + hook in `_update_preview_ui`; `%HBoxContainer_DebugTools` row hosting both tools.
- **Modify** `lib/src/deep_link.rs`, `lib/src/godot_classes/dcl_parse_deep_link.rs`, `godot/src/global.gd` — the `scene-stats=true` deep-link parameter.
- **Modify** `lib/src/tools/memory_monitor.rs`, `lib/src/scene_runner/scene_manager.rs` — passive `get_process_memory_mb()` (RSS on linux/android via `/proc/self/statm`, `phys_footprint` on iOS, `mach_task_basic_info` on macOS).

## Test plan (mobile)

Works in **any build flavor** (dev, staging, and prod) — the overlay only appears via the deeplinks below. Grab the Android APK from this PR's build-report comment, or the TestFlight build (`build` label). Open the links below **on the phone** (they are universal links), or push one via adb:

```bash
adb shell am start -a android.intent.action.VIEW -d "'https://mobile.dclexplorer.com/?preview=raft.dcl.eth'"
```

### 1. Preview Worlds (overlay auto-shows)

`preview=<name>.dcl.eth` loads the World **as the preview realm**, which is what enables the overlay — no extra param needed.

- [ ] `https://mobile.dclexplorer.com/?preview=dclexperience2.dcl.eth`
- [ ] `https://mobile.dclexplorer.com/?preview=raft.dcl.eth`
- [ ] One or two more Worlds of the tester's choice (`https://mobile.dclexplorer.com/?preview=<world>.dcl.eth`) — pick scenes of different weight to compare readings.

For each: the overlay appears on load; the red monitor button toggles it (navbar order [Console] [Reload Scene] [monitor] [camera] [profile]); rows refresh (~1.5 s); over-hard rows go red with an overpass % + ⚠.

### 2. Compare against Genesis Plaza (forced overlay outside preview)

- [ ] `https://mobile.dclexplorer.com/?position=0,0&scene-stats=true` — overlay shows in Genesis City on the dev build; compare its numbers with the Worlds above.

### 3. Row-level checks (on any of the scenes above)

- [ ] **Content size** is non-zero and converges as assets download (counts the `-mobile.zip` optimized packs — regression: used to read ~0 on mobile).
- [ ] **External content** populates on scenes that pull url-textures / `fetch()` (Genesis Plaza reads ~54 MB) and stays 0 on scenes that don't.
- [ ] **Memory** shows real process RSS (regression: read 0 on mobile) and is the same across scenes — it's a whole-app row.
- [ ] Per-scene rows (triangles/entities/meshes/…) differ between the scenes — confirms per-scene scoping.
- [ ] Performance (FPS) row inverse coloring: green ≥ 50, yellow 30–50, red < 30, no overpass %.

### 4. Gating & safe area

- [ ] Genesis City **without** `scene-stats=true` → no overlay and no monitor button.
- [ ] Production/store build: the deeplinks above DO show the overlay (works in every flavor); without them, nothing is instantiated.
- [ ] Notched / punch-hole device: the stats row sits fully inside the safe area in landscape (not under the camera cutout).

## Notes

- Limit numbers in `scene_limits.gd` are **tunable defaults** flagged for a review pass.
- Global metrics (memory/draw calls/FPS) are whole-app, not per-scene; in a single-scene preview they closely track the scene + its avatars.

image


leanmendoza:
### Review: overlaps with existing code

The panel/UX and the per-scene *budget* framing are genuinely new, and the deep-link plumbing is clean. But a good chunk of the collection logic reimplements code that already exists — flagging it before it becomes a 4th copy.

**1. Process-memory reader (Rust) — near copy-paste.** `used_memory_mb()` re-writes all 4 platforms, each already present:
- iOS `phys_footprint` via `proc_pid_rusage(RUSAGE_INFO_V2)` → identical to `read_footprint_mb()` in the *same file* (`memory_monitor.rs:276`).
- Android/Linux `/proc/self/statm` RSS → same as `read_available_mb()` (`memory_monitor.rs:299`).
- macOS `mach_task_basic_info`/`task_info` → near-verbatim copy of `benchmark_report.rs:569` `get_process_memory_usage_mb` (same struct, externs, `MACH_TASK_BASIC_INFO = 20`).

Suggest one shared `used_memory_mb()` in `memory_monitor.rs` and have `benchmark_report` call it. The new `get_process_memory_mb()` `#[func]` is fine — no RSS entry point to GDScript existed before.

**2. Scene-resource walk (GDScript) — already implemented ~3x.** Mesh/material/texture counting via subtree walk + `Performance.get_monitor(...)` exists in `addons/dcl_dev_tools/.../resource_counter/resource_counter.gd`, `tools/gp_benchmark_runner.gd` (`_walk_for_dedup`/`_count_node_types`), persisted by `benchmark_report.rs`. Whole-app HUD (FPS/draw calls/VRAM/mem) is also `tool/performance_monitor.gd` (`class_name PerformanceMonitor`), just not wired into the app. Per-scene entity count already lives in `tool/debug_server/debug_collector.gd:32` (`collect_scenes_summary`). Genuinely new: triangles, colliders, and single-scene scoping — could be layered onto `DebugCollector`/`resource_counter` instead of a new walker.

**3. Content size (GDScript) — re-scans disk each tick.** `content_bytes()` re-stats `user://content` every 1.5s, but `resource_provider.rs` already holds per-file sizes in memory (`FileMetadata.file_size`, `existing_files`, `total_size()`), exposed as `get_cache_folder_total_size()`/`get_cache_size_mb()` (`content_provider.rs:1865/1871`, used in `settings.gd:406`). Only the per-scene filter is new — better as a per-scene accessor on `ResourceProvider` summing cached sizes for the scene's hashes.

Nit: `label_ram` in `explorer.gd` is dead (only ever `.visible = false`) — good candidate to retire alongside this.

Net: the feature is worth having; the ask is to build it on the existing collectors rather than duplicating them (memory reader is the quickest win).

leanmendoza (changes requested):
See comment above

rodrigo685:
Triangles--> Let's go for easy to follow/memorable numbers
- Soft: 1M
- Hard 1.2 M

rodrigo685 (approved):
Looking great product side, super useful plug-in

leanmendoza (approved):

PR #2148, opened 2026-05-20
feat: populate screen_inset_area on PbUiCanvasInformation

kuruk-mm:
## Summary
- `PbUiCanvasInformation.screen_inset_area` was being sent as `None` ever since the protocol bump added the field (see d8f7c0741c). Scenes that read it currently get nothing.
- Mirror the same `BorderRect` we already compute for `interactable_area` into `screen_inset_area` so scenes can use it.
- The two values match today (the explorer's safe area is the only inset we know about). If hardware-imposed insets diverge from the HUD-driven interactable area in the future we can split them.

## Test plan
- [ ] `cd lib && cargo check` — verifies the struct field is populated.
- [ ] Run the client and confirm scenes that consume `UiCanvasInformation.screenInsetArea` (notch/safe-area aware SDK7 UIs) receive non-null values matching `interactableArea`.
- [ ] Resize the window / toggle HUD elements — both rects should update together.


EibrielInv (approved):
Good work!!

PR #2157, opened 2026-05-21
feat(sentry): attach user, context, tags, and dist to events

kuruk-mm:
## Summary

Enriches Sentry events with user identity, contextual state, and filterable tags so issues can be triaged by who/what/where.

- **User identity** — `SentrySDK.set_user(...)` was missing (a `SentryUser` was built in `global.gd` and dropped). Now wired. `user.id` stays pinned to `analytics_user_id` for the lifetime of the install; `user.username` tracks the wallet address after `wallet_connected` and the display name after `profile_changed`, cleared on logout. Keeping `id` stable means "Users affected" stays attributed to one install across auth changes.
- **Contexts** — `graphics` (adapter name/vendor/version) at init; `realm` and `location` (parcel + scene entity id) refresh from existing signals.
- **Tags** — `platform`, `build_type`, `gpu_vendor`, `locale` at init; `realm`, `comms_adapter`, `is_guest` refreshed from signals.
- **`dist`** set to `<platform>-<commit_hash>` so cross-platform builds of the same release don't collide.
- `send_default_pii = false` pinned explicitly.

All wiring lives inside the existing `telemetry_enabled` gate, so asset-server mode and the `disable_telemetry` feature still skip Sentry. No new files — edits confined to `project_main_loop.gd` and `global.gd`.

## Test plan

Staging build (dev is dropped by `_before_send`):

- [ ] Captured error → event has `user.id = <analytics_user_id>` and a `graphics` context card.
- [ ] Connect wallet → `user.username` = address, `is_guest:false`. Sign out → `username` cleared, `is_guest:true`, `id` unchanged.
- [ ] Change realm → `realm` context + tag reflect new realm.
- [ ] Walk to new parcel → `location` context shows new parcel + scene entity id.
- [ ] Tag filters present: `platform`, `build_type`, `gpu_vendor`, `locale`, `is_guest`, `realm`, `comms_adapter`.
- [ ] Same commit on two platforms → distinct `dist` values under one release.
- [ ] `cargo run -- run --features asset_server` boots cleanly.


regenesis-claw (approved):
Clean, well-scoped PR — the dropped `SentryUser` was a real gap and the context/tags additions will make triage significantly easier.

**One minor observation (non-blocking):**

In `_sentry_on_profile_changed`, if `new_profile` is null or `get_name()` returns empty, the `sentry_user.username` is left unset — which replaces whatever `_sentry_on_wallet_connected` had stored (the wallet address). So for wallets with no display name set yet, you'd briefly lose the address as username. Not a big deal since `user.id` stays stable and `is_guest` is tracked as a tag, but a fallback like keeping the previous address when display name is empty could be nice.

Everything else looks solid:
- `user.id` pinned to `analytics_user_id` across auth changes ✅
- Signal connections are inside the `telemetry_enabled` gate and after the connected nodes are ready ✅
- `send_default_pii = false` pinned explicitly ✅
- `dist` disambiguates cross-platform builds ✅
- No debug prints, no `.claude/` files, no platform-specific paths needing iOS build ✅

LGTM — approving. CI (Linux/Android) still running but static checks + clippy are green.

Requested by Mateo via Slack

leanmendoza (changes requested):
Could we move _sentry funcs to SentrySeeder?
LGTM anyways

PR #2168, opened 2026-05-26
fix: AvatarModifierArea excludeIds reliably excludes remote avatars (#2166)

### LINKED ISSUE #2166: [Bug]: AvatarModifierArea excludeIds do not correctly affect remote avatars on mobile
### 🕹️Platform

iOS

### 🔢App Version

490

### 📱Device Information

Iphone 15

### 📄Issue description

On mobile, AvatarModifierArea with AMT_HIDE_AVATARS works, but excludeIds does not appear to be applied correctly to remote avatars.

The base hide behavior works as expected. However, when players are included in excludeIds, those remote avatars are not reliably restored/shown on mobile clients.

This makes excludeIds unreliable for multiplayer gameplay logic where some avatars should remain visible while others are hidden.

**Additional Information**

- The issue appears to be specifically related to excludeIds, not to AMT_HIDE_AVATARS itself.
- AMT_HIDE_AVATARS works correctly when all avatars are hidden.
- The problem appears when using excludeIds to keep selected players visible.
- In practice, this can look like the exclusion logic works for the local avatar, but not correctly for remote avatars.

### ✅Expected behavior

Any avatar included in excludeIds should remain visible, including remote avatars.

### 🖼️Screenshots / Media

In carrito.dcl.eth, there is a large AvatarModifierArea covering the whole parcel. Before the game starts, the system adds all player avatar IDs to excludeIds so everyone should remain visible to each other.

As shown in these screenshots, I am connected from two different mobile clients, and they still cannot see each other. The excludeIds logic appears to work for the local player’s own avatar, but it does not correctly respect the other players.

Image
Image

### ▶️Steps to reproduce

1. Open a scene using AvatarModifierArea with AMT_HIDE_AVATARS on mobile.
2. Include one or more players in excludeIds.
3. Observe avatar visibility from a mobile client.
4. Compare local avatar behavior vs remote avatar behavior.

### 👁️Occurrence

100% of the time

manuelmaceira:
Fixes #2166

## Summary

`AvatarModifierArea.excludeIds` was failing to keep remote avatars visible on mobile — three independent gaps were stacking up to make the failure look flaky:

1. **Address casing** — profiles delivered over comms can carry checksummed eth addresses while scenes typically push lowercase. Direct `==` comparison missed those. Extracted into `AvatarExcludeIdMatcher` with normalization (lowercase + tolerant `0x` prefix).
2. **Race on profile arrival** — a remote avatar could enter the area before its profile (and thus `avatar_id`) was populated, hiding it permanently. `async_update_avatar_from_profile` now calls `try_show()` after `avatar_id` changes so the area is re-evaluated.
3. **Stale excludeIds on already-overlapping avatars** — when the scene updates `excludeIds` after avatars are already inside the area, the Rust `set_exclude_ids` setter alone doesn't retrigger detection. `avatar_modifier_area.rs` now invokes `refresh_overlapping_detectors`, which iterates `Global.avatars` (rather than `Area3D.get_overlapping_areas()`, whose state is stale mid-CRDT-tick) and asks each avatar to re-evaluate via `try_show()`.

In `carrito.dcl.eth` (1 large AMA + scene adds player IDs to excludeIds) two clients now consistently see each other. Without this fix the user that joined first never saw the second.

## Files

- `godot/src/decentraland_components/avatar/avatar_exclude_id_matcher.gd` — new pure helper with `is_excluded(avatar_id, exclude_ids) -> bool`.
- `godot/src/decentraland_components/avatar/avatar.gd` — `_on_set_avatar_modifier_area` uses the matcher; `async_update_avatar_from_profile` calls `try_show()` after `avatar_id` changes (both wire-profile and AvatarShape branches).
- `godot/src/decentraland_components/avatar_modifier_area.gd` — new `refresh_overlapping_detectors()` that iterates `Global.avatars`.
- `lib/src/scene_runner/components/avatar_modifier_area.rs` — calls `refresh_overlapping_detectors` after updating an existing area's properties.
- `godot/src/test/avatar/test_avatar_modifier_exclude_ids.gd` — regression test covering case/prefix mismatches and the empty-`avatar_id` race.

## Test plan

- [x] `cargo run -- full-tests --skip-visual` passes (cargo fmt/clippy, gdformat/gdlint, gdscript validation, rust + integration tests)
- [x] Regression test passes:
  ```
  .bin/godot/godot4_bin --headless --path godot \
    --script res://src/test/avatar/test_avatar_modifier_exclude_ids.gd
  # → [test_avatar_modifier_exclude_ids] PASS (all 7 cases)
  ```
- [x] Manual repro on `carrito.dcl.eth`, two clients (macOS + iOS): both see each other regardless of join order.
- [ ] CI green

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar.gd
+++ b/godot/src/decentraland_components/avatar/avatar.gd
@@ -263,9 +263,8 @@ func try_show():
 func _on_set_avatar_modifier_area(area: DclAvatarModifierArea3D):
 	_unset_avatar_modifier_area()  # Reset state
 
-	for exclude_id in area.exclude_ids:
-		if avatar_id == exclude_id:
-			return  # the avatar is not going to be modified
+	if AvatarExcludeIdMatcher.is_excluded(avatar_id, area.exclude_ids):
+		return  # the avatar is not going to be modified
 
 	for modifier in area.avatar_modifiers:
 		if modifier == 0:  # hide avatar
@@ -348,6 +347,7 @@ func async_update_avatar_from_profile(profile: DclUserProfile):
 		new_avatar_name += "#" + profile.get_ethereum_address().right(4)
 	nickname_ui.name_claimed = profile.has_claimed_name()
 
+	var avatar_id_changed := avatar_id != profile.get_ethereum_address()
 	avatar_id = profile.get_ethereum_address()
 	has_connected_web3 = profile.has_connected_web3()
 	prints("Async update avatar from profile", avatar_id)
@@ -355,6 +355,12 @@ func async_update_avatar_from_profile(profile: DclUserProfile):
 	# Update metadata with the new avatar_id
 	if click_area:
 		click_area.set_meta("avatar_id", avatar_id)
+
+	# Re-evaluate AvatarModifierArea exclusion: the area may have triggered
+	# before the profile arrived (issue #2166 race), leaving the avatar hidden
+	# even though its id is in excludeIds.
+	if avatar_id_changed:
+		try_show()
 
 	await async_update_avatar(avatar, new_avatar_name)
 
@@ -378,6 +384,7 @@ func async_update_avatar(
 			avatar_id = shape_id
 			if click_area:
 				click_area.set_meta("avatar_id", avatar_id)
+			try_show()
 
 	# Update metadata for raycast detection
 	if click_area:
--- a/godot/src/decentraland_components/avatar/avatar_exclude_id_matcher.gd
+++ b/godot/src/decentraland_components/avatar/avatar_exclude_id_matcher.gd
@@ -1,0 +1,23 @@
+class_name AvatarExcludeIdMatcher
+
+
+# Decides whether `avatar_id` is present in `exclude_ids` for the purposes of
+# AvatarModifierArea visibility. Treats Ethereum addresses case-insensitively
+# and tolerates a missing `0x` prefix on either side: profiles delivered over
+# comms may carry the checksummed form, while scenes commonly push lowercase
+# addresses. See issue #2166.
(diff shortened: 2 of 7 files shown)


regenesis-claw (approved):
Clean, well-decomposed fix tackling all three root causes of the flaky excludeIds behaviour.

**What's right:**
- `AvatarExcludeIdMatcher` is a clean pure-static helper with robust normalization (case + `0x` prefix tolerance).
- The race-condition fix (calling `try_show()` after `avatar_id` changes in `async_update_avatar_from_profile`) is correctly guarded by `avatar_id_changed`.
- `refresh_overlapping_detectors` wisely avoids `Area3D.get_overlapping_areas()` (stale mid-CRDT-tick) and iterates `Global.avatars` instead — the comment explains the rationale clearly.
- No infinite recursion risk: `try_show()` → `check_areas()` → emits signal → `_on_set_avatar_modifier_area` which does NOT call back into `refresh_overlapping_detectors`.
- The regression test covers 7 cases including the edge case of empty `avatar_id` and mixed-entry lists. Good coverage.

**One observation (non-blocking):**
`refresh_overlapping_detectors` iterates *all* avatars, not just those overlapping this specific area. This is correct since `check_areas()` re-filters by scene_id and overlapping areas anyway, so it's effectively a no-op for avatars not in the area — but worth noting it's O(total_avatars) per excludeIds update. Totally fine for normal counts.

LGTM — approving. CI is still running (static checks + iOS build in progress, Android build was cancelled on the previous push and should re-run). No platform-specific code paths touched (all GDScript/Rust, no iOS/Android branches), so the iOS build is nice-to-have but not blocking.

kuruk-mm (approved):
looks great!

PR #2178, opened 2026-05-28
fix: TextShape font_auto_size to match Unity client

### LINKED ISSUE #1956: [Bug]: SDK Text entity renders much smaller on mobile than desktop
### Platform
Mobile (iOS and Android)

### App Version
Not specified

### Device Information
Not specified

### Issue Description
Text rendered via SDK Text entity (on a large coffin object) appears significantly smaller on mobile compared to desktop. The same text renders at normal, readable size on desktop but is tiny/unreadable on mobile.

**Scene:** -99,103 (https://mobile.dclexplorer.com/open?position=-99,103)
**Reported by:** JayKaying (scene owner)
**Object:** Large coffin text

This is a cross-platform rendering parity issue similar to other mobile SDK UI rendering differences (related to #1760, #1807).

### Expected Behavior
Text entities should render at the same size/scale on both mobile and desktop clients when using the same SDK component configuration.

### Steps to Reproduce
1. Visit scene -99,103 on desktop client and observe the text on the large coffin (normal size)
2. Visit the same scene on mobile client
3. Observe that the same text appears tiny/much smaller than on desktop

### Occurrence
100% - consistently reproducible for text on large coffin in this scene

### Related Issues
- #1760 (SDK UI rendering parity issues - normals/backfaces flipped, borders missing)
- #1807 (rendering differences - colors and transparency)

---

EibrielInv:
## Summary                                                                                                                                                                                                       
                                                                                                                                                                                                                   
  - Implement `PBTextShape.font_auto_size` (previously ignored) to match the Unity reference client, fixing the small-text rendering at scene `-99,103`.                                                         
  - Fix outline rendering so it overlays the glyph (`outline_render_priority = 1`) instead of expanding the silhouette and visually thickening the font.                                                           
  - Guard a latent integer-underflow panic in `GrowOnlySet::to_binary` surfaced while writing tooling against the CRDT state.                                                                                      
                                                           
  ### `font_auto_size`                                                                                                                                                                                             
                                                                                                                                                                                                                   
  `text_shape.rs` now replicates Unity TMP's behavior when `font_auto_size = true`:                                                                                                                                
  - `text_wrapping = true` → fit-to-rect inside `width × height`, no word wrap, clamped to `[fontSizeMin, fontSizeMax]`.                                                                                           
  - `text_wrapping = false` → Unity's `sizeDelta = Vector2.zero` quirk collapses the rect and TMP falls back to `fontSizeMin`. We replicate the fallback rather than implementing the proto's "fit in width/height"
   contract literally, since published scenes were authored against the Unity behavior. The proto divergence is documented inline (link to `text_shape.proto:22`).                                                 
                                                                                      
  A single `DCL_TMP_SIZE_FACTOR = 17/18` constant drives `TMP_TO_LABEL3D_FONT_SIZE`, `UNITY_TMP_FONT_SIZE_MIN`, and `UNITY_TMP_FONT_SIZE_MAX`, tuned empirically against the live Unity client. Retuning is a      
  one-line change.                                                                                                                                                                                                 
                                                                                                                                                                                                                   
  ### Outline                                                                                                                                                                                                      
                                                                                                                                                                                                                   
  - `outline_render_priority = 1` so the outline draws on top of the glyph fill, matching TMP's "ring around the silhouette" behavior instead of Godot's default "bolden the glyph by N pixels".                   
  - `TMP_TO_LABEL3D_OUTLINE_WIDTH = 0.8` scales the computed `outline_size` so the ring doesn't eat the glyph interior.
                                                                                                                                                                                                                   
  ### GOS panic guard                                                                 
                                         
  `GrowOnlySet::to_binary` did `entry.len() - reverse_element_index - 1` and panicked on out-of-range probes. Replaced with `checked_sub`, returning the same `Err("Index out of range")` callers already handle.  
  No behavior change for existing callers in `message.rs`.                                                                                                                                                         
                                                                                                                                                                                                                   
  Closes #1956.                                                                                                                                                                                                    

## Images

<img width="894" height="374" alt="Captura de pantalla 2026-05-28 a la(s) 15 50 47" src="https://github.com/user-attachments/assets/742682b2-f1fd-4d7b-8cfd-59c5eff68a07" />

<img width="1265" height="780" alt="Captura de pantalla 2026-05-28 a la(s) 12 34 35" src="https://github.com/user-attachments/assets/bda4b828-838e-4e88-b736-8f71c2b02f94" />



## Test plan                                                                                                                                                                                                     
- Go to scene -99,103 (https://mobile.dclexplorer.com/open?position=-99,103), "Tip" text should have the same size on Godot and Unity.

### DIFF
--- a/lib/src/scene_runner/components/text_shape.rs
+++ b/lib/src/scene_runner/components/text_shape.rs
@@ -13,10 +13,50 @@ use crate::{
     scene_runner::scene::Scene,
 };
 use godot::{
-    classes::{label_3d::AlphaCutMode, text_server::AutowrapMode, Label3D, Node},
+    classes::{
+        label_3d::AlphaCutMode,
+        text_server::{AutowrapMode, JustificationFlag, LineBreakFlag},
+        Label3D, Node,
+    },
     global::{HorizontalAlignment, VerticalAlignment},
     prelude::*,
 };
+
+/// Empirically-derived scale factor between Unity TextMeshPro's `fontSize`
+/// (world-space units per em) and Godot `Label3D.font_size` (glyph pixels):
+/// our baseline test confirmed that SDK `font_size = 1.4` renders identically
+/// in both clients when Label3D `font_size = 30` and `pixel_size = 0.005`,
+/// i.e. 30 / 1.4 ≈ 21.4. Kept at 22.0 to round-trip cleanly with `i32` sizes.
+const TMP_TO_LABEL3D_FONT_SIZE: f32 = 22.0;
+
+/// Unity TextMeshPro's autosize bounds. TMP documentation lists `fontSizeMin = 18`
+/// and `fontSizeMax = 72` as the component defaults, but empirical measurement
+/// against the live Unity client (autosize-fallback case: degenerate rect,
+/// `text_wrapping = false`) gave a Godot:Unity render ratio that implies the
+/// DCL TMP prefab scales those bounds down to about 16/64. Express that as a
+/// single scale factor applied to both bounds so future tuning is one number.
+const DCL_TMP_AUTOSIZE_SCALE: f32 = 17.0 / 18.0;
+const UNITY_TMP_FONT_SIZE_MIN: f32 = 18.0 * DCL_TMP_AUTOSIZE_SCALE;
+const UNITY_TMP_FONT_SIZE_MAX: f32 = 72.0 * DCL_TMP_AUTOSIZE_SCALE;
+
+/// Unity TMP's `_OutlineWidth` is a 0..1 shader-domain SDF distance — the
+/// outline appears as a thin ring drawn outside the glyph silhouette without
+/// expanding the glyph itself.
+///
+/// Godot's `Label3D.outline_size` is raw pixels and, by default, the outline
+/// is rasterized underneath the glyph so the silhouette grows by
+/// `outline_size` on each side — visually equivalent to making the font bolder.
+///
+/// To match Unity's "ring around the existing silhouette" look we:
+///   1. Set `outline_render_priority = 1` so the outline is drawn on top of
+///      the glyph fill — the outline color now overlays the glyph edge instead
+///      of stretching the silhouette outward.
+///   2. Scale the computed `outline_size` way down so the outline covers only
+///      a thin ring at the glyph edge and doesn't eat the glyph interior.
+///
+/// (1) + (2) together approximate Unity's outline visual without needing a
+/// custom shader. Tune (2) empirically per the test scene.
+const TMP_TO_LABEL3D_OUTLINE_WIDTH: f32 = 0.8;
 
 pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
     let godot_dcl_scene = &mut scene.godot_dcl_scene;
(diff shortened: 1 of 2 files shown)


kuruk-mm (commented):
_(review withdrawn)_

lib/src/scene_runner/components/text_shape.rs
```diff
@@ -86,15 +132,74 @@ pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
+                        UNITY_TMP_FONT_SIZE_MIN * TMP_TO_LABEL3D_FONT_SIZE
+                    } else {
+                        compute_auto_font_size(
+                            &font_resource,
+                            &display_text,
+                            fit_w,
+                            fit_h,
+                            label_3d.get_pixel_size(),
+                            (UNITY_TMP_FONT_SIZE_MIN * TMP_TO_LABEL3D_FONT_SIZE) as i32,
+                            (UNITY_TMP_FONT_SIZE_MAX * TMP_TO_LABEL3D_FONT_SIZE) as i32,
```
kuruk-mm: The `17/18` factor is applied twice here: it's already inside both `UNITY_TMP_FONT_SIZE_MIN` and `TMP_TO_LABEL3D_FONT_SIZE`, so this product squares it (`18·22·(17/18)²`). The non-autosize path applies it once, so autosize text ends up ~5.5% smaller for the same nominal size. Was the autosize size tuned visually (then just fix the comments) or assumed to track the non-autosize calibration (then make `MIN`/`MAX` raw `18.0`/`72.0`)?

lib/src/scene_runner/components/text_shape.rs
```diff
@@ -86,15 +132,74 @@ pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
+                        ) as f32
+                    }
+                } else {
+                    (TMP_TO_LABEL3D_FONT_SIZE * new_value.font_size.unwrap_or(3.0)).max(1.0)
+                };
+                let outline_size = font_size
+                    * new_value.outline_width.unwrap_or(0.0)
+                    * TMP_TO_LABEL3D_OUTLINE_WIDTH;
```
kuruk-mm: `font_size * outline_width * 0.8` — with autosize `font_size` reaching ~1400px, a small `outline_width` still yields a large outline, not the "thin ring" the comment describes. Worth confirming against the reference scene.

lib/src/scene_runner/components/text_shape.rs
```diff
@@ -163,6 +260,54 @@ pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
+    let fits = |fs: i32| -> bool {
+        let measured = font
+            .get_multiline_string_size_ex(text)
+            .max_lines(-1)
+            .width(-1.0)
+            .font_size(fs)
+            .justification_flags(JustificationFlag::NONE)
+            .brk_flags(LineBreakFlag::MANDATORY)
+            .done();
+        measured.x <= rect_w_px && measured.y <= rect_h_px
```
kuruk-mm, line 306: `pixel_size` is never set so it defaults to `0.01` (comment assumes `0.005`); with the default fit width the rect is ~100px and nothing fits, so the search always returns the min. Confirm this matches the reference.

kuruk-mm (commented):
Review notes inline. `grow_only_set.rs` looks good. A few things to confirm on the TextShape font sizing.

kuruk-mm (changes requested):

lib/src/scene_runner/components/text_shape.rs
```diff
@@ -86,15 +132,74 @@ pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
+                };
+                let outline_size = font_size
+                    * new_value.outline_width.unwrap_or(0.0)
+                    * TMP_TO_LABEL3D_OUTLINE_WIDTH;
                 label_3d.set_font_size(font_size as i32);
                 label_3d.set_outline_size(outline_size as i32);
                 label_3d.set_outline_modulate(outline_color);
+                // Draw the outline on top of the glyph fill so it overlays the
+                // glyph edge (matching Unity TMP) instead of expanding the
+                // silhouette outward. See `TMP_TO_LABEL3D_OUTLINE_WIDTH` notes.
```
regenesis-claw, line 214: The old code used `22.0` as the conversion factor:
```rust
let font_size = (22.0 * new_value.font_size.unwrap_or(3.0)).max(1.0);
```
The new code uses `TMP_TO_LABEL3D_FONT_SIZE = 22.0 * 17/18 ≈ 20.78`. This means **all** non-autosize TextShapes in production got ~5.6% smaller — not just the autosize ones this PR targets.

Was this intentional (i.e. the old 22.0 was slightly off and you re-calibrated everything together)? If so, the PR summary should mention it as a separate behavioral change. If not, the non-autosize path should keep using `22.0` directly (or `DCL_TMP_SIZE_FACTOR` should only apply to the autosize constants, not `TMP_TO_LABEL3D_FONT_SIZE`).

lib/src/scene_runner/components/text_shape.rs
```diff
@@ -149,7 +246,7 @@ pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
 
                 if add_to_base {
                     label_3d.set_name("TextShape");
-                    label_3d.set_font(&new_font.get_font_resource());
+                    label_3d.set_font(&font_resource);
                     node_3d.add_child(&label_3d.upcast::<Node>());
                 }
```
regenesis-claw, line 267: `set_font` is only called in the `add_to_base` branch. If a scene updates the `font` field on an existing TextShape entity (e.g. switching from SansSerif to Monospace at runtime), the `font_resource` computed above is used for autosize measurement but never applied to the label. Pre-existing issue, but since you refactored the font loading path, might be worth fixing in this PR:
```rust
label_3d.set_font(&font_resource);
```
moved outside the `if add_to_base` block (just the `set_font` call — `set_name` and `add_child` stay gated).

lib/src/scene_runner/components/text_shape.rs
```diff
@@ -86,15 +132,74 @@ pub fn update_text_shape(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
                 label_3d.set_outline_modulate(outline_color);
+                // Draw the outline on top of the glyph fill so it overlays the
+                // glyph edge (matching Unity TMP) instead of expanding the
+                // silhouette outward. See `TMP_TO_LABEL3D_OUTLINE_WIDTH` notes.
+                label_3d.set_outline_render_priority(1);
                 label_3d.set_alpha_cut_mode(AlphaCutMode::DISCARD);
 
-                let text_wrapping = new_value.text_wrapping.unwrap_or_default();
-
                 let (width_meter, height_meter) = if text_wrapping {
```
regenesis-claw, line 218: Nit: `set_outline_render_priority(1)` is called on every dirty update even when `outline_width` is 0 (i.e. no outline). Harmless, but could be gated on `outline_size > 0` to avoid unnecessary GPU state changes on text-heavy scenes.

regenesis-claw (commented):
Solid work implementing `font_auto_size` — the Unity quirk documentation is excellent and the binary search approach is clean. `grow_only_set.rs` fix is correct and safe.

Agreed with @kuruk-mm's three inline findings (double `DCL_TMP_SIZE_FACTOR`, outline scaling at large autosize pixels, `pixel_size` default mismatch). Adding a few more observations below.

Also +1 to Mateo's Slack suggestion: extracting the conversion math into named helper functions (e.g. `unity_tmp_to_label3d_font_size(tmp_size: f32) -> i32`) would make this much easier to reason about and audit — the current inline arithmetic requires holding several constants in your head simultaneously.

Requested by Mateo via Slack

kuruk-mm (approved):
looking great!

Issue #1862, opened 2026-04-17
Genesis Plaza Profiling


rodrigo685:

kuruk-mm:
fixed by https://github.com/decentraland/godot-explorer/pull/1992

PR #2191, opened 2026-06-01
perf(rendering): mobile GPU optimization stack — perf passes, ETC2 pipeline

manuelmaceira:
Full mobile GPU rendering-optimization stack for Genesis Plaza, rebased onto `main` as a **linear, one-commit-per-change** history (no merge commits). This **supersedes and absorbs #2197** (`feat/mobile-perf-stack`), which is now closed — everything from that stack plus the octahedral-impostor work lives here in 13 reviewable commits.

## Commits (each a self-contained change)

**Runtime perf passes — ship on-device, default ON**
1. `feat(content)` optimized-content-base-url override (CLI + deeplink)
2. `perf(gltf)` coalesce GltfContainer load-timeouts into one Node (replaces ~1419 per-container Timers in GP)
3. `perf(dcl)` split CRDT recv into wait+drain to reuse the JS recv buffer
4. `perf(tween)` emit TweenState only on state transition (~51% fewer dirty LWW entries/frame)
5. `perf(avatar)` encode + write impostor PNG off the render thread
6. `perf(graphics)` retune default profile + Mali shadow atlas

**Asset-server bake — device-only, via pre-baked optimized `.scn`**
7. `perf(content)` bake LOD chain + dedup shared-mesh VBOs at import
8. `perf(asset-server)` auto-bake BoxOccluder3D for large opaque meshes

**Bench harness — dev only**
9. `feat(bench)` Genesis Plaza profiling harness + diagnostic knobs (#1862)

**Texture pipeline + octahedral impostors**
10. `feat(content)` ETC2 texture pipeline (ImageTexture) end-to-end + main-thread resource-pack load
11. `feat(impostors)` octahedral impostors + packed normal atlas + cap High shadow_quality (Mali)

**Tooling / hygiene**
12. `chore(bench/asset-server)` bake hygiene + GP bench memory/settling instrumentation
13. `chore(repo)` raise gdlint max-public-methods 40→45 + bump iOS build timeout

> **Texture note:** textures decode as **ETC2 via `ImageTexture`**, not `PortableCompressedTexture2D`. On the Vulkan path (`textureCompressionETC2=false`) PCT2 forced a GPU readback that decompressed ETC2→RGBA8 mid-load and shipped magenta avatars/scenes on device.

### Bake-time memory recovery — Samsung A54 / Mali-G68, HIGH, Genesis Plaza
VBO/material dedup at bake recovers shared-mesh sharing, so low-end devices no longer OOM-crash:

| metric | before dedup | after dedup | Δ |
|---|---:|---:|---:|
| GL mtrack | 1,882 MB | **715 MB** | **−62%** |
| TOTAL Pss | 3,556 MB | **1,328 MB** | **−63%** |
| `.scn` on disk | 447 MB | **160 MB** | **−64%** |

### Shadow tuning — A54 / Mali-G68, release, fixed pose, fixed skybox
| Profile | fps mean | GPU ms | shadow_prim | total prim | draws |
|---|---:|---:|---:|---:|---:|
| High shadows on (`gfx-shadow=2`) | 21.6 | 44.1 | 1,018 K | 1.61 M | 1,643 |
| **High shadows Low (this stack)** | **26.3** | **33.4** | **792 K** | **1.39 M** | **1,379** |
| Shadows off (ceiling) | 27.0 | 34.8 | 0 | 0.57 M | – |

Capping High's shadow cascade at Low keeps near-field shadows visually present while recovering ~5 fps (+22%) — the shadow pass is 62% of the scene's geometry.

### End-to-end perf-pass progression — A54, HIGH, GP `ea9cc738`
| state | fps | render_gpu_ms | primitives |
|---|---:|---:|---:|
| A — pre-perf baseline | 13.88 | 67.27 | 2,967,296 |
| **D — + this stack** | **18.51** | **51.66** | **1,920,326** |

**A→D total: +33% fps, −23% GPU frame time.**

## Verification
- Linear history, **no merge commits**; the 9 perf-stack commits retain #2197's per-commit independent-compile property.
- Final tree verified **byte-identical** to a 3-way merge of the original work into latest `main` (v71 incorporated); `cargo metadata --locked` consistent; `gdlint` clean.

Supersedes #2197.


godot/assets/shaders/dcl_octahedral_impostor.gdshader
```diff
@@ -0,0 +1,161 @@
...
+uniform sampler2D normal_atlas : hint_default_black, filter_linear, repeat_disable;
+uniform int n_cells = 8;
+uniform bool is_hemi = true;
+
+// Sun direction + colors for the manual Lambert lighting. Driven by
+// the same global shader params as the sky / main directional light
+// (sky_base.gd pushes them every frame from the day/night cycle).
+// Staying in `unshaded` keeps Godot's PBR pipeline out of the picture
+// (which empirically choked Mali shader compilation when 288 unique
+// ShaderMaterials hit the load queue at once).
```
regenesis-claw: `sun_direction` is read here as a `global uniform` but it's never declared in `project.godot`'s `[shader_globals]` section, and `sky_base.gd` never calls `RenderingServer.global_shader_parameter_set("sun_direction", ...)`. The half-Lambert term will get a zero vector → `normalize(vec3(0))` → NaN → impostors render flat/ambient-only.

Add `sun_direction` to `[shader_globals]` with a sensible default, and push it from `sky_base.gd._process()` alongside `sun_color`/`ambient_color` (e.g. from `main_light.rotation` or `main_light.global_transform.basis * Vector3.FORWARD`).

godot/project.godot
```diff
@@ -70,7 +71,7 @@ window/handheld/orientation=6
 
 [editor]
 
-run/main_run_args="--emulate-ios"
+run/main_run_args="--preview --realm http://localhost:8000 --optimized-content-base-url http://localhost:9092 --position 0,0 --skip-lobby"
```
regenesis-claw: Per REVIEW.md §3 Tier 1 #5 — personal local dev settings (`--preview --realm http://localhost:8000 ...`) leaked into project.godot. Revert to the previous value or a neutral default.

lib/src/content/scene_saver.rs
```diff
@@ -31,6 +31,15 @@ pub fn save_node_as_scene(node: Gd<Node3D>, file_path: &str) -> Result<(), Strin
         return Err(format!("Failed to pack scene: {:?}", err));
     }
 
+    // DIAGNOSTIC: also save a .tscn (text) sibling so we can grep
+    // for VisibilityRange / dcl_impostor / ShaderMaterial without
+    // dealing with the RSCC binary format.
+    let tscn_path = format!("{}.tscn", file_path);
+    let _ = ResourceSaver::singleton()
+        .save_ex(&packed.clone().upcast::<Resource>())
+        .path(&tscn_path)
```
regenesis-claw: This `.tscn` diagnostic write runs unconditionally on every asset-server bake. It doubles disk I/O and the text sibling will end up in the production ZIP (packer walks `output/*`). Consider gating behind an env var or CLI flag, or at least ensuring the packer excludes `.tscn` files.

godot/assets/shaders/dcl_octahedral_impostor.gdshader
```diff
@@ -0,0 +1,161 @@
+shader_type spatial;
+render_mode unshaded, cull_disabled, depth_prepass_alpha;
+
+uniform sampler2D atlas : source_color, filter_linear, repeat_disable;
+uniform sampler2D normal_atlas : hint_default_black, filter_linear, repeat_disable;
```
regenesis-claw: `normal_atlas` uniform is declared but never set from Rust — the packed atlas approach puts albedo+normal into the single `atlas` texture. This is dead code; remove to avoid confusion.

lib/src/godot_classes/dcl_cli.rs
```diff
@@ -720,6 +731,7 @@ impl INode for DclCli {
                 &optimized_content_base_url.to_string(),
             );
         }
+        let cheap_pbr_enabled = !args_map.contains_key("--no-cheap-pbr");
```
regenesis-claw: This defaults `cheap_pbr_enabled` to `true` (only `--no-cheap-pbr` disables it), but the `ArgDefinition` above declares `--cheap-pbr` with description "Default OFF". The arg name, the documented default, and the parsing logic all disagree. Since the field is never consumed in this PR, consider removing the flag entirely and adding it when the consumer lands.

godot/src/decentraland_components/gltf_container.gd
```diff
@@ -137,6 +137,27 @@ func _async_load_optimized_asset(gltf_hash: String):
...
+	if Global.deep_link_obj != null:
+		var v: String = Global.deep_link_obj.params.get("pvs-runtime", "")
+		pvs_enabled = v.to_lower() in ["true", "1", "yes"]
+	if pvs_enabled:
+		var pvs_path := "res://glbs/" + gltf_hash + ".pvs.bin"
+		var f := FileAccess.open(pvs_path, FileAccess.READ)
+		if f != null:
+			var buf := f.get_buffer(f.get_length())
+			f.close()
+			print("[pvs-hook] hash=", gltf_hash, " loaded buf=", buf.size(), " bytes")
```
regenesis-claw: Debug `print()` leftover. Remove or gate behind `OS.is_debug_build()`.

regenesis-claw (changes requested):
Excellent impostor pipeline — the architecture (worker→main-thread SubViewport bake→PCT2 stitch→.scn embed) is clean, the v2 mitigations against the 17-bake crash are well-reasoned, and the +22% fps headline on Mali-G68 speaks for itself. The packed 256×512 atlas solving the two-texture .scn serialization bug is a nice win. A few things to address before merge:

Followed `REVIEW.md` conventions for this review.

---

### Findings

| # | Sev | File | Finding |
|---|-----|------|---------|
| 1 | 🔴 | `dcl_octahedral_impostor.gdshader` / `project.godot` / `sky_base.gd` | **`sun_direction` global uniform is never declared or written.** The shader reads `global uniform vec3 sun_direction` but it's missing from `[shader_globals]` in `project.godot` (only `sun_color` and `ambient_color` are declared), and `sky_base.gd` never calls `RenderingServer.global_shader_parameter_set("sun_direction", ...)`. Result: the half-Lambert term `dot(nrm, normalize(vec3(0,0,0)))` is NaN/zero — impostors will render flat with ambient-only lighting, losing the directional response the normal bake was built for. Add the `sun_direction` vec3 to shader_globals and push it from `sky_base.gd._process()` alongside `sun_color`. |
| 2 | 🔴 | `project.godot` | **Editor run args are local dev settings.** `run/main_run_args` changed from `--emulate-ios` to `--preview --realm http://localhost:8000 --optimized-content-base-url http://localhost:9092 --position 0,0 --skip-lobby`. Per REVIEW.md §3 Tier 1 #5, these are personal local paths that should not be committed. |
| 3 | 🟡 | `scene_saver.rs` | **`.tscn` diagnostic sibling saved unconditionally.** `save_node_as_scene` now writes a `.tscn` text copy alongside every `.scn` on the asset server. This doubles disk I/O for every bake and ships debug artifacts in production zips. Gate behind a feature flag, env var, or CLI flag (e.g. `--save-tscn-debug`). |
| 4 | 🟡 | `dcl_octahedral_impostor.gdshader` | **Orphan `normal_atlas` uniform.** `uniform sampler2D normal_atlas` is declared but the Rust code never sets it — the packed atlas (albedo+normal in one texture) is set as `atlas`. Harmless (`hint_default_black`) but dead code. Remove it. |
| 5 | 🟡 | `dcl_cli.rs` | **`cheap_pbr_enabled` CLI mismatch.** The `ArgDefinition` declares `--cheap-pbr` with description "Default OFF", but parsing uses `!args_map.contains_key("--no-cheap-pbr")` which defaults it ON. The `--no-cheap-pbr` flag has no `ArgDefinition` either (won't appear in `--help`). Also, the field is never consumed in this PR (`apply_pre_generate_material_overrides` is `#[allow(dead_code)]`). If this is intentionally preparatory, align the arg parsing with the definition or remove the flag entirely. |
| 6 | 🟡 | `gltf_container.gd` | **`print()` in PVS hook.** Line `print("[pvs-hook] hash=", gltf_hash, ...)` — debug leftover per REVIEW.md Tier 1 #3. |
| 7 | nit | `common.rs` | **~440 lines of new `#[allow(dead_code)]` functions.** `apply_pre_generate_material_overrides`, `apply_pre_generate_mesh_simplification`, `apply_post_generate_lod_chain`, `build_lods_for_surface` are all dead code. If these are intentionally parked for follow-up, a brief doc-comment noting that (or a tracking issue) would help future reviewers. |
| 8 | nit | `metrics.rs` | Uses `std::sync::Mutex` — rest of the codebase uses both `std::sync` and `parking_lot` depending on context. Fine here since contention is negligible, just noting. |

---

### CI
All checks green (static checks, clippy, Linux, Android, Docker). No iOS build triggered — PR touches rendering/shaders but the impostor pipeline is bake-time only (asset server) and the runtime shader is read-only. No iOS-specific code paths changed, so an iOS build is not blocking.

### Positives
- Packed 256×512 atlas (albedo top + normal/ORM bottom) solving the .scn sub-resource drop is a much better approach than two separate textures.
- Rollback logic on bake failure (detach impostor MI, clear visibility_range_end) is defensive and correct.
- The coverage check (`image_alpha_coverage`) filtering out sparse meshes prevents wasted billboard draws.
- Unit tests for the octahedral mapping roundtrip are welcome.
- The benchmark runner memory snapshots and dedup walker are useful diagnostic additions.

godot/assets/shaders/dcl_octahedral_impostor.gdshader
manuelmaceira: False alarm — `sun_direction` is pre-existing infrastructure for the sky shader:
- Declared in `project.godot` (`[shader_globals]`) with default `Vector3(0, 0.7, 0.7)` (non-zero → no `normalize(vec3(0))` NaN).
- Pushed every frame from `sky_base.gd:97` (`RenderingServer.global_shader_parameter_set("sun_direction", sun_dir)`).

This PR only adds `sun_color` and `ambient_color` globals; the impostor shader just reuses the existing `sun_direction`, transformed to view space via `VIEW_MATRIX`.

godot/project.godot
manuelmaceira: Good catch. Reverted to `--emulate-ios` (matches `feat/mobile-perf-stack`).

lib/src/content/scene_saver.rs
manuelmaceira: Removed. The packer doesn't walk the dir (it references `{hash}.scn` by name, so the `.tscn` siblings never made it into the zip), but the diagnostic served its purpose during impostor bring-up — no reason to keep the doubled I/O.

godot/project.godot
manuelmaceira: Correction: cleared to empty string (`run/main_run_args=""`) instead — neither `--emulate-ios` nor any other personal default is appropriate as a checked-in editor run-arg.

godot/assets/shaders/dcl_octahedral_impostor.gdshader
manuelmaceira: Correct — dead since the move to the packed 256×512 atlas (albedo in top half, normal+ORM in bottom half, both sampled from `atlas` via UV offset). Removed the uniform declaration.

lib/src/godot_classes/dcl_cli.rs
manuelmaceira: All three points correct (name/default/parsing disagreed, and nothing reads the field). Removed entirely: `cheap_pbr_enabled` field, `--cheap-pbr` `ArgDefinition`, the `--no-cheap-pbr` parsing, the struct init, the bench-runner deeplink override, and a stale doc reference in `scene.rs`. Will re-introduce when the consumer lands.

godot/src/decentraland_components/gltf_container.gd
manuelmaceira: Removed. The block is opt-in (`pvs-runtime` deeplink param), but the print fires per GLB — noisy even when intentional, and the PVS hook is self-documenting.

regenesis-claw (changes requested):
## Re-review (delta from `134b37c` → `de549ea`)

5 of 6 actionable findings resolved — nice cleanup commit. One blocker remains.

| # | Prior Finding | Status |
|---|---|---|
| 1 | 🔴 `sun_direction` global uniform never declared/written | ⚠️ **Still open** — shader reads `global uniform vec3 sun_direction` but it's not in `project.godot` `[shader_globals]` and `sky_base.gd` doesn't push it. Impostors will render flat (ambient-only). |
| 2 | 🔴 Editor run args leak | ✅ Fixed — set to `""` |
| 3 | 🟡 `.tscn` diagnostic writes | ✅ Fixed — removed |
| 4 | 🟡 Orphan `normal_atlas` uniform | ✅ Fixed — removed |
| 5 | 🟡 `cheap_pbr_enabled` CLI mismatch | ✅ Fixed — entire feature removed |
| 6 | 🟡 `print()` in PVS hook | ✅ Fixed — removed |

CI all green on new head ✅

Once `sun_direction` is declared + written (add to shader_globals + push from `sky_base.gd._process()` alongside `sun_color`/`ambient_color`), this is good to go.

regenesis-claw (approved):
My bad on the `sun_direction` flag — confirmed it's pre-existing in the base branch (`project.godot` shader_globals + `sky_base.gd:97` pushing it every frame). Only looked at the PR diff, not the base.

All prior findings resolved. CI green. Approving.

godot/src/decentraland_components/gltf_container.gd
```diff
@@ -139,6 +137,26 @@ func _async_load_optimized_asset(gltf_hash: String):
...
+	if Global.deep_link_obj != null:
+		var v: String = Global.deep_link_obj.params.get("pvs-runtime", "")
+		pvs_enabled = v.to_lower() in ["true", "1", "yes"]
+	if pvs_enabled:
+		var pvs_path := "res://glbs/" + gltf_hash + ".pvs.bin"
+		var f := FileAccess.open(pvs_path, FileAccess.READ)
+		if f != null:
+			var buf := f.get_buffer(f.get_length())
+			f.close()
+			var rt = load("res://src/tools/dcl_pvs_runtime.gd").new()
```
kuruk-mm: 🔴 **Critical — guaranteed crash.** `res://src/tools/dcl_pvs_runtime.gd` does not exist anywhere in the repo (nor does the `init_with_buffer` method called below). `load()` returns `null` for a missing resource, so `.new()` raises *"Cannot call method 'new()' on a null value"* and aborts the load.

Gated behind the off-by-default `pvs-runtime` deeplink + a `.pvs.bin` sidecar — but the comment above notes sidecars still ship in zips, so the flag is the only guard. This is dead-on-arrival: enabling the feature crashes instantly. Ship the runtime script or remove this path before merge.

lib/src/content/content_provider.rs
```diff
@@ -81,6 +81,29 @@ struct ImageSize {
...
+/// enqueue (zip_path, reply) here and await the reply; `process()` drains
+/// it on the main thread where the call belongs.
+#[allow(clippy::type_complexity)]
+static MAIN_THREAD_PACK_QUEUE: once_cell::sync::Lazy<
+    std::sync::Mutex<Vec<(String, oneshot::Sender<bool>)>>,
+> = once_cell::sync::Lazy::new(|| std::sync::Mutex::new(Vec::new()));
+
+/// Enqueue a resource-pack mount for the main thread and await its result.
+/// Safe to call from any tokio worker.
+async fn load_resource_pack_on_main(zip_path: String) -> bool {
```
kuruk-mm, line 98: 🟠 **Threading race reintroduced.** Previously `load_resource_pack` ran on a worker *while holding* `godot_single_thread` (the semaphore that also serializes worker GLTF generation / Godot API access — its comment cites "COW string reference counting corruption"). Routing it through `MAIN_THREAD_PACK_QUEUE` → `process()` runs it on the main thread **without** that semaphore, so it can now execute concurrently with worker GLTF gen + `ResourceSaver.save`, touching Godot's ResourceLoader/filesystem internals from two threads.

Consider acquiring `godot_single_thread` in the main-thread drain too. (The deadlock risk I checked is *not* real — promises resolve via `call_deferred`, so the frame loop keeps ticking and the oneshot reply always fires.)

src/run.rs
```diff
@@ -806,16 +804,38 @@ fn deploy_and_run_android(_release: bool, _extras: Vec<String>) -> anyhow::Resul
...
+        // GodotAppLauncher is the exported launcher activity; GodotApp itself
+        // is not exported, so `am start -n .../GodotApp` is denied from shell.
+        "org.decentraland.godotexplorer/com.godot.game.GodotAppLauncher".to_string(),
+    ];
+
+    // Forward CLI flags as Android intent extras. Godot's Android launcher maps
+    // an extra `foo=true` to the `--foo` command-line arg picked up by DclCli.
+    // For example `--skip-lobby` / `--clear-cache-startup` / `--debug-ws`.
+    for extra in &extras {
+        let arg = extra.trim_start_matches('-');
```
kuruk-mm: 🟠 **Android arg forwarding is non-functional.** Godot's `GodotActivity` only reads the `command_line_params` string-array extra — it never maps `-e <key> <value>` intent extras into the command line, and `DclCli` reads only `Os::get_cmdline_args()`. So **every** forwarded flag is silently dropped (not just value-bearing ones like `--rust-log=debug`). Also, `--rust-log=debug` becomes key `rust-log=debug` = `"true"`, mangling the value.

This breaks the workflow documented in CLAUDE.md: `cargo run -- run --target android --only-lib -- --skip-lobby`. Forward via `--es command_line_params "<space-joined args>"` (or `--esa`) instead. Same pattern at the `start_args` sites (~1176).

lib/src/content/texture.rs
```diff
@@ -420,6 +421,23 @@ fn pad_image_to_multiple_of_4(image: &mut Gd<Image>) -> bool {
 
 /// Creates a texture from a compressed image, resizing if needed.
 /// Uses ETC2 compression for better memory usage on mobile platforms.
+///
+/// The ETC2 image is wrapped in a `PortableCompressedTexture2D` with
```
kuruk-mm, line 426: 🟠 **PCT2 path silently coupled to a fork build.** This `PortableCompressedTexture2D` bake is correct only on DCL Godot fork `gh.9ee6af7ab` (2026-06-03) or newer — older binaries ship PCT2 with no compressed buffer and render **magenta on device**. The only protection is this comment; if the asset-server/CI image rolls back, every mesh/material texture bakes magenta with nothing to catch it. Worth a runtime version assertion.

(Two things I verified as *not* problems: this doesn't contradict the `.res` `Image`/`ImageTexture` path — different serialization mechanisms, PackedScene-embedded vs standalone Resource; and there's no double-ETC2-encode — PCT2's `create_from_image` detects the already-compressed image and skips its own encode.)

lib/src/scene_runner/components/tween.rs
```diff
@@ -276,14 +284,22 @@ pub fn update_tween(scene: &mut Scene, crdt_state: &mut SceneCrdtState) {
...
-                state: current_tween_state as i32,
-            }),
-        );
+        // Update TweenState only on transitions (TsActive → TsPaused / TsCompleted,
+        // or first frame after creation). SDK7 userland reads `state`, not
+        // `current_time` mid-flight — emitting the same TsActive every frame
+        // pumps ~135 dirty entries/frame into the V8 recv path for nothing.
+        // See docs/bench/profile-deep-2026-05.md for measurement.
+        let new_state = current_tween_state as i32;
+        if tween.last_emitted_state != Some(new_state) {
```
kuruk-mm, line 303: 🟡 **`TweenState.currentTime` no longer streams mid-flight.** Old code `put()` every frame with live `current_time`; now it's gated on a state transition, so during `TsActive` only the first frame's ~0 progress is written until completion sets `1.0`. The proto field `PBTweenState.current_time` is userland-visible. No in-repo SDK consumer reads it (the standard helper keys off `state`), so likely fine — but it's a CRDT-contract change worth a deliberate confirmation. Also note `last_emitted_state` is reset only on tween creation (line 229), never on the existing-tween reset path.

godot/src/decentraland_components/gltf_container.gd
```diff
@@ -223,35 +241,14 @@ func _async_load_and_instantiate(scene_path: String) -> Node3D:
...
-	# Get the loaded resource
-	var resource := ResourceLoader.load_threaded_get(scene_path)
+	# Synchronous load on the MAIN thread. The optimized .scn embeds
+	# ETC2 ImageTexture atlases (impostors) and mesh textures that upload
+	# to the GPU during load; doing that on Godot's WorkerThreadPool
+	# (load_threaded_request) raced with the render thread (VkThread) over
+	# the RenderingServer command lock and intermittently DEADLOCKED the
+	# load on Mali (both threads parked in futex_wait). A main-thread load
+	# serializes the GPU upload with the frame loop.
+	var resource := ResourceLoader.load(scene_path)
```
kuruk-mm, line 231: 🟡 **Synchronous load on the main thread.** This replaces the old threaded `load_threaded_request` + `await process_frame` poll loop. `ResourceLoader.load()` has no suspension point, so the enclosing `await` doesn't yield — the main thread stalls for the full `.scn` load (the heavy ETC2/GPU-upload work) once per container, and the granular `THREAD_LOAD_FAILED`/`THREAD_LOAD_INVALID_RESOURCE` branches + mid-load `is_inside_tree()` abort are gone. Understood this is the deliberate Mali-deadlock fix; flagging the frame-hitch / lost-diagnostics tradeoff.

lib/src/scene_runner/components/asset_preprocessor/vertex_strip.rs
```diff
@@ -0,0 +1,116 @@
...
+//! size — UV2 is 8 bytes/vert, tangents 16 bytes/vert, colors 16 bytes/vert.
+//! On Mali-G68 vertex bandwidth is on-die but the payload from main RAM
+//! still costs DDR bandwidth.
+
+use godot::classes::base_material_3d::TextureParam;
+use godot::classes::mesh::{ArrayType, PrimitiveType};
+use godot::classes::{ArrayMesh, BaseMaterial3D};
+use godot::prelude::*;
+
+pub fn strip_unused(source: &Gd<ArrayMesh>) -> Option<(Gd<ArrayMesh>, u64)> {
```
kuruk-mm: 🧹 **Dead module.** `strip_unused` (and `colors_are_all_white`, plus `metrics::record_stripped` and the `meshes_stripped`/`stripped_bytes_total` fields it feeds) have zero callers anywhere in `lib/src` — the whole 116-line module compiles and ships doing nothing. Either wire it into `scene.rs::walk_and_preprocess` or delete it + the `stripped_*` metrics.

lib/src/scene_runner/components/asset_preprocessor/mesh_occluder.rs
```diff
@@ -0,0 +1,121 @@
...
+) -> bool {
+    // Auto-occluder spawn disabled. The per-mesh BoxOccluder3D produced
+    // by the size/opacity filter below over-culls in real DCL scenes
+    // (signs, decals, foliage near walls disappear at runtime — Godot
+    // sees them as fully occluded by the wall's BoxOccluder even with
+    // the per-axis inset tuning). Remove this short-circuit once we
+    // either tighten the filter further or switch to ArrayOccluder3D
+    // built from the actual mesh hull.
+    let _ = (mi, mesh, scene_root);
+    return false;
```
kuruk-mm: 🧹 **No-op with a live caller.** `try_spawn_for` short-circuits to `return false` (everything below is `#[allow(unreachable_code)]`), but `scene.rs::apply_asset_server_optimizations` → `walk_and_preprocess` still recurses the *entire* scene tree calling this on every `MeshInstance3D` — pure wasted traversal in the bake pass. Consistent with the recent "disable auto BoxOccluder3D spawn" commits, but the walk should short-circuit too (or gate the whole pass behind a flag) until the occluder path is re-enabled.

kuruk-mm (commented):
## Review — mobile GPU optimization stack

Reviewed `main...HEAD` (~3,150 insertions / 70 files) with multi-angle correctness + cleanup passes and per-finding verification. Solid perf work overall; the boundary renames (CRDT recv protocol, `session_hide_ui_options_sync`, the `GltfLoadTimeoutCoalescer` replacing the old `Timer`) are all correctly paired. Findings below, inline.

### 🔴 Must-fix before merge
- **`gltf_container.gd:155` — guaranteed crash.** `load("res://src/tools/dcl_pvs_runtime.gd").new()` references a script (and an `init_with_buffer` method) that don't exist anywhere in the repo → null-call crash whenever the `pvs-runtime` deeplink is enabled and a `.pvs.bin` sidecar is present.

### 🟠 Worth fixing
- **`content_provider.rs` — Godot threading race.** `load_resource_pack` was moved off the `godot_single_thread` semaphore onto the main-thread `process()` drain, so it can now run concurrently with worker GLTF generation / `ResourceSaver` — the exact COW/refcount race that semaphore documents. (The deadlock concern I checked is **not** real — promises resolve via `call_deferred`, frame loop keeps ticking.)
- **`run.rs:823` — Android arg forwarding is non-functional.** Godot's Android launcher only reads the `command_line_params` string-array extra; `-e <key> <value>` extras are never mapped to the command line, so *all* forwarded flags are silently dropped — this breaks the documented `cargo run -- run --target android --only-lib -- --skip-lobby` workflow. Use `--es command_line_params "<space-joined args>"`.
- **`texture.rs` — PCT2 path silently coupled to a fork build.** Correct only on fork `gh.9ee6af7ab` (2026-06-03)+; older binaries render magenta on device, with no runtime guard. (NB: the `.res` path uses `Image`/`ImageTexture` via a different serialization mechanism, so the two are *not* contradictory; and there is **no** double-ETC2-encode — PCT2 detects the already-compressed image and skips re-encoding.)

### 🟡 Intentional behavior changes — confirm desired
- **`tween.rs:293` — `TweenState.currentTime` no longer streams mid-flight** (emit-on-transition only). No in-repo SDK consumer reads it, but it's a CRDT-contract change.
- **`gltf_container.gd:251` — optimized `.scn` load is now synchronous on the main thread** (deliberate Mali deadlock fix, but stalls the frame loop per container and drops the granular error branches + `is_inside_tree()` abort).
- **`config_data.gd` — default `graphic_profile` 0 (Very Low) → 3 (High) + ~3× tighter benchmark thresholds.** Desktop stays High permanently (intended); fresh mobile renders at High until the async benchmark downgrades it — a real first-scene spike on weak GPUs. Matches the perf-retune commits, so likely by design.

### 🧹 Dead code to remove or gate
- `vertex_strip.rs` (whole 116-line module) + `metrics::record_stripped` — zero callers.
- `mesh_occluder.rs::try_spawn_for` is `return false`, yet its `scene.rs::walk_and_preprocess` chain still walks every node of every asset-server GLTF.
- `gltf/common.rs` carries ~400 lines of `#[allow(dead_code)]` LOD functions duplicating the one live path; bone-detection block copy-pasted 4×.

Minor: `surface_get_arrays(s)` fetched 3× per surface in the live LOD bake path; `processor.rs` docstring promises per-tier `.res` but writes one hash-keyed file; LOD pixel thresholds in GDScript vs SSE keys hardcoded in Rust (drift risk).

_🤖 Automated review via Claude Code._

manuelmaceira:
Addressed your review:

| # | Fix | Commit |
|---|---|---|
| 🟠 #1 PCT2 fork-version | `assert_pct2_serialization_ok()` runs once at first `create_compressed_texture` call — packs/reloads a 4x4 ETC2 PCT2 via `.res`, panics with a clear message if the fork binary is missing the serialization fix. Asset-server bake either ships correctly or fails loud at startup. | 516fdb75 |
| 🟡 #2b tween reset | `existing_tween.last_emitted_state = None` in the `reset_tween` branch — the "emit only on transition" perf path was leaving re-armed completed tweens stuck in `TsCompleted` for SDK consumers. | 516fdb75 |
| 🧹 #4 vertex_strip dead | Whole `asset_preprocessor` module deleted (`vertex_strip.rs`, `metrics.rs`, `mesh_occluder.rs`, dead `stripped_*` counters). | af1f70e5 |
| 🧹 #5 mesh_occluder no-op caller | Same as #4 — the entire `apply_asset_server_optimizations` → `walk_and_preprocess` traversal in `scene.rs` is gone. | af1f70e5 |

Two intentional WONTFIX:
- **#2a `current_time` no longer streams** — that's the perf change motivating #2b's transition-only pattern (~135 dirty entries/frame less into V8 recv). No in-repo SDK consumer reads `current_time` per-frame, the standard helper keys off `state`.
- **#3 synchronous `ResourceLoader.load` in gltf_container.gd** — confirmed by your review note as the deliberate Mali deadlock fix. The frame-hitch tradeoff is real but smaller than the alternative.

Also unrelated to your review, this PR went through a few late-stage decisions worth flagging:
- Octahedral impostors removed entirely (16da5eec) — too many on-device validation issues across the asset-pipeline ↔ container ↔ R2 chain.
- Auto-occluder spawn disabled (70047448), runtime culler off (04fdcdfc), then the whole preproc module dropped (af1f70e5).
- Asset-server bake reverted to `--headless` (1f6275d3) — no more SubViewport rasterizer means the Mesa llvmpipe + Xvfb stack was pure overhead (was tripping the orchestrator's 10-min `ASSET_SERVER_TIMEOUT_MS`).

Net effect of the PR is now: ETC2 source-texture pipeline (`fb979dd1` PCT2 fix + #1 self-test), Mali-friendly shadow profile, mesh splitter/LOD chain in common.rs, all the rest of the pre-impostor mobile-perf work.

kuruk-mm (commented):
## Re-review — after `516fdb75c` / `1f6275d38` / `af1f70e5f`

Thanks for the quick turnaround. Status of the 8 findings from my first pass:

### ✅ Resolved
- **Dead code (`vertex_strip` / `mesh_occluder` / `metrics`)** — modules deleted, `scene.rs` walk removed, **no dangling references** anywhere (grepped clean). 👍
- **`tween.rs` `last_emitted_state` reset** — `existing_tween.last_emitted_state = None` in the `reset_tween` block is exactly right; a re-armed entity now re-emits `TsActive` on the next tick.
- **PCT2 fork-version coupling** — the `assert_pct2_serialization_ok()` self-test is a solid upgrade over a bare comment. It runs once, and it sits **inside the existing `GodotSingleThreadSafety` / `GodotThreadSafetyGuard`** (both `load_image_texture` and the GLTF bake hold the guard), so it introduces **no new threading race**. Two small notes inline.

### 🟡 Acknowledged / WONTFIX (agree)
- **Synchronous `.scn` load** — deliberate Mali deadlock fix, documented tradeoff. Fine to leave.
- **`TweenState.current_time` no longer streams mid-flight** — intentional perf change, no in-repo consumer. Fine.

### 🔴 Still open — not addressed by these commits
The fix commit renumbered findings and its "#1" was the PCT2 item — but my original **#1 critical was a different issue and is still in the tree:**

- **🔴 `gltf_container.gd:155` — the PVS null-call crash is UNCHANGED.** `load("res://src/tools/dcl_pvs_runtime.gd").new()` still references a script that doesn't exist → hard crash whenever the `pvs-runtime` deeplink is enabled with a `.pvs.bin` sidecar. This is the one I'd block merge on.
- **🟠 `content_provider.rs:98` — the main-thread pack-mount threading race is UNCHANGED.** `load_resource_pack` still runs off the `godot_single_thread` semaphore via `MAIN_THREAD_PACK_QUEUE` → `process()`.
- **🟠 `run.rs:823` — Android arg forwarding still non-functional** (extras never reach the command line; `--es command_line_params` is the fix).

Net: the cleanup + correctness-of-fixes look good; the **PVS crash is the remaining must-fix**, with the threading race and Android-args as the next two. Inline re-flags below.

_🤖 Automated re-review via Claude Code._

godot/src/decentraland_components/gltf_container.gd
kuruk-mm: 🔴 **STILL OPEN (my original #1 critical).** This wasn't touched by the fix commits — the `fix(review)` commit's "#1" referred to the PCT2 item, not this. `res://src/tools/dcl_pvs_runtime.gd` still does not exist in the repo, so this is a hard null-call crash on the `pvs-runtime` path. Please ship the runtime script or remove this branch before merge.

lib/src/content/content_provider.rs
kuruk-mm, line 98: 🟠 **STILL OPEN.** `load_resource_pack` continues to run via `MAIN_THREAD_PACK_QUEUE` → `process()` without the `godot_single_thread` semaphore, so it can still execute concurrently with guarded worker GLTF gen / `ResourceSaver`. Acquiring `godot_single_thread` in the main-thread drain (or documenting why it's safe here) would close it.

src/run.rs
kuruk-mm: 🟠 **STILL OPEN.** Godot's Android launcher reads only the `command_line_params` string-array extra; `-e <key> <value>` extras never reach `DclCli`, so all forwarded flags are dropped (and `--rust-log=debug` is mangled to key `rust-log=debug`=`true`). Forward via `--es command_line_params "<space-joined args>"` instead.

lib/src/content/texture.rs
```diff
@@ -420,8 +422,26 @@ fn pad_image_to_multiple_of_4(image: &mut Gd<Image>) -> bool {
...
+/// serialization fix is in (PR decentraland/godotengine#14, merged
+/// 2026-06-03 as `4.6.2.stable.gh.9ee6af7ab`). Older fork binaries
+/// (`stable.gh.7e5bfa428` and earlier) ship PCT2 with no compressed
+/// buffer and would render magenta on device — confirm the container's
+/// `decentraland.godot.client.x86_64 --version` is `gh.9ee6af7ab` or
+/// newer before deploying.
+///
 /// Falls back to uncompressed texture if compression fails.
 pub fn create_compressed_texture(image: &mut Gd<Image>, max_size: i32) -> Gd<Texture2D> {
+    assert_pct2_serialization_ok();
```
kuruk-mm, line 444: ✅ Good fix — and verified it runs under the existing thread guard, so no new race. Two small notes, neither blocking:

1. **It also runs on-device (mobile), not just the asset-server.** `create_compressed_texture` is reached at runtime on iOS/Android (texture.rs:387 + the GLTF path), so this self-test fires once per mobile session on the first texture, adding a small disk round-trip. On a (hypothetical) broken device binary it converts magenta-rendering into a panic — arguably fine as fail-fast, just worth being aware it's not asset-server-only.
2. **`panic!` here runs inside a tokio worker task**, not at startup. gdext/tokio catch the unwind, so it likely aborts just that one texture/GLTF task rather than halting the process — so the bake may continue with other assets rather than "failing fast at startup" as the doc-comment says. Worth confirming the bake actually aborts on this (e.g. a process-level check) if hard-stop is the intent.

manuelmaceira:
Addressed the three re-flagged findings in 486cf4b2:

| # | Fix | What |
|---|---|---|
| 🔴 PVS null-call crash | gltf_container.gd:140-158 removed | The whole runtime block deleted. The `.pvs.bin` sidecar load was the call site; the loaded script doesn't exist. The bake side was already a no-op too — no `apply_pvs_bake` exists in lib/src, so the packer's existence-gated sidecar attach (packer.rs:184) never ran. Removed both sides. |
| 🟠 pack-mount threading race | content_provider.rs:310 wrapped in `try_acquire_owned()` on `godot_single_thread` | Skips this tick if a worker is mid-Godot-call; runs again next frame. Can't `await` from sync `process()` and a blocking acquire would deadlock the engine. |
| 🟠 Android arg forwarding | run.rs `extras` → single `--es command_line_params "<joined>"` | The per-flag `-e foo true` form was a no-op — Godot's GodotApp Android launcher only reads `command_line_params` (whitespace-split into argv). |

Apologies for renumbering on the previous response — should've checked the original ordering. The PCT2 self-test + tween reset (516fdb75) and the dead-code cleanup (af1f70e5) stand from before.

kuruk-mm (approved):
looking great!

PR #2192, opened 2026-06-01
fix: avatar impostors freeze instead of throttling near screen edges

### LINKED ISSUE #2177: [Bug]: Remote avatars looking like they’re floating when using the isometric camera
### 🕹️Platform

iOS

### 🔢App Version

490

### 📱Device Information

Iphone 15

### 📄Issue description

When playing on mobile with the isometric camera enabled, remote avatars do not properly display their walking/running animations. Instead, other players appear to glide or float across the map while moving.

**Actual Behavior:**
Remote avatars slide/float while moving instead of properly animating.

### ✅Expected behavior

**Expected Behavior:**
Remote avatars should correctly play their walking/running animations while moving.

### 🖼️Screenshots / Media

https://github.com/user-attachments/assets/6ffda30a-dde2-4acd-9d7d-a8525b78d000

### ▶️Steps to reproduce

1. Open toastedllama.dcl.eth on two mobile devices
2. Join a match in Dead Surge with both players and keep the ISO camera enabled
3. Move around with one player
4. Observe the other player from the second device
5. Notice the remote avatar appears to float/glide instead of playing walking/running animations properly

### 👁️Occurrence

100% of the time

manuelmaceira:
Closes #2177

## Problem

Avatar impostors at MID/CROSSFADE range sometimes **froze on a static pose instead of throttling** — the bug reported as "se clava sin animación en vez de throttlear" when going from fully-animated to throttled.

### Root cause

Throttled avatars drive the `AnimationTree` in `MANUAL` callback mode and advance it every N frames. The off-screen animation freeze cleared `_anim_throttle_active` **without resetting the callback mode**, leaving the tree in `MANUAL` with nothing to advance it. The per-frame re-activation (`if not active: active = true`) then forced `active = true`, producing the stuck state:

> `active = true` + callback `MANUAL` + `_anim_throttle_active = false` → tree "active" but never advanced → **frozen pose**.

It only showed up "sometimes" because the freeze was driven by the LOD coordinator's **approximate, 6-frame-lagged** frustum test (a single `is_position_in_frustum` on a sphere point). Avatars at **screen edges** or sweeping in **during a camera pan** got flagged off-screen while Godot still drew them — so they sat frozen-while-visible and never hit the clean restore path.

## Fix

- **Drive the freeze off the avatar's real on-screen state** via `VisibleOnScreenNotifier3D` (`screen_entered`/`screen_exited`) instead of the coordinator's approximate flag, so freeze ⟺ actually not drawn (no edge-band or lag false positives). The coordinator's `_off_frustum` now only governs impostor-slot release.
- **Route every throttle change through `set_animation_throttle`** so callback mode and the throttle flag always move together. The freeze now resets the callback to `IDLE`, so even a stray re-activation animates normally instead of freezing.
- **Centralize the animation drive** in `AvatarLODHelpers` (`resolve_anim_drive` / `apply_screen_freeze` / `ensure_anim_active`) as a single source of truth, replacing the scattered per-LOD-state animation toggles.
- The local player is never frozen.

## Test

Adds `godot/src/test/avatar/test_avatar_anim_throttle.gd` — a headless `SceneTree` regression test (same style as `test_avatar_modifier_exclude_ids.gd`) that walks the real transition sequences (FULL→MID→off-screen→ticks→on-screen + the full on/off × LOD matrix) over a duck-typed avatar with a real `AnimationTree`, using the real `AvatarLODHelpers` functions, and pins the invariant:

> an on-screen avatar in FULL/MID/CROSSFADE is **never** left `active + MANUAL + throttle-off` (frozen while drawn).

Verified the test **fails on the old behavior** (reintroduced the bug → `FROZEN-WHILE-DRAWN` detected, exit 1) and **passes on the fix** (exit 0). Run it with:

\`\`\`
.bin/godot/godot4_bin --headless --path godot --script res://src/test/avatar/test_avatar_anim_throttle.gd
\`\`\`

## Verification

- Pre-commit `full-tests` green (fmt / gdlint / cargo check / clippy / rust tests / GDScript validation).
- Deployed and confirmed working on physical iPhone 16e (iOS) and Galaxy A54 (Android).

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar.gd
+++ b/godot/src/decentraland_components/avatar/avatar.gd
@@ -12,6 +12,12 @@ const DEBUG_SAVE_AVATAR_DATA = false
 # Useful to filter wearable categories (and distinguish between top_head and head)
 const WEARABLE_NAME_PREFIX = "__"
 
+# AABB for the off-screen freeze notifier — sized to cover the avatar including
+# arms-out emote poses. Erring large is the safe direction: a too-eager "on
+# screen" only animates an avatar that might be off-screen, whereas a too-eager
+# "off screen" freezes a drawn one.
+const SCREEN_NOTIFIER_AABB: AABB = AABB(Vector3(-1.0, -0.3, -1.0), Vector3(2.0, 2.8, 2.0))
+
 const TOON_SHADER = preload("res://assets/avatar/dcl_toon.gdshader")
 const TOON_SHADER_ALPHA_CLIP = preload("res://assets/avatar/dcl_toon_alpha_clip.gdshader")
 const TOON_SHADER_ALPHA_BLEND = preload("res://assets/avatar/dcl_toon_alpha_blend.gdshader")
@@ -174,10 +180,19 @@ var _impostor_layer_is_overflow: bool = false
 # entirely — no multimesh instance, no real layer, no capture. Disk cache makes
 # re-entry fast (texture rehydrates from PNG without recapture).
 var _off_frustum: bool = false
-# Latched while off-frustum: the AnimationTree was paused regardless of LOD
-# state, so when we come back in-frustum we know we have to restore the
+# Driven by _screen_notifier's screen_entered/screen_exited signals: Godot's
+# exact draw state for this avatar. The animation freeze keys off this, NOT the
+# coordinator's approximate _off_frustum sphere test — otherwise an avatar near
+# the screen edge (still drawn, but flagged off-frustum) or one sweeping into
+# view during the coordinator's 6-frame update lag freezes on a stale pose
+# instead of throttling. Default true so a freshly spawned avatar animates until
+# the notifier reports its real state.
+var _on_screen: bool = true
+var _screen_notifier: VisibleOnScreenNotifier3D = null
+# Latched while frozen off-screen: the AnimationTree was paused regardless of LOD
+# state, so when we come back on-screen we know we have to restore the
 # state-driven anim setup (active/manual/throttle).
-var _anim_frozen_off_frustum: bool = false
+var _anim_frozen_off_screen: bool = false
 # Wall-clock ms when the freeze started. Used to advance the AnimationTree by
 # the elapsed time on re-entry so the emote phase matches what it would have
 # been had we not paused — single one-shot recompute, not a frame-by-frame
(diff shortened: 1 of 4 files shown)


regenesis-claw (approved):
Clean, well-reasoned fix. The root cause analysis is excellent — the stale `MANUAL` callback mode + re-activation producing the frozen-while-drawn state is clearly traced, and the fix addresses it structurally rather than with another edge-case patch.

**Pre-flight**

- Branch is 2 commits behind `main` — minor, but worth merging before final merge to ensure CI signals are trustworthy.
- Android build was cancelled on the first push; latest CI run still in progress (static checks, Linux, Clippy, Docker all green on previous run).
- No iOS build needed — pure GDScript avatar animation logic, no platform-specific code paths.

**What I like**

- `VisibleOnScreenNotifier3D` is the right tool here — Godot's actual draw state vs. the coordinator's approximate 6-frame-lagged sphere test eliminates the false-positive edge cases by design.
- `resolve_anim_drive` as a pure function is great — makes the policy testable and readable in one place.
- The `_on_lod_state_changed` collapse from a repetitive 4-arm match into a compact `resolve_anim_drive` call is a nice cleanup that also closes the invariant gap.
- `ensure_anim_active` guarding against undoing the freeze is the minimal, correct fix for the `_process` re-activation that was the final ingredient of the bug.
- The test is thorough: covers the full on/off × LOD matrix, the exact bug-trigger sequence (MID → off-screen → tick → on-screen), and even verifies the invariant check itself detects the old bug state (the synthetic old-bug-state test). The `FakeAvatar` duck-type approach keeps it headless without needing a full scene.
- `SCREEN_NOTIFIER_AABB` comment explaining why erring large is safe is the kind of "why" comment that helps future readers.
- Default `_on_screen = true` so freshly spawned avatars animate until the notifier reports — correct, avoids a flash-of-frozen.

**Minor notes (non-blocking)**

- The `LOD_*` constants in `AvatarLODHelpers` duplicate `Avatar.LODState` to avoid a cyclic dependency — the test asserts they match, which is good. If GDScript ever gets `const` imports or the enum is extracted to its own file, this can be unified. Fine as-is.
- One `VisibleOnScreenNotifier3D` per avatar is lightweight but worth keeping in mind if avatar counts scale significantly. Not a concern at current mobile caps (~20 avatars).

Approving — merge after CI is green and the branch is up to date with `main`. Solid fix.

PR #2036, opened 2026-05-11
perf(rendering): shadow size 512 + 30m range + 2-split cascade (split_1=0.3)

manuelmaceira:
## What

Config-only shadow-pass cost tweaks:

| setting | before | after |
|---|---:|---:|
| `rendering/lights_and_shadows/directional_shadow/size` | 2048 | **512** |
| MainLight `directional_shadow_max_distance` | 50.0 | 30.0 |
| MainLight `directional_shadow_mode` | default (`SHADOW_PARALLEL_4_SPLITS=2`) | `SHADOW_PARALLEL_2_SPLITS=1` |
| MainLight `directional_shadow_split_1` (cascade-1 ratio of max_distance) | 0.1 (≈3m) | 0.3 (≈9m) |

No Rust code, no shader code. Only `godot/project.godot` and `godot/assets/environment/sky_lights.tscn`.

## Why

A54-class Mali-G68 spends significant fragment-pass time rasterizing into the directional shadow map atlas. Reducing atlas size, max distance, and cascade count all cut shadow-pass cost.

**Why 512 (not 256):** at 256 the near-field shadows were too coarse and Genesis Plaza had visible artifacts (oversized cascade-0 footprint on close objects). 512 keeps the perf win and the visual quality is good.

**Why push cascade-1 split to 0.3:** the default 0.1 ratio puts the boundary at ~3m from the camera, which is right where the player's avatar shadow lives — the boundary was visible as a sharpness step. Pushing it to 0.3 (~9m) hides the transition past the avatar.

## Bench (A54 HIGH, GP preview, warm cache)

> Numbers below are from the 256/0.1 configuration. The current branch tip uses 512/0.3 which trades back a small amount of GPU for the visual fix — rerun pending; expected delta vs baseline is still ≈ −30% render_gpu_ms.

| metric | baseline | after (256/0.1) | Δ |
|---|---:|---:|---:|
| fps.mean | 7.14 | 8.17 | +14.4% |
| render_gpu_ms.mean | 95.33 | 54.41 | −42.9% |
| render_cpu_ms.mean | 7.35 | 7.06 | −3.9% |
| video_mem_mb | 778.67 | 756.77 | −2.8% |
| texture_mem_mb | 630.43 | 605.42 | −4.0% |
| load_seconds | 26.28 | 24.80 | −5.6% |

### Visual diff

| baseline | this PR |
|---|---|
| baseline | after |

## Optimization opportunities spotted but NOT addressed here

[...]

PR #2197, opened 2026-06-02
perf(rendering): mobile GPU optimization stack — one commit per feature

manuelmaceira:
Clean re-roll of #2025 with **one commit per feature** for easier review. Same shipping
content as #2025 **minus the experimental `--rs-gltf-direct` MultiMesh path** (left for a
follow-up PR). Branched fresh off `main`; each commit compiles independently (cargo
check/clippy + gdlint per commit).

#2025 stays open for reference.

### Commits (each a self-contained feature)
1. **feat(content): optimized-content-base-url override** — `--optimized-content-base-url` CLI/deeplink → DclCli, resolved by content_provider.
2. **perf(gltf): coalesce GltfContainer load-timeouts** — one `GltfLoadTimeoutCoalescer` Node replaces ~1419 per-container Timers in Genesis Plaza.
3. **perf(dcl): split CRDT recv into wait+drain** — reuse a persistent JS recv buffer instead of per-frame `Vec<Vec<u8>>` churn.
4. **perf(tween): emit TweenState only on transition** — cuts ~51% of dirty LWW entries/frame in GP.
5. **perf(avatar): async impostor PNG encode/save** — moves the ~200–500ms encode+write off the render thread.
6. **perf(graphics): retune default profile + Mali shadow atlas** — default profile High; low-tier shadow atlas 1024+16-bit.
7. **perf(content): bake LOD chain + dedup shared-mesh VBOs** — native generate_lods at import + per-profile `mesh_lod_threshold`.
8. **perf(asset-server): auto-bake BoxOccluder3D** — coarse occluders baked into the optimized `.scn` (asset-server only).
9. **feat(bench): Genesis Plaza profiling harness (#1862)** — GP benchmark runner + deeplink/CLI diagnostic knobs (`--skip-gltf`, `--kill-sky`), device orchestration, content-author guide.

### Shipping surface
- **Runs on-device at runtime (default ON):** timer coalescer, CRDT buffer reuse, TweenState gate, async impostor I/O, graphics-profile retune.
- **Baked at the asset server (device only via pre-baked optimized `.scn`):** LOD chain + VBO dedup, BoxOccluder, `mesh_lod_threshold`.
- **Dev/bench only:** GP benchmark harness + diagnostic knobs.

### Verification
`cargo check`/`cargo clippy -- -D warnings` and `gdlint` pass per commit; full `cargo clippy -D warnings` + `gdlint godot/` clean at the tip. Diff vs #2025's branch is exactly the excluded `rs-gltf-direct` experiment, nothing else.

---

## Benchmarks — Samsung A54, HIGH profile, GP `ea9cc738` (same hardware, same content)

- **Low-end devices:** no longer OOM-crashes. VBO/material dedup at bake time recovers shared-mesh sharing → **total Pss 3.56 GB → 1.33 GB (−63 %)**, GL mem 1.88 GB → 715 MB (−62 %).
- **Performance:** end-to-end **+33 % fps (13.9 → 18.5)**, **−23 % GPU frame time (67.3 → 51.7 ms)**; this stack alone adds **+5 % fps / −5 % GPU** over current `main`.
- **Visual regressions:** none relevant — the LOD bake skips transparent/alpha-tested materials so foliage/billboards stay intact; visible meshes cast shadows normally.

### A/B/C/D progression (each row = a different APK vs the same GP `ea9cc738`)

| state | fps | render_gpu_ms | primitives | video_mem |
|---|---:|---:|---:|---:|
| **A** — pre-perf baseline | 13.88 | 67.27 | 2,967,296 | 762 MB |
| **B** — + #2036 shadow tweaks | 17.06 | 55.36 | 2,278,718 | 710 MB |
| **C** — + current `main` render fixes | 17.60 | 54.26 | 2,181,687 | 735 MB |
| **D** — + this stack | **18.51** | **51.66** | **1,920,326** | **690 MB** |

| step | Δ fps | Δ render_gpu_ms |
|---|---:|---:|
| A → B (#2036 shadow tweaks, already merged) | **+3.18 (+23 %)** | −11.9 ms |
| B → C (render-side fixes on main) | +0.54 | −1.1 ms |
| C → D (this stack) | **+0.91 (+5 %)** | −2.6 ms |
| **A → D total** | **+4.62 (+33 %)** | **−15.6 ms (−23 %)** |

### Device memory (`dumpsys meminfo`, end-of-bench)

| metric | pre-dedup | post-dedup (this stack) | Δ |
|---|---:|---:|---:|
| **GL mtrack** | 1,882 MB | **715 MB** | **−1,167 MB (−62 %)** |
| EGL mtrack | 71 MB | 23 MB | −48 MB |
| Native Heap | 1,107 MB | 476 MB | −631 MB |
| **TOTAL Pss** | 3,556 MB | **1,328 MB** | **−2,228 MB (−63 %)** |

### Asset-server output size

| | before dedup | after dedup |
|---|---:|---:|
| `.scn` total disk | 447 MB | **160 MB (−64 %)** |


regenesis-claw (approved):
## Review — perf(rendering): mobile GPU optimization stack

Followed `REVIEW.md`. All CI checks green ✅, iOS build present and passing ✅, branch not stale. No submodule drift. No `.claude/` files. No stray `print()` debug leftovers.

Excellent PR — the one-commit-per-feature structure makes a 1824-line diff genuinely reviewable, and the A/B/C/D benchmark table in the body is the gold standard for perf PRs. The −63% Pss / +33% fps numbers on the A54 speak for themselves. Nine distinct optimizations, each self-contained and well-commented.

### Findings

#### 🟡 `enable_frame_pacing=false` ships to production (project.godot)

`window/frame_pacing/android/enable_frame_pacing=false` is set project-wide, not gated on bench mode. Android frame pacing (via Swappy) helps smooth frame delivery on devices with variable refresh or DVFS-driven clock scaling. Disabling it is correct for benchmarking (avoids the pacing library capping to vsync), but shipping it to all production users may cause micro-jank on devices that previously benefited from it. Was this intentional for production, or should it be a bench-only toggle?

#### 🟡 Default graphic profile: Very Low → High

`config_data.gd` now defaults to `graphic_profile = 3` (High) instead of 0 (Very Low). The HardwareBenchmark will downgrade on first launch — but there's a brief window between app start and benchmark completion where the device renders at High (FPS_60, full bloom, high shadows). On truly low-end devices this could cause a thermal spike or dropped frames during the benchmark itself, potentially skewing its own result. If this was measured and found acceptable on the A54, no action needed — just flagging the interaction.

#### 🟡 Shadow proxy code is dead

`create_scene_colliders` sets `let shadow_proxy = false;` and the entire `create_scene_colliders_inner` branch for `shadow_proxy == true` (SHADOWS_ONLY + front-cull material + visible cast_shadow OFF) is unreachable. The comment explains the zeppelin animation break — but ~40 lines of dead code (including `build_shadow_proxy_material()`) ship in the binary. Either remove it or gate behind a flag/cfg if you plan to revisit. Low priority — not blocking.

#### 🟢 `MAIN_THREAD_PACK_QUEUE` — nice fix

The deadlock from calling `load_resource_pack` on a tokio worker was a real production hazard. Draining via `process()` on the main thread with `oneshot` reply channels is the right pattern. Clean.

#### 🟢 CRDT recv split (wait + drain)

The two-op design avoids per-frame V8 BackingStore allocation. The power-of-two grow strategy in `EngineApi.js` and the `slice()` vs `subarray()` note are both correct. One minor observation: `op_crdt_recv_drain` returns 0 for both "no pending buffer" and "buffer too small" — since JS always calls drain immediately after wait, the ambiguity is harmless in practice, but a comment in the Rust side noting this would help future readers.

#### 🟢 GltfLoadTimeoutCoalescer

Replacing 1419 Timer nodes with a single Dictionary + `_process` is a clean win. The note about the VkThread NULL-deref crash when adding `_process` to DclGlobal itself is valuable (and bisected — thank you).

#### 🟢 Async impostor PNG encode

Moving zlib compression off the render thread (13.8% inclusive in GP profile) via the `png` crate on a tokio blocking worker is well-motivated. The atomic write (`.tmp` → rename) prevents partial-read races. The `try_load_cached_texture` switch from `Image::load_from_file` to `FileAccess::get_file_as_bytes` + `load_png_from_buffer` fixing the Android `user://` resolution issue is a good catch.

#### 🟢 LOD + VBO dedup pipeline

The skip guards (blend shapes, skinned, tiny indices, transparent materials) are thorough. The `baked_cache` reuse across MeshInstance3Ds sharing the same source ArrayMesh preserves VBO dedup — this is the key to the −63% GL mtrack number. Asset-server-only gate via `apply_optimizations` is correct.

#### 🟢 PortableCompressedTexture2D for ASTC

Fixes a real silent-corruption bug where NVIDIA Vulkan decompresses ETC2 back to RGBA8 during the GPU readback that `ImageTexture` serialization triggers. PCT2's CPU-side buffer retention sidesteps this entirely. Good root-cause analysis in the comment.

### Summary

Strong PR. The three 🟡 items are non-blocking — the frame pacing one is the most worth confirming before merge. Approving.

kuruk-mm:
## 📱 Avatar-impostor benchmark — on-device run

Ran the avatar-impostor benchmark on a physical **Motorola Edge 60 Pro** from this branch (fresh APK built + exported from `feat/mobile-perf-stack`).

**How it was triggered** (Android wires this benchmark via deeplink, not the xtask):
```bash
adb shell am start -a android.intent.action.VIEW \
  -d "decentraland://open?benchmark=avatar-impostors" org.decentraland.godotexplorer
```

### Result — 200 avatars, all in frustum (5–60 m), 15 s measured per phase

| Phase | avg FPS | avg ms | p99 ms | p1 ms |
|-------|--------:|-------:|-------:|------:|
| Impostors **OFF** | 5.50 | 181.9 | 296.1 | 155.1 |
| Impostors **ON**  | 12.38 | 80.8 | 200.4 | 74.0 |

**Delta FPS: +125.3%** — impostors more than double framerate under this stress load. Values match between logcat and the on-device `user://impostor_benchmark.log`.

### ⚠️ Real issue observed on this GPU
The impostor **capture pipeline failed heavily** during pre-fetch:

- **782×** `RGBA buffer too small: got 131072 bytes, expected 524288` (`avatar_scene.rs:1273`)
- **781×** `Failed to generate impostor mipmaps: ERR_UNAVAILABLE` (`avatar_scene.rs:611`)
- Pre-fetch only reached **102/160 impostor slots loaded** (~36% never got a texture)

The got/expected ratio is exactly **1:4** (131072 = ¼ of 524288), which looks like the capture render target is coming back at half-resolution per axis on this device. So **+125.3% is a conservative floor** — that's the gain with a third of impostors missing their texture. Worth a follow-up on the capture-size assumption.

kuruk-mm:
Superseded by #2191. The mobile GPU optimization stack and the octahedral-impostor work have been consolidated into a single linear, one-commit-per-change branch (`feat/octahedral-impostors-v2`) rebased on `main`. All 9 perf-stack commits from here are preserved verbatim there (commits 1–9), followed by the texture-pipeline + impostor work. Closing this in favor of merging #2191.

PR #2202, opened 2026-06-03
fix: phantom impostors on hidden avatars + catalyst body snapshot discarded

### LINKED ISSUE #2203: Phantom impostors on the Tower of Madness podium when no winner is visible
## Symptom
On the Tower of Madness podium, when no winner avatar should be visible, moving away from the podium spawns **3 identical phantom impostors** where the winner `AvatarShape`s sit. They look like an invisible person rendered as a billboard (and, depending on state, a single floating wearable).

## Root cause
The podium `AvatarShape`s are hidden by their **parent entity** via a `VisibilityComponent` (entity `visible=false`, `scale≈0`) while the inner `Avatar` node keeps `visible=true`. The LOD code gated impostor participation on the Avatar node's own `visible` flag (which stays `true`), so the avatar still entered the impostor system and `AvatarScene`'s MultiMesh drew a full-size impostor at its world position — ignoring the parent's tree visibility/scale (the impostor lives on a separate MultiMesh, not under the avatar node).

Confirmed live via the debug WS inspector: the 3 `podium-avatar` entities report `VisibilityComponent.visible=false` while the Avatar node reports `visible=true`.

## Fix
Gate impostor participation on `is_visible_in_tree()` (which folds in every ancestor's visibility) instead of the node's own `visible` flag, in both `_update_lod` (`avatar.gd`) and the LOD coordinator.

Fixed in #2202.

### LINKED ISSUE #2204: Frame hiccups when avatars appear: catalyst body snapshot discarded, impostor always baked locally
## Symptom
Frame hiccups when remote avatars appear / cross the impostor (FAR) distance threshold.

## Root cause
`ContentProvider.fetch_avatar_body_texture` resolves a **`TextureEntry`** (`image` + `texture` + `failed`), but `impostor_capturer.gd::_async_fetch_catalyst_body` checked `result is Texture2D` — which **never matches** — so every catalyst body snapshot was discarded **even when the download succeeded**, and impostor capture always fell back to the expensive ~300ms **main-thread** off-screen local bake (`_async_capture_local`, which re-assembles the full avatar + renders it off-screen).

## Measured (Genesis Plaza, instrumented build)
| | catalyst-hit | local-bake (main thread) |
|---|---|---|
| before | 0 / 14 | 14 × ~307 ms |
| after  | 32 / 32 | 0 |

The catalyst fetch (~296 ms) runs off the main thread (async tokio), so it no longer blocks the frame.

## Fix
Read `TextureEntry.image` directly (and skip the placeholder marked by `TextureEntry.failed`) instead of the never-matching `is Texture2D` check.

Fixed in #2202.

manuelmaceira:
Closes #2203
Closes #2204

Two avatar/impostor fixes found and validated locally (desktop, Genesis Plaza + Tower of Madness), plus a safer impostor fallback for `AvatarShape`s.

## Bug 1 — phantom impostors on the podium (#2203)

**Symptom:** On the Tower of Madness podium, when no winner avatar should be visible, moving away spawned 3 identical phantom impostors where the `AvatarShape`s sit.

**Root cause:** The podium `AvatarShape`s are hidden by their **parent entity** via a `VisibilityComponent` (entity `visible=false`, `scale≈0`) while the inner `Avatar` node keeps `visible=true`. The LOD code gated impostor participation on the Avatar node's own `visible` flag, which stayed `true`, so the avatar entered the impostor system and `AvatarScene`'s MultiMesh drew a full-size impostor at its world position — ignoring the parent's tree visibility/scale.

**Fix:** Gate on `is_visible_in_tree()` (folds in every ancestor's visibility) in both `_update_lod` (`avatar.gd`) and the LOD coordinator.

Confirmed live via the debug WS inspector: the 3 `podium-avatar` entities have `VisibilityComponent.visible=false` + the Avatar node `visible=true`, and after the fix they no longer hold an impostor slot.

## Bug 2 — hiccups when avatars appear (always baking locally) (#2204)

**Symptom:** Frame hiccups when remote avatars cross the impostor (FAR) threshold.

**Root cause:** `fetch_avatar_body_texture` resolves a `TextureEntry` (`image` + `texture` + `failed`), but `_async_fetch_catalyst_body` checked `result is Texture2D` — which never matched — so **every** catalyst body snapshot was discarded even when the download succeeded, and impostor capture always fell back to the ~300ms **main-thread** off-screen local bake.

**Fix:** Read `TextureEntry.image` directly and skip the placeholder marked by `TextureEntry.failed`.

**Measured (Genesis Plaza, instrumented build):**
| | catalyst-hit | local-bake (main thread) |
|---|---|---|
| before | 0 / 14 | 14 × ~307ms |
| after | 32 / 32 | 0 |

The catalyst fetch (~296ms) runs off the main thread (async tokio), so it no longer blocks.

## AvatarShape fallback — no main-thread bake

Following team feedback: the impostor capture fallback (scene `AvatarShape`s with no eth id, or profiles without a body snapshot) **must never** run the off-screen local bake, which re-assembles the whole avatar on the main thread (~300ms hiccup). It now fetches one of the catalyst's 160 stock default body snapshots (`profiles/default{N}`) instead — chosen stably per avatar so it doesn't flicker, cached per slot (downloaded at most once each), off the main thread. The local-bake path is removed.

Real generation will return behind an async queue — tracked in **#2206** (implemented in a follow-up branch stacked on this one).

## Tests
- New headless regression test `test_avatar_impostor_hidden_parent.gd` pinning the bug-1 invariant (a child under a hidden parent reports `is_visible_in_tree() == false`, accounting for Node3D's deferred visibility propagation).
- `gdformat` / `gdlint` clean; `cargo fmt` / `cargo clippy` clean; existing Rust unit suite (145) passing.

## Notes
- A third, separate animation bug ("remote avatar runs without animation", a refinement of #2192's freeze logic) was investigated but is **not** included here — it needs a live repro to pin down. A confirmed secondary gap was noted: the compressed-movement path (`update_avatar_transform_with_movement_compressed`) doesn't call `apply_wire_movement_state`, so `is_grounded`/`jump_count`/`glide_state` go stale for remotes on that path.
- Draft: pending on-device (Android/iOS) validation of the hiccup improvement, and in-app validation of the default-body fallback (only exercised when an `AvatarShape` becomes visible / a profile has no body snapshot).


godot/src/decentraland_components/avatar/impostor/impostor_capturer.gd
```diff
@@ -64,8 +64,13 @@ func _async_fetch_catalyst_body(user_id: String) -> Image:
...
 	if result is PromiseError:
 		return null
-	if result is Texture2D:
-		return result.get_image()
+	# fetch_avatar_body_texture resolves a TextureEntry (image + texture + failed),
+	# not a bare Texture2D. The old `is Texture2D` check never matched, so every
+	# catalyst body — even when the download succeeded — was discarded and fell
+	# back to the ~300ms off-screen local bake. Read the image directly and skip
+	# the placeholder that TextureEntry.failed marks.
+	if result != null and result.get("image") != null and not result.get("failed"):
```
regenesis-claw, line 78: Minor: since `fetch_avatar_body_texture` resolves with a `TextureEntry` (Rust GDExtension class with `#[var]` fields), you could use the class name directly:
```gdscript
if result is TextureEntry and not result.failed:
    return result.image
```
This is a bit more self-documenting about what the promise actually resolves with, and you'd get a clearer error if the return type ever changes. The `.get()` approach works fine too — just a readability preference.

regenesis-claw (approved):
Solid pair of bug fixes — both root-cause analyses are sharp and the regression test pins the invariant well.

**Pre-flight notes:**
- Branch is 3 commits behind `main`. The changes are self-contained enough that this isn't risky, but worth rebasing before merge so CI signals are trustworthy.
- No iOS-sensitive code paths touched — iOS build not required.
- Android build was _cancelled_ (not truly failed); static checks + Linux + Clippy all green.

**Bug 1 (phantom impostors):** `visible` → `is_visible_in_tree()` is the correct fix. The comment in `avatar.gd` is a model "why" comment — it captures the exact scenario (parent `VisibilityComponent` vs. child `visible` flag) so future readers don't re-derive it.

**Bug 2 (catalyst body snapshot discarded):** The `is Texture2D` → `TextureEntry` type mismatch is a nasty silent bug — 100% local-bake fallback with zero log noise. Nice catch via instrumented build.

**Minor note on the `_async_fetch_catalyst_body` fix:** I left one inline comment — the duck-typed `.get()` approach works, but `is TextureEntry` would give you a compile-time-ish safety net and make the code self-documenting about the actual return type.

**Test:** Good use of `await process_frame` to match Godot's deferred visibility propagation — matches the real `_process` timing.

kuruk-mm (approved):
looking great!

PR #2205, opened 2026-06-03
fix: connection-lost modal shows too frequently + health-check request leak

### LINKED ISSUE #2174: [Bug] Connectivity modal coming up too frequently
## Description
In a bad connectivity environment, users are getting logged out randomly after the no connection modal appears. The modal seems to be triggering too frequently and leading to unexpected logouts.

## Steps to Reproduce
1. Use the app in a bad connectivity environment
2. No connection modal appears
3. User gets logged out randomly

## Expected Behavior
The connectivity modal should handle poor network conditions gracefully without logging the user out unnecessarily.

## Reported By
Gino Cingolani via Slack
Requested by Rodri via Slack

EibrielInv:
## Problem

  The "Connection lost" modal popped far too often. Two root causes:
  
  1. **Counter + retry reset.** Health was a count of consecutive failed pings, and tapping **Retry** reset that counter and forced state back to GOOD. Since Retry doesn't actually reconnect anything (it only
  closes the modal), one fresh failure immediately re-tripped the modal — so a single sustained outage produced a rapid string of modals.
  2. **Leaked health-check requests.** The `/about` poll raced a 3s timeout but never cancelled the underlying request, which lingered on the default 60s HTTP timeout while holding one of the shared request 
  queue's 10 concurrency slots. During an outage these zombies saturated the queue, so after reconnect new checks couldn't get a slot and health stayed pinned at `0.00` (the `Request timed out … health 0.00` 
  spam).

  ## Changes

  **`connection_quality_monitor.gd` — decoupled into two layers:**
  - **Connection health** — a truthful, ping-driven 0–1 score with hysteresis bands (latency-aware: a near-timeout success heals less). The UI never touches it, so a single lucky ping can't flap the modal.
  - **Feedback policy** — toast on GOOD→POOR, modal on LOST, dismiss only on genuine recovery. **Retry is now a snooze**, not a fake reconnect: it hides the modal for a window that backs off exponentially on 
  each retry and decays back to baseline during calm. Health stays objectively LOST underneath.

  **HTTP request timeout (`rust_http_queue_requester.rs`):**
  - New `request_json_with_timeout(...)` exposing the already-plumbed `RequestOption.timeout` to GDScript. The health check now aborts at the HTTP layer (~4s) instead of lingering 60s, so it releases its queue 
  slot promptly and recovery is detected within 2–3 polls of reconnection. This also stops the shared queue from being starved app-wide during outages.

  ## Testing
  - Manual: toggle Wi-Fi off ~60s (toast → modal), back on → health recovers within 2–3 polls, modal dismisses, no lingering timeout spam.

  Closes #2174


regenesis-claw (approved):
## Approved ✅

Clean, well-designed rewrite. The two-layer split (truthful health score vs. feedback policy) is the right architecture — it eliminates the old counter-reset-on-retry flapping and makes each concern independently testable.

### What works well

- **Hysteresis bands** with separate enter/exit thresholds prevent single-ping flapping between states.
- **Snooze-with-exponential-backoff + decay** is a nice touch — it naturally adapts to intermittent vs. sustained outages without an explicit state machine.
- **HTTP-layer timeout** on health-check requests (`request_json_with_timeout`) directly fixes the queue starvation root cause. The extra 1s margin over the async race timeout is sensible.
- `_reset_health()` now properly dismisses any showing modal on realm change — fixes a class of orphaned-modal bugs the old `_resume_polling` missed.
- Recovery logging (`prints`) removed; only `printerr` remains for genuine failures. ✓

1. **Unrelated `.claude/skills/` change** — the `focus` debug command docs are useful but orthogonal to the connectivity fix. Could have been a separate commit/PR for cleaner history. Not blocking.

2. **`request_json_with_timeout` body-parsing duplication** — the `match body.to_string().as_str()` block is copy-pasted from `request_json`. If more timeout-aware callers appear, extracting the body parsing into a shared helper would reduce the surface. Fine for now with only two call sites.

3. **Near-timeout successes heal almost nothing** — a 2.9s success with a 3s timeout yields `latency_factor ≈ 0.033`, so `heal ≈ 0.011`. This means a server responding slowly (but successfully) barely improves health. That's arguably correct for a health-check context (barely making it ≈ still degraded), but worth being aware of during tuning.

Requested by Mateo via Slack

kuruk-mm (approved):

PR #2207, opened 2026-06-03
feat: budgeted, prioritized impostor capture queue (#2206)

### LINKED ISSUE #2206: Implement async queue system for avatar profile generation (body/face snapshots)
## Context

PR #2202 fixed the catalyst body snapshot fetch (corrected the `TextureEntry` type mismatch), which eliminated the bug where **every** catalyst fetch was being discarded and falling back to local rendering. However, when catalyst snapshots are unavailable (profile has no snapshots, fetch fails, etc.), the system still falls back to `_async_capture_local`, which:

1. Instantiates an `AvatarPreview` scene
2. Re-assembles the full avatar with wearables
3. Renders it off-screen via `async_get_viewport_image`
4. The viewport rendering (~300ms) still **blocks the main thread** during capture

While `_async_capture_local` is marked as async and uses `await`, the actual rendering operation is synchronous and causes frame hiccups.

## Proposed Solution

Implement a proper **async/background queue system** for avatar profile generation:

### Requirements

1. **True async rendering** — Move the viewport capture off the main thread or use a frame-budgeted approach that spreads rendering across multiple frames
2. **Persistent queue** — Queue requests for body/face snapshot generation that can be processed in the background
3. **Priority system** — Visible/nearby avatars get higher priority than distant ones
4. **Cache integration** — Generated snapshots should be cached with the same key structure as catalyst snapshots
5. **Graceful degradation** — Impostors should render with a placeholder/silhouette until the snapshot is ready, rather than blocking

### Implementation Ideas

- **Option A**: Render snapshots at reduced resolution (e.g. 128x256 instead of full size) to reduce main-thread time
- **Option B**: Use multiple small SubViewports and render one avatar part per frame, compositing the final image over several frames
- **Option C**: Offload to a background Godot instance (server mode) that renders snapshots and returns them via IPC/file
- **Option D**: Pre-generate snapshots for common avatar configurations during idle time

### Related

- Closes #2204 (catalyst fetch fixed, but local fallback still blocks)
- Related to #1880 (profile generation errors — 829 entities)
- Builds on the impostor system in `godot/src/decentraland_components/avatar/impostor/`

### Benefits

- Eliminates remaining frame hiccups from avatar appearance
- Better UX for users with profiles that lack catalyst snapshots
- More predictable performance in crowded scenes
- Allows for progressive loading of avatar quality

Requested by Mateo via Slack

manuelmaceira:
Closes #2206. **Stacked on #2202** (base branch = `fix/avatar-impostor-podium-bake`) — review/merge that first.

## What

Re-enables real local impostor generation for scene `AvatarShape`s that #2202 had switched to a stock default body, but does it **budgeted and prioritized** so it never bursts into a frame stall.

The `ImpostorCapturer` autoload becomes a two-stage queue:

- **Stage 1 — fetch (cheap, user-visible):** drained closest-in-frustum first. `0x` avatars get the catalyst body; everyone else gets a stock **default body snapshot uploaded immediately as a placeholder**, and (only if they have custom wearables) are handed to stage 2.
- **Stage 2 — gen (expensive, budgeted):** runs only when the fetch queue is idle, **one at a time**, with a frame cooldown (`GEN_MIN_FRAMES_BETWEEN`), **never for off-frustum avatars**, at **reduced resolution** (`128×256`; `set_impostor_texture` upscales to the 256×512 layer). The real snapshot then **swaps** the placeholder.

**Cache-poisoning guard:** placeholders are uploaded with an **empty cache key** so they are never persisted to `user://impostor_cache` under the avatar's identity — only real looks (catalyst / generated) are disk-cached. (A persisted placeholder makes `impostor_needs_capture` stay false forever, silently disabling all future captures for that avatar — found and fixed during live validation.)

Priority pick extracted to `ImpostorCapturePriority` (preloaded, not a global class — no asset-reimport dependency) and unit-tested.

## Requirements coverage (#2206)
- ✅ Persistent queue, ✅ priority (distance + frustum), ✅ cache integration (existing `set_impostor_texture` path, real looks only), ✅ graceful degradation (default placeholder → real swap), ✅ budgeted/spread generation.
- ⚠️ "True off-thread rendering": the dominant per-generation cost is **avatar re-assembly** (`async_update_avatar`: mesh duplicate, toon shader compile) — reduced resolution shrinks render+readback but not assembly. This change removes the *burst* (one-at-a-time, closest first, off-frustum deferred); fully eliminating the assembly cost (reuse assembled nodes / off-process render — issue Option C) is the **noted follow-up**.

## Tests / validation
- New headless unit test `test_impostor_capture_queue.gd` pinning the priority ordering (in-frustum before off-frustum, nearest first, drain order). PASS.
- `gdformat` / `gdlint` clean.
- ✅ **Validated live (desktop):** visible `AvatarShape`s with custom wearables show the default placeholder immediately and swap to their real reduced-res snapshot one at a time (`swapped=true` ×4 in the instrumented run); only real snapshots written to the disk cache.
- ⏳ Pending on-device (Android/iOS) validation of frame pacing.

## Notes
- `.uid` siblings for the two new `.gd` files are generated on editor import (only `preload`-by-path is used, no `uid://` refs).
- Anyone who ran an earlier build of this branch may have placeholder PNGs persisted in `user://impostor_cache/`; they self-expire via the existing cache TTL (or clear the folder manually).


godot/src/decentraland_components/avatar/impostor/impostor_capturer.gd
```diff
@@ -16,48 +35,126 @@ func request_capture(avatar) -> void:
...
-		var key: String = (
-			avatar._get_impostor_cache_key() if avatar.has_method("_get_impostor_cache_key") else ""
-		)
-		Global.avatars.set_impostor_texture(iid, image, key)
+		# The default placeholder must NOT be disk-cached under the avatar's
+		# identity key: a persisted placeholder makes request_impostor_layer
+		# treat the avatar as already captured (impostor_needs_capture stays
+		# false), so the real snapshot would never be generated. Empty key =
+		# upload without persisting; only real looks (catalyst / generated)
+		# are cached.
```
regenesis-claw, line 80: Minor: when `_take_next` removes freed/invalid avatars from the queue (lines 90-94), the corresponding `_enqueued[iid]` entry is never cleaned up since we don't have the instance ID at that point. Over a long session with many avatar churn this dictionary slowly leaks stale keys.

Not blocking — the leak is small (one int key per freed avatar). If you want to clean it up: store the `iid` alongside the avatar ref (e.g. wrap in a dict `{avatar, iid}`) so you can erase `_enqueued` during the cleanup pass. Or just accept the tiny leak — it's bounded by total avatars ever seen in the session.

regenesis-claw (approved):
Clean, well-documented two-stage queue. The priority extraction + unit test is a great pattern. Cache-poisoning guard with the empty key is a smart catch.

One minor observation below — not blocking.

**Positives:**
- Excellent PR description with full design rationale and cache-poisoning diagnosis.
- Priority logic extracted into a preloaded, testable unit — easy to verify and refactor.
- Reduced-res capture + frame cooldown is a pragmatic approach that addresses the burst without over-engineering.
- Placeholder → swap pattern gives instant visual feedback while the real snapshot is budgeted.

Approving — the approach is solid and the code is clean.

PR #2208, opened 2026-06-03
feat: GDScript eval command on the debug WS server + auto-start in debug builds

manuelmaceira:
## What

Extends the debug WebSocket inspector (#2181) with an `eval` command so agents/tools can run arbitrary GDScript against the live client and get the serialized result back — the agent-facing equivalent of the old devtools console expression box, but supporting multi-line statement bodies.

- `{"id":N,"cmd":"eval","code":"..."}` compiles the snippet on the fly (`GDScript.new()` + `reload()`) and calls a generated `_run(tree, global, server)` with the `SceneTree`, the `Global` autoload and the server node in scope.
- Bare single-line expressions are auto-wrapped in `return` (`"1 + 1"` → `2`); multi-line bodies use explicit `return`. A shape heuristic with compile-failure fallback keeps the common case free of spurious parse errors.
- Results are serialized through the collector's existing `_variant_to_json` (primitives, `Vector*`, `Color`, arrays, dicts; `str()` fallback for Objects).
- **Hard-gated out of production builds** — `Global.is_production()` short-circuits the command with an error reply.
- The server now **auto-starts in debug builds** (`OS.is_debug_build()` and not production) so agents can attach without the Settings toggle. Release builds keep the manual toggle; the Developer-tab checkbox now reflects the live server state on open.
- `debug-ws-inspector` SKILL.md documents the command, the locals, the production gate and the limitations (synchronous only — no `await`; no `print()` capture; runtime errors log to the client console and return `null`).

## Verification

Tested against the live client (debug build, macOS):
- Server auto-started on boot: `DebugWsServer: listening on ws://127.0.0.1:9230` with the settings toggle untouched.
- `"1 + 1"` → `{"data":2,"ok":true}`; `return Engine.get_frames_per_second()` → `159.0`.
- Multi-line body iterating `tree.get_root().get_children()` → full autoload list.
- State mutation persists across evals (`server.set_meta(...)` → readback in a second eval).
- Syntax error → `{"ok":false,"error":"compile failed (err=43)"}`.
- `gdformat`/`gdlint` clean; pre-commit `full-tests --skip-visual` PASS.


kuruk-mm (approved):
great work!

PR #2210, opened 2026-06-03
fix: full session teardown on sign out (fresh lobby like an app restart)

kuruk-mm:
## Problem

After signing out, the previous session kept running. **All the heavy subsystems are added to `/root` as siblings of the active scene, not children of `explorer.tscn`** (`global.gd:398-411`), so `change_scene_to_file("lobby.tscn")` freed the explorer UI but left `scene_runner`, `comms`, `realm`, `player_identity`, `avatars` and `scene_fetcher` fully alive with the old user's state.

Concretely, `sign_out()`:
- never killed the running DCL scenes (`kill_all_scenes()` was only reached on a realm change),
- never disconnected comms (MainRoom/SceneRoom/LiveKit stayed open),
- never cleared the Rust player identity (wallet + ephemeral keys persisted in memory; only the on-disk `session_account` dict was cleared).

On top of that, the **Profile** and **Backpack** logout buttons bypassed `sign_out()` and **quit the app** via `comms.disconnect(true)` → `player_identity.logout` → `explorer._on_player_logout()` → `get_tree().quit()` (the "It's crashing" `TODO`).

### The non-obvious trap
Even calling `kill_all_scenes()` right before the swap was **not enough**: it only *marks* scenes `ToKill`. The reaping loop in `scene_runner_update` sits behind two early-returns — `if self.pause` and `if !player_avatar_node.is_instance_valid()` — and **both trip the moment the lobby loads** (`lobby._ready` pauses the runner; the freed explorer invalidates the avatar). So killed scenes would have lingered forever with live V8/Deno threads.

## Fix

**Rust — `scene_manager.rs`:** Extracted the kill state machine into `reap_dying_scenes()` and call it at the **top** of `scene_runner_update`, *before* the `pause`/avatar early-returns, so killed scenes tear down in the background even while the lobby pauses the runner. Shared the node-freeing tail into `finalize_scene_removal()`.

**`global.gd`:** Rewrote `sign_out()` into the single canonical, re-entrancy-guarded teardown:
1. `explorer.prepare_for_logout()` (disconnect persistent-emitter signals + stop timers while still in tree)
2. stop notifications polling + social streams + clear blacklist
3. `scene_runner.kill_all_scenes()`
4. `realm.async_clear_realm()` + `scene_fetcher.reset_for_logout()`
5. `comms.disconnect(true)` (closes comms **and** clears the Rust wallet/ephemeral identity)
6. `scene_runner.recreate_base_ui()`
7. clear `session_account` + save, portrait orientation
8. **deferred** `change_scene_to_file` to a fresh `lobby.tscn`

**`explorer.gd`:** `_on_player_logout()` now funnels into `Global.sign_out()` instead of quitting; added `prepare_for_logout()` / `_disconnect_persistent_signals()` to sever connections to persistent emitters (including the `profile_changed → avatars.update_primary_player_profile` Global→Global connection that leaked one callback per login).

**`scene_fetcher.gd`:** new `reset_for_logout()` clears `loaded_scenes` + island/wall/floor visuals and resets position/version state.

**`profile_settings.gd` / `backpack.gd`:** logout buttons now call `Global.sign_out()` (no more app-quit). **`lobby.gd`:** clears the re-entrancy guard on entry.

## Deliberate scope
- Trimmed the explorer signal-disconnect set to what genuinely needs manual handling — Godot auto-severs receiver-side connections on free, and the file has a hard 1800-line lint cap. The principle is documented in the code.
- Does **not** clear the disk content cache or other settings — this is a sign-out, not a factory reset.

## Out of scope (follow-up)
- The LiveKit adapter's `_clean()` is a no-op (`livekit.rs:134`), so its worker thread isn't explicitly joined on disconnect. Separate, deeper fix (the obvious snippet doesn't compile against current field types).

## Validation
- `cargo build` ✅ and `cargo clippy` ✅ — no warnings.
- `gdformat` ✅ + `gdlint` ✅ on all changed files.
- `check-gdscript` validates all 6 changed GDScript files **✓ OK**. (The command's overall exit-1 is from **pre-existing** parse errors in generated `android/build/src/instrumented/assets/` artifacts, unrelated to this change.)

## How to test
1. Sign in, walk into a populated scene.
2. Sign out (Settings, Profile, or Backpack logout).
3. You should land on a clean `lobby.tscn` and **not** be auto-logged back in.
4. Query the debug WS `scenes` tree (`ws://127.0.0.1:9230`) — it should report **0 scenes**.

### DIFF
--- a/godot/src/global.gd
+++ b/godot/src/global.gd
@@ -727,10 +727,14 @@ func sign_out() -> void:
 	#    second logout() no-ops once the address is already cleared.
 	comms.disconnect(true)
 
-	# 6. The Explorer reparents scene_runner.base_ui into its tree; freeing the
-	#    Explorer leaves a dangling pointer. Recreate it now so nothing touches a
-	#    freed node before the next Explorer load.
+	# 6. Reset the scene_runner's Godot-node handles before the Explorer (and its
+	#    Player) is freed. The #[var] player-node getters PANIC on a freed instance
+	#    and survivors poll them after sign-out (e.g. the analytics first-move poll
+	#    reads player_body_node); base_ui is reparented into the Explorer and would
+	#    dangle too. Reset them to safe placeholders / a fresh instance now.
+	scene_runner.clear_player_nodes()
 	scene_runner.recreate_base_ui()
+	player_camera_node = null
 
 	# 7. Erase the persisted session (ephemeral keys / wallet bytes) from disk.
 	#    Everything else (graphics, last realm, content cache) is intentionally
--- a/godot/src/global.gd
+++ b/godot/src/global.gd
@@ -139,6 +139,12 @@ var _safe_margin_debug_overlay: SafeMarginDebugOverlay = null
 # Startup instrumentation timestamp (set once at load time)
 var _startup_time: int = Time.get_ticks_msec()
 
+# Guards sign_out() against re-entrancy. comms.disconnect(true) emits the
+# player_identity.logout signal deferred, which can route back into sign_out() a
+# frame later; without this the whole teardown + scene swap would run twice.
+# Cleared in lobby._ready once we're back on a clean screen.
+var _signing_out: bool = false
+
 
 func is_xr() -> bool:
 	return OS.has_feature("xr") or get_viewport().use_xr
@@ -675,7 +681,25 @@ func get_explorer() -> Explorer:
 	return null
 
 
+## Single canonical sign-out / logout entry point. Tears down the live session —
+## explorer signals & timers, social streams, every running DCL scene, comms,
+## realm, scene-fetcher state and the Rust player identity — then returns to a
+## fresh lobby so the next login starts as if the app had restarted. Every UI
+## logout button and the disconnect handler funnel through here.
 func sign_out() -> void:
+	if _signing_out:
+		return
+	_signing_out = true
+
+	# 1. Tear down the live Explorer first, while it is still in the tree, so its
+	#    autoload-signal callbacks and retry timers are severed before the steps
+	#    below re-emit any of those signals (orientation, realm clear, scene kill)
+	#    or free the Explorer node.
+	var explorer := get_explorer()
+	if explorer != null:
+		explorer.prepare_for_logout()
+
+	# 2. Stop session pollers and tear down the social gRPC streams.
 	NotificationsManager.stop_polling()
 	social_service.unsubscribe_from_updates()
 	social_service.unsubscribe_from_connectivity_updates()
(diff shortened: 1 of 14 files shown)


kuruk-mm:
### Follow-up: fixed two post-sign-out crashes found in testing (commit 4e49119)

Once `sign_out()` actually frees the Explorer + Player, two teardown-survivors crashed:

1. **Analytics first-move poll panic** — `analytics_controller`'s `FirstMovePoller` Timer lives under the `Global` autoload, so it survives sign-out and reads `Global.scene_runner.player_body_node`. That handle is a `#[var] Gd<Node3D>` whose auto-getter calls `to_godot()` → **panics on the freed node** (`access to instance ... after it has been freed`), spamming every poll tick. Fix: `SceneManager.clear_player_nodes()` swaps in valid orphan placeholders on sign-out (with `free_if_orphan_placeholder()` so they don't leak across login cycles) — makes *every* player-node getter safe, not just analytics. Also null `Global.player_camera_node`.

2. **Chat coroutine null `get_tree()`** (`chat.gd:227`) — `set_orientation_portrait()` → `Global.orientation_changed` → `chat_panel` relayout starts a 3-frame `await get_tree().process_frame` coroutine; the deferred scene swap frees the chat, so on resume `get_tree()` is null. Fix: a guarded `_async_await_frames(count)` helper that bails when the node leaves the tree mid-await; all four chat `process_frame` awaits route through it.

Validation: `cargo build` + `cargo clippy` clean; `gdformat`/`gdlint` clean; both changed scripts validate ✓ in Godot's parser.

kuruk-mm:
### Follow-up 2: fixed `reap_dying_scenes` panic spam (commit 44887ed)

Killing scenes on sign-out exposed a double-free → permanent panic loop (`Option::unwrap() on a None value` at `reap_dying_scenes`, every frame).

**Chain:** each scene's `root_node_ui` is a child of `SceneManager.base_ui`, which is reparented into the Explorer. `sign_out()` recreated `base_ui` (and the deferred `change_scene_to_file` frees the Explorer), so `base_ui` and the dying scenes' UI roots were freed **before** `reap_dying_scenes()` finalized them. `finalize_scene_removal()` → `free_root_nodes()` → `queue_free()` on the already-freed UI root **panicked between `scenes.remove()` and `dying_scene_ids.retain()`**, leaving the id in `dying_scene_ids` but not in `scenes` → `unwrap()` re-panicked on it every subsequent frame.

**Fixes:**
- `free_root_nodes()`: `is_instance_valid()` guard before each `queue_free()` (and removed the logging that also touched freed nodes) — no more double-free panic.
- `finalize_scene_removal()`: prune bookkeeping lists **first**, then `let-else` remove (no `unwrap`) — an id can never get stuck even if a later step fails.
- `reap_dying_scenes()`: `let-else` skip + prune ids whose scene is already gone — **self-heals** any pre-existing stuck state instead of spamming.
- `sign_out()`: stopped recreating `base_ui` (redundant — `explorer._ready()` recreates it on next login; freeing it early is what orphaned the UI roots). `base_ui` now lives until the scenes are reaped.

`cargo build` / `clippy` clean; `gdformat`/`gdlint` clean.

kuruk-mm:
### Follow-up 3: reverted the player-node placeholder (it was wrong); fixed spam + loading-screen tween (commit f87dfa5)

The previous `clear_player_nodes()` placeholder approach was a mistake. An orphan `Node3D` is **"valid but not in the tree"**, which:
- spammed `get_global_transform: Condition "!is_inside_tree()" is true` every physics frame, and
- **defeated SceneManager's own `is_instance_valid()` guards** (`physics_process:2245`, `scene_runner_update`) — the original design relies on the freed Player reporting `is_instance_valid()==false` after sign-out so those guards return early.

**Fixes:**
- **Reverted** `clear_player_nodes()` / `free_if_orphan_placeholder()` / the `set_player_node()` changes. The player handles go back to reporting `is_instance_valid()==false` once the Explorer is freed, so the existing guards skip them cleanly — no spam, no transform reads on a not-in-tree node.
- Fixed the original analytics panic at its **only surviving reader** instead: `Global.sign_out()` now calls `AnalyticsController.cancel_first_move_poll()`, stopping the first-move poll Timer before the Player is freed (it restarts on next login). No more `player_body_node` getter panic.
- `loading_screen.gd`: guard `set_progress()` and `set_bg_shader_color()` with `is_inside_tree()` before `get_tree().create_tween()` — sign-out can free the loading screen between frames while a progress update is in flight, making `get_tree()` null (the `loading_screen.gd:144` crash).

The reap/`free_root_nodes`/double-free robustness from the prior commit stays (unrelated and still needed). `cargo build`/`clippy` clean; `gdformat`/`gdlint` clean; changed scripts validate ✓ in Godot.

regenesis-claw (approved):
Thorough, well-structured teardown. The PR description alone is worth the review — it explains the root cause chain (autoload siblings surviving `change_scene_to_file`, kill state machine stuck behind `pause`/avatar early-returns, the `TODO: It's crashing` quit workaround) and the fix is cleanly layered.

Followed `REVIEW.md`. Pre-flight: branch CI all green ✅, iOS build present ✅, no submodule drift ✅.

## Key strengths

- **Re-entrancy guards** (`_signing_out` / `_returning_to_discover`) with clear ownership (guard set in `sign_out()`, cleared in `lobby._ready()`) — the deferred scene swap + deferred logout signal combo would absolutely re-enter without this.
- **`reap_dying_scenes()` hoisted above the `pause`/avatar early-returns** — the core fix for lingering V8 threads. The `let-else` + `stale_scene_ids` self-healing pattern in the Rust side is particularly nice: if anything ends up in an inconsistent state, it recovers instead of panic-looping.
- **`finalize_scene_removal()` prunes bookkeeping first** — ensures an id can never get stuck even if a later step fails. Defensive in exactly the right place.
- **`free_root_nodes()` validity guards** — prevents the double-free cascade described in the follow-up commits.
- **`_async_await_frames()` in chat.gd** — clean, reusable pattern for coroutines that can outlive their node. All four call sites migrated consistently.
- **`_disconnect_persistent_signals()`** correctly identifies the Global→Global connection (`profile_changed → update_primary_player_profile`) that Godot's auto-sever wouldn't catch. The comment explaining *why* only these signals need manual handling is exactly the kind of "why not what" this repo values.
- **`return_to_discover()`** nicely mirrors `sign_out()` with the `disconnect(false)` / keep-social-streams difference clearly called out.

## Minor observations (non-blocking)

1. **Step numbering gap in `sign_out()` comments** — steps go 1→2→3→4→5→6 → (unnumbered orientation) → 8. Missing a `# 7.` or the orientation block should be 7. Cosmetic only.

2. **`_signing_out` / `_returning_to_discover` are `_`-prefixed (private convention) but accessed externally** from `lobby.gd` and `menu.gd` (`Global._signing_out = false`). The `_` prefix in GDScript conventionally signals "don't touch from outside". Consider dropping the underscore since these are intentionally part of the cross-scene contract, or adding a `clear_sign_out_guard()` setter. Not blocking — the current approach works and the comments explain the intent.

Solid work. The iterative follow-up commits (placeholder revert, analytics fix, double-free guard) show careful in-context debugging.

Requested by Mateo via Slack

leanmendoza (approved):

PR #2212, opened 2026-06-04
fix: cinematic camera stuck after teleporting

### LINKED ISSUE #2160: [Bug]: Leaving a scene while in cinematic camera breaks the camera
Teleporting to a new scene while the camera is in cinematic mode. Breaks the camera for the new scene (it remains in cinematic mode)

To test:

- go to Genesis Plaza, go to the Theater
- look at the screen and enter Focus mode
- Open the Discover and jump to any other scene

The camera will be broken. 100% of the time.

Image

EibrielInv:
## Problem

  Entering a scene's cinematic (virtual) camera, then teleporting (or walking) to another
  scene, left the camera frozen — on mobile you could still move but not rotate the camera.

  Closes #2160

  ## Root cause

  In `dcl_global_camera_controller.gd`, once the virtual camera reaches its target it is
  reparented onto a scene entity. A scene change frees that entity (and the camera under it);
  `_process`'s re-init branch then recreates the camera with `last_camera_reached = true` /
  `last_virtual_camera_entity_node = null`, so every subsequent frame hits the
  `elif last_camera_reached: return` early-out. The only site that resets
  `raycast_use_cursor_position` and the camera mode never runs, so both stay latched
  (`raycast_use_cursor_position = true`, `current_camera_mode = CINEMATIC`).
  `mobile_camera_input.gd` then routes every touch to the cinematic cursor handler instead of
  camera rotation. Confirmed live on-device via the debug WS (port 9230).

  ## Fix

  A scene should always start non-cinematic. The controller now listens to
  `SceneManager::on_change_scene_id` and, **only when actually latched in cinematic state**,
  forces a clean return to the player camera (clears the cursor flag, restores the real
  camera mode, makes the player camera current, resets tracking vars). This mirrors the
  existing pattern where Rust already resets input modifiers / skybox time on the same
  signal (`scene_manager.rs:1768-1772`). GDScript-only — no Rust change. The shared
  visibility/cursor cleanup is extracted into a helper reused by both the transition-complete
  and reset paths.

  ## Verification

  Reproduced the teleport-out-of-cinematic flow on Android and confirmed over the debug WS:

  | Signal | Before | After |
  |---|---|---|
  | `raycast_use_cursor_position` | `true` | `false` |
  | `current_camera_mode` | `2` (CINEMATIC) | `1` (FIRST_PERSON) |
  | `Player/Mount/Camera3D.current` | `true` | `true` |
  | controller `transition_time_counter` | `0.89` (stale) | `0.0` |

  Camera rotation works again. Guard prevents any reset (no fade sound / re-tween) on normal
  non-cinematic parcel crossings.

### DIFF
--- a/godot/src/decentraland_components/dcl_global_camera_controller.gd
+++ b/godot/src/decentraland_components/dcl_global_camera_controller.gd
@@ -18,6 +18,12 @@ func _ready():
 	global_virtual_camera.clear_current()
 	global_virtual_camera.cull_mask = 0x7fff
 
+	# A scene always starts non-cinematic. Reset the cinematic/virtual-camera state
+	# whenever the current parcel scene changes so the latch can't survive a teleport
+	# (mirrors how Rust resets input modifiers/skybox time on the same event).
+	if is_instance_valid(Global.scene_runner):
+		Global.scene_runner.on_change_scene_id.connect(_on_parcel_scene_changed)
+
 
 # in the process fn, this node has the responsibility of transitioning the camera between player one and virtual one
 #  virtuals can change between scenes or in the same scene, computed as desired_target
@@ -153,12 +159,7 @@ func _process(delta: float) -> void:
 			Global.scene_runner.raycast_use_cursor_position = false
 
 			# When coming back from virtual camera, we always show outline system and set the first/third person camera behaviour
-			var explorer = Global.get_explorer()
-			if is_instance_valid(explorer):
-				explorer.reset_cursor_position()
-				explorer.player.outline_system.show()
-				if explorer.player.camera.get_camera_mode() == Global.CameraMode.FIRST_PERSON:
-					explorer.player.avatar.set_hidden(true)
+			_restore_player_camera_visibility()
 	else:
 		# Transitioning to virtual camera target
 		var target_transform = desired_target.global_transform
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
Clean, well-scoped fix. The root-cause analysis is excellent.

**Pre-flight:**
- ✅ CI — all checks green (static checks, clippy, Linux, Android, iOS, Docker)
- ✅ iOS build present and green (TestFlight artifact posted)
- ✅ Single GDScript file — no platform-specific branches, no Rust/proto changes, no submodule drift

**Review (following REVIEW.md):**

The approach is sound: hook into the existing `on_change_scene_id` signal (same pattern used by `camera_mode_area.gd`, `scene_music_changer.gd`, `player.gd`, `explorer.gd`, etc.) and only reset when actually latched in cinematic state. This avoids spurious resets on normal parcel crossings.

Specific things I checked:
- **Latch guard logic** — the `or` in the latched check covers both symptoms (camera mode stuck at CINEMATIC *or* cursor flag stuck true). After `_force_return_to_player_camera` runs, both flags are cleared, so rapid scene-id emissions won't re-trigger.
- **Reparent safety** — `global_virtual_camera` is reparented back to `self` only if it's somewhere else; the `is_instance_valid` check handles the freed-camera case correctly (re-init in `_process` will catch it on the next frame if the camera was already freed).
- **State reset completeness** — `last_virtual_camera_entity_node = null`, `last_camera_reached = true`, `transition_time_counter = 0.0`, camera mode, cursor flag, player camera made current, visibility restored. All the variables that could latch are explicitly cleared.
- **Extracted helper** — `_restore_player_camera_visibility()` deduplicates the explorer/outline/avatar-hide logic between the normal transition-complete path and the forced reset path. No behavior change on the existing path.
- **No debug prints, no `.claude/` files, no `TODO`s.**

Good fix. 👍

Requested by Mateo via Slack

kuruk-mm (approved):
looking great!

PR #2216, opened 2026-06-05
fix: onLeaveScene not triggered when players disconnect

EibrielInv:
## What

  `@dcl/sdk/players` `onLeaveScene` (and `onEnterScene`) never fired for remote players. The renderer signalled a player leaving by **killing the avatar entity**, but entity deaths never reach the scene's JS
  (they're stripped from the dirty set, and #1444 deliberately stops them reaching the SDK). The SDK detects presence purely from components — `onEnterScene` when an entity gains `PlayerIdentityData` +
  `AvatarBase`, `onLeaveScene` when `AvatarBase` is removed — so the entity-kill was invisible to it.

  ## Fix

  Communicate departure as **component deletions** (set the player components to `None`), which propagate to the scene exactly like the component PUTs the enter path already uses.

  - **`avatar_data.rs`** — on avatar removal (disconnect), clear the player components instead of `entities.kill`. Entity-slot recycling still happens via `AvatarScene`'s own CRDT in `remove_avatar`.
  - **`avatar_scene.rs` (`on_avatar_changed_scene`)** — on walk-out, clear the components for the scene being left; on walk-in, repopulate them from the cached avatar data. This makes scene-boundary crossings 
  fire `onLeaveScene`/`onEnterScene` too.

  Grounded in the proto contract (`PlayerIdentityData` is "written by the engine using the communications transports' data") and the SDK's actual detection logic in `@dcl/sdk/src/players`.

  ## Testing

  Requires two players in the same realm (a second desktop/Bevy/Unity client, or two instances).

  1. Build & run: `cargo run -- run`.
  2. Go to parcel **`-31,19`** and select the **"Avatars"** test block. It shows live counters: `Connected Avatars`, `onEnterScene`, `onLeaveScene`, and `PlayerIdentityData entities`.
  3. Bring a **second player into the scene**. Confirm every counter reads the full count (e.g. `(2)`, `onEnterScene: 2`).
  4. **Walk the second player out** of the parcel (stays connected): `onLeaveScene` increments and `PlayerIdentityData entities` decrements; "Recent leaves" records the address.
  5. **Walk back in**: `onEnterScene` increments and the counts restore.
  6. **Fully disconnect** the second player: `onLeaveScene` increments, `PlayerIdentityData entities` decrements, and the avatar is removed — with no double-firing.

  > Note: restart the client between full test runs — the SDK's player-tracking map can retain stale entries from a prior session, which masks subsequent events.

  Closes #1818

### DIFF
--- a/lib/src/avatars/avatar_scene.rs
+++ b/lib/src/avatars/avatar_scene.rs
@@ -1071,40 +1071,68 @@ impl AvatarScene {
         // maybe it's better to cache the last parcel here instead of using prev_scene_id
         // the state of to what parcel the avatar belongs is stored in the avatar_scene
 
+        // Cached component values for this avatar, used to (re)populate the scene
+        // being entered. `@dcl/sdk/players` derives onEnterScene from an entity having
+        // both PlayerIdentityData and AvatarBase, and onLeaveScene from AvatarBase
+        // being removed — so scene membership must add/remove these components.
+        let avatar_base = SceneCrdtStateProtoComponents::get_avatar_base(&self.crdt_state)
+            .get(&avatar_entity_id)
+            .and_then(|v| v.value.clone());
+        let player_identity_data =
+            SceneCrdtStateProtoComponents::get_player_identity_data(&self.crdt_state)
+                .get(&avatar_entity_id)
+                .and_then(|v| v.value.clone());
+        let avatar_equipped_data =
+            SceneCrdtStateProtoComponents::get_avatar_equipped_data(&self.crdt_state)
+                .get(&avatar_entity_id)
+                .and_then(|v| v.value.clone());
+
         let mut scene_runner = DclGlobal::singleton().bind().scene_runner.clone();
         let mut scene_runner = scene_runner.bind_mut();
+
+        // Leaving the previous scene: clear the avatar's components there so the SDK's
+        // onLeaveScene fires and the player drops from getEntitiesWith(PlayerIdentityData).
+        // Routed through `deleted_entities`, which sets these components to None — the
+        // departure cannot be communicated as an entity death (the renderer→scene path
+        // never carries entity deaths). See `update_avatar_scene_updates`.
         if let Some(prev_scene) = scene_runner.get_scene_mut(&prev_scene_id) {
             prev_scene
                 .avatar_scene_updates
-                .transform
-                .insert(avatar_entity_id, None);
-            prev_scene
-                .avatar_scene_updates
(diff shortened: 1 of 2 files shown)


EibrielInv:
### Why a client restart is needed between full test runs

This is a quirk of the **SDK's** player tracking, not of this fix — but it's worth knowing so test results aren't misread.

`@dcl/sdk/players` keeps an in-memory `playerEntities` map that lives for the **whole lifetime of the scene instance** (it only resets when the scene reloads, i.e. on client restart). The relevant logic (from `@dcl/sdk/src/players`):

```ts
engine.addSystem(() => {
  const players = Array.from(engine.getEntitiesWith(PlayerIdentityData, AvatarBase))
  if (players.length === playerEntities.size) return          // (1) early-return
  for (const [entity, identity] of players) {
    if (!playerEntities.has(entity)) {
      playerEntities.set(entity, identity.address)
      onEnterSceneCb.forEach(...)                              // onEnterScene
      AvatarBase.onChange(entity, (value) => {                 // (2) registered lazily, per entity
        if (!value && playerEntities.get(entity)) {
          onLeaveSceneCb.forEach(...); playerEntities.delete(entity)
        }
      })
    }
  }
})
```

Two details combine to create the trap:

1. **`if (players.length === playerEntities.size) return`** — a size-based optimization. If the live count equals the tracked-map count, the loop is skipped entirely, even if the *membership* differs.
2. **The leave/cleanup handler (`AvatarBase.onChange`) is only registered when an entity first passes through the enter loop.** An entity that's never registered can never fire `onLeaveScene` or be removed from the map.

So if a *leave* is ever missed, the map keeps a **stale entry**. That's exactly the pre-fix bug (the entity was killed, the SDK never saw `AvatarBase` go null, so the player stuck in the map), and it also happened transiently while iterating on broken builds. When a player then reconnects they get a **new entity** (same entity number, bumped version), and the size check can now coincidentally match (one stale + the new one vs. the previous total) → the system early-returns → the new entity is **never registered** → all of its subsequent enter/leave events are silently dropped for the rest of that scene's life.

A client restart reloads the scene and starts `playerEntities` empty, clearing the poison. With this PR, leaves are reliable (every departure clears `AvatarBase` → `onChange` fires → the entry is removed), so the map no longer accumulates stale entries in normal operation. The restart note in the testing steps is purely to avoid carrying poisoned state from a previous (pre-fix) run into a verification run.

leanmendoza (approved):
code looks clear, tested and working as expected :)

PR #2223, opened 2026-06-08
fix: clear notifications on sign-out + repair attended-events sync

### LINKED ISSUE #2104: [Bug] Notification History of old profile persists after logout and new account creation
## Description
Notification history is not cleared when logging out and creating a new account, causing old notification state to persist incorrectly.

## Platform
- Mobile

## Steps to Reproduce
1. Log in with an account that has notification history
2. Log out
3. Create a new account
4. Check notification history

## Expected Behavior
Notification history should be wiped clean when logging out and switching accounts.

## Actual Behavior
Old notification state from previous account remains visible in new account.

Image

---
Requested by Rodri via Slack

EibrielInv:
## Summary

  Three related notification fixes, surfaced while investigating #2104:

  1. **Clear in-memory notification history on sign-out (closes #2104).**
     `NotificationsManager` is an autoload that survives the lobby scene swap, and
     `sign_out()` only stopped the poll timer — the cached `_notifications` array (and the
     bell badge count) leaked into the next account. New `clear_notification_history()`
     wipes the in-memory history/queue and refreshes the panel + bell via
     `notifications_updated`.

  2. **Drop the previous account's local event reminders on sign-out.**
     New `clear_event_local_notifications()` cancels the OS notifications and deletes the
     `event_`-prefixed rows from the local SQLite queue, so a signed-out/guest session can't
     fire a previous account's reminders. The per-install `day1_welcome` is preserved.

  3. **Repair the attended-events re-sync (prerequisite for #2 not regressing).**
     `async_sync_attended_events()` was fetching `mobile_events_api()` (mobile-bff)
     `?only_attendee=true`, which ignores the param and returns `attending=false` for every
     event — so the sync never scheduled anything. Re-pointed it at `events_api()`
     (events.decentraland.org), the canonical service that reports a correct per-user
     `attending` flag (the same source the REMIND ME button reads/writes). This was a
     regression from #1424, which moved the sync onto mobile-bff in the same PR that
     documented mobile-bff "doesn't return per-user attending state."

  ## Verified on-device (Android)
  
  Via debug-WS (in-memory) and `adb run-as` sqlite snapshots of `dcl_notifications.db`:

  - Sign-out clears in-memory history (33 → 0) and `event_` rows (1 → 0), `day1_welcome` preserved.
  - After fix, logout → login → enter world restores attended/future reminders automatically
    (4 attending detected; 3 scheduled, 1 correctly skipped as already-started) — no manual re-set.

  ## Files

  - `godot/src/notifications_manager.gd` — `clear_notification_history()`,
    `clear_event_local_notifications()`, `async_sync_attended_events()` URL fix.
  - `godot/src/global.gd` — both clears wired into `sign_out()`.

  Closes #2104

### DIFF
--- a/godot/src/global.gd
+++ b/godot/src/global.gd
@@ -711,6 +711,15 @@ func sign_out() -> void:
 
 	# 2. Stop session pollers and tear down the social gRPC streams.
 	NotificationsManager.stop_polling()
+	# Wipe the previous account's in-memory notification history so it can't leak
+	# into the next session's panel/bell badge (issue #2104).
+	NotificationsManager.clear_notification_history()
+	# Drop the previous account's event reminders so they can't fire on the
+	# device after sign-out. Only "event_" entries (per-account attended-event
+	# reminders) are cleared; the per-install day1 welcome is preserved. The
+	# sync-on-next-login REMOVE pass only runs for an authenticated account, so
+	# without this a signed-out/guest session keeps the old reminders scheduled.
+	NotificationsManager.clear_event_local_notifications()
 	# The analytics first-move poll (a Timer under this autoload) reads
 	# scene_runner.player_body_node; left running it would poll the freed Player
 	# after the swap, panicking the #[var] getter. It restarts on the next login.
--- a/godot/src/notifications_manager.gd
+++ b/godot/src/notifications_manager.gd
@@ -147,6 +147,18 @@ func stop_polling() -> void:
 		_poll_timer.stop()
 
 
+## Clear the in-memory server notification history and pending toast queue.
+## Used on sign-out so a previous account's notifications don't leak into the
+## next session (issue #2104). NotificationsManager is an autoload that survives
+## the lobby scene swap, so without this the cached list persists across accounts.
+## Emitting notifications_updated refreshes the open panel and the bell badge.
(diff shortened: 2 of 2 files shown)


kuruk-mm (approved):
great!

regenesis-claw (approved):
Clean, well-scoped fix. Three things done right:

1. **Sign-out teardown ordering** — `clear_notification_history()` and `clear_event_local_notifications()` are called right after `stop_polling()`, before the analytics timer teardown. Correct sequencing within the `sign_out()` cascade.

2. **Selective local notification cleanup** — only `event_`-prefixed rows are cleared, preserving the per-install `day1_welcome`. The loop through `get_queued_local_notifications()` + `cancel_queued_local_notification()` handles both OS cancellation and DB deletion, so no stale reminders can fire post-logout.

3. **events_api() URL fix** — switching from `mobile_events_api() + /?only_attendee=true` (which silently ignores the param) to the canonical `events_api()` that actually reports per-user `attending` state on signed requests. Good root-cause analysis in the PR description tracing this back to #1424.

All CI green (Linux, Android, iOS, Docker, static checks, clippy). iOS build present and passing. LGTM.

Requested by Mateo via Slack

Issue #2215, opened 2026-06-05
[Bug] Avatar speech bubble freezes after first message on Android


regenesis-claw:
**Reported by:** Skatlionrock (Discord)

**Bug Description:**
Avatar speech bubble breaks/freezes after sending the first message in chat and stays stuck showing the same text until the app is restarted. The bug reoccurs every time a new message is sent after restarting.

**Platform:** Android  
**Device:** Motorola 60 Edge Fusion  
**OS Version:** Android 16  
**App Version:** v0.68.0-cffbefe-prod  
**Location:** Genesis Plaza

**Steps to Reproduce:**
1. Open the app and navigate to Genesis Plaza
2. Send a message in chat
3. Observe the speech bubble above your own avatar

**Expected Behavior:**
Speech bubble should update with new messages or disappear after a short time.

**Actual Behavior:**
Speech bubble freezes/breaks after the first message and displays the same text indefinitely. Only restarting the app clears it, but the bug reoccurs on the next message sent.

**Screenshots:**
- Screenshot 1: Chat showing messages from user
- Screenshot 2: Frozen speech bubble persisting above avatar
- Screenshot 3: Speech bubble still stuck
- Screenshot 4: Different view showing persistent broken bubble

**Workaround:** Restart the app (temporary - bug reoccurs)

**Mobile Deeplink:** https://mobile.dclexplorer.com/open?position=0,0

---
Requested by ToxicWaifu via Slack

manuelmaceira:
Fixed by #2224 (merged in e1c68b03).

Avatar nameplates no longer render through per-avatar SubViewport textures — they're now drawn in a shared 2D screen-space CanvasLayer, projecting each avatar's head position to screen. This structurally eliminates the uninitialized-VRAM garbage seen on the reporter's device (a render-target that some mobile drivers don't guarantee is cleared), and also removes the speech-bubble freeze / resize-race from the same subsystem.

The 2D nameplates keep distance fade, frustum culling, depth sorting, and raycast-based occlusion against world geometry and avatar bodies (including the local player's). XR keeps the previous viewport+quad path.

PR #2224, opened 2026-06-08
fix(avatar): render nameplates in a 2D layer, not per-avatar viewports (#2215)

manuelmaceira:
## Problem (#2215)

Avatar nametags showed **uninitialized-VRAM garbage** on mobile (noise / partial text / color artifacts). Each avatar rendered its `NicknameUI` into a **per-avatar `NicknameViewport` (SubViewport)**, sampled live by a `Sprite3D` `ViewportTexture` and bumped with `UPDATE_ONCE`. On mobile tile GPUs that single-shot render can be dropped/not-cleared, freezing the texture on uninitialized memory. This is the **2nd** bug from this architecture (see #2039, the "garbled pixels" / freeze fix).

Repro note: the bug is driver-specific — the A54 (Mali) clears its render targets, so it does **not** reproduce there; it manifests on the reporter's device (Motorola Edge Fusion). The fix is therefore **structural**, not a tweak to the viewport heuristics.

## Fix: render nameplates in a shared 2D screen-space layer

Stop using viewport textures entirely. Each avatar's `NicknameUI` (already a pure `Control` tree) is **reparented out of its SubViewport into a shared `CanvasLayer`** (`NameplateLayer`, `layer = -1` → above the 3D world, below the HUD) and **projected onto the head anchor every frame**. The per-avatar `SubViewport` + `ViewportTexture` are dropped; the `Sprite3D` is kept (invisible) only as the head anchor for the SDK `NAME_TAG` attach point.

This **eliminates the whole bug class** — no render targets → no uninitialized VRAM — and as a side effect also removes the chat-bubble freeze and the resize race that #2039 was patching around.

### Behaviour
- **Distance fade**: full < 10 m → faded out and hidden by 15 m; depth-sorted by distance (closer draws on top).
- **Smooth fade in/out**: off-frustum, behind-camera, far and occluded all drive the alpha toward 0 (and back), so tags fade instead of popping when they enter/leave the screen.
- **Depth occlusion**: a throttled raycast (every 6 physics frames, staggered per avatar, in `_physics_process`) hides tags behind solid world geometry (`CL_PHYSICS`) and remote avatar bodies (`CL_AVATAR`), excluding the avatar's own collider.
- **XR** keeps the legacy viewport path (screen-space 2D doesn't apply in VR).

## Files
- `godot/src/decentraland_components/avatar/nameplate_layer.gd` (new): owns the shared layer + per-frame projection/fade/occlusion runtime.
- `godot/src/decentraland_components/avatar/avatar.gd`: branch to the 2D path (non-XR), reparent on `_ready`, drive it from `_process` / `_physics_process`, free it on `_exit_tree`; `_request_nickname_redraw` becomes XR-only.

## Verification
- Runs clean on desktop (Forward Mobile / Vulkan) with `--spawn-avatars`: nameplates project, scale, fade, sort and occlude against geometry + remote bodies; chat bubbles animate natively.
- gdformat / gdlint / Godot script validation / cargo check / clippy / rust+integration tests all pass (pre-commit).
- On-device validation of the original VRAM artifact must happen on an affected device (not the A54).

## Known follow-up
- The **local player's own body** does not occlude others' tags yet: the local player frees its `click_area`, so its only collider is the `CharacterBody` on a broad layer (4) that, when added to the occlusion mask, over-occludes (catches every avatar's TriggerDetector + the camera-mode area). A targeted fix (re-layering the kept click_area, mindful of the ground raycast) is left as a follow-up.

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar.gd
+++ b/godot/src/decentraland_components/avatar/avatar.gd
@@ -163,6 +163,11 @@ var _free_bone_pool: Array[int] = []
 var _stale_bone_counter: int = 0
 
 var _lod_state: int = LODState.FULL
+# 2D screen-space nameplate (non-XR) vs legacy viewport quad (XR). Runtime lives in
+# NameplateLayer; these cache the hide-flag/FAR gate and the depth-occlusion result.
+var _use_2d_nameplate: bool = false
+var _nametag_gate_visible: bool = true
+var _nameplate_occluded: bool = false
 var _impostor_layer: int = -1
 var _lod_phase: int = 0
 var _mesh_lod_visibility_captured: bool = false
@@ -269,6 +274,9 @@ func _ready():
 	# Hide mic when the avatar is spawned
 	nickname_ui.mic_enabled = false
 	Global.on_chat_message.connect(on_chat_message)
+	_use_2d_nameplate = not Global.is_xr()
+	if _use_2d_nameplate:
+		NameplateLayer.attach(self)
 	_apply_nickname_visibility()
 
 	_lod_phase = int(self.unique_id) % AvatarImpostorConfig.DISTANCE_CHECK_PERIOD_FRAMES
@@ -282,6 +290,11 @@ func _ready():
 
 func _exit_tree() -> void:
 	AvatarLODCoordinator.unregister(self)
+
+	# The 2D nameplate lives in the shared layer, not under this avatar, so it
+	# won't be auto-freed — free it explicitly.
+	if _use_2d_nameplate:
+		NameplateLayer.detach(self)
 
 	# For local player and remote avatars, trigger detection is setup later via setup_trigger_detection()
 	# For AvatarShapes (scene NPCs), remove_trigger_detection() is called from avatar_shape.rs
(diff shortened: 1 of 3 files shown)


leanmendoza (approved):
Code looks good

regenesis-claw (approved):
## Review — fix(avatar): render nameplates in a 2D layer (#2215)

Clean architectural fix that eliminates the whole class of uninitialized-VRAM bugs by removing per-avatar SubViewport textures in favor of a shared screen-space CanvasLayer. Well-structured, well-commented, follows established repo patterns.

### Pre-flight
- ✅ Branch is up to date with main (0 behind)
- ✅ No iOS-sensitive changes (pure GDScript, no platform branches)
- ✅ No submodule drift
- ✅ Static checks + Clippy pass; Android + Linux builds pending

### What I checked
- **Lifecycle**: `attach()` properly reparents NicknameUI, frees the viewport, nulls the texture, hides the quad (preserved for NAME_TAG SDK attach point). `detach()` in `_exit_tree` correctly queue_frees the reparented UI that won't auto-free with the avatar. ✅
- **Projection + fade**: `update()` in `_process` handles behind-camera, off-screen, distance fade, depth-sorting by z_index — all clean. `move_toward` gives smooth fade instead of popping. ✅
- **Occlusion**: Correctly placed in `_physics_process` (direct_space_state crashes from idle frames). Frame-staggered via `unique_id % OCCLUSION_PERIOD`, own colliders + camera-mode area excluded. ✅
- **Gate logic**: `_apply_nickname_visibility` properly short-circuits for 2D mode, sets `_nametag_gate_visible` flag, and hides immediately when gated off. Show path is driven by `update()` alpha convergence — correct. ✅
- **XR fallback**: Legacy viewport path preserved for XR via `_use_2d_nameplate = not Global.is_xr()`. All modified methods (`_request_nickname_redraw`, `_apply_nickname_visibility`, `_process`, `_physics_process`) correctly branch or early-return. ✅
- **Static var staleness**: Both `_root` and `_camera_area` guarded by `is_instance_valid` — handles scene-change / sign-out teardown. ✅
- **Mouse filter**: `MOUSE_FILTER_IGNORE` on root Control and reparented NicknameUI — won't block scene or HUD input. ✅
- **No debug leftovers**: No `print()` / `prints()` / `.claude/` files. ✅

### Known follow-up (acknowledged)
Local player body occlusion noted in PR description — layer 4 CharacterBody over-occludes when added to the mask. Documented and scoped out intentionally.

LGTM — solid structural fix with clean separation between the 2D runtime and the legacy XR path.

PR #2228, opened 2026-06-09
feat: bake textures at multiple resolutions for runtime quality selection

### LINKED ISSUE #2076: Asset pipeline: bake textures at multiple resolutions for runtime quality selection

manuelmaceira:
Closes #2076

## Problem

The offline asset pipeline baked a single `.res` per texture at one fixed quality (`ctx.texture_quality`, hardcoded Medium). At runtime the optimized fast-path loaded that `.res` blindly, so:

- the user's `TextureQuality` setting was silently ignored (High request → got Medium, no warning);
- `fetch_texture_by_hash_with_quality` / `fetch_texture_by_url_with_quality` could **never** use the fast-path and always paid full network + decode (e.g. every UI image asking for Source).

### Bake side (`lib/src/asset_server/`)
- `process_texture` now bakes one variant per quality tier — **Low (256) / Medium (512) / High (1024) / Source** — each resized from a fresh `image.duplicate()` of the pristine decode to avoid lossy resize cascades. Source is capped at `MAX_TEXTURE_DIMENSION` (4096) and collapsed into the top variant when the original already fits in High (identical pixels).
- Variants are saved as `content/{hash}_q{N}.res`. New manifest field `textureQualities: {hash: [ordinals]}` records which variants exist; `hashSizeMap` for textures is the sum of all variant sizes. Packer + job plumbing carry N files per texture.

### Runtime side (`lib/src/content/content_provider.rs`)
- `select_texture_variant` picks the **smallest baked variant ≥ the requested quality**; if only lower variants exist it falls back to network + decode — never silently degrades (the core bug).
- `fetch_texture_by_hash` (uses the runtime setting) and `fetch_texture_by_hash_with_quality` (explicit) are unified on a shared internal that hits the fast-path when a variant satisfies the request, sharing a single load via a `{hash}_v{N}` cache key.

### Versioning / back-compat
- Bumped the optimized bucket to **`/v4`**. The runtime tries versioned buckets **newest-first (v4 → v3)** for scene manifests (`scene_fetcher.gd`), per-hash zips (`async_fetch_optimized_asset`), and wearable/emote zips. So **v4 clients keep serving v3 content until the v4 bucket is populated** — no regression while the re-bake rolls out. Legacy manifests without `textureQualities` parse fine and fall back to network.

## Out of scope (per the issue)
- Exact resolution set, cache eviction / storage budget, CDN-side negotiation.
- GLTF-embedded textures stay single-quality (baked into the `.scn`).
- **Note on packaging:** variants live inside the existing per-hash zip, so a Low client still downloads the Source bytes (disk + bandwidth, not RAM/VRAM). Splitting into per-variant zips is a possible follow-up.

## Testing
- Unit tests: variant selection rule (exact / High→Source collapse / "only lower" → None / empty manifest), and serde round-trip of `textureQualities` (camelCase, optional on read for legacy manifests).
- `cargo fmt`, `clippy`, full Rust unit suite, `gdformat`/`gdlint` pass.
- **Bake e2e**: ran the asset server locally (`--rendering-driver opengl3`), processed 3 textures of different sizes → zip contents verified byte-for-byte (1536px → q0/q1/q2/q3; 936px & 192px → q0/q1/q3 with High collapsed; valid `RSCC` headers).
- `gdformat`/`gdlint` and headless script validation pass (0 errors).
- Not yet exercised live: runtime quality matrix against re-baked v4 content, and the `/process-scene` manifest flow end-to-end.

### DIFF
--- a/lib/src/asset_server/processor.rs
+++ b/lib/src/asset_server/processor.rs
@@ -18,12 +18,13 @@ use crate::content::gltf::{
 };
 use crate::content::packed_array::PackedByteArrayFromVec;
 use crate::content::resource_provider::ResourceProvider;
+use crate::content::texture::MAX_TEXTURE_DIMENSION;
 use crate::content::thread_safety::GodotSingleThreadSafety;
 use crate::godot_classes::dcl_config::TextureQuality;
 use crate::utils::infer_mime;
 
 use super::job_manager::JobManager;
-use super::types::{AssetRequest, AssetType, JobStatus};
+use super::types::{AssetRequest, AssetType, JobStatus, TextureVariant};
 
 /// Context for asset processing, similar to ContentProviderContext but standalone.
 #[derive(Clone)]
@@ -89,6 +90,8 @@ pub struct ProcessResult {
     pub optimized_file_size: Option<u64>,
     /// GLTF dependencies - texture hashes (for GLTFs only)
     pub gltf_dependencies: Option<Vec<String>>,
+    /// Baked quality variants (for textures only)
+    pub texture_variants: Option<Vec<TextureVariant>>,
 }
 
 /// Process an asset request.
@@ -139,6 +142,9 @@ pub async fn process_asset(
             }
             if let Some(deps) = process_result.gltf_dependencies {
                 job_manager.set_gltf_dependencies(&job_id, deps).await;
+            }
+            if let Some(variants) = process_result.texture_variants {
+                job_manager.set_texture_variants(&job_id, variants).await;
             }
 
             job_manager
@@ -250,6 +256,7 @@ async fn process_scene_gltf(
         original_size: None,
         optimized_file_size,
         gltf_dependencies: Some(gltf_dependencies),
+        texture_variants: None,
     })
 }
 
@@ -305,6 +312,7 @@ async fn process_wearable_gltf(
         original_size: None,
         optimized_file_size,
         gltf_dependencies: Some(gltf_dependencies),
+        texture_variants: None,
     })
 }
 
@@ -360,6 +368,7 @@ async fn process_emote_gltf(
         original_size: None,
         optimized_file_size,
         gltf_dependencies: Some(gltf_dependencies),
+        texture_variants: None,
     })
 }
 
@@ -374,8 +383,6 @@ async fn process_texture(
     tracing::debug!("Processing texture: {}", request.hash);
 
     let raw_file_path = format!("{}{}", ctx.content_folder, request.hash);
-    // Use .res extension for compressed ImageTexture (Godot binary resource format)
-    let res_file_path = format!("{}{}.res", ctx.content_folder, request.hash);
 
     // Download the raw image file (or read from cache if cache_only)
     let bytes_vec = if request.cache_only {
@@ -476,10 +483,86 @@ async fn process_texture(
         image_format
     );
 
-    // Resize if needed based on texture quality
-    let max_size = ctx.texture_quality.to_max_size();
-    resize_image_if_needed(&mut image, max_size);
+    // Bake one variant per quality tier. Each variant is resized from a fresh
+    // duplicate of the pristine decoded image to avoid lossy resize cascades.
+    //
+    // Tiers whose max size covers the original resolution all produce identical
+    // pixels, so when the original fits within High (1024) the top variant is
+    // baked once and named at the Source ordinal: a High request then selects
+    // the q3 file via the "smallest variant >= requested" rule and gets exactly
+    // the pixels a dedicated High bake would have produced.
+    let original_max = original_width.max(original_height) as i32;
+    let mut texture_variants: Vec<TextureVariant> = Vec::new();
 
+    for quality in [
+        TextureQuality::Low,
+        TextureQuality::Medium,
+        TextureQuality::High,
+        TextureQuality::Source,
+    ] {
+        if quality == TextureQuality::High && original_max <= TextureQuality::High.to_max_size() {
+            continue; // identical to the Source variant, which is always baked
+        }
+        let max_size = quality.to_max_size().min(MAX_TEXTURE_DIMENSION);
+
+        let mut variant_image = image
+            .duplicate()
+            .ok_or_else(|| anyhow::anyhow!("Failed to duplicate image for variant"))?
+            .cast::<Image>();
+        resize_image_if_needed(&mut variant_image, max_size);
+
+        let variant_path = format!(
+            "{}{}_q{}.res",
+            ctx.content_folder,
+            request.hash,
+            quality.to_i32()
+        );
+        let file_size = compress_and_save_variant(variant_image, &variant_path)?;
+
+        tracing::debug!(
+            "Texture variant saved: {} q{} -> {} ({} bytes)",
+            request.hash,
+            quality.to_i32(),
+            variant_path,
+            file_size
+        );
+
+        texture_variants.push(TextureVariant {
+            quality: quality.to_i32(),
+            path: variant_path,
+            file_size,
+        });
+    }
+
+    let top_variant = texture_variants
+        .last()
+        .ok_or_else(|| anyhow::anyhow!("No texture variants baked"))?;
+    let optimized_path = top_variant.path.clone();
+    let optimized_file_size = Some(top_variant.file_size);
+
+    tracing::debug!(
+        "Texture {} baked ({}x{}, {} variants)",
+        request.hash,
+        original_width,
+        original_height,
+        texture_variants.len()
+    );
+
+    Ok(ProcessResult {
+        optimized_path,
+        original_size: Some((original_width, original_height)),
+        optimized_file_size,
+        gltf_dependencies: None,
+        texture_variants: Some(texture_variants),
+    })
+}
+
+/// Pad to ETC2 block size, compress, and save an image as a `.res` ImageTexture.
+/// Returns the saved file size in bytes.
+fn compress_and_save_variant(
+    mut image: Gd<Image>,
+    res_file_path: &str,
+) -> Result<u64, anyhow::Error> {
     // Compress the image using ETC2 (mobile-optimized format)
     // Pad dimensions to multiples of 4 (ETC2 operates on 4x4 blocks)
     if !image.is_compressed() {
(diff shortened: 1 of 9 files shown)


lib/src/content/content_provider.rs
```diff
@@ -1258,85 +1363,71 @@ impl ContentProvider {
...
 
+        if let Some(variant) = selected_variant {
+            let optimized_data = self.optimized_data.clone();
             let original_size = self.optimized_original_size.get(&hash_id).copied();
 
             TokioRuntime::spawn(async move {
                 let _ = ContentProvider::async_fetch_optimized_asset(
                     hash_id.clone(),
-                    ctx,
+                    ctx.clone(),
```
regenesis-claw, line 1428: When `selected_variant` is `None` but the hash is in `self.optimized_assets` (from a v3 zip mount), this falls through to network download instead of loading the already-mounted `res://content/{hash}.res`. During the v3→v4 transition, this means textures that were previously served from the optimized zip will be re-downloaded. Consider a middle branch:
```rust
} else if self.optimized_assets.contains(&hash_id) {
    // v3 legacy: single-quality bake already mounted
    let godot_path = format!("res://content/{}.res", hash_id);
    // ... load from mounted resource
}
```
Not blocking — the tradeoff is documented in the PR description.

lib/src/asset_server/processor.rs
```diff
@@ -476,10 +483,86 @@ async fn process_texture(
...
+    // baked once and named at the Source ordinal: a High request then selects
+    // the q3 file via the "smallest variant >= requested" rule and gets exactly
+    // the pixels a dedicated High bake would have produced.
+    let original_max = original_width.max(original_height) as i32;
+    let mut texture_variants: Vec<TextureVariant> = Vec::new();
+
+    for quality in [
+        TextureQuality::Low,
+        TextureQuality::Medium,
+        TextureQuality::High,
```
regenesis-claw, line 500: Nit for a follow-up: when `original_max <= TextureQuality::Medium.to_max_size()` (512), the Medium and Low variants produce identical pixels. Similarly when `original_max <= 256`. You already collapse High→Source; extending the same pattern to lower tiers would avoid baking redundant identical files for small textures.

regenesis-claw (approved):
Thorough work — the multi-resolution bake pipeline, the v4/v3 cascading fallback, cache-key unification, and the variant selection rule are all clean. Tests cover the important edge cases (collapsed High, legacy manifests, empty qualities). CI is green across all checks. Two observations, neither blocking:

**1. V3 texture regression during rollout** — The old code served textures from the mounted v3 zip via `optimized_asset_exists` → `res://content/{hash}.res`. The new code only enters the optimized path when `select_texture_variant` returns `Some`, which requires `textureQualities` in the manifest. V3 manifests lack that field, so **all standalone textures in v3-only scenes will fall back to network + decode** even though the zip was already downloaded and mounted. GLTFs are unaffected (`.scn` embeds textures). The PR description acknowledges this (`"Legacy manifests without textureQualities parse fine and fall back to network"`), but during the transition period before v4 content is fully populated this could measurably increase scene load times and bandwidth for texture-heavy scenes. A small fallback — checking `optimized_asset_exists` + trying `res://content/{hash}.res` when `select_texture_variant` returns `None` — would preserve v3 texture serving. Not blocking since the author accepted this tradeoff explicitly, but worth considering if the v4 re-bake will take a while.

**2. Identical variants for small textures** — A 256px texture bakes q0(256), q1(256), q3(256) — three identical files since `resize_image_if_needed` only downsizes. The High→Source collapse is already there; extending it to collapse Medium and Low when the original is at or below their respective thresholds would save disk/bandwidth. Minor; fine as a follow-up.

The `connection_quality_monitor.gd` headless-mode fix is a nice catch — infinite re-defer during script validation is a real footgun.

kuruk-mm:
### Possible regression: legacy (v3) standalone textures lose the optimized fast-path

Heads up on `fetch_texture_by_hash` for the rollout window. I think the "v4 clients keep serving v3 content … no regression" guarantee holds for the **zip-level** paths (scenes, wearables/emotes, per-hash zip deps — those got the v4→v3 bucket fallback), but **not** for per-hash textures.

The new fast-path decision is keyed entirely on `select_texture_variant`, which reads `optimized_texture_qualities` — populated only from the manifest's new `textureQualities` field. Legacy v3 manifests don't have that field (`#[serde(default)]` → empty map), so:

```rust
fn select_texture_variant(&self, hash, requested) -> Option<i32> {
    select_variant_from(
        self.optimized_texture_qualities.get(hash)?,  // None for v3 → bails out
        requested.to_i32(),
    )
}
```

…returns `None`, and in `fetch_texture_by_hash_internal` control falls to the `else` branch → `load_image_texture(raw_url)`.

The catch is what the **old** code keyed off — `optimized_asset_exists` (hash ∈ `optimizedContent`), which v3 manifests *do* have:

```
OLD v3:  optimized_asset_exists(hash) == true  → load res://content/{hash}.res   (pre-baked ETC2, no decode)
NEW v3:  select_texture_variant(hash) == None  → load_image_texture(raw url)      (download raw + recompress on-device)
```

So during the whole re-bake window (= all current content until v4 is populated), every texture fetched by hash — SDK `Material`-component textures (`material.rs`, `gltf_node_modifiers/material.rs`) and UI images — regresses from "load a GPU-ready `.res`" to "download the raw PNG/JPEG and recompress to ETC2 at runtime," which is exactly the mobile cost the offline pipeline exists to avoid. The "falls back to network" note in the description undersells this: the prior behavior wasn't network, it was the baked `.res`.

**Suggested fix** — keep the legacy path as a fallback when there's no per-quality info; only the *new* quality selection needs `textureQualities`:

```rust
let selected_variant = self.select_texture_variant(&file_hash, &quality);

if selected_variant.is_some() {
    // v4: pick the baked quality variant → res://content/{hash}_q{N}.res
} else if self.optimized_asset_exists(file_hash_godot.clone()) {
    // v3 legacy: still serve the single baked res://content/{hash}.res — no runtime decode
} else {
    // truly no optimized data: raw download + decode
}
```

That keeps the rollout guarantee for everything already in the v3 bucket while v4 content gets the new quality-aware behavior. (The legacy branch needs its own `godot_path = res://content/{hash}.res` and `load_baked_texture_entry` call, mirroring the variant branch.)

lib/src/content/content_provider.rs
manuelmaceira: Fixed in dd6d69cf. Added the middle branch: when `select_texture_variant` returns `None` but `optimized_asset_exists` (v3 manifests carry `optimizedContent`), it now serves the legacy `res://content/{hash}.res` instead of falling through to network. The v4 variant path and the raw download+decode fallback are unchanged; the legacy bake shares a `{hash}_legacy` cache slot since it's quality-independent.

lib/src/asset_server/processor.rs
manuelmaceira: Agreed, leaving this as a follow-up as you suggested — it's a disk/bandwidth optimization, not correctness (the selection rule already handles arbitrary recorded ordinal sets). Will track it separately.

manuelmaceira:
Thanks both — @regenesis-claw @kuruk-mm independently flagged the same v3 texture regression, and you're right it's the more impactful of the two (raw download + on-device recompress for *all* current content during the rollout window, not just network).

Fixed in dd6d69cf with the exact middle branch you both suggested:
```rust
if selected_variant.is_some() {
    // v4: res://content/{hash}_q{N}.res
} else if optimized_asset_exists(hash) {
    // v3 legacy: res://content/{hash}.res  (GPU-ready ETC2, no decode)
} else {
    // truly no optimized data: raw download + decode
}
```
The legacy single bake is quality-independent so it shares one `{hash}_legacy` cache slot across default/explicit-quality fetches. Full local pre-commit suite passes (incl. headless validation).

The identical-small-texture collapse (regenesis #2) I'm leaving as a follow-up per the suggestion — it's disk/bandwidth only, and the selection rule already handles arbitrary recorded ordinal sets.

kuruk-mm (approved):
great job

PR #2232, opened 2026-06-09
fix: scene emotes not playing / never changing on the local player

### LINKED ISSUE #2219: [Bug] Scene Animations Not Working in Mobile
### 🕹️Platform
Mobile (iOS and Android)

### 🔢App Version
v67

### 📱Device Information
Not specified (needs confirmation from reporter)

### 📄Issue description
GLB models with multiple animations are completely frozen on mobile - they display but remain stuck on a single static frame. All animations fail to play, affecting NPCs, avatars, and any scene objects with animations. The same scene works correctly on desktop client where all animations play as expected.

**Affected elements:**
- NPCs with multiple animations
- Custom avatars with walk/jump/bonk animations
- All scene objects using `Animator` component with GLB animations

**Scene for reproduction:**
https://github.com/carlosmu/goal-legends-arena
https://github.com/MetaLiveStudio/Doge-Hunt-Proof-of-Concept

/goto goallegendsarena.dcl.eth

### ✅Expected behavior
GLB animations should play on mobile the same way they do on desktop client. Models should cycle through their animation states (idle, walk, jump, etc.) instead of remaining frozen on one frame.

### 🖼️Screenshots / Media
https://github.com/user-attachments/assets/a7ae22cb-4210-46f9-b119-9463d4fd0a9f

### ▶️Steps to reproduce
1. Clone/deploy the test scene from https://github.com/MetaLiveStudio/Doge-Hunt-Proof-of-Concept
2. Visit the scene on desktop client - observe animations working correctly
3. Visit the same scene on mobile device
4. Observe that all GLB models are frozen on a single animation frame
5. Models display correctly but animations do not play at all

### 👁️Occurrence
100% of the time (reported as consistent in this specific scene)

### Additional Context
**Reporter:** MetaDoge via Discord (relayed by ToxSam)

**Related but different issues:**
- #1888 - Selective GLTF animation failures (some animations play, others don't)
- This issue is more severe: ALL animations are frozen, not selective failures

**Questions to clarify:**
- Which mobile platform specifically? (iOS/Android/both)
- Device model and OS version?
- App version being tested?
- Does this happen in other scenes with GLB animations or only this specific scene?

---
Requested by A Daniel Garcia Aranda via Slack

kuruk-mm:
Closes #2219

## Summary

Scene-triggered avatar emotes (`triggerSceneEmote`) did not play on the **local/main player**, while the very same clips animated correctly on scene-spawned avatars and `GltfContainer`s (the NPC / training bot). Reproduced in the **Goal Legends Arena** scene: the kicker/goalkeeper (you) stayed frozen while the bot played its animations.

Root cause was **two** independent bugs in the local-player scene-emote path. The NPC/bot dodges both because it uses the `Animator` component on a single GLB with distinctly-named clips, not the avatar emote pipeline.

### Bug 1 — scene emote never loads when a content hash contains `-`
The scene-emote URN is `urn:decentraland:off-chain:scene-emote:{scene_id}-{glb_hash}-{loop}`. `EmoteSceneUrn` parsed it by splitting on `-`, but **preview** content hashes are `b64-<base64-of-path>` and contain a dash. The split dropped the `b64-` prefix from `glb_hash` and corrupted `scene_id`, so the content-registry lookup missed and the GLB never loaded:

```
Failed to load scene emote: urn:...:scene-emote:b64-...=-b64-...K_intro_emote.glb...-false
Emote urn:...K_intro_emote...-false not found from player
```

Deployed CID hashes (`bafkrei...`) have no dashes, so this only surfaced in preview.

**Fix:** resolve the URN against the registry that Rust populates via `register_scene_emote_content()` immediately before `async_play_emote()` — an exact, unambiguous match regardless of dashes. The old dash-split parse is kept as a fallback for dashless CID hashes / remote players.

### Bug 2 — every scene emote plays the first-loaded clip ("animation never changes")
Independent emote GLBs frequently ship the **same internal animation name**. In this scene every clip is named either `Armature` (K_intro, GK_shoot_L/C/R) or `Animation` (GK_intro, K_shoot). They all share one `AnimationLibrary` on the avatar, keyed by that raw name, so after the first `Armature`/`Animation` is added, every later emote with the same name hit `has_animation()` and silently reused the first clip.

**Fix:** namespace the library key by the per-GLB content hash, so every emote is distinct. This also fixes a latent cleanup bug where removing one emote could delete a clip another emote still referenced.

## Verification (desktop, `--emulate-ios`, preview)
Reproduced live against the hosted scene. Captured the restricted-action + emote-load logs before/after:

| | triggers | "Failed to load scene emote" | "not found from player" |
|---|---|---|---|
| before | 4 | 4 | 4 |
| after  | 7 | **0** | **0** |

With temporary instrumentation, distinct scene emotes now register under distinct library keys (e.g. `…GK_intro_emote.glb…_…` vs `…GK_shoot_L_emote.glb…_…`) and play distinct animations. Confirmed working in-game.

## Scope / risk
- GDScript only (no Rust rebuild). `gdformat` + `gdlint` clean.
- `default_anim_name` is only used internally (store / play / cleanup), so re-keying the shared library is contained to the emote controller.

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar.gd
+++ b/godot/src/decentraland_components/avatar/avatar.gd
@@ -1714,6 +1714,35 @@ func get_scene_emote_info(scene_id: String, glb_hash: String) -> Dictionary:
 	return {"base_url": Global.realm.content_base_url, "audio_hash": ""}
 
 
+## Resolve a scene-emote URN back to its registered (scene_id, glb_hash) and content.
+## The URN format `scene-emote:{scene_id}-{glb_hash}-{loop}` is ambiguous to dash-split
+## when the ids contain '-' — which is exactly the case for preview content, whose hashes
+## look like `b64-<base64-of-path>`. Splitting on '-' then drops the `b64-` prefix from
+## glb_hash and corrupts scene_id, the registry lookup misses, and the emote never loads.
+## Instead we match the URN against the registry that Rust populated via
+## register_scene_emote_content() right before async_play_emote(), which is unambiguous.
+## Returns {scene_id, glb_hash, base_url, audio_hash, looping} or {} when no entry matches.
+func find_scene_emote_by_urn(emote_urn: String) -> Dictionary:
+	for scene_id in _scene_emote_registry:
+		var scene_data = _scene_emote_registry[scene_id]
+		for glb_hash in scene_data["emotes"]:
+			for looping in [false, true]:
+				var loop_str := "true" if looping else "false"
+				var candidate := (
+					"urn:decentraland:off-chain:scene-emote:%s-%s-%s"
+					% [scene_id, glb_hash, loop_str]
+				)
+				if candidate == emote_urn:
+					return {
+						"scene_id": scene_id,
+						"glb_hash": glb_hash,
+						"base_url": scene_data["base_url"],
+						"audio_hash": scene_data["emotes"][glb_hash],
+						"looping": looping,
+					}
+	return {}
+
+
 ## Play emote triggered by AvatarShape's expression_trigger_id field.
 ## Supports: default emotes (e.g. "wave"), URN emotes, and local scene emotes (.glb/.gltf)
 func _async_play_expression_trigger(emote_id: String) -> void:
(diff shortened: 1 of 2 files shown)


EibrielInv (approved):
Looks good! Would be nice to have a way to find an emote without looping so much. But I suppose is a limitation of the SDK.

PR #2252, opened 2026-06-12
fix: scene emotes not playing on remote players (multiplayer)

### LINKED ISSUE #2240: [Bug] Scene Emotes Rotated 180 Degrees - Avatar Attachments Offset
## Platform
Mobile (iOS & Android)

## App Version
Current release branch

## Device Information
Reproduced on multiple devices

## Issue Description
Scene emotes are rotated 180 degrees, causing all objects attached to AvatarAttach points to be positioned incorrectly (offset/mirrored). This bug has apparently always existed in the mobile client.

When using scene emotes with avatar attachment points, the attachments appear on the opposite side from where they should be.

## Expected Behavior
Scene emotes should match Unity client behavior - attachments should appear in the correct position relative to the avatar, not rotated/mirrored.

## Screenshots/Media
See Slack thread video: ScreenRecording_06-10-2026 22-56-33_1.MP4
(User must upload video manually to GitHub)

## Steps to Reproduce
1. Load a scene that uses emotes with avatar attachment points
2. Trigger an emote
3. Observe that attached objects appear offset/rotated 180 degrees from expected position
4. Compare with Unity client - works correctly there

## Occurrence
Always - affects all scene emotes with avatar attachments

## Additional Context
- Bug identified by Mateo during World Cup Prediction Game testing
- Fix implemented, but may require reprocessing all emotes through asset pipeline
- Blocking release of scene with emote features

---
Requested by Rodri via Slack

kuruk-mm:
Closes #2240 

Follow-up to #2232, which fixed scene emotes for the **local** player. In **multiplayer** the same clips still never played on the *other* clients: in Goal Legends Arena, the opponent (and any spectator) saw the kicker/goalkeeper frozen in Idle for the whole match, while their own avatar animated correctly.

## Root cause

When a scene calls `triggerSceneEmote`, Rust registers the emote content (`register_scene_emote_content`) **only on the triggering player's own avatar** before playing, then broadcasts the URN over comms. On receiving clients that registry is empty, so `_async_load_scene_emote_as_wearable` fell back to dash-splitting the URN payload:

```
urn:decentraland:off-chain:scene-emote:{scene_id}-{glb_hash}-{loop}
```

Preview ids/hashes are `b64-<base64>` and contain `-` (and `=`), so the split stripped the `b64-` prefix off `glb_hash` and corrupted `scene_id`. The content lookup missed and every remote emote failed:

```
Failed to load scene emote: urn:...:scene-emote:b64-...=-b64-...K_intro_emote.glb...-false
Emote urn:... not found from player 'zsDYw8UjVo#6287'
```

The training bot was never affected (it's a `GltfContainer` + `Animator`, not an avatar emote), which is why PvE looked fine on the player's own screen. `AvatarShape` `expressionTriggerId` scene emotes had the same gap (URN built in Rust, never registered).

## Fix

- **`SceneManager::resolve_scene_emote_urn()`** (new `#[func]`): resolves the URN against the **loaded scenes' entity ids** — an exact prefix match, unambiguous no matter how many dashes the ids contain — and returns `{scene_id, glb_hash, base_url, audio_hash, looping}` from that scene's content mapping.
- **`ContentMappingAndUrl::get_scene_emote_hash_by_glb_hash()`**: reverse lookup (hash → glb file → sibling audio), so remote players now also get the emote **audio**, which the old realm-URL fallback never provided — even on deployed scenes.
- **`avatar_emote_controller.gd`**: resolution chain is now per-avatar registry (local trigger) → loaded-scenes resolver (remote players, AvatarShape) → legacy dash-split (only for dashless deployed CID hashes when the emote's scene isn't loaded locally).
- **`debug_ws_server.gd`**: `DCL_DEBUG_WS_PORT` env override so a second local instance can bind its own debug inspector — needed to verify this fix, generally useful for multiplayer debugging.

## Verification (live, two desktop clients vs LAN preview)

Ran two explorer instances on one machine (separate guest identities) against the goal-legends-arena preview; the scene's auto-PvE played full matches while the other client observed, inspected over the debug WS:

| | before | after |
|---|---|---|
| `Failed to load scene emote` / `not found from player` on observer | 4+ per match | **0** across multiple matches |
| Remote avatar state machine during rounds | pinned to `Idle` | cycles `Run → Emote → Idle` per round |
| Scene emotes loaded on the remote avatar | 0 | 6–8 (distinct library keys) |
| Ball prop on remote kicker | never appears | visible during intro/kick, hidden on idle |

Resolver edge cases checked live via `eval`: preview `b64-…=` ids resolve with correct `looping`; unknown scene returns `{}` and falls through to the legacy parse unchanged.

4 new unit tests for the reverse lookup. `cargo fmt` / `clippy -D warnings` / `gdformat` / `gdlint` clean.

## Scope / risk

Local-player emotes still resolve through the registry exactly as before — only the remote/NPC resolution path changed. If a remote player's scene isn't loaded on the observer, behavior is identical to today's fallback.

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar_emote_controller.gd
+++ b/godot/src/decentraland_components/avatar/avatar_emote_controller.gd
@@ -497,14 +497,20 @@ func _async_load_scene_emote_as_wearable(scene_emote_urn: String):
 	var looping: bool
 
 	var resolved = avatar.find_scene_emote_by_urn(scene_emote_urn)
+	if resolved.is_empty():
+		# Remote players and AvatarShape triggers never populate the per-avatar
+		# registry (it's filled by Rust only on the avatar that ran
+		# triggerSceneEmote), so resolve the URN against the loaded scenes —
+		# unambiguous even when ids/hashes contain '-' (preview `b64-` hashes).
+		resolved = Global.scene_runner.resolve_scene_emote_urn(scene_emote_urn)
 	if not resolved.is_empty():
 		glb_hash = resolved["glb_hash"]
 		base_url = resolved["base_url"]
 		audio_hash = resolved["audio_hash"]
 		looping = resolved["looping"]
 	else:
-		# Fallback: dash-split parse (correct for dashless CID hashes, e.g. deployed
-		# content) and the realm-URL fallback for remote players not in the registry.
+		# Last resort: dash-split parse (correct for dashless CID hashes, e.g. deployed
+		# content, when the emote's scene is not loaded locally).
 		var parsed = EmoteSceneUrn.new(scene_emote_urn)
 		if not parsed.is_valid:
 			printerr("Failed to parse scene emote URN: %s" % scene_emote_urn)
--- a/godot/src/tool/debug_server/debug_ws_server.gd
+++ b/godot/src/tool/debug_server/debug_ws_server.gd
@@ -72,7 +72,13 @@ func _ready() -> void:
 	# manual toggle. Release/exported builds stay off until the Settings →
 	# Developer toggle. Never in production.
 	if OS.is_debug_build() and not Global.is_production():
-		start()
+		# DCL_DEBUG_WS_PORT lets a second local instance get its own inspector
+		# (the default port can only be bound by one process).
+		var port := DEFAULT_PORT
+		var env_port := OS.get_environment("DCL_DEBUG_WS_PORT")
+		if env_port.is_valid_int():
+			port = env_port.to_int()
+		start(port)
 
 
 func is_running() -> bool:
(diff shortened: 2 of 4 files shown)


regenesis-claw (approved):
Clean fix for a well-diagnosed problem. The prefix-matching approach in `resolve_scene_emote_urn()` is much more robust than dash-splitting — correctly handles `b64-` hashes with embedded dashes. Resolution chain (per-avatar registry → loaded-scenes → legacy fallback) is well-layered with a graceful degradation path.

A few notes:

- **Branch is 1 commit behind main** — not blocking, but worth merging before final merge to keep CI signals trustworthy.
- **iOS build present and green** ✅ — TestFlight artifact posted.
- `get_scene_emote_hash_by_glb_hash()` does a linear scan over content mappings — fine given typical scene sizes, and consistent with the existing `get_scene_emote_hash()` pattern.
- 4 unit tests cover the reverse lookup well (with audio, without audio, unknown hash, non-gltf filtering).
- The `DCL_DEBUG_WS_PORT` env override is a nice quality-of-life addition for multiplayer debugging; correctly gated behind `is_valid_int()`.
- No debug prints, no `.claude/` files, no orphaned code.

Approving — solid root-cause analysis, targeted fix, good test coverage.

Requested by Mateo via Slack

PR #2262, opened 2026-06-15
fix: scene emotes not playing on remote players (multiplayer) (#2252) [release cherry-pick]

kuruk-mm:
Cherry-pick of 7791944b (#2252) from `main` into `release`.

## What this fixes
Scene-emote URNs broadcast over comms arrive on other clients without the per-avatar content registration that `triggerSceneEmote` performs locally, so the receiver dash-split the URN payload. Preview ids/hashes are `b64-<base64>` and contain `-`, so the split corrupted `scene_id`/`glb_hash`, the content lookup missed, and remote avatars stayed frozen in Idle while the local player animated fine.

The fix resolves the URN against the loaded scenes' entity ids (exact prefix match, unambiguous regardless of dashes), reading `base_url` and the emote's sibling audio hash from the scene's content mapping. Also covers `AvatarShape` expression triggers (same gap) and gives remote players emote audio the old realm-URL fallback never provided. Dash-split parsing remains only as a last resort for dashless deployed hashes when the scene isn't loaded.

## Files
- `lib/src/content/content_mapping.rs` (+74) — scene-emote URN resolver + tests
- `lib/src/scene_runner/scene_manager.rs` (+48) — resolve emote URNs against loaded scenes
- `godot/src/decentraland_components/avatar/avatar_emote_controller.gd` (+8/-2)

## Deviation from the original commit
The original commit also touched `godot/src/tool/debug_server/debug_ws_server.gd` (a `DCL_DEBUG_WS_PORT` env override on the debug-WS auto-start path). That hunk was **intentionally dropped** here: the auto-start block it modifies comes from PR #2208, which is not on `release`. Pulling the hunk in would have imported a slice of #2208's debug-build behavior. The dev-only override has no bearing on the user-facing emote fix and never runs in production builds.

### DIFF
--- a/godot/src/decentraland_components/avatar/avatar_emote_controller.gd
+++ b/godot/src/decentraland_components/avatar/avatar_emote_controller.gd
@@ -497,14 +497,20 @@ func _async_load_scene_emote_as_wearable(scene_emote_urn: String):
 	var looping: bool
 
 	var resolved = avatar.find_scene_emote_by_urn(scene_emote_urn)
+	if resolved.is_empty():
+		# Remote players and AvatarShape triggers never populate the per-avatar
+		# registry (it's filled by Rust only on the avatar that ran
+		# triggerSceneEmote), so resolve the URN against the loaded scenes —
+		# unambiguous even when ids/hashes contain '-' (preview `b64-` hashes).
+		resolved = Global.scene_runner.resolve_scene_emote_urn(scene_emote_urn)
 	if not resolved.is_empty():
 		glb_hash = resolved["glb_hash"]
 		base_url = resolved["base_url"]
 		audio_hash = resolved["audio_hash"]
 		looping = resolved["looping"]
 	else:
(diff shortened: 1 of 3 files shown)


regenesis-claw (approved):
Clean cherry-pick of #2252 onto `release`. The intentional omission of the `debug_ws_server.gd` hunk (from #2208, not on release) is the right call — avoids importing unrelated debug-build behavior.

Code looks good:
- `resolve_scene_emote_urn` correctly strips the loop suffix first, then prefix-matches scene entity IDs — handles the `b64-` dash problem without ambiguity.
- `get_scene_emote_hash_by_glb_hash` reverse-lookup is clean; properly filters to `.glb`/`.gltf` only.
- GDScript fallback chain is well-structured: per-avatar registry → scene runner resolution → dash-split last resort.
- Good test coverage on the new `content_mapping` method (happy path, no audio, unknown hash, non-gltf filter).

All CI checks green. No iOS-sensitive changes — iOS build not required.

Requested by Mateo via Slack

PR #2307, opened 2026-06-18
feat: highlight effect on objects (pointer-events show_highlight)

### LINKED ISSUE #857: Implement 'Highlights' effect on objects, and pointer events property 'show highlight'
Similar to outline, but for objects that have show_highlight=true

implementing the following protocol change: https://github.com/decentraland/protocol/pull/224

manuelmaceira:
## What

Implements the **"Highlights" effect** on interactive scene objects, and wires up the new `show_highlight` pointer-events property.

Closes #857. Implements protocol change decentraland/protocol#224.

When the crosshair hovers an entity whose `PBPointerEvents.Info` has `show_highlight = true` (default) and `show_feedback = true` within `max_distance`, the entity's meshes get the **same outline effect already used for avatars** — no new shader, no new camera/viewport.

Semantics (per the protocol comments):
- `show_feedback = false` → no hover text **and** no highlight.
- `show_highlight = false` → hover text still shows, **no** highlight.

## How

Minimal change, reusing existing machinery (the layer-20 depth camera + Sobel post-process and `_set_layers_recursive`):

- **`lib/src/scene_runner/scene_manager.rs`** — new `#[var] highlighted_entity: Option<Gd<Node3D>>`, computed inside the existing hover/tooltip loop (same place `show_feedback`/`max_distance` are read) and surfaced to GDScript on the existing `pointer_tooltip_changed` signal.
- **`godot/src/logic/player/outline_system.gd`** — `set_outlined_entity()` reuses `_set_layers_recursive` to toggle the outline render layer; a shared `_update_effect_visibility()` keeps the post-process active while **either** an avatar or an entity is highlighted (the crosshair only hits one at a time).
- **`godot/src/ui/explorer.gd`** — drives the entity highlight from `change_tooltips()`, tracking `_last_outlined_entity` and respecting the hide-world-interactions session toggle.

No protocol bump needed — `show_highlight = 5` is already in the pinned protocol; prost generates the field automatically.

## Testing

- `cargo run -- full-tests --skip-visual` passes (fmt, gdformat, gdlint, clippy, GDScript validation, Rust unit + integration tests).
- Manual: a scene with `PBPointerEvents` on a mesh entity — outline appears on hover by default; disappears with `show_highlight:false` (tooltip stays); neither tooltip nor outline with `show_feedback:false`; nothing beyond `max_distance`; avatar vs object never both outlined at once.

### DIFF
--- a/godot/src/logic/player/outline_system.gd
+++ b/godot/src/logic/player/outline_system.gd
@@ -5,6 +5,7 @@ const OUTLINE_LAYER = 20  # Using layer 20 for outlined objects (bit 19, value 5
 
 var main_camera: Camera3D = null
 var current_outlined_avatar: Node3D = null
+var current_outlined_entity: Node3D = null
 
 # Cached so we only push to the shader on size changes.
 var _last_pushed_viewport_size: Vector2i = Vector2i.ZERO
@@ -68,11 +69,35 @@ func set_outlined_avatar(avatar: Node3D):
 	if avatar:
 		_set_avatar_layers(avatar, true)
 
+	_update_effect_visibility()
+
+
+func set_outlined_entity(entity: Node3D):
+	if current_outlined_entity == entity:
+		return
+
+	# Clear previous outline (guard against a freed entity node)
+	if is_instance_valid(current_outlined_entity):
+		_set_layers_recursive(current_outlined_entity, false)
+
+	current_outlined_entity = entity
+
+	# Set new outline
+	if entity:
+		_set_layers_recursive(entity, true)
+
+	_update_effect_visibility()
+
+
+# The outline post-process is shared by the avatar and entity paths; keep it
+# rendering while either has a target (the crosshair hits only one at a time).
+func _update_effect_visibility():
+	var active := current_outlined_avatar != null or current_outlined_entity != null
 	if outline_quad:
-		outline_quad.visible = avatar != null
+		outline_quad.visible = active
 	if sub_viewport:
 		sub_viewport.render_target_update_mode = (
-			SubViewport.UPDATE_ALWAYS if avatar != null else SubViewport.UPDATE_DISABLED
+			SubViewport.UPDATE_ALWAYS if active else SubViewport.UPDATE_DISABLED
 		)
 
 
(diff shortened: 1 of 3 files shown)


kuruk-mm (approved):
looks great!

godot/src/logic/player/outline_system.gd
```diff
@@ -68,11 +69,35 @@ func set_outlined_avatar(avatar: Node3D):
 	if avatar:
 		_set_avatar_layers(avatar, true)
 
+	_update_effect_visibility()
+
+
+func set_outlined_entity(entity: Node3D):
+	if current_outlined_entity == entity:
+		return
+
```
regenesis-claw, line 78: Minor: `current_outlined_avatar != null` doesn't catch a freed reference (freed ≠ null in GDScript). Consider:
```gdscript
var active := is_instance_valid(current_outlined_avatar) or is_instance_valid(current_outlined_entity)
```
Non-blocking — the callers already guard properly, so this would only matter if a node is freed externally without going through the setter.

regenesis-claw (approved):
Clean, well-scoped change. Reuses the existing outline infrastructure (layer-20 depth camera + Sobel post-process) without adding any new shaders or viewports — exactly the right approach.

**What I checked:**
- Protocol semantics: `show_feedback=false` → skips the loop body → no highlight and no tooltip. `show_highlight=false` → tooltip still shows, no outline. Matches the PR description and protocol spec.
- Freed-node safety: `set_outlined_entity` correctly uses `is_instance_valid()` before clearing, and `explorer.gd` guards with `is_instance_valid()` before reading `highlighted_entity`. `_exit_tree` was also upgraded to `is_instance_valid` for both avatar and entity — good fix.
- Signal coalescing: `pointer_tooltip_changed` fires when either tooltips or highlight state changes — no redundant emissions.
- Mutual exclusion: the crosshair only hits one target at a time, so `_update_effect_visibility()` correctly OR's the two states.
- The `should_highlight` flag accumulates across all `PointerEventInfo` entries on the entity — "highlight if any active event wants it" is the correct semantic since each entry represents a different interaction.

**One minor suggestion** (non-blocking): In `_update_effect_visibility()`, `current_outlined_avatar != null` and `current_outlined_entity != null` don't check instance validity — a freed reference is non-null. If either node is freed externally (without going through the setter), `active` stays `true` and the Sobel post-process keeps running on an empty depth buffer. No visual artifact, just unnecessary GPU work. Using `is_instance_valid()` in that check would be slightly more robust. Not blocking since the callers guard properly.

Requested by Mateo via Slack

PR #2311, opened 2026-06-19
perf(avatar): kill per-spawn mesh clone via surface override materials (#2269)

manuelmaceira:
## Spike #2269 — Avatar hiccups investigation

Investigates and fixes the main-thread frame-time spike ("hiccup") when spawning avatars.

### Tooling added
- **`avatar_spawn` benchmark** (reuses the existing `--spawn-avatars` flag): rewrote `TestSpawnAndMoveAvatars` to spawn random avatars **one at a time**, wait for `avatar_loaded`, sample the per-frame spike + per-phase cost, and write a JSON report to `user://output/avatar-spawn-bench.json` (per-avatar rows + p50/p95/p99 aggregates).
- **`AvatarBuildProfiler`** (`godot/src/tools/avatar_build_profiler.gd`): lightweight static phase timer, **no-op when disabled** (zero overhead in production). `begin()`/`mark()` calls inserted at the synchronous phase boundaries of `avatar.gd::async_load_wearables`.

### Diagnosis (desktop M4 Pro, 45 avatars)
Per-avatar **synchronous** build cost ≈ **9.5 ms p50**, of which **`child.mesh.duplicate_deep()` was ~92% (8.8 ms)**. Toon/facial/hide were negligible (the toon-material cache is already `static`/shared). The remaining ~76–92 ms wall time is **off-thread** GLTF loading (not a hiccup).

### Fix
The mesh was shallow-cloned only so per-avatar `surface_set_material` wouldn't leak across avatars. Replaced mesh-level material mutation with **`MeshInstance3D.set_surface_override_material`** (per-instance override, no mesh clone) in `apply_toon_material` / `apply_color_and_facial` / `apply_texture_and_mask`, and removed the `duplicate_deep` loop.

### Result (same benchmark, before → after)
| Metric | Before | After |
|---|---|---|
| Synchronous build cost p50 | 9.5 ms | **0.55 ms** (~17×) |
| `mesh_duplicate` phase | 8.8 ms | ~0 |
| Frames >33 ms over 45 spawns | 14 | 2 |
| Hiccup max-frame p95 | 52 ms | 26 ms |

GDScript only — **no Rust changes, no new CLI flag**.

### Follow-ups (out of scope)
- Visual spot-check on screen/device.
- Re-measure on A54 (savings expected to be proportionally larger on mobile).

> Note: pre-commit `cargo check` was skipped for the commit — this branch was built in a worktree reusing a sibling's prebuilt lib, so the Rust build-script check fails on an environmental missing-dir, unrelated to these GDScript-only changes. `gdformat`/`gdlint` pass.


regenesis-claw (approved):
**LGTM — clean, well-diagnosed performance fix.**

Following `REVIEW.md`.

### Pre-flight
- ✅ Branch is up to date with `main`
- ✅ iOS build not required (GDScript-only changes, no platform branches)
- ✅ No submodule drift

### Review

The root-cause analysis is solid: `duplicate_deep()` on every avatar mesh was ~92% of the synchronous spawn cost, and the replacement with `set_surface_override_material()` is the correct Godot pattern — per-instance material overrides without touching the shared mesh resource.

**Core fix (avatar.gd)**
- `apply_toon_material`: `mesh.surface_set_material` → `node.set_surface_override_material` ✅
- `apply_color_and_facial`: correctly reads override first (`get_surface_override_material`), falls back to mesh material. The new `material.duplicate()` + `set_surface_override_material` for `StandardMaterial3D` is important — without the mesh clone, the shared material would leak `metallic=0` / tint across avatars. Good catch. ✅
- `apply_texture_and_mask`: same override pattern ✅
- `duplicate_deep` loop removed ✅

**Profiler (avatar_build_profiler.gd)**
- Static, no-op when `enabled = false`, zero production overhead ✅
- Note: static state means concurrent avatar builds would interleave marks, but that's fine — only the benchmark enables it, and it spawns one at a time

**Benchmark (spawn_and_move.gd)**
- One-at-a-time spawn with `avatar_loaded` signal gate, warmup discard, percentile aggregates, JSON report output — solid benchmarking methodology ✅
- `_pending_avatar_node` capture via sync signal is clean (relies on `avatar_added` being emitted non-deferred, with null-check fallback)
- Moving-avatars phase removed in favor of pure spawn measurement — appropriate for the hiccup investigation scope

17× improvement on synchronous build cost is excellent. Nice work.

kuruk-mm (approved):
great!

PR #2329, opened 2026-06-24
ci(ios): publish unsigned iOS export to R2 (decouple ASC handoff)

leanmendoza:
First step of decoupling App Store Connect from this repo.

Adds a standalone workflow **`ios_r2_artifact.yml`** that builds the iOS app (Rust lib + Godot export) and publishes the **unsigned** export tree to the R2 mobile-artifacts bucket. A separate deploy repo will pull this artifact, sign it, set the build number and upload to TestFlight — so this repo holds no Apple signing material or ASC keys.

- `workflow_dispatch` only, **no `environment:`** → no reviewer-approval gate (unlike `ios_builds.yml`, which routes through the gated `ios-*` environments).
- Builds → `cargo run -- export --target ios` → `tar` of `exports/` → `aws s3 cp` to `ios/<branch>/<sha>/export.tar.gz` (+ `latest/`). Mirrors the Android R2 publish pattern.
- No cert / no ASC / no fastlane here. `libdclgodot` dSYM still goes to Sentry from here (built here).
- `ios_builds.yml` and `fastlane/Fastfile` are untouched — the production TestFlight flow is unaffected.

Note: `workflow_dispatch` only becomes available once this lands on `main` (GitHub registers manual triggers from the default branch).


kuruk-mm (approved):
Looking greatt!!

PR #2337, opened 2026-06-24
perf(ci): thin LTO + WarpCache (restore-all / save-on-main) for iOS & Android

leanmendoza:
Speeds up the iOS **and** Android CI builds substantially, plus a small Slack fix.

## What
- **`lto = "thin"`** in `[profile.release]` (`lib/Cargo.toml`): the dominant build cost was the fat-LTO whole-program link, which a cache can't skip — thin LTO nearly halves it. This profile is **shared**, so it speeds up *every* platform (iOS, Android, Linux, Windows, macOS), not just iOS.
- **Cache the Rust build** so the dep tree (v8/deno, ffmpeg, livekit, godot-rust) isn't recompiled from scratch every run. Both legs use the **WarpBuild-native cache** (raw `target/`, no cleaning), split into two composite actions:
  - `set-up-warp-cache` → **restore**, runs on **every branch**.
  - `save-warp-cache` → **save**, runs **only on `main`** (see below).
  - iOS (`ios_r2_artifact.yml`) and Android (`android_builds.yml`). The Android build runs on **every push/PR**, so it benefits far more often.
- **Android runs inside a Docker container**, so two extra bits are needed for WarpCache there:
  - forward `WARPBUILD_RUNNER_VERIFICATION_TOKEN` into the container (without it the cache action silently no-ops with `Authentication token is invalid`);
  - ensure `wget` (the action shells out to it).
- **`dry_run` input** on `ios_r2_artifact` (build only: skip the R2 upload / deploy dispatch / Sentry) — used for the timing runs below and handy for future build validation.
- **Fix**: gate the Slack thread posts on `SLACK_TS` (a thread exists) instead of `SLACK_CHANNEL`. Since the channel is now read straight from the secret it's always set, so standalone `workflow_dispatch` runs were posting messages outside any thread.

## Why save only on main
The cache **save** is the expensive half: it uploads the full multi-GB `target/` and takes **~5 min**. If every push saved, that ~5 min would cancel almost the entire build speed-up on a typical (code-changing) PR push — a wash. So PR builds **restore** from main's cache (via `restore-keys` prefix match) but **don't save**; only the less-frequent `main` build refreshes it. PRs get the full speed-up for free.

## Why WarpCache and not rust-cache (Android)
`Swatinem/rust-cache`'s cleanup **strips the prebuilt `librusty_v8.a`** (the v8 crate's downloaded static lib under `target/.../build/v8-*/out`) while keeping v8's fingerprint, so warm builds skip rebuilding v8 but then **fail to link** (`could not find native static library rusty_v8`). The WarpBuild-native cache keeps the raw `target/`, so the lib survives — matching the iOS leg.

## Measured — iOS (build-only timing runs, `warp-macos-26`)
| | Build MacOS | Build iOS | TOTAL |
|---|---|---|---|
| fat LTO (cold) | 8:27 | 7:59 | 20:19 |
| fat LTO (warm cache) | 6:47 | 6:19 | 16:35 |
| **thin LTO (cold)** | **4:26** | **4:22** | **12:28** |

## Measured — Android (`Build Rust libraries`, `warp-ubuntu` container)
| | Build Rust | Restore | Save |
|---|---|---|---|
| fat LTO, no cache (main today) | 13:48 | — | — |
| thin LTO, cold (no cache hit) | ~11:23 | — | — |
| **thin LTO, warm (WarpCache)** | **5:21** | 1:16 | 0 on PRs / ~5:15 on main |

Net on a warm PR push: `1:16 + 5:21 + 0 ≈ 6:37` vs `11:23` uncached → **~5 min faster**, and the v8 link error is gone.

## Why thin LTO is worth it (the trade-off)
Thin LTO recovers ~90–95% of full (fat) LTO's optimization, so the **runtime/battery difference vs fat is ~1–2% — imperceptible** for this app, whose hot path is the Godot engine (C++) + V8 + GPU, not the Rust gdextension's cross-crate inlining. In exchange it nearly halves the Rust compile time on every build. The imperceptible perf delta is well worth the large, recurring CI speedup, and it's reversible in one line if an on-device benchmark ever shows otherwise.

### DIFF
--- a/.github/actions/save-warp-cache/action.yml
+++ b/.github/actions/save-warp-cache/action.yml
@@ -1,0 +1,37 @@
+name: Save WarpBuild cache (main only)
+description: >
+  Persist the Rust build + cargo registry caches, but ONLY on a push to main. PR builds
+  restore the cache (via set-up-warp-cache) without paying the multi-minute save on every
+  push; only the less-frequent main build refreshes it, which PRs then restore from. Keys
+  come from set-up-warp-cache's outputs (captured before the build, so generated .rs files
+  under target/ can't shift the key between restore and save).
+inputs:
+  cargo-key:
+    description: cargo-key output from set-up-warp-cache.
+    required: true
+  rust-key:
+    description: rust-key output from set-up-warp-cache.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: WarpCache save - Cargo Registry (main only)
+      if: github.ref == 'refs/heads/main'
+      uses: WarpBuilds/cache/save@v1.4.6
+      with:
+        path: |
+          ~/.cargo/bin/
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          ~/.cargo/git/db/
+        key: ${{ inputs.cargo-key }}
+
+    - name: WarpCache save - Rust Build (main only)
+      if: github.ref == 'refs/heads/main'
+      uses: WarpBuilds/cache/save@v1.4.6
+      with:
+        path: |
+          target/
+          lib/target/
+        key: ${{ inputs.rust-key }}
--- a/.github/actions/set-up-warp-cache/action.yml
+++ b/.github/actions/set-up-warp-cache/action.yml
@@ -1,15 +1,27 @@
-name: Set up WarpBuild cache
-description: Set up cache with WarpBuild optimized caching
+name: Set up WarpBuild cache (restore)
+description: >
+  Restore the Rust build + cargo registry caches (WarpBuild-native, raw target/ with NO
+  cleaning so the v8 prebuilt static lib survives). Restore runs on every branch; the SAVE
+  is handled separately by save-warp-cache and only writes on main — PR builds get the
+  restore speed-up without paying the multi-minute save on every push.
 inputs:
   cache-name:
-    description: The cache base name (job name by default).
+    description: The cache base name (job name by default). Currently informational; keys are keyed by runner.os.
     default: ${{ github.job }}
+outputs:
+  cargo-key:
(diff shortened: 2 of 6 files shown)


kuruk-mm (approved):

regenesis-claw (approved):
Clean, well-documented CI optimization. Reviewed against REVIEW.md.

**Pre-flight:**
- Branch is mergeable, no conflicts.
- iOS build not required — changes are purely CI/workflow infrastructure + one `Cargo.toml` profile line. No platform-sensitive runtime code.
- No submodule pointer drift.

**Findings:** None blocking.

The restore-all / save-on-main cache split is the right pattern — PRs get the speed-up without paying the save cost. Key details are all handled correctly:

- `WARPBUILD_RUNNER_VERIFICATION_TOKEN` forwarded into the Docker container (silent no-op otherwise ✓)
- `wget` installed as a fallback for the WarpCache action dependency ✓
- Cache keys computed *before* the build (so generated `.rs` under `target/` can't shift the key between restore and save) ✓
- `save-warp-cache` placed right after the build step so a later export/sign failure can't lose a valid cache ✓
- `dry_run` input gates R2 upload, deploy dispatch, and Sentry — useful for future timing runs ✓
- Slack thread fix (`SLACK_TS` instead of `SLACK_CHANNEL`) prevents orphan messages on standalone `workflow_dispatch` runs ✓
- Thin LTO tradeoff is well-justified: ~1–2% runtime delta (negligible for this app's hot path) in exchange for nearly halving Rust compile time, and it's a one-line revert if benchmarks ever show otherwise ✓

The measured numbers are compelling — ~5 min faster on a warm Android PR push, iOS cold build cut from 20 min to 12.5 min.

LGTM. Approving.

PR #2342, opened 2026-06-25
tool: improve logging and dev connection

leanmendoza:
# What

Unifies logging + remote debugging into a single **scene-inspector channel** and adds a desktop **debug-hub** that brokers it, so `cargo run -- run --target ios/android` gives the same rich on-screen log + inspector experience on a physical device as on desktop — **without changing production behaviour**.

Goals:
- `cargo run -- run --target ios/android` produces the same behaviour as desktop (unified logs, scene inspector, MCP-friendly channel).
- One-command deploy + debug for mobile: auto-starts the hub, wires the scene-inspector, streams device logs.
- iOS **dev relay** (`adb reverse` equivalent) so localhost-only flows (OAuth callbacks, dev servers) reach the Mac from a physical device.
- All of the above **strictly gated out of production builds**.

### xtask (`cargo run -- …`)
- **`debug-hub`** *(new)* — rendezvous broker: device connect-out ↔ many local consumers (logs + tree queries + eval). Flags: `--device-port` (9231, device-facing), `--consumer-port` (9230, **loopback only**), `--bind` (0.0.0.0).
- **`run`** — new flags:
  - `--launch-only` — relaunch the installed app, no build/export/install.
  - `--deeplink "<url>"` — bake a deeplink applied at startup with no CLI args (mobile / TestFlight).
  - `--no-hub` — don't auto-start the hub/scene-inspector; fall back to plain device-log streaming (`adb logcat` / iOS `--console`).
  - `--hub-viewer` — make the hub the sole on-screen log view (no logcat/console duplication).
  - *Behaviour:* android/ios device deploys now **auto-start the debug-hub and wire the scene-inspector** (`adb reverse` / iOS export plugin) by default.
- **`build`** — new flag `--deeplink "<url>"` (same baking as `run`).

### Client runtime args (after `--`, or via deeplink)
- `--test-logging` / `?test-logging=true` *(new)* — logging self-test across every stack (Rust / GDScript / Swift / ObjC / Kotlin) to verify the unified channel + Sentry.
- `--dev-relay=host[:port]` *(new)* — iOS-only, **debug-only** localhost relay to your Mac.
- `--scene-inspector`, `--rust-log` already existed; `run` now injects them automatically on device deploys.

> The debug-WS / scene-inspector protocol verbs (`ping`/`focus`/`scenes`/`scene`/`entity`/`ui_*`/`avatars`/`avatar`/`app_ui`/`eval`) are **unchanged** — the channel just now also carries **logs + network**.

## Production safety (how this stays out of prod)

- **xtask (`src/`) is the dev build tool**, a crate separate from the shipped client (`lib/`). `debug-hub`, the log-server and deeplink-baking **never ship**.
- **debug-hub** keeps the command/eval channel on **loopback (9230)**; only the device-facing port (9231) listens on the LAN.
- **Debug WS server (9230)** auto-starts only on `OS.is_debug_build() && !is_production()`, bound to loopback; **`eval` is hard-gated by `is_production()`**.
- **iOS dev relay + Local Network Info.plist keys** are injected only into **debug exports** — release/TestFlight/App Store IPAs carry neither, so **no Local Network prompt in production**. The relay also needs an explicit `--dev-relay=` host (off by default even in debug).
- The streamed scene-inspector entries carry only `url`/`method`/`status` — **no headers/bodies** (no token leakage); logs/network are opt-in, and capture is inert while no consumer is connected.
- The baked deeplink (`build_config.gd`) is **gitignored and ignored in `-prod` builds**, so a stale file on the `clean:false` iOS self-hosted runner can't leak into a production IPA. Live `?scene-inspector=` deeplinks still work in prod (needed by content tooling).
- Per-component `[INIT]` boot logs sit behind a single flag per stack (`InitLog.ENABLED` / `INIT_COMPONENT_LOGS`), default on, with a TODO to disable.
- The scene-inspector perf snapshot only runs while a consumer is attached.

## Check
- [ ] No regression in Sentry logging for production/staging builds
- [ ] No overhead introduced for production/staging builds (perf snapshot gated by consumer; `[INIT]` logs flag-gated)
- [ ] Scene Inspector still working on Android and iOS
- [ ] MCP connection is seamless (`debug-hub` 9230 ↔ consumer)
- [ ] Release/TestFlight iOS IPA: **no Local Network** keys in Info.plist, no prompt on launch
- [ ] `-prod` build: debug WS off, `eval` returns "eval disabled in production builds", stale `build_config.gd` ignored
- [ ] `--dev-relay=host` is a no-op in release and tunnels on a debug iOS build


godot/src/global.gd
```diff
@@ -194,6 +200,87 @@ func is_gp_benchmark() -> bool:
...
+	if _selftest_done:
+		return
+	var requested := cli.test_logging
+	if not requested and deep_link_obj != null:
+		requested = str(deep_link_obj.params.get("test-logging", "")).to_lower() == "true"
+	if requested:
+		_selftest_done = true
+		_run_logging_selftest.call_deferred()
+
+
```
regenesis-claw, line 267: 🔴 This self-test is reachable in production via `decentraland://open?test-logging=true` — the deeplink param is checked without an `is_debug_build()` or `is_production()` guard. The test deliberately fires `push_error`, `tracing::error!`, `ERR_PRINT`, and `Log.e`, all of which ship to Sentry. Add `if not OS.is_debug_build(): return` at the top of this function to prevent accidental Sentry quota consumption in production.

docs/UNIFIED_DEBUG_CHANNEL_HANDOFF.md
```diff
@@ -0,0 +1,127 @@
+# Unified Debug Channel — Handoff
```
regenesis-claw: 🟡 This reads like internal session handoff notes ("context ran out", "nothing is committed — commit when ready", "how to continue"). It's useful during development but shouldn't ship as permanent documentation. Consider removing it — `docs/UNIFIED_DEBUG_CHANNEL.md` already covers the architecture comprehensively.

godot/src/tool/init_log.gd
```diff
@@ -0,0 +1,18 @@
...
+## Per-component "[INIT] <name>" boot trace, emitted as each autoload / subsystem
+## comes up so the wiring can be verified through the unified debug channel.
+##
+## Static (no instance) so it can be called before the `Global` autoload is ready
+## — `UiSounds` initializes before `Global` — and so the whole set toggles from a
+## single flag.
+##
+## TODO: set ENABLED = false once the boot-wiring verification is no longer needed.
+## These are info-level and currently print on every startup, production included.
+const ENABLED := true
```
regenesis-claw: 🟡 `ENABLED = true` with a TODO to disable means these `[INIT]` lines print on every production startup. They're `print` level (not Sentry), but they add noise. Before final merge, consider gating with `OS.is_debug_build()` or setting `ENABLED = false` for production.

regenesis-claw (commented):
Solid architectural work unifying the logging + scene-inspector + network inspector into a single protocol over one transport — the connect-out model is the right call for iOS, and the production safety design (connection-gated, opt-in capture, eval hard-gated) is well thought out. Reviewed against `REVIEW.md`.

---

### Pre-flight

- ✅ CI: all checks green (Linux, Android, iOS, static checks, clippy)
- ✅ iOS build: present and green — required since the PR touches `plugins/dcl-godot-ios/**`, `plugins/dcl-swift-lib/**`, and the iOS export plugin
- ⚠️ Draft status: PR is *not* marked draft, but author noted it's not final

---

### Findings

**🔴 `test_logging` not gated out of production — Sentry quota risk**

`_maybe_run_logging_selftest()` in `global.gd` is triggered by `--test-logging` CLI or `?test-logging=true` deeplink, with **no `is_debug_build()` or `is_production()` guard**. The self-test deliberately fires `push_error`, `tracing::error!`, `ERR_PRINT` (ObjC), and `Log.e` (Kotlin) — all of which ship to Sentry. A `decentraland://open?test-logging=true` deeplink in production would fire multiple Sentry error events.

Suggested fix: wrap the body of `_maybe_run_logging_selftest` with `if not OS.is_debug_build(): return` (or `is_production()` guard).

**🟡 `docs/UNIFIED_DEBUG_CHANNEL_HANDOFF.md` is internal session state**

This file is a Claude session handoff document ("context ran out", "nothing is committed", "how to continue"). It contains implementation notes useful during development but not reference documentation. Consider removing it from the PR — the companion `docs/UNIFIED_DEBUG_CHANNEL.md` already serves as the architectural reference.

**🟡 `[INIT]` logs print on every startup in production**

`InitLog.ENABLED = true` (GDScript) and `INIT_COMPONENT_LOGS = true` (Rust) emit `print`/`info!` for every component on every launch, including production. These are console-level (not Sentry), so not a quota issue, but they add noise to every user's log output. Both have a TODO to disable. Since the PR description says this isn't final — fine for now, but these should be behind `is_debug_build()` or disabled before merge to main.

**🟡 `DclLogHubLogger::log_message` — `log_error` intentionally not overridden**

The comment explains this is due to a gdext panic with `Array<Gd<ScriptBacktrace>>` on iOS. This means GDScript `push_error`/`printerr` won't flow through the log capture as separate error entries (they'll arrive via `log_message` without the error flag if Godot routes them there, or via `LogHubLayer` for Rust errors). Worth a follow-up issue to revisit once gdext fixes the panic — the unified channel would benefit from capturing GDScript errors with full metadata.

---

### Observations (non-blocking)

- The `debug!`-level `[HTTP]` traces scattered across auth, resource_provider, fetch, and http_queue_requester are well-placed — `debug!` never hits Sentry, and they'll be valuable for network debugging via the unified channel.
- Production safety looks solid: `_generated_deeplink()` returns empty in prod, `eval` is `is_production()`-gated, iOS local-network plist keys are debug-export-only, the perf snapshot is now correctly gated behind `is_consumer_connected()`.
- `EARLY_LOG_RING` uses `std::sync::Mutex` — consistent with the module's existing `std::sync` usage (OnceLock, atomics). The cold-path-only access pattern makes this fine.
- The refactor of `debug_ws_server.gd` into `run_command()` → shared backend is clean — one implementation, two transports, no duplication.
- The Android deeplink-based arg forwarding (replacing the sanitized `command_line_params`) is a good find — the old path was silently broken.

---

### Summary

One blocker: gate `test_logging` out of production to prevent Sentry quota incidents. Two suggestions: remove the handoff doc, and plan to disable `[INIT]` logs before final merge. Overall the architecture is sound and production safety is well-designed.

Requested by Mateo via Slack

kuruk-mm (approved):

regenesis-claw (approved):
Delta re-review (13 new commits since `7660316a`). Reviewed against `REVIEW.md`.

---

### Previous findings — status

| # | Finding | Status |
|---|---------|--------|
| 🔴 | `test_logging` not gated out of production | ✅ Fixed — `_maybe_run_logging_selftest()` now returns early on `is_production()` (commit `2d1d287`). Staging/TestFlight still run it. |
| 🟡 | `docs/UNIFIED_DEBUG_CHANNEL_HANDOFF.md` is session state | ✅ Removed (commit `2d1d287`) |
| 🟡 | `[INIT]` logs print on every production startup | ✅ Removed entirely — `InitLog` class, all `.emit()` call sites, and the Rust `INIT_COMPONENT_LOGS` loop all deleted (commit `2d1d287`). Clean approach: the `test_logging` self-test already covers unified-channel verification. |

All three review items fully addressed. 👍

---

### New changes

**Loopback DebugWs removal (`db9f5d7`)** — clean surgery. The TCPServer listener, peer management, `_handle_message`/`_reply`/`_send`, `start`/`stop`/`is_running`/`get_port`, the `--debug-ws` CLI flag, the Settings → Developer toggle + `.tscn` nodes, and the deeplink handler are all removed consistently. `DebugWsServer` now only holds `run_command()` + focus tracking — the shared backend the scene-inspector bridge calls. No orphan references found.

**iOS remote-debug baking (`b415f84`)** — the export plugin now injects `--remote-debug tcp://<host>:<port>` into `godot_cmdline` when the `REMOTE_DEBUG_FLAG` bit is set. Loopback/wildcard hosts (`127.0.0.1`, `localhost`, `0.0.0.0`, `*`) are rewritten to the machine's LAN IP via `_lan_ip()`. Sensible — the editor's `remote_host` is often set to `0.0.0.0` for the bind-all case, but the device needs a real address.

**Auto-dial hub from debug builds (`global.gd`)** — debug non-prod builds now default to `ws://127.0.0.1:9231` even with no `--scene-inspector` arg. The WS has exponential backoff reconnect (1s → 30s), and capture is still connection-gated, so this is safe. Good DX improvement — `cargo run -- debug-hub` in one terminal is enough to start inspecting.

**Godot engine SHA bump** — `6ddcadb64` → `36d9ae5fa` (remote-debug bind-all + multi-IP fix merged to 4.6.2 fork). The TEMP feature-branch detour (`9b06af7`) was reverted cleanly (`ca954e1`).

**Bug fixes (bundled):**
- `ftue_screen.gd`: null `contact_name` from the destinations API no longer crashes the carousel — properly handles null → hides the creator row instead of showing a raw wallet address. Good defensive fix.
- `nameplate_layer.gd` + `avatar.gd`: start nametag UI at `modulate.a = 0.0` so the stale placeholder doesn't flash; `_apply_nickname_visibility` also zeroes alpha on hide to prevent the 6-frame fade-out artifact.
- `avatar_shape.rs`: set `is_avatar_shape` before `add_child` so the nameplate gate applies from frame 1 — prevents the NPC tag flash.
- `virtual_joystick.gd`: regrab explorer focus on joystick press after dismissing a panel — fixes #2361 (avatar won't move after closing passport).
- `dcl_floating_islands_manager.rs`: destroy orphan islands that are no longer candidates (teleport / scene load), not just by distance — fixes lingering geometry.

All bug fixes are correct and well-commented.

**`unified.sh` race fix** — keeps stdin open with a `sleep` timeout so the WebSocket stays alive long enough for on-device round-trips (loopback was sub-ms but USB/LAN is not). Fallback error message when no reply arrives. Good.

---

### CI

⚠️ iOS build is still in progress at time of review. Linux/Android/static checks from the previous push were all green. The PR touches iOS native plugins so a green iOS build is required before merge — hold for that.

---

### Summary

All previous review items addressed. New changes are clean — the DebugWs loopback removal simplifies the codebase without losing functionality, the bug fixes are solid, and the engine SHA bump is properly pinned. Approving — merge once the iOS build goes green.

Requested by Lean via Slack

PR #2364, opened 2026-06-26
feat(versioning): converge marketing version to SemVer 1.10.1

leanmendoza:
## What

Converges the **marketing version** to a single SemVer and weaves the **store build number** into the baked version string. Two independent version axes, fully decoupled.

> ℹ️ The branch name stays `feat/marketing-version-1.10.0`, but the shipped marketing version is now **1.10.1**.

### Marketing version → `1.10.1`
- Source of truth: `lib/Cargo.toml` `version = "1.10.1"` (+ `Cargo.lock`).
- Mirrored to Android `version/name` and iOS `application/short_version` in `export_presets.cfg`.
- `version_check.rs` enforces the **marketing trinity** — `Cargo.toml == Android version/name == iOS short_version` — replacing the old `Cargo-minor == versionCode == application/version` coupling (gone now that the store build number is allocated independently).
- `bump_version.yml` is a marketing-only `major|minor|patch` bump (no longer touches `version/code`).

### Store build number → woven into the version string
The store build number (Android `versionCode` / iOS `CFBundleVersion`) is a separate, opaque axis allocated per commit from the Cloudflare Worker (shipped in #2365). This PR bakes it into `GODOT_EXPLORER_VERSION`:
- `build.rs` reads `DCL_BUILD_NUMBER` and inserts it as a 4th version segment → `{version}.{build}-{hash}{mode}-{env}` (e.g. `1.10.1.1007-abc1234-prod`). Unset on local/fork builds → the segment is omitted (`1.10.1-abc1234-dev`).
- The `compute-build-number` CI step now runs **before** the lib build in `android_builds.yml` and `ios_r2_artifact.yml`, so `DCL_BUILD_NUMBER` is set when `build.rs` runs. On iOS the numbered string also flows to `godot-asc-deploy` via the "Capture build version" step.
- The force-update gate (`version_gate.gd`) parses only `major.minor.patch`, so the extra `.{build}` segment is ignored.

### Placeholders
`version/code` / `application/version` in `export_presets.cfg` are set to a clear placeholder `1` (overwritten at export from `DCL_BUILD_NUMBER`; the committed value never ships), with an inline comment.

## Verified
- `cargo run -- version-check` → `Cargo.toml=1.10.1, Android version/name=1.10.1, iOS short_version=1.10.1`.
- Local build with `DCL_BUILD_NUMBER=12345 DECENTRALAND_STAGING_BUILD=1` bakes `1.10.1.12345-<sha>-staging`; unset bakes `1.10.1-<sha>-dev`.


kuruk-mm (approved):
looks great!

PR #2365, opened 2026-06-26
feat(ci): store build number from the Cloudflare Worker allocator

leanmendoza:
## What

The **store build number** (Android `versionCode` / iOS `CFBundleVersion`) is now allocated from a small **Cloudflare Worker + Durable Object**, **idempotent per commit SHA**, so Android and iOS building the same commit converge on the same number.

- At export time, `export.rs` stamps `DCL_BUILD_NUMBER` into `version/code` / `application/version` inside its backup→restore window (the committed file keeps placeholders).
- The `compute-build-number` action `curl`s the allocator with `Authorization: Bearer ${{ secrets.BUILD_NUMBER_SECRET }}` and the commit SHA, writing the result to `DCL_BUILD_NUMBER`. Wired into `android_builds` (APK + AAB) and `ios_r2_artifact`; the secret is plumbed via `runner.yml`.
- **No clock fallback**: when no `--build-number` / `DCL_BUILD_NUMBER` is present (fork PRs / local builds), the committed placeholder `versionCode` / `CFBundleVersion` is kept as-is — those builds are never shipped to a store, so no number is invented.
- iOS forwards the number to `godot-asc-deploy` so the signed TestFlight build keeps the same number.

### Why a Worker (not a date formula, not the BFF)

- Android builds on push (`runner.yml`), iOS later (`mobile_distribute`) — different CI runs that never share a run number, so the commit **SHA** is the only key that makes them converge exactly. The Durable Object serializes allocation → atomic & strictly monotonic, no races.
- Self-deployed via `wrangler`, no Postgres, no ops/SSM; decoupled from the client-facing BFF.

### Endpoint: workers.dev (not the custom domain)

CI hits the Worker's **`*.workers.dev`** route. The custom domain (`build-numbers.dclregenesislabs.xyz`) sits behind the zone's **Bot Fight Mode**, which edge-blocks the datacenter CI runner IPs with a `403` before the request ever reaches the Worker. `workers.dev` is Cloudflare-managed and not bot-protected. The custom domain stays live for humans / non-datacenter IPs.

## Security (independent review baked in)

- **Fail-closed**: the allocator returns `503` if no secret is configured — no fail-open window.
- The bearer secret is `::add-mask::`-ed so it can't surface in CI logs.
- The payload is built with `printf` and parsed with `grep`/`tr` (no `jq` — the Android build container doesn't ship it).
- The public read-only `GET /build-numbers` exposes only opaque `sha → number` bindings.

## Backend

The allocator Worker is **already deployed** (DCL Regenesis Labs Cloudflare account) and verified live; the `BUILD_NUMBER_SECRET` Actions secret is set. It's infra with no app repo, so there's no backend PR.

## ⚠️ Merge order

Merge **decentraland/godot-asc-deploy#1** (honor the forwarded `build_number`) **first** — the iOS dispatch sends `-f build_number=`, which must already exist on its `main` or `gh workflow run` 422s. Then this PR (Android converges immediately; full iOS convergence needs the asc-deploy PR).

Independent of the marketing-version PR **#2364** (different files — either order).


kuruk-mm (approved):
great!

PR #2369, opened 2026-06-26
fix(mobile): allow SceneUI interaction during joystick/camera, and swipe-off handoff

### LINKED ISSUE #2172: [Bug] Can't interact with Scene or Controls UI when moving with joystick (lacks multitouch, only works for Jump/Glide)
**Issue:**
When the player is moving using the joystick, they cannot interact with the scene or controls UI elements. The system lacks proper multitouch support - only Jump and Glide buttons work while moving.

**Expected Behavior:**
Players should be able to use the movement joystick and interact with UI elements or scene objects simultaneously (multitouch support).

**Current Behavior:**
- Movement joystick blocks all other touch interactions except Jump/Glide buttons
- Cannot click on scene objects while moving
- Cannot interact with controls UI while moving

Requested by Rodri via Slack

EibrielInv:
Closes #2172

  ## What

  On mobile, makes Decentraland SDK **SceneUI** usable while moving, and lets a touch flow from a UI element straight into camera/joystick control.

  1. **Interact with SceneUI while using the joystick or camera-swipe.** `DclUiControl` now handles real `InputEventScreenTouch` on mobile instead of relying on single-pointer mouse-emulation, so a second
  finger on scene UI registers while another finger drives movement/look.
  2. **Swipe-off handoff.** Pressing an interactive SceneUI element and swiping off it fires `PetUp` at the element boundary, then continues the gesture without lifting the finger — routed by the press
  position: joystick zone → movement (base seeded at the press point), otherwise → camera rotation.

  ## How

  - `DclUiControl` (Rust): platform-gated touch handling; detects drag breakout from the element rect, emits `PetUp`, and forwards the adopted gesture (per touch index) to `MobileCameraInput`.
  - `MobileCameraInput`: `adopt_touch` / `update_adopted_touch` / `release_adopted_touch` with zone routing; adopted gestures keyed per finger; `_input` safety-net ends a gesture when its finger lifts even if 
  the source control was freed mid-gesture (prevents stuck movement). Reuses new `Player.apply_look_delta`.
  - `VirtualJoystick`: `external_begin` / `external_update` / `external_end` + `get_active_area_global_rect` to drive the stick from an arbitrary screen point.

  ## How to test

  Run on a touchscreen device (or desktop with `--force-mobile`, which routes mouse through touch emulation). Load a scene with an interactive SceneUI element that listens for pointer down/up (e.g. an SDK7 UI 
  button).

  1. **SceneUI works while moving** — hold the joystick with one finger, tap the UI button with another → the button registers. Repeat while drag-rotating the camera with the first finger.
  2. **Tap (no regression)** — tap the button and release on it → fires pointer down then up; no camera/joystick movement.
  3. **Swipe-off → camera** — press the button, drag off it toward the right/center (camera zone) and keep holding → pointer-up fires when leaving the element and the camera rotates, smoothly, without lifting 
  the finger.
  4. **Swipe-off → joystick** — press a button in the left (joystick) zone, drag off it staying on the left → the joystick engages with its base at the press point and the avatar moves.
  5. **No stuck movement** — start a swipe-off into the joystick zone, then let the scene hide/rebuild that UI (or lift after the element disappears) → movement stops when the finger lifts; the avatar never 
  keeps walking with no finger down.
  6. **Desktop regression** — on a normal desktop build, mouse clicks on SceneUI still fire pointer down/up.


regenesis-claw (approved):
Clean, well-structured PR. The three-layer architecture (Rust touch handling → GDScript handoff router → camera/joystick consumer) is nicely separated, and the per-finger `_adopted` dictionary + `_input` safety net in MobileCameraInput is a solid defense against stuck gestures when a scene UI control is freed mid-drag.

**Reviewed against REVIEW.md.**

### Pre-flight
- **Branch**: 2 commits behind main (CI fix + protocol tracking) — unrelated to touch handling, not blocking.
- **iOS build**: ✅ Present (build 1012, TestFlight triggered).
- **CI**: ✅ All green (static checks, clippy, Linux, Android, iOS).

### Findings

No blockers. A few notes:

1. **`pressed_touch_index` overwrite on multi-press (minor):** If two fingers press the same DclUiControl simultaneously, `handle_touch` overwrites `pressed_touch_index` with the second finger's index — the first finger's adopted gesture won't be released through this control. The `_input` safety net in `mobile_camera_input.gd` catches the orphaned finger-up, so this is safe in practice (and two-finger press on one small UI button is an extremely unlikely user gesture). Just documenting it's understood.

2. **`apply_look_delta` coupling (fine):** `Player.apply_look_delta` references `MobileCameraInput.HORIZONTAL_SENS` / `VERTICAL_SENS`, tying Player to MobileCameraInput's constants. This is the right call — the sensitivity *must* match between normal touch and adopted gestures — but worth noting if sensitivity ever becomes per-mode or user-configurable.

3. **`DclGlobal.is_mobile` in Rust (fine):** REVIEW.md convention (`OS.get_name()`) applies to GDScript. Using the cached Rust-side `is_mobile` flag in `dcl_ui_control.rs` is correct — avoids a cross-language call on a hot input path.

### Positives
- `push_pointer_result` helper eliminates the duplicated pointer event construction — clean refactor.
- `EXTERNAL_TOUCH_INDEX = -2` sentinel avoids collision with both real indices (≥0) and the default `-1` — no ambiguity.
- The swipe-off zone routing (joystick vs camera) based on `start_position` is intuitive and matches user expectation.
- PR description and test plan are thorough — every edge case (stuck movement, desktop regression) is covered.

Approving.

kuruk-mm (approved):
looks great

PR #2412, opened 2026-07-01
Release 1.10.1: promote main → release

kuruk-mm:
Promotes **all of `main`** into `release` for the **1.10.1** release.

This branch is now **identical to `main`** (`git diff main...release-1.10.1` is empty). The three hotfixes that were previously cherry-picked on top — `floating-islands`, `nametags`, `ftue` — have since landed in `main` (via #2413, which also brought `fix(mobile): regrab explorer focus on joystick press`), so there is **nothing extra on top** and no commit is listed twice.

- Previous release: `0.70.0` (marketing v70 / iOS 1.8)
- This release: **`1.10.1`** — marketing version converged to SemVer (#2364)
- ~121 commits since the last release.

Supersedes #2328.

---

### Versioning
- Marketing version converged to **SemVer 1.10.1** (#2364); internal Android/iOS bumps (v67→v72) folded in.

### Scene UI & SDK
- Scene UI **scroll (`UiScroll`)** with touch drag + Unity-parity layout (#2119)
- **Border** properties for Scene UI (#2049)
- **Highlight** effect on objects via pointer-events `show_highlight` (#2307)
- Populate `screen_inset_area` on `PbUiCanvasInformation` (#2148)
- **TextShape** rendered via enhanced `Label3D` (#2315); `font_auto_size` matches the Unity client (#2178)
- Resolve `@dcl/protocol` from `@next` via the npm registry (#2151)

### Avatars
- `AvatarAttach` extended to all **26 anchor points** (#2158)
- `AMT_HIDE_NAMETAGS` modifier on `AvatarModifierArea` (#2079)
- Apply `PBPhysicsCombinedImpulse/Force` to the player (#2140)
- `CL_MAIN_PLAYER` collision layer to distinguish main player from remotes (#2234)
- Fixes: `excludeIds` reliably excludes remote avatars (#2168); `HIDE_AVATARS` no longer blocks passport interaction (#2096); nameplates rendered in a 2D layer instead of per-avatar viewports (#2224); camera-distance fade, physics-only occlusion, pending-profile nameplates (#2321)

### Mobile & input
- Unified joystick input via `gui_input` (#2083)
- Reworked Hide UI with granular sub-toggles (#2055) + Hide Scene Interface toggle & improved controls while hidden (#2253)
- SceneUI interaction allowed during joystick/camera + swipe-off handoff (#2369)
[...]

PR #2413, opened 2026-07-01
Forward-port 1.10.1 hotfixes to main

kuruk-mm:
Forward-ports the hotfixes from the **1.10.1** release onto `main`, so they aren't lost in the next release cycle. These currently live only on the release line — `main` does not have them.

Commits (same SHAs as on the release branch — no re-duplication):
- `0dbf2240da` fix(floating-islands): destroy island tiles that leave the candidate set
- `36376eadc0` fix(nametags): stop NPC# 2D nameplate flash on AvatarShape spawn
- `61511b86c3` fix(ftue): hide creator row when place has no contact_name
- `06f393a3bb` fix(mobile): regrab explorer focus on joystick press

Companion to #2412 (`release-1.10.1` → `release`).

> Note: `06f393a3bb` (mobile focus) is currently on this branch and heading to `main`, but is **not** on `release-1.10.1` yet — add it there too if the release should ship it.


leanmendoza (approved):
remember this to be a merge commit :)

PR #2417, opened 2026-07-01
feat: add account deletion guest

leanmendoza:


leanmendoza:
This change lives in https://github.com/decentraland/godot-explorer/pull/2421

PR #2423, opened 2026-07-01
ci: map any release* branch to the prod version suffix

leanmendoza:
## What

Make **any branch whose name starts with `release`** (e.g. `release`, `release-1.10`, `release/1.10`) bake the `-prod` environment suffix into the version string — not just the exact branch named `release`.

## Why

The branch→env decision lives in two places, both of which used an **exact** string match against `release`:

1. **`.github/actions/set-prod-flag/action.yml`** — the single source used by every platform build (windows/macos/linux/android/ios). It sets `PROD_FLAG=--prod|--staging`, which `src/main.rs` turns into `DECENTRALAND_PROD_BUILD`, which `lib/build.rs` bakes as the `-prod`/`-staging`/`-dev` suffix in `GODOT_EXPLORER_VERSION`.
2. **`.github/workflows/mobile_distribute.yml`** — reconstructs `build_version` deterministically in the iOS reuse path (must mirror `lib/build.rs`).

Because both only matched `release` exactly, a release cut like `release-1.10` would silently produce `-staging`.

## Change

Swap the exact comparison for a `release*` prefix match in both spots, accepting both the `refs/heads/<branch>` and bare `<branch>` forms.

- `set-prod-flag`: `if [ "$REF" = "release" ]…` → `case "$REF" in refs/heads/release* | release*)`
- `mobile_distribute.yml`: `if [ "$REF" = "release" ]…` → `case "$REF" in release*)`

The xtask (`src/main.rs`) and `lib/build.rs` need no change — they already just honor whatever `--prod`/`--staging` flag CI passes.

## Out of scope (left as exact `release` match on purpose)

These gate **store/registry publishing**, not the version suffix, so they were intentionally left untouched:
- `android_builds.yml` / `mobile_distribute.yml` Play Store AAB upload (`ref == 'main' || ref == 'release'`)
- `docker_build_test.yml` Docker `latest` tag/publish

### DIFF
--- a/.github/actions/set-prod-flag/action.yml
+++ b/.github/actions/set-prod-flag/action.yml
@@ -1,9 +1,9 @@
 name: Set Prod Flag
-description: Set PROD_FLAG environment variable based on the current branch (--prod for release, --staging for main and other branches)
+description: Set PROD_FLAG environment variable based on the current branch (--prod for any release* branch, --staging for main and other branches)
 
 inputs:
   ref:
-    description: 'Override ref to check (defaults to github.ref). Accepts both refs/heads/release and bare release formats.'
+    description: 'Override ref to check (defaults to github.ref). Accepts both refs/heads/release* and bare release* formats.'
     required: false
     default: ''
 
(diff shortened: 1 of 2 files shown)


regenesis-claw (approved):
Clean, well-scoped CI fix. The `case` pattern matching is correct for both `refs/heads/release*` and bare `release*` forms. Good call leaving the store/registry publishing gates (`android_builds.yml`, `docker_build_test.yml`) on exact `release` — those should stay intentional.

LGTM — approving.

kuruk-mm (approved):
looks great

PR #2428, opened 2026-07-02
feat(mobile): graceful memory management with low-memory warning (#2002)

### LINKED ISSUE #2002: Implement graceful memory management: kill scenes and show modal instead of crashing app
## Problem

Currently, when the mobile app exceeds available RAM, it crashes abruptly, resulting in a poor user experience. The app simply closes without any warning or explanation.

## Proposed Solution

Instead of crashing the entire application, implement a resource tracking system that:

1. **Monitors memory usage** across scenes
2. **Proactively kills scenes** when memory threshold is exceeded (before OS kills the app)
3. **Shows a user-friendly modal** explaining the situation (reuse the scene crash modal design from #1566 )

This provides a graceful degradation experience — users understand what happened instead of experiencing an unexpected crash.

## Implementation Notes

- Resource tracking mechanism needs to monitor RAM usage in real-time
- Define memory thresholds that trigger scene cleanup
- Implement scene killing logic that frees resources properly
- Reuse the cross-platform disconnection/crash modal design by Eli (#1566 )
- Consider showing which scene caused the issue and offer options (retry, go back, etc.)

## Related Issues

#1566 

## Context

This has been a long-standing pain point that was partially considered during the asset loading refactor but never fully prioritized. While crash frequency has decreased, graceful failure handling is still needed for a production-quality mobile experience.

---
Requested by Rodri via Slack  
Proposed by leanmendoza

leanmendoza:
Closes #2002 (feature half — see companion engine PR for the freeze fix).

## What & why

On memory-constrained iOS devices the client was being **jetsam-killed** by the OS with no warning once its physical footprint crossed the platform ceiling. This adds graceful memory management: a background monitor, scene eviction under pressure, and a user-facing modal as a last resort.

## Changes

- **`lib/src/tools/memory_monitor.rs`** (new): a background OS thread that samples memory off the main thread and publishes to atomics.
  - iOS: `os_proc_available_memory()` for headroom + `proc_pid_rusage(RUSAGE_INFO_V2).ri_phys_footprint` for the real footprint the OS jetsams on.
  - Android: `/proc` based available memory.
  - Derives a **footprint ceiling** (50% of device RAM on iOS) and a pressure `level` (WARNING / CRITICAL) from both the available-memory thresholds and the footprint hard-stop.
- **`scene_manager`**: on CRITICAL pressure, evicts the **farthest non-current, non-global** scene via `kill_scene`; if there is nothing safe to evict, emits a new `low_memory_warning` signal. Also bumps a main-thread heartbeat each physics frame (used to characterize stalls).
- **`modal_manager` + `scene_fetcher`**: wire `low_memory_warning` to a modal offering **Continue anyway** (opts into verbose memory logging) or **Back to Discover**.

## Validation

Deployed to a physical **iPhone 13**. In Genesis Plaza the footprint climbs to the ceiling (`footprint≈1939MB ≥ ceiling 1831MB`), the monitor raises `level=CRITICAL`, and the low-memory modal is surfaced — exactly as designed. Host `cargo fmt` / `clippy`, `gdformat` and `gdlint` are clean.

## Companion PR

The underlying main-thread **freeze** (a pipeline-compilation deadlock, separate from OOM) is fixed in the engine fork: decentraland/godotengine#17. Both are needed for a fully graceful heavy-scene load on iOS.

### DIFF
--- a/lib/src/tools/memory_monitor.rs
+++ b/lib/src/tools/memory_monitor.rs
@@ -1,0 +1,351 @@
+//! Background memory-pressure monitor (issue #2002).
+//!
+//! Why a dedicated OS thread: on mobile the app is killed abruptly when it
+//! exceeds its per-process memory limit (iOS jetsam / Android lowmemorykiller).
+//! The dangerous moment — loading a heavy scene — is exactly when the *main
+//! thread freezes* (a big synchronous upload blocks the frame, or the kernel
+//! compresses/pages memory). A watchdog living on the SceneTree / main loop
+//! cannot react while the main thread is frozen.
+//!
+//! This monitor runs on its own thread, sampling memory every `POLL_MS`
+//! independently of the main thread, so detection keeps working even during a
+//! freeze. It publishes the result into atomics that the main loop
+//! (`SceneManager::handle_memory_pressure`) consumes to evict scenes / show the
+//! low-memory modal the instant it gets a slice of CPU.
+//!
+//! ## Memory signal (two metrics, deliberately)
+//!
+//! * **Primary — headroom:** how many MB the app can still allocate before the
+//!   OS kills it.
+//!   - iOS: `os_proc_available_memory()` — bytes left before jetsam. Cheap,
+//!     off-thread-safe, and already reflects the *dynamic* per-device limit.
+//!   - Android: heuristic `budget(MemTotal) - RSS` from `/proc` (no precise
+//!     per-app headroom call exists, and JNI is not thread-safe to call here;
+//!     `/proc` files are readable from any thread).
+//!
+//! * **Secondary hard-stop — `phys_footprint`** (iOS only, via
+//!   `proc_pid_rusage`). This is the value jetsam actually compares against and,
+//!   unlike `os_proc_available_memory`, it *includes* GPU / IOSurface / texture
+//!   memory. On GPU-heavy scenes `os_proc_available_memory` was measured to
+//!   over-report headroom by ~180 MB on an iPhone 13 (that GPU memory is
+//!   invisible to it). So we also read `phys_footprint` and force CRITICAL once
+//!   it crosses a device-scaled ceiling — catching the GPU-driven OOM that the
+//!   headroom signal alone would miss.
+//!
+//! On desktop there is no OOM killer to anticipate, so the monitor is not
+//! started and both readers return -1.
+
+use std::sync::atomic::{AtomicBool, AtomicI32, AtomicU32, AtomicU8, Ordering};
+
+/// Available headroom in MB (bytes the app can still allocate before the OS
+/// kills it). -1 when unknown / not yet sampled.
+pub static AVAILABLE_MB: AtomicI32 = AtomicI32::new(-1);
+
+/// `phys_footprint` in MB — the real total the OS charges the process, GPU
+/// memory included (the value jetsam compares against). iOS only; -1 elsewhere
+/// or before the first sample.
+pub static FOOTPRINT_MB: AtomicI32 = AtomicI32::new(-1);
+
+/// Pressure level published by the monitor thread: 0 = OK, 1 = WARNING, 2 = CRITICAL.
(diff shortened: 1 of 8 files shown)


lib/src/tools/memory_monitor.rs
```diff
@@ -0,0 +1,351 @@
...
+}
+
+#[cfg(target_os = "ios")]
+fn read_footprint_mb() -> i32 {
+    // `phys_footprint` via proc_pid_rusage(RUSAGE_INFO_V2) — the same value Xcode
+    // and jetsam use, GPU/IOSurface memory included. Pure syscall, off-thread
+    // safe.
+    let mut info = std::mem::MaybeUninit::<libc::rusage_info_v2>::zeroed();
+    // SAFETY: proc_pid_rusage fills a v2 rusage struct for our own pid; the
+    // buffer is a correctly-sized, zeroed rusage_info_v2.
```
regenesis-claw, line 283: Nit (non-blocking): `page_size` is hardcoded to 4096, but newer ARM64 Android kernels (Android 15+) can use 16KB pages. On such a device RSS would read 4x too low, making the budget look generous when it isn't. `libc::sysconf(libc::_SC_PAGESIZE)` would be trivially more accurate and is safe to call from any thread. Follow-up material.

regenesis-claw (approved):
Solid work. Well-documented, well-structured, and the dual-metric memory approach (headroom + footprint) is exactly right for catching GPU-driven OOM that os_proc_available_memory alone would miss.

Pre-flight:
- Branch is 2 commits behind main, acceptable, no conflicting changes detected.
- iOS build: present and green (TestFlight dispatch + R2 export).
- No submodule pointer drift.

Followed REVIEW.md.

What I reviewed:
- memory_monitor.rs (351 lines, the new core): background OS thread design, platform readers (iOS os_proc_available_memory + proc_pid_rusage, Android /proc), pressure-level derivation, main-thread stall detection, logging.
- scene_manager.rs: handle_memory_pressure() eviction policy, settle-frame guard, signal flow.
- modal_manager.gd + scene_fetcher.gd: GDScript wiring of the low-memory modal.
- consts.rs / export.rs / install_dependency.rs: temporary GODOT_USE_BRANCH engine-fork override.
- Companion engine PR godotengine#17: inline pipeline compilation fix.

Observations (non-blocking):

1. Android page_size hardcoded to 4096 (read_available_mb): most Android devices use 4KB pages, but newer ARM64 kernels (Android 15+) can use 16KB pages. Since this is already a heuristic (ANDROID_BUDGET_FRACTION), the practical impact is small, but the RSS reading would be 4x too low on a 16KB-page device, making the budget look generous when it isn't. A libc::sysconf(libc::_SC_PAGESIZE) call would be trivially more accurate. Follow-up material, not a blocker.

2. GODOT_USE_BRANCH reminder: the comment is clear about resetting to None once godotengine#17 merges and GODOT_BUILD_SHA is bumped. Just flagging for visibility, leaving it pinned after merge would silently pull an unmerged engine build for every dev/CI run.

3. Monitor thread never joins: the std::thread::spawn in start() has no join handle stored. Acceptable for a process-lifetime watchdog, just noting it does not participate in graceful shutdown (not that Godot has one on mobile jetsam).

Positives:
- The dual stdout+stderr emit_log that survives a main-thread freeze is a smart diagnostic choice, exactly the logging that matters when the app is about to die.
- Settle-frame guard (90 frames / ~1.5s) prevents thrashing after an eviction.
- The stall detector is a bonus: it does not fix the freeze but makes it observable in device logs, invaluable for triage.
- Clean separation: Rust publishes to atomics, GDScript consumes via signals, no cross-language locking.
- The engine PR (godotengine#17) correctly releases local_mutex before running the creation function, so dependent pipelines can safely re-enter the map.

Approving both PRs. The companion engine fix is already approved by kuruk-mm.

Requested by Lean via Slack

kuruk-mm (approved):
Works great!

Issue #2425, opened 2026-07-02
iOS/Android store build number can go backwards → TestFlight rejects the upload (allocator is store-blind)


leanmendoza:
## Problem

The store build number can go **backwards**, and TestFlight/Play reject the upload.

Store versions (`CFBundleVersion` / `versionCode`) must be **strictly increasing**. Our build number comes from the `godot-build-number-allocator` Worker, which is **idempotent per commit SHA** (so iOS and Android converge on the same number and rebuilds are reproducible) — but it has **no knowledge of what has actually been shipped to a store**. So an out-of-order deploy (an older commit re-deployed, or a reused export shipped after a newer build already landed) gets a number **≤ the current store max** and is rejected.

### Incident

godot-asc-deploy run [28581987863](https://github.com/decentraland/godot-asc-deploy/actions/runs/28581987863): stamped `CFBundleVersion 1077`, but TestFlight was already at `1082` (uploaded ~9h earlier). Apple rejected it:

```
altool -19232 / ENTITY_ERROR.ATTRIBUTE.INVALID.DUPLICATE
The bundle version must be higher than the previously uploaded version: '1082'.
```

`1077` was never actually used — the real problem is it's **below** the current max. Retrying does nothing: the allocator is idempotent, so it hands back `1077` every time.

## Root cause

Three requirements are in tension:

- **I1 — Coincidence:** a SHA maps to the same number on iOS **and** Android.
- **I2 — Reproducibility:** rebuilding a SHA yields the same number.
- **I3 — Store monotonicity:** each upload's version must be **strictly greater** than the last one a store accepted.

The allocator satisfies I1 + I2 but is blind to I3. The fix makes **I2 conditional**: reproducible *until a store ships past it*, then re-allocated above the store high-water mark (which forces a rebuild, so I1 still holds).

## Design

Close the loop — the deploy pipeline **reports every accepted upload back to the allocator**, and allocation becomes **store-aware**.

**1. Allocator (`dcl-regenesislabs/godot-build-number-allocator`)**
- New `POST /build-numbers/published {sha, buildNumber, platform}` → raises a persisted `publishedHighWater`.
- `allocate()` reuses the SHA's number only while `buildNumber >= publishedHighWater` (`>=` so the SHA that *set* the mark keeps its number → I1); otherwise re-allocates above the mark.
- `publishedHighWater` defaults to `0`, so **deploying this alone is a no-op** until reporting starts — this is what makes the rollout safe.

[...]

PR #2436, opened 2026-07-03
ci: unify iOS/Android mobile distribution behind one store-aware build number

leanmendoza:
## What

Unify the iOS and Android mobile-distribution legs behind **one** store-aware build number, so the two platforms can never diverge on the same commit.

`prepare` now allocates the build number **once**; each leg resolves reuse-vs-rebuild against it with a shared guard. Android gains the resolve/rebuild path it was missing — until now it only surfaced the push-built APK, so when iOS refreshed a superseded number (#2429) Android kept its old `versionCode` and the two diverged. The full baked version string is also shown on the Slack card from the start.

## Changes

- **new** `check-build-number-current` composite — shared sidecar-vs-N guard used by both legs
- `android_builds.yml` — `ref`/`sha`/`concurrency_key` inputs (on-demand rebuild) + `build_number.txt` sidecar
- `mobile_distribute.yml` — single allocation in `prepare`; mirrored `resolve → build | reuse` on both legs
- `compute-build-number` — retry the allocator with exponential backoff (2→4→8→16s) and surface its error body

## Allocator resilience (de paso)

The 2026-07-03 Cloudflare Durable Objects incident (ENAM) made the build-number allocator flap `500 object reset` bursts, failing both mobile builds on #2440 — and `curl -fsS` swallowed the body, so the CI log only showed `last response: <empty>`. This drops `-f` to capture the Worker's real error, gates on the HTTP status, and retries transient failures (5xx/429/connect-fail) with backoff while failing fast on non-retryable 4xx. Idempotent per SHA → safe to retry. Fixes all three callers of the composite (mobile `prepare`, `android_builds`, `ios_r2_artifact`) at once.

The primary self-heal (in-Worker retry with a fresh DO stub) lives in the separate `godot-build-number-allocator` repo and is tracked as a follow-up.

## Behavior

| case | iOS | Android |
|---|---|---|
| number current | reuse | reuse |
| superseded on a store | rebuild w/ new N | rebuild w/ new N |
| artifact missing / build failed | rebuild | rebuild *(new: was fail-only)* |

Backward-compatible: fork PRs (no allocator secret) skip the guard; the cron no-op is unchanged. The allocator and `godot-asc-deploy` are untouched.

Follow-up of #2425.

### DIFF
--- a/.github/actions/check-build-number-current/action.yml
+++ b/.github/actions/check-build-number-current/action.yml
@@ -1,0 +1,62 @@
+name: Check build number is current
+description: >
+  Decide whether an artifact already in R2 is safe to REUSE, by comparing the build number it was
+  BAKED with (an R2 `build_number.txt` sidecar) against the authoritative number allocated for this
+  commit (passed in as `expected-number`). Shared by the iOS and Android resolve legs so both make
+  the identical reuse-vs-rebuild decision. Outputs `current=true` only when the sidecar exists and
+  equals the expected number; otherwise `current=false` (missing sidecar or superseded on a store →
+  rebuild). When `expected-number` is empty (fork PR / allocator skipped, no authoritative number to
+  compare against) it is a no-op and returns `current=true` — nothing is shipped to a store in that
+  case anyway. Reads AWS credentials from the ambient job environment (AWS_ACCESS_KEY_ID / _SECRET),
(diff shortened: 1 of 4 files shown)


kuruk-mm (approved):
looks great!

PR #2438, opened 2026-07-03
ci: distribute Android APK to Firebase App Distribution

leanmendoza:
## What

Adds a **Firebase App Distribution** step to the Android leg of `mobile_distribute.yml` (`android-notify` job): once the signed APK for a commit is in R2, it's pushed to Firebase App Distribution testers — the Android counterpart of the existing iOS→TestFlight leg.

- Free (Firebase Spark plan), no app changes — distributes the already-signed APK.
- Keyless auth (Workload Identity Federation, no service-account keys).
- Gated so it's a clean no-op until configured — can't break existing CI.
- Runs only on intentional distributions (`build` label / release push / main cron / dispatch), never on every PR.

## How to trigger

Add the **`build`** label to a PR (or release push / main cron / dispatch). Note: `build` also triggers the iOS TestFlight leg (existing behavior).


kuruk-mm (approved):
looks great!

PR #2440, opened 2026-07-03
feat(analytics): per-event timestamps, messageId, and send retry for Segment

leanmendoza:
## Why

Investigating a `SCREEN_VIEWED` export (renderer `1.10.1.1074-…-staging`) surfaced that our Segment events carry **no client-side time at all**:

- `Original Timestamp` was `null` in **148/148** rows; the canonical `Timestamp` equalled `Received At` in **148/148** — i.e. we were recording *when Segment received the batch*, not when the screen was viewed.
- Events flushed together collapsed onto one instant: **62%** of events shared a millisecond with another event in the same session (biggest tie cluster: **11** events), so per-event chronology within a session was unrecoverable.
- We sent **no `sentAt`** (so Segment couldn't correct device clock skew) and **no `messageId`** (so Segment couldn't deduplicate re-sends — a documented requirement for the self-managed HTTP API).
- On a failed `POST` the batch was **silently dropped** (no retry, no buffering) — real data loss on flaky mobile networks.

Root cause: events were serialized at **flush** time, not at creation, and the payload never included any of Segment's timing/idempotency fields.

## What changed

`lib/src/analytics/{metrics.rs,data_definition.rs}`:

1. **Per-event `timestamp` + `messageId`, captured at queue time.** Every emit path now goes through `Metrics::queue_event`, which stamps `Utc::now()` and a UUID v4 the moment the event is created — not at flush — so a batch no longer shares a single timestamp. `messageId` makes re-sends deduplicable (Segment stores it stored as `originalTimestamp`, dedup window ≥24h / ~170d in practice).
2. **Batch-level `sentAt`**, stamped at the actual moment of sending, so Segment corrects device clock skew: `timestamp = receivedAt - (sentAt - originalTimestamp)`.
3. **`client_timestamp` custom property** (next to `renderer_version`) — the ground-truth chronology that no skew correction or warehouse mapping can rewrite.
4. **Retry / offline buffer.** On a send failure the batch is buffered and retried on the next flush (cap 1000, drop-oldest) instead of dropped. HTTP 4xx/5xx arrive as `Ok` and are *not* retried, so a permanently-rejected payload can't spin forever. Re-sends keep their original `timestamp`/`messageId`.

Not changed by design: no `context` object (would require a data-model change, out of scope). The retry buffer is in-memory (survives network blips during a session, not an app kill).

## Resulting payload

```jsonc
{
  "writeKey": "…",
  "sentAt": "2026-07-02T13:38:10.512Z",            // batch, moment of send
  "batch": [{
    "type": "track",
    "event": "Screen Viewed",
    "userId": "…",
    "messageId": "3f2a…-uuid",                       // dedup + retry idempotency
    "timestamp": "2026-07-02T13:38:08.243Z",         // creation → originalTimestamp
    "properties": {
      "renderer_version": "1.10.1…",
      "client_timestamp": "2026-07-02T13:38:08.243Z" // untouchable ground-truth time
    }
  }]
}
```

## Verification

- `cargo fmt --all`, `cargo check --lib`, and `cargo clippy --lib -- -D warnings` all pass.
- On-device: with `set_debug_level(1)` the `[Metrics] Event queued` log now prints the full payload including the three new fields; batch `sentAt` is visible via the network inspector.

### DIFF
--- a/lib/src/analytics/data_definition.rs
+++ b/lib/src/analytics/data_definition.rs
@@ -1,3 +1,4 @@
+use chrono::{DateTime, SecondsFormat, Utc};
 use godot::{classes::Os, obj::Singleton};
 use serde::Serialize;
 
@@ -8,6 +9,14 @@ pub struct SegmentMetricEventBody {
     event: String,
     #[serde(rename = "userId")]
     user_id: String,
+    // Client-generated unique id. Segment deduplicates on this (>=24h window, ~170d in
+    // practice), which makes retries idempotent. Must be <100 chars — a UUID v4 is 36.
+    #[serde(rename = "messageId")]
+    message_id: String,
+    // Segment-reserved event time, captured when the event was created on the device (NOT at
+    // flush). Segment stores this as `originalTimestamp` and, together with the batch-level
+    // `sentAt`, corrects device clock skew into the canonical `timestamp` column.
+    timestamp: String,
     properties: serde_json::Value,
 }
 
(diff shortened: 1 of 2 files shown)


kuruk-mm (approved):
looks great

Issue #2450, opened 2026-07-06
[Bug] Loading Stuck at 25% Endlessly


regenesis-claw:
### Platform
IOS

### App Version
1138

### Device Information
IPHONE 16 E

### Issue Description
Loading gets stuck at 25% endlessly and never progresses, blocking the user from entering the scene/world.

### Expected Behavior
Loading should progress normally past 25% and complete, allowing the user to enter the scene.

### Screenshots / Media
Screenshot shared in Slack thread (loading screen stuck at 25%).

<img width="2532" height="1170" alt="Image" src="https://github.com/user-attachments/assets/e15d1492-56f5-4b02-891a-8ba35ff00027" />

### Steps to Reproduce
1. Launch the mobile app
2. Attempt to load into a scene/world
3. Observe loading bar gets stuck at 25% and never advances

### Occurrence
Not specified

---
Requested by Rodri via Slack

leanmendoza:
We tracked this down during the #1602 and #1640 loading investigation. There are two things going on, a display artifact and several real hangs.

The 25 to 30% number is partly by design. `LoadingSession::calculate_progress` caps the bar at 30% for the first 20s and holds the 60% Assets bucket at 0 until floating islands finish. A healthy Genesis load therefore sits at 25 to 30% for a while, and on a slow link that is indistinguishable from a hang. On device the bar held at 30% for about 25s straight. Tracked in #2468.

On top of that there are real infinite loading causes that keep it there:
- No HTTP timeout on the asset download path, so a stalled connection hangs forever (#2462).
- A notify leak that poisons an asset hash after one failed download, so retries never succeed (#2463).
- A confirmed accounting bug: deleting an entity mid load never bumps the finished counter, so the Assets phase can never reach 100% and the screen hangs with nothing actually stuck (#2475).
- No retry after an asset error, so on a throttled link about 734 assets timed out and stayed missing (#2476).

On device (cold cache, 1 Mbps) the load took 97.7s, the first scene spawned at 27.6s, and no scene ever reached ready. The screen only went away via the 90s wall clock modal, never real completion.

The first three fixes are implemented in #2461.

leanmendoza:
Closing this in favor of the single tracking issue **#2536**, where this symptom and its root causes are tracked (progress-bar artifact + the underlying hangs: no HTTP timeout, `pending_downloads` poisoning, deleted-while-loading accounting — several already shipped in #2461).

[...]

Issue #1640, opened 2026-03-06
Scene Lifecycle Issues: Investigation and Root Cause Analysis


regenesis-claw:
## Overview

Multiple scene lifecycle issues have been detected affecting scene loading and runtime behavior:

1. **Infinite scene loading** - Scenes that never complete loading
2. **Larger scene loading issues** - Problems specifically affecting larger/complex scenes  
3. **Scenes with unexpected behaviors** - Runtime behaviors that are difficult to track and reproduce

## Objective

Investigate and document the **root causes** of each of these issues without assuming what the underlying problems are.

## Deliverables

For each issue category, provide:

1. **Compact and precise description** of what is causing the problem
2. **Reproduction steps** - How to reliably force the issue to occur
3. **Reproduction chronology** - Timeline/sequence of events that leads to the problem
4. Documentation of findings to enable targeted fixes

## Approach

- Focus on observation and data collection first
- Document the actual behavior vs expected behavior
- Identify patterns and conditions that trigger each issue
- No assumptions about solutions until root causes are clearly understood

---

Requested by Lean via Slack

leanmendoza:
Root cause analysis done, covering the three failure modes and the five suspicions from the thread, each with a compact cause, repro and chronology.

Infinite scene loading. A scene never finishes when at least one GLTF entity never leaves `scene.gltf_loading`. Causes:
- No HTTP timeout on the asset path, so a stalled connection never resolves (#2462).
- The pending_downloads notify leaks on failure and poisons the hash permanently (#2463).
- The coordinator holds a fetch slot across the whole await, so 24 zombies stall all loading (#2464).
- Confirmed: deleting an entity mid load never bumps the finished counter, so the Assets phase never completes and the screen hangs with nothing stuck (#2475).
- No retry after an error leaves models permanently missing (#2476).

No destination infinite loading, which matches the reporter's hypothesis and is confirmed. The loading screen is shown before any session exists, so if `/about` fails or is malformed, nothing can dismiss it. `realm_change_failed` is only a toast on the main entry paths. Tracked in #2471, #2472, #2473 and #2474.

Large scene issues are dominated by the O(N) rescan (#2465, #2466) plus tween spikes. Weird behaviour covers a tween versus scene transform pop race, the avatar impostor retry storm (#2467) and a nameplate access after free (#2469).

[...]

Issue #1602, opened 2026-03-02
Add progressive scene loading with priority-based asset streaming


regenesis-claw:
## Description

Implement progressive scene loading that allows users to start interacting with a scene before all assets are fully downloaded, improving perceived load times on mobile devices with limited bandwidth.

## Motivation

Mobile users often experience longer wait times when entering high-detail scenes due to:
- Limited bandwidth on cellular connections
- Large asset sizes (textures, models, audio)
- Current all-or-nothing loading approach

Progressive loading would improve UX by:
- Showing low-res placeholder content immediately
- Streaming high-res assets in the background
- Prioritizing assets based on user proximity and viewport

### 1. Asset Priority System
- **Critical**: Ground mesh, skybox, nearby objects (<5m)
- **High**: Visible objects in viewport (5-20m)
- **Medium**: Objects outside viewport but in scene
- **Low**: Audio, decorative elements, distant objects

### 2. LOD Strategy
- Load low-poly versions first (or simplified placeholder geometry)
- Stream high-res textures progressively (mip-map levels)
- Defer audio loading until needed

### 3. Content System Changes
```rust
// lib/src/content/
- Add priority queue for asset requests
- Implement bandwidth estimation
- Track asset dependencies for critical path
```

### 4. Scene Runner Integration
```rust
// lib/src/scene_runner/
- Notify scene when assets become available
- Handle missing asset gracefully (show placeholder)
- Emit events for progressive load states
```

## Technical Considerations

- **Cache Management**: Need smart eviction for partially-loaded scenes
- **IPFS Streaming**: May need chunked asset fetching from content servers
- **Memory Budget**: Careful not to load too many LOD levels simultaneously
- **Scene SDK Compatibility**: Ensure SDK7 scenes handle missing assets gracefully

## User Experience

**Before**:
1. User clicks teleport
2. Loading screen shows
3. Wait for ALL assets (10-30s on slow connection)
4. Scene appears

**After**:
1. User clicks teleport
2. Ground + skybox appear immediately (1-2s)
3. Nearby objects stream in (3-5s)
4. Background details load (5-10s)
5. User can start exploring while loading continues

## Success Metrics

- Reduce time-to-first-interaction by 60%+
- Maintain smooth FPS during streaming
- < 5% increase in peak memory usage
- No increase in total bandwidth consumed

## Related Work

- Desktop client may benefit from similar approach for large scenes
- Could extend to prefetching scenes near current position

## Demo Context

[...]

PR #2461, opened 2026-07-08
fix: infinite loading hangs (delete accounting, HTTP timeouts, download notify leak, malformed /about)

### LINKED ISSUE #2462: Add HTTP timeouts to the asset download path
`ResourceProvider` builds its HTTP client with `Client::new()` and no timeout (`resource_provider.rs`). A stalled or half open connection hangs the download future forever, which leaves the entity stuck in `gltf_loading` and the scene loading with no end.

The fix cannot be a plain total timeout. On a 1 Mbps link a legitimate asset was measured taking 198s of wire time, so a fixed request timeout would break exactly the slow network case we care about. We set a connect timeout, a response (headers) timeout and a per chunk idle timeout, so only dead connections get killed. A stall then surfaces as an error and the container reaches `FinishedWithError` instead of hanging.

Root cause RC-1 in the #1602 and #1640 investigation. Implemented in #2461.

### LINKED ISSUE #2463: Release the pending_downloads slot on every path to avoid poisoning a hash
When a download fails, the `Arc<Notify>` stored in `pending_downloads` is never removed or fired, because cleanup only runs on the success path (`resource_provider.rs`). Two things break: every duplicate waiter for that hash hangs forever, and the hash stays poisoned so even a later retry registers as a waiter and blocks. One failed download can keep an asset broken until the app restarts.

Fix: run the cleanup on every exit path, success or failure, across all three fetch variants. A regression test covers it: a failed download must release its slot and a following retry must succeed, which hangs forever before the fix.

Root cause RC-2. Implemented in #2461.

### LINKED ISSUE #2472: Guard malformed /about content before committing realm state
A realm whose `/about` returns `content: null` or a missing `publicUrl` throws at `content_base_url` (`realm.gd`) after realm state was already partially committed and without emitting `realm_change_failed`. The result is a silent stuck loading screen with no toast.

Fix: validate `content.publicUrl` before committing realm state, and emit `realm_change_failed` on missing or invalid values. Root cause RC-6. Implemented in #2461.

### LINKED ISSUE #2475: Count a deleted while loading GLTF as finished so the screen does not hang
Deleting an entity while its GLTF is still loading removes it from `gltf_loading` but never bumps `gltf_loading_finished_count` (`deleted_entities.rs`). The component removal and normal completion paths both bump it, so this path leaves `started > finished` forever. The Assets phase then never reaches 100% and the loading screen hangs even though nothing is actually stuck.

Fix: bump the finished counter when removing a loading entity, same as the other two paths. One line. Confirmed end to end, root cause RC-10. Implemented in #2461.

leanmendoza:
Fixes a group of infinite loading and "stuck at 25 to 30%" root causes found during the #1602 and #1640 scene loading investigation. Each fix is small, isolated and independently verified. The user facing symptom is tracked in #2450.

## What this fixes

**Deleted while loading GLTFs never counted as finished (RC-10, confirmed)**
Deleting an entity while its GLTF was still loading removed it from `gltf_loading` but never bumped `gltf_loading_finished_count` (`deleted_entities.rs:48`). The component removal and normal completion paths both bump it (`gltf_container.rs:68`), so this path left `started > finished` forever. The Assets phase then never reached 100% and the loading screen hung with nothing actually stuck. One line, mirrors the existing pattern.

**No HTTP timeout on the asset download path (RC-1)**
`ResourceProvider` used `Client::new()` with no timeout, so a stalled or half open connection hung the download future forever and left the entity stuck in `gltf_loading`. This adds bounds that kill dead connections without killing slow but alive ones. A legitimate asset was measured taking 198s of wire time on a 1 Mbps link, so there is deliberately no total request timeout. Instead it sets a connect timeout, a response (headers) timeout and a per chunk idle timeout. A stall now surfaces as an error and the container reaches `FinishedWithError`.

**pending_downloads notify leak (RC-2)**
On download failure the `Arc<Notify>` was never removed or fired, because cleanup only ran on the success path. Duplicate waiters hung forever and the content hash stayed poisoned, so even retries could never succeed. Cleanup now runs on every exit path via `finish_pending_download`, across all three fetch variants. Includes a regression test where a failed download must release its slot and a following retry must succeed. Before the fix that retry hangs forever.

**Malformed /about content left the screen stuck silently (RC-6)**
A realm whose `/about` returns `content: null` or a missing `publicUrl` threw at `content_base_url` after realm state was already partially committed and without emitting `realm_change_failed`. The loading screen just sat there. It is now validated before the state commit, so the failure is surfaced and recoverable.

## Verification
- `cargo clippy --features use_livekit,use_deno -- -D warnings` clean
- `cargo test --features use_livekit,use_deno resource_provider` 4 of 4, including the new leak regression test
- `gdformat --check` and `gdlint` clean on `realm.gd`

Closes #2475. Closes #2462. Closes #2463. Closes #2472.
Refs #1640, #1602, #2450.

### DIFF
--- a/godot/src/logic/realm.gd
+++ b/godot/src/logic/realm.gd
@@ -178,6 +178,18 @@ func async_set_realm(new_realm_string: String, search_new_pos: bool = false) -> 
 		_emit_realm_change_failed(new_realm_string, reason)
 		return false
 
+	# Guard malformed /about `content` BEFORE committing realm state. A null/missing
+	# `content` or a non-String `content.publicUrl` would otherwise throw at
+	# `content_base_url` below — after realm state was already partially committed and with
+	# NO `realm_change_failed` emitted — leaving a silently stuck loading screen (F-11 / RC-6).
+	# (`or` short-circuits, so `.get()` is never called on a non-Dictionary `content`.)
+	var about_content = json.get("content")
+	if not about_content is Dictionary or not about_content.get("publicUrl") is String:
+		var reason := "/about missing or malformed content.publicUrl"
+		printerr("[REALM] ", reason, " for ", new_realm_string)
+		_emit_realm_change_failed(new_realm_string, reason)
+		return false
+
 	# /about was validated — now it's safe to commit the new realm state.
 	realm_string = new_realm_string
 	realm_url = candidate_realm_url
--- a/lib/src/content/resource_provider.rs
+++ b/lib/src/content/resource_provider.rs
@@ -455,6 +455,17 @@ impl ResourceProvider {
         Ok(())
     }
 
+    /// Release the `pending_downloads` slot for `file_hash`: remove the entry and wake any
+    /// waiters. Must run on EVERY exit path of a fetch (success or failure), otherwise a
+    /// failed download leaks the `Notify` — duplicate waiters hang forever and the hash is
+    /// poisoned permanently so retries can never succeed (RC-2 / F-2).
+    async fn finish_pending_download(&self, file_hash: &str) {
+        let mut pending_downloads = self.pending_downloads.write().await;
+        if let Some(notify) = pending_downloads.remove(file_hash) {
+            notify.notify_waiters();
+        }
+    }
+
     pub async fn file_exists(&self, file_hash: &str) -> bool {
         let existing_files = self.existing_files.read().await;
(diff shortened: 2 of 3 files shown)


kuruk-mm (approved):
looks great

PR #2478, opened 2026-07-08
fix: prevent profile screen state changes when returning from background

### LINKED ISSUE #2378: [Bug] Profile screen auto-scrolls and scaling breaks when returning from background
### Platform
iOS

### App Version
Not specified

### Device Information
Not specified

### Issue Description
When the app is backgrounded and then returned to the foreground while on the Profile screen, unexpected behavior occurs:
- The screen auto-triggers user input (as if the user interacted with it)
- Automatically scrolls all the way to the bottom
- Scaling breaks, displaying the UI as if zoomed in by 200%

### Expected Behavior
The Profile screen should remain in the same state (scroll position, zoom level) when returning from background. No automatic user input should be triggered.

### Screenshots / Media
Two screen recordings demonstrating the issue were shared in Slack.

https://github.com/user-attachments/assets/d205c874-2889-48b3-a80c-97e53d014e31

https://github.com/user-attachments/assets/c784a135-652a-4f24-9c9c-ce468e92ca45

### Steps to Reproduce
1. Open the Profile screen in the app
2. Background the app (e.g., switch to another app or home screen)
3. Return to the app

### Occurrence
Reproducible (observed by multiple users)

---

Requested by Rodri via Slack

sdilauro:
Fixes #2378

## Problem

Two independent bugs triggered when returning the app from background on iOS, both visible on the Profile screen.

### 1. Bottom sheet card changes state when returning from background

On iOS, two separate mechanisms can cause the bottom sheet to change state unexpectedly:

**A — Phantom touches after FOCUS_OUT:** When iOS backgrounds the app while a touch is active, it cancels it. On foreground resume, the cancelled touch arrives as a `pressed=false` event. If `dragging` was still `true`, the sheet computed a drag distance from the stale `start_pos` and could fire an UP/DOWN gesture.

**B — Swipe-up-to-home gesture:** On iPhone (no home button), the gesture to go home delivers `pressed=true` + `pressed=false` to the app and **fully completes before `FOCUS_OUT` fires**. By the time we handle `FOCUS_OUT`, `drag_state` may already reflect the gesture's outcome (e.g. HALF→FULL), so simply saving `drag_state` on `FOCUS_OUT` saves the wrong value.

**Fix:**

- On `NOTIFICATION_APPLICATION_FOCUS_OUT`: set `dragging = false` and `_ignore_input = true` (blocks phantom touches). To detect the home gesture, record the drag state at each touch start (`_state_at_touch_start`) and the timestamp at each touch end (`_last_touch_end_ms`). If a touch ended less than 300ms before `FOCUS_OUT`, use `_state_at_touch_start` (pre-gesture) as the saved state instead of the current `drag_state`.
- On `NOTIFICATION_APPLICATION_FOCUS_IN`: if `drag_state` differs from the saved state, call `_snap_to_state()` to silently restore the card — the snap happens during the iOS transition animation and is invisible. A 100ms timer then clears `_ignore_input` to cover any phantom touches that arrive between `applicationWillEnterForeground` and `applicationDidBecomeActive`.

### 2. Card appears wider than the display

On iOS, returning from background can briefly trigger a `size_changed` event with the window dimensions **swapped** (height reported as width). `apply_ui_zoom` was responding to this, computing an incorrect `content_scale_factor` that made the UI re-layout as if in landscape, causing the card to overflow the portrait screen width.

**Fix:** Added a guard in `apply_ui_zoom` for real mobile devices: if the window dimensions imply a different orientation than `Global.is_orientation_portrait()` (the authoritative orientation state, always set before any real rotation fires `size_changed`), the event is skipped as a transient erroneous report.

## Test plan

- [x] Open the Profile screen (portrait mode)
- [ ] Drag the card to PEEK state, then HALF, then use the iOS swipe-up gesture to go home — on return the card should be at HALF (not FULL)
- [x] With the card at FULL, background the app normally (home button / swipe) — on return the card should still be FULL
- [x] Background the app mid-drag-animation — on return the card should be in its target state
- [x] Verify the card width matches the screen width on return (no overflow wider than the display)


regenesis-claw (approved):
Clean, well-documented fix for two independent iOS background/foreground bugs. The approach is sound on both fronts.

**Bottom sheet state restoration** — the dual-mechanism analysis (phantom touches vs swipe-up-to-home gesture completing before FOCUS_OUT) is thorough, and the fix correctly addresses both:
- `_ignore_input` blocks phantom touch-cancel events between FOCUS_OUT/IN
- `_state_at_touch_start` + 300ms heuristic detects the home-gesture race
- `_snap_to_state` restores the card instantly (during iOS transition animation, so invisible to the user)
- The 100ms post-FOCUS_IN timer with `CONNECT_ONE_SHOT` is a reasonable safety margin for late-arriving phantom events

**UI zoom orientation guard** — skipping `size_changed` events where dimensions contradict the authoritative orientation state is the right level to fix this. The `not Global.is_virtual_mobile()` guard ensures desktop/editor testing is unaffected. Consistent with the existing `Global.is_mobile()` usage in `apply_window_config()` just below.

No debug prints, no non-English comments, no dead code. Comments explain *why* throughout. LGTM.

*Note: iOS build is still queued at review time. Since this is an iOS-specific fix, a green iOS build should be confirmed before merge.*

Requested by Seba via Slack

kuruk-mm (approved):
code looks great, wait qa review

Ludmilafantaniella:
Tested on iOS (v1.11.0 build 1175). ✅

Verified repro steps from #2378:
- Opened the Profile screen, backgrounded the app and returned'; screen stays in the same state, no unexpected scrolling or input triggered ✅
- Card width renders correctly on return from background, no landscape overflow ✅

LGTM 🚀


https://github.com/user-attachments/assets/a03ea93e-9dc3-4e9e-81e9-46a41eb0f4f9

cc @kuruk-mm

PR #2482, opened 2026-07-08
feat: allow deleting an upgraded (email-linked) guest via double unlink

### LINKED ISSUE #2479: Allow Deletion of upgraded guest user accounts via deeplink or debug panel
## Description

Add an option to facilitate deletion of guest users and upgraded guest users directly from mobile. This will help QA/support/dev processes for testing guest onboarding flows without needing manual backend intervention.

## Proposed Solution
- Add a "Delete Account" (or similar) action via deeplink for:
  - Upgraded guest users (removes the guest-to-web3 upgrade link, resetting to a clean state)
- Confirm destructive action with a confirmation dialog before executing
- Ensure the app returns to a clean login state after deletion (similar to sign-out teardown flow)

leanmendoza:
Fix #2479 

## What

Adds an opt-in path to fully delete an **upgraded** thirdweb guest (a guest that
has linked an email) from the account-deletion flow.

Today the on-device guest deletion only handles a **non-upgraded** guest: it
unlinks the sole `guest` profile and explicitly refuses if any email/social
profile is linked (those fall back to the manual server `/deletion` request).
This PR lets an upgraded guest be deleted on-device too, via a **double unlink**.

## How

When the gate is on, an upgraded guest takes the same automatic on-device
deletion path and unlinks **both** the `guest` and the linked `email` profile.
Removing all profiles deletes the whole thirdweb user (instead of stripping it
back to a bare guest), which frees the deterministic `sessionId` so the next
guest login mints a brand-new wallet.

- Email profiles are unlinked by `details.email`, everything else by `details.id`.
- The `guest` profile is unlinked **last**, so the "last-profile" quirk (the API
  deletes the user and then answers `500` because it can't re-fetch it — which we
  treat as success) lands on a single known request.
- Best-effort: if a network step fails, the client still wipes local storage and
  signs out.

## Gate (both required)

- `not is_production()` — **non-prod only** (hard-disabled on release cuts)
- deeplink `decentraland://open?enable-upgraded-deletion=true`

Off by default, so a recoverable (email-linked) account is never force-deleted.
Without the gate an upgraded guest keeps the existing manual `/deletion` flow,
and a non-upgraded guest is unchanged.

## Files

- `lib/src/auth/thirdweb_guest.rs` — extract the unlink POST + 500-as-success into
  a shared `post_unlink`; add `unlink_upgraded_account` (unlinks every profile,
  guest last). `unlink_guest_profile` is unchanged and still refuses upgraded
  accounts as a safety net.
- `lib/src/auth/dcl_player_identity.rs` — `async_delete_upgraded_account` (mirrors
  `async_delete_guest_account`, best-effort resolves true/false).
- `godot/src/global.gd` — `_enable_upgraded_deletion` flag +
  `is_upgraded_deletion_enabled()` (flag AND `is_production()`).
- `godot/src/deep_link_router.gd` — set the flag from the deeplink param.
- `godot/src/ui/components/organisms/menu/account_deletion_popup.gd` — route an
  upgraded guest to the double-unlink path only when the gate is on.

## Testing

- `cargo check` + `cargo clippy --lib` clean (no warnings); `gdformat` + `gdlint`
  clean on the whole `godot/` folder.
- The confirmation-dialog copy is unchanged (same as the guest deletion).

### DIFF
--- a/godot/src/deep_link_router.gd
+++ b/godot/src/deep_link_router.gd
@@ -51,6 +51,14 @@ func process_deep_link(url: String) -> void:
 		Global.cli.set_kill_sky(kill_sky_value.to_lower() in ["true", "1", "yes"])
 		print("[DEEPLINK] kill-sky=", Global.cli.get_kill_sky())
 
+	# Opt-in gate for deleting an UPGRADED (email-linked) guest. Sticky-on for the
+	# session; only takes effect on a NON-production build (see
+	# Global.is_upgraded_deletion_enabled() + account_deletion_popup.gd).
+	var upgraded_deletion_value = Global.deep_link_obj.params.get("enable-upgraded-deletion", "")
+	if upgraded_deletion_value.to_lower() in ["true", "1", "yes"]:
+		Global._enable_upgraded_deletion = true
+		print("[DEEPLINK] enable-upgraded-deletion=true")
+
 	# Genesis Plaza profiling benchmark (issue #1862). The CLI path spawns the
 	# runner from Global._ready, but on mobile the deep link is not parsed by
 	# then — spawn here once the deeplink lands and only if no runner exists.
--- a/godot/src/global.gd
+++ b/godot/src/global.gd
@@ -169,6 +169,13 @@ var attestation: AttestationService = null
 
 var _is_portrait: bool = true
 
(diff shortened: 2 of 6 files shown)


rodrigo685 (approved):
Reviewed and Approved.

PR #2517, opened 2026-07-12
feat: add TextEncoder/TextDecoder scene globals

eordano:
Scene runtimes have no `TextEncoder`/`TextDecoder` because the embedded deno_core runs without deno_web, so SDK scenes and libraries relying on the WHATWG text codecs crash at load (or fall back to slow bundled polyfills).

This adds two std-only ops (`op_utf8_encode`/`op_utf8_decode`) and a small JS shim module registered like the existing fetch/ws shims, installing WHATWG-shaped `TextEncoder` and `TextDecoder` on `globalThis` when absent.

- `TextDecoder` accepts the WHATWG utf-8 label set with `fatal` and `ignoreBOM` support, throws `RangeError` for any other encoding and a loud `TypeError` for the unsupported `{ stream: true }` option.
- `TextEncoder` implements `encode` and `encodeInto`, including correct UTF-8 boundary truncation for `encodeInto` on short destinations.
- Decoding uses Rust's `from_utf8_lossy`, whose maximal-subparts U+FFFD substitution matches the WHATWG spec exactly; encoding inherits USVString lone-surrogate replacement from deno_core's string conversion. No new dependencies.

Unit tests cover the Rust decode paths (`cargo test text_encoding`) and the JS shim (13 `node:test` cases mirroring `ws.test.cjs`). Full crate suite passes (170 tests), `cargo fmt` clean. The ops are pure `std` string/slice operations with no OS APIs, so this ships identically on desktop and the Android/iOS builds (default features include `use_deno` on all targets).

Scenes built with the current SDK keep working unchanged: the SDK installs its polyfill only when the global is absent, so the native-backed implementation wins automatically, including for already-published content.

Related: decentraland/js-sdk-toolchain#1452 (slims the SDK polyfill to a validating utf-8 codec), decentraland/bevy-explorer#958 (same exposure via deno_web).

Note for maintainers: the shim tests are runnable with `node --test lib/src/dcl/js/js_modules/text_encoding.test.cjs` and are meant to join `ws.test.cjs` in the static-checks `node --test` step; that one-line workflow edit is left out of this PR (pushed from a token without workflow scope).

### DIFF
--- a/lib/src/dcl/js/js_modules/main.js
+++ b/lib/src/dcl/js/js_modules/main.js
@@ -310,6 +310,14 @@ globalThis.Headers = fetchModule.Headers;
 globalThis.Response = fetchModule.Response;
 globalThis.WebSocket = require('ws').WebSocket;
 
+const textEncodingModule = require('text_encoding');
+if (globalThis.TextEncoder === undefined) {
+    globalThis.TextEncoder = textEncodingModule.TextEncoder;
+}
+if (globalThis.TextDecoder === undefined) {
+    globalThis.TextDecoder = textEncodingModule.TextDecoder;
+}
+
 
 globalThis.UnityOpsApi = undefined
 globalThis.global = globalThis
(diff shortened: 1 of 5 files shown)


kuruk-mm (approved):
looks great! thanks

kuruk-mm:
Closing in favor of #2542, which contains the exact same changes re-committed as a signed commit (this PR couldn't be merged because its commits are unsigned). Authorship is preserved via Co-authored-by.

PR #2539, opened 2026-07-14
fix: loading screen layout and trimmed label height

### LINKED ISSUE #2495: Loading Screen: Review load details text and % load boxing layout
## Description

The loading screen's middle-bar layout needs a design review. The separation between the load details text and the load percentage components appears to be static rather than auto-adjusting. This results in excessive whitespace on the left side of the separator in some cases.

Both parts of the middle-bar are dynamic components that change width depending on content, so the spacing between them should adjust accordingly.

Image

## Current Behavior

Static separation causes an unbalanced layout with too much space on the left side of the separator when the load details text is shorter.

## Expected Behavior
TBD

sdilauro:
Closes #2495

## Summary
- Removes the static VSeparator between load details text and percentage, so spacing auto-adjusts to content width
- Aligns percentage label to the right side of the row
- Reduces TrimmedRichTextLabel extra vertical padding: for text that fits within `max_lines`, lets `fit_content` handle height naturally instead of using a fixed computed height with a 0.2 multiplier

## Test plan
- Open the loading screen and verify the load details text and % are properly spaced without excess whitespace on the left
- Verify 1-line titles (e.g. CozyFarm), 2-line titles, and truncated long titles all render without extra gap below the text


regenesis-claw (approved):
Clean fix — the `TrimmedRichTextLabel` height logic is much simpler now that `fit_content` handles single-line cases naturally, and removing the static `VSeparator` in favor of `size_flags_horizontal = EXPAND` + right-alignment is the right call.

Two small things:

1. **Unclosed `[b]` tag** in `loading_screen.gd:265` — `"[color=#DF9CFF]By[/color] [b]" + creator` opens a bold tag but never closes it. It works fine here since the creator name is the rest of the text, but closing it with `+ creator + "[/b]"` would be more robust (e.g. if content is ever appended later).

2. **Close button container hidden** — `SafeMarginContainer2` is now `visible = false` in the `.tscn`, which hides the close button on the loading screen. This wasn't mentioned in the Summary — is it intentional? If so, worth a line in the PR description for reviewers/QA.

Neither blocks merge. CI is all green (Linux, Android, iOS, static checks, Clippy). LGTM.

Requested by Seba via Slack

regenesis-claw (approved):
Re-reviewed after `ed35b98`: both notes from my previous review are addressed — `[b]` tag is now closed, and the close button container is visible again. LGTM 🚀

ElzathRusso (approved):

Issue #2536, opened 2026-07-14
Scene loading & lifecycle — infinite loads, stalls & root causes


leanmendoza:
Umbrella for scene-loading & lifecycle reliability — the loading / lifecycle work items surfaced by the #1602 · #1640 · #2450 investigation, kept in one place so progress is visible across all of them.

As of **#2540** the loading funnel ships a single Segment `Loading Event`, so the remaining open items are now **measured from field data** — real frequency and severity — instead of being blocked on a repro that most of them never had. Read the `Loading Event` schema and the pilot-dashboard / next-steps notes on **#2540**.

Each item keeps its original number for history; from now on reference an item as **`#<this-issue> D)`**.

**Legend:** `[x]` shipped · `[ ]` open · **Prio** High / Med

- [x] **A)** Add HTTP timeouts to the asset download path — **High** — shipped in #2461 (was #2462)
- [x] **B)** Release the `pending_downloads` slot on every path (no hash poisoning) — **High** — shipped in #2461 (was #2463)
- [ ] **C)** Bound or reap zombie download slots in the GLTF coordinator — **High** — was #2464
- [ ] **D)** Turn `sync_gltf_loading_state` from a per-tick rescan into a push model — **Med** — was #2465
- [ ] **E)** Cache the `GltfContainer` node handle per entity — **Med** — was #2466
- [ ] **F)** Fix the avatar impostor capture buffer size + stop the retry storm — **Med** — was #2467
- [ ] **G)** Rework the loading bar so a healthy load doesn't look frozen at 25–30% — **Med** — was #2468
- [ ] **H)** Tie freed-node cleanup to `NOTIFICATION_PREDELETE` (nameplate + LOD coordinator) — **Med** — was #2469
- [ ] **I)** Reconcile asset accounting when a stuck scene is force-marked ready — **Med** — was #2470
- [ ] **J)** Dismiss the loading screen + show an error when realm resolution fails — **High** — was #2471
- [x] **K)** Guard malformed `/about` content before committing realm state — **High** — shipped in #2461 (was #2472)
- [ ] **L)** Add a global loading watchdog + fix the cached empty-teleport dismissal — **High** — was #2473
- [ ] **M)** Retry `/about` with backoff + fall back to a known-good realm — **Med** — was #2474
- [x] **N)** Count a deleted-while-loading GLTF as finished (no hang) — **High** — shipped in #2461 (was #2475)
[...]

Issue #2477, opened 2026-07-08
Send loading telemetry to Segment so the loading decisions are data driven


leanmendoza:
The whole loading diagnosis currently rests on one device capture. The open decisions (drop or keep the 30% cap, coordinator slot management, asset retry, timeout tuning, no destination hangs) need field distributions across the fleet, not a single run. The `LoadingProfiler` autoload already computes the exact milestones we want (first spawn, first ready, time in the 25 to 30% band, queue versus wire split, timeout count, dismissed by completion versus wall clock), it just logs them locally. The Segment pipeline already exists and was hardened in #2440.

Proposed events, bucketed and with no PII:
- `scene_loading_started`: realm, is world, cold cache, network estimate, device tier.
- `scene_loading_completed`: duration, reached ready, dismissed by (completion, wall clock, error), time in the 25 to 30% band, first spawn, first ready, phase breakdown, assets expected, loaded, errored, timed out, queue wait fraction, peak loading length.
- `scene_loading_asset_failure` (sampled): count, reason.
- `realm_change_failed`: reason.

What each field unblocks: the time in band distribution plus reached ready versus wall clock tells us whether the 30% plateau is perception or a real hang and how often. The timeout and error rates size the retry and timeout work. The queue wait fraction by network tier sizes the slot management and priority queue work. The realm_change_failed volume sizes the no destination hangs.

Source the numbers from the loading profiler summary, emit through the existing Segment path, bucket and round to avoid PII and high cardinality, respect the analytics opt in, and sample the high frequency per asset event. Can ship as one issue or split into a core funnel event and the per asset diagnostics.

leanmendoza:
Consolidated into #2536 — item **P)**. Closing in favor of the single tracker; the work continues there.

PR #2456, opened 2026-07-07
fix: tap on featured card opens detail panel

sdilauro:
Fixes #2105

## Summary
- Tapping a featured card now opens the detail/jump-in panel instead of teleporting directly
- The Jump In button on the card still teleports directly (existing behavior preserved)
- Fixed carousel cards getting stuck at intermediate position when tapping during auto-scroll or snap animation (`_is_animating` reset + `_layout_cards()` called on tap)
- Fixed detail panel image not filling its container (`TextureRect_Image` changed to anchor-based full rect layout)
- Deferred panel open to avoid `SidePanelWrapper` closing itself when receiving the same touch-release event that triggered the open

## Test plan
- [x] Tap a featured card body → detail panel opens with place info
- [x] Tap the Jump In button on the featured card → jumps in directly
- [x] Swipe between featured cards, tap during snap animation → card repositions correctly, panel opens
- [x] Auto-scroll running, tap a card → card snaps to correct position, panel opens
- [ ] Open detail from a regular place item → still works as before

PR #2540, opened 2026-07-14
feat(loading): scene-loading funnel in Rust → single Segment "Loading Event" + GLTF loading coordinator

### LINKED ISSUE #2098: [Bug]: Several Gltf instances with one unique source delays the loading queue
### 🕹️Platform

Android

### 🔢App Version

All versions

### 📱Device Information

All devices

### 📄Issue description

If a scene create several entities with the same GLB source, each instance consumes one slot of concurrency in the queue (MAX = 10) so it can delay the loading more than necessary (imagine several GLBs with several instances, like Genesis Plaza)

### ✅Expected behavior

One queue slot per GLB source

### 🖼️Screenshots / Media

_No response_

### ▶️Steps to reproduce

This is not verified, it needs further investigation.

### 👁️Occurrence

100% of the time

leanmendoza:
## What & why

1. **Scene-loading funnel → a single Segment `Loading Event`** (item **P** of #2536). Most loading hangs in #2536 have **no reliable repro**, so their DONE criterion is reframed from *"reproduced + fixed"* to **"instrumented + measurable from field data"**.
2. **GLTF loading coordinator** — dedups GLTF loads by source hash and decouples download from the main-thread load/realize pump, replacing the per-instance queue bottleneck (#2098). On-device: Genesis Plaza queue-wait ~42s → ~12s, screen time ~72s → ~55s.

## The Segment model

One event name, `Loading Event`, discriminated by `type` and correlated across a whole load by **`loading_id`**. A "load" is one loading-screen-bounded episode (it may span several internal loading sessions — a realm change mid-load supersedes the previous one). Group on `loading_id` and the funnel reassembles.

| `type` | Emitted | Properties |
|---|---|---|
| `started` | load begins | `loading_id`, `schema`, `realm_bucket`, `is_world`, `when`, `cpu_count`, `platform` |
| `progress` | every 10s **while a load is still in flight** | the accumulated snapshot below — everything `completed` carries **except** `dismissed_by` and `frames` |
| `completed` | load ends, **always** | `started`'s fields, plus `duration_ms`, **`dismissed_by`**, `complete_ms`, `first_spawn_ms`, `scene_rendered_ms`, `about_end_ms`, `discovery_ms`, `time_in_25_30_band_ms`, `phase_breakdown{…}`, **`network_avg_mb_s`**, `network_peak_mb_s`, **`network_stall_ms`**, **`process_assets_pending_at_end`**, `process_assets_pending_peak`, `assets_errored`, `frames` |
| `asset_failure` | with `completed`, only if any asset group failed | `loading_id`, `realm_bucket`, `reason`, `count` |
| `realm_change_failed` | on realm resolution failure | `loading_id`, `realm_bucket`, `reason` |

- `dismissed_by` ∈ `completion` / `wall_clock_timeout` / `error` / `superseded` / `cancelled`. Precedence: superseded → cancelled → error → completion → timeout. **`completion` is decided by the pipeline reaching the `done` phase, nothing else.**
- `realm_bucket` ∈ `world` / `genesis` / `other` / `unknown`. No PII: no parcel coords, no URLs, no addresses.
- `when` — `on_explorer_ready` is the one **cold lobby→world entry (pre-world)**; every other value happens **in-world**: `on_teleport`, `on_world`, `on_moveto` (walk-teleport), `on_goto_realm` / `on_changerealm` (chat commands), and `auto` (background parcel streaming). So **pre-world ⟺ `when == 'on_explorer_ready'`** — no separate field needed. `-`/absent `when` appears only on the `asset_failure` / `realm_change_failed` side events. **Filter `when != 'auto'`** to drop background streaming from user-visible loads.
- Milestones are ms-since-begin, **`-1` = never happened**. `phase_breakdown` is the entry ms of each phase (`metadata`, `spawning`, `assets`, `ready`, `floating_islands`, `done`), same rule.

### Expected chronologies

Between `started` and `completed` the funnel accumulates in memory; a `progress` pulse ships every 10s **only while a load is still in flight**. A healthy load finishes before the first pulse, so it emits just `started` + `completed`; slow/stuck/abandoned loads emit pulses.

**Healthy teleport to Genesis**
```
t=0        started    id=42  realm_bucket=genesis  is_world=false  when=on_teleport
             │  ├─ about_end_ms=180        realm /about resolved
             │  ├─ discovery_ms=340        scene entity discovered
             │  ├─ first_spawn_ms=520      first scene node spawned
             │  ├─ phase_breakdown: metadata=90 spawning=500 assets=760 ready=9200
             │  │                   floating_islands=11100 done=11800
             │  ├─ complete_ms=11800       pipeline reached `done`
             │  └─ scene_rendered_ms=9400  first render (best-effort, see limitations)
t=12.0s    completed  id=42  dismissed_by=completion  duration_ms=12000
                             network_avg_mb_s=3.69  network_stall_ms=~0
                             process_assets_pending_at_end=0
```

**Infinite loading** — the bug #2450/#1640 describe, now one query
```
t=0        started    id=43  when=on_teleport
t=90s      completed  id=43  dismissed_by=wall_clock_timeout  complete_ms=-1
                             process_assets_pending_at_end=72   ← still owed content
                             network_stall_ms=11000             ← and nothing was arriving
```

**Throttled connection** — hangs, never fails, so `assets_errored` stays 0. Pulses ship as it drags on
```
t=0        started    id=44
t=10s      progress   id=44  duration_ms=10000  network_avg_mb_s=0.55  network_stall_ms=6000
t=20s      progress   id=44  duration_ms=20000  network_stall_ms=11000
t=55s      completed  id=44  dismissed_by=completion  network_avg_mb_s=0.60
                             network_stall_ms=11000  assets_errored=0
                             process_assets_pending_peak=72
```

**User quits mid-load (abandoned)** — the pulse is the only record; no `completed` ever ships
```
t=0        started    id=50  when=on_teleport
t=10s      progress   id=50  duration_ms=10000  complete_ms=-1  process_assets_pending_at_end=64
t=20s      progress   id=50  duration_ms=20000  network_stall_ms=9000
           ← user kills the app: no `completed`
```
Query: a `started` (or `progress`) with no matching `completed` on its `loading_id` = an abandoned load; the last `progress` shows how far it got. Each event is flushed immediately, so these survive the quit.

**User cancels**
```
t=0        started    id=45
t=4.2s     completed  id=45  dismissed_by=cancelled  complete_ms=-1  duration_ms=4200
```

**Teleport during a load (supersede)** — two loads, two ids, never merged
```
t=0        started    id=46
t=61ms     completed  id=46  dismissed_by=superseded  duration_ms=61
t=61ms     started    id=47                                     ← the new destination
```

**Asset failures** — a second event carries the aggregate, same `loading_id`
```
t=0        started       id=48
t=8s       completed     id=48  dismissed_by=completion  assets_errored=3
t=8s       asset_failure id=48  count=3  reason="…"
```

**Realm resolution fails**
```
t=0        started             id=49
t=2.0s     realm_change_failed id=49  reason="invalid /about response"
t=2.1s     completed           id=49  dismissed_by=error
```

### Three signals this exists to produce

- **Infinite loading** — `dismissed_by = wall_clock_timeout`: the screen was dismissed while the pipeline never reached `done`. That *is* the stuck-loading-screen bug, field-measurable instead of anecdotal.
- **Slow connection** — bucket on `network_avg_mb_s`; `network_stall_ms` is time with work outstanding and ~nothing arriving. Throttled downloads **hang rather than fail**, which is why they never appear in `assets_errored` and needed their own signal.
- **Completed but incomplete** — `process_assets_pending_at_end > 0` with `dismissed_by = completion`: the screen went away while the client still owed content. See *Follow-up*.
- **Abandoned load** — a `started`/`progress` with no `completed`: the user quit or the client was killed mid-load, before the wall-clock timeout would fire. The 10s `progress` pulse (flushed immediately) preserves the last-known state, so the highest-signal failure — a user giving up on a stuck load — is countable instead of invisible.

## Design

`lib/src/scene_runner/loading_funnel.rs` is a **pure accumulator**: no Godot objects, no `async`/`await`, unit-tested. `DclSceneManager` drives it from the loading-session lifecycle it already owns and does all Godot boundary I/O; the funnel's methods just **return** a typed `SegmentEventLoading`. GDScript keeps only the boundary push-calls for marks originating in UI/realm code (`loading_begin_episode` / `loading_end_episode` / `loading_mark_about_end` / `loading_mark_discovery` / `loading_realm_change_failed` / `loading_note_asset_failure`).

This **deletes `loading_profiler.gd`** (562 lines) and the GDScript→JSON→Rust round-trip: four `track_scene_loading_*` `#[func]`s and the `#[serde(flatten)] Value` payload are replaced by one typed struct + one `Loading(Box<SegmentEventLoading>)` variant. **The GLTF coordinator stays in GDScript by design** — it is `async`/`await`- and Godot-object-bound, so it does not belong in the pure-Rust funnel.

Each loading event is **flushed to Segment immediately** rather than waiting for the periodic batch, and `completed`/`progress` share one `snapshot_event` builder so their field-set can't drift. Together with the 10s pulse, that means an abandoned load leaves its `started` + last `progress` on record instead of nothing — the queue is in-memory and would otherwise lose an un-flushed event on a background/quit.

## Validation

18 `loading_funnel` unit tests, 0 failed. `clippy`, `gdformat`, `gdlint` clean.

On device (physical Android), four rounds: **realm bucketing** 11/11 loads correct across world→world, world→genesis, genesis→genesis, genesis→world; **cancel** reports `cancelled`; **slow connection** separates cleanly (`0.60 MB/s` / `11s` stalled / 72 pending vs `3.69 MB/s` / `~0` / 0); **delivery** confirmed end to end.

Four defects were found this way, none by the unit tests — they passed throughout, because they fed the funnel inputs the client never actually produces:

- **Cancelling reported `superseded`** — `return_to_discover()` tears the Explorer down without the hide effect, the only caller of `loading_end_episode`, so `cancelled` was unreachable in practice.
- **Genesis loads could never bucket as `genesis`** — the loading screen forwarded a realm narrowed to a bare dcl ENS for the places API (so non-world destinations fell back to the realm being *left*), and `realm_bucket` only knew short forms while the client resolves to URLs.
- **Throttled connections were invisible** — the network estimate was a single sample taken at begin, before any fetch is in flight, reading `0.0` on exactly the loads it needed to describe.
- **Six fields would have shipped noise** and were cut rather than documented around: `assets_expected`/`assets_loaded` (cumulative process counters, `loaded > expected` on 8 of 18 observed loads), `first_ready_ms` (`-1` on 15 of 18), `reached_ready`, `cold_cache` (an in-memory counter, not the disk cache), `network_estimate_mbs`. The schema is unpublished, so cutting them now is free.

The last round also confirmed a prediction from #2456 (merged): with the featured-card double-tap fixed at its source, the sub-100ms `superseded` episodes it produced are gone (0 of 11 loads).

## Known limitations

- **`process_assets_pending_*` measure the whole process, not the load** — hence the names. The ContentProvider exposes two cumulative `fetch_add` counters with no record of who requested each resource. Per-session accounting can't substitute: the session gates its own `Assets → Ready` transition on every asset it tracks having loaded, so it is zero by construction on any completed load. TODO on the field.
- **`scene_rendered_ms` is best-effort** (`-1` on a fair share of loads) — the session is dropped the instant it completes, so milestones depending on a later tick race the teardown. Evidence when set, not a rate.
- **`auto` episodes can bucket `unknown`** — they begin and end inside a realm switch, before any realm resolves.

## Issues

**Closes #2098** — the coordinator dedups N instances of one GLTF source into one shared download + one main-thread load, so they no longer each take a queue slot. Ticks item **P** of **#2536** (was #2477, already closed at consolidation).

**Left open as tracking — DONE criterion is now field data, not a repro.** This PR does not fix these; it makes them countable, so they can be prioritized by real frequency and later verified as fixed from the same query.

| Item | Measured with |
|---|---|
| **#2536 C)** zombie download slots | `network_stall_ms` + `process_assets_pending_peak` |
| **#2536 I)** asset accounting on force-ready | `process_assets_pending_at_end` with `dismissed_by=completion` |
| **#2536 L)** loading watchdog + cached empty-teleport dismissal | `dismissed_by=wall_clock_timeout` rate |
| **#2536 O)** retry failed/timed-out asset loads | `assets_errored` |
| **#2450** stuck at 25% | `time_in_25_30_band_ms` + `dismissed_by` |

**To open after merge — 2 issues**

1. **ContentProvider: attribute each requested resource to its requester.** Unblocks turning `process_assets_pending_*` into a real per-load figure, which C and I both need to be conclusive.
2. **The loading session completes while the process still owes content.** Measured here, not fixed: on device the session declared itself complete owing ~72–83 resources and the screen went away, with `scene_rendered_ms = -1` on the same load. The half-loaded-world symptom. Blocked on (1) for precise attribution.

**Before the data is trustworthy:** let one release bake, then read the `wall_clock_timeout` rate and the `network_avg_mb_s` distribution before touching C / I / L / O — the point of shipping this first is that their priority comes from the numbers.


godot/src/decentraland_components/gltf_container.gd
```diff
@@ -70,203 +61,88 @@ func async_load_gltf():
...
+		self, file_hash, dcl_gltf_src, dcl_scene_id, use_optimized, content_mapping
+	)
 
-	var content_mapping := Global.scene_runner.get_scene_content_mapping(dcl_scene_id)
-	var promise = Global.content_provider.load_scene_gltf(dcl_gltf_src, content_mapping)
 
-	if promise == null:
-		_finish_with_error("failed to start loading")
-		return
+#endregion
```
regenesis-claw, line 88: Very minor: if `change_gltf` fires during the frame between `_instantiate_and_add` and `_complete_shared_load` (the coordinator's `await get_tree().process_frame` gap), the old group's `_complete_shared_load` will see `dcl_gltf_loading_state == LOADING` (now belonging to the *new* request) and call `_complete_load`, marking the container FINISHED before the new load resolves.

A cheap guard: accept the group's hash and check `_requested_hash == hash` before completing:
```gdscript
func _complete_shared_load(hash: String, gpu_ms: float = -1.0) -> void:
    if _requested_hash != hash:
        return
    if dcl_gltf_loading_state != GltfContainerLoadingState.LOADING:
        return
    _complete_load(gpu_ms)
```
Extremely unlikely in practice (requires a CRDT source change on the exact realize frame), but closes the window.

regenesis-claw (approved):
Followed `REVIEW.md`.

## Pre-flight
- **Branch vs main**: BLOCKED (needs approval), CI all green.
- **iOS build**: iOS → R2 build present and passing. ✅
- **Submodule drift**: none.

## Verdict: Approve ✅

Excellent work. The two-piece architecture — pure Rust accumulator (`LoadingFunnel`) for the Segment schema, GDScript coordinator for the Godot-bound async pipeline — is clean, well-documented, and well-tested (17 unit tests covering healthy loads, throttled connections, cancels, supersedes, realm failures, plateau bands, and counter isolation between loads). The on-device validation methodology in the PR description is exemplary.

The GLTF coordinator's two-stage pipeline (wide network downloads → paced single-source-per-frame main-thread load+realize) is a smart decomposition. Retiring groups after realize to free `PackedScene` memory is the right call for mobile OOM sensitivity.

1. **Subtle race in `_complete_shared_load` across `change_gltf`**: if a CRDT update calls `change_gltf` (which unregisters from the old hash, sets state back to `LOADING`, and requests a new hash) during the frame between `_instantiate_and_add` and `_complete_shared_load`, the old group's `_complete_shared_load` sees `dcl_gltf_loading_state == LOADING` (now owned by the *new* request) and marks FINISHED prematurely. In practice this requires an entity's GLTF source to change during the exact realize frame — extremely rare — but passing the group hash into `_complete_shared_load` and checking `_requested_hash == hash` would close it cleanly.

2. **PR body says "deletes `loading_profiler.gd` (562 lines)"** but the file doesn't appear in the diff or on `main`. Was it already removed in an earlier PR, or is this a planned follow-up?

3. **No formal `## Test plan` with QA-executable steps** (per REVIEW.md §4). The `## Validation` section is thorough for engineering review but doesn't have the step-by-step format QA can run on a device. Since this is analytics instrumentation (no user-visible UI change), this is acceptable — but worth adding a brief case like "teleport to Genesis → loading screen appears and dismisses → check Segment dashboard for a `Loading Event` with `dismissed_by=completion`" if QA will validate the pipeline end-to-end.

### What I checked
- `loading_funnel.rs`: all 17 tests, state machine transitions, milestone set-once semantics, `classify_dismissal` precedence, realm bucketing with real on-device strings, network stall logic, plateau band accumulation.
- `gltf_loading_coordinator.gd`: download stage throttling + prioritization, load pump sequencing (one source/frame), group lifecycle (PENDING→DOWNLOADING→FETCHED→REALIZING→DONE/ERROR), late-waiter handling, memory cleanup via `_retire_group`.
- `gltf_container.gd`: simplified flow (delegate to coordinator), `_exit_tree` cleanup via `unregister`, `change_gltf` detach-before-reattach.
- `scene_manager.rs`: funnel integration points (begin/end episode, phase marks, progress band, scene spawn/render milestones, per-tick network sampling), `loading_content_counts` fallback.
- `loading_screen.gd` / `realm.gd` / `scene_fetcher.gd`: GDScript push-calls for begin/end episode, about-end, discovery, realm-change-failed.
- `data_definition.rs`: schema shape, `skip_serializing_if` on all optional fields, `Box<SegmentEventLoading>` to keep enum size down.
- No debug prints, no non-English comments, no `.claude/` files, no leaked editor settings.

Requested by Mateo via Slack

rodrigo685 (changes requested):
- [ ] Add a `fired_from` property to identify where the user was when loading (`in-world`  vs `pre_game`), as firing from in-world is done under a context of resource-heavy utilization and after phone has been actively working for some time, so we can analyze failures with that context.

## Concern
If we are dispatching the events only on `"completed" as stated on the PR, wouldn't we be blind if the load is unsuccessful or crashes and never completes or the client never dispatches it ?

If so,  let's add a "processing" or "progress" stage that forces dispatch based on a decremental pulse, with higher frequency at the beginning of the load and reduced (static) as it progresses past 20 seconds.

> _Nothing is emitted between started and completed — the funnel accumulates in memory and the milestones land as fields on completed..

kuruk-mm (approved):
Looks great

rodrigo685 (approved):

leanmendoza:
## Post-merge: analyzing the `Loading Event` data + a pilot dashboard

Now that the funnel is live, here is (1) the data model to query against, (2) the questions it answers, (3) a pilot dashboard to visualize what's already instrumented, and (4) next steps. This is the umbrella follow-up for **#2536** — the remaining open items there are prioritized off these numbers, not off a repro.

### 1. Data model

One event name — **`Loading Event`** — discriminated by `type` and correlated across a whole load by **`loading_id`** (one loading-screen episode = one id; every event of that load shares it). Group on `loading_id` and the funnel reassembles.

| `type` | Emitted | Key properties |
|---|---|---|
| `started` | load begins | `loading_id`, `schema`, `realm_bucket`, `is_world`, `when`, `cpu_count`, `platform` |
| `progress` | every 10s **while still in flight** | the `completed` snapshot **minus** `dismissed_by` / `frames` |
| `completed` | load ends, **always** | `+ duration_ms`, **`dismissed_by`**, `complete_ms`, `first_spawn_ms`, `scene_rendered_ms`, `about_end_ms`, `discovery_ms`, `time_in_25_30_band_ms`, `phase_breakdown{…}`, **`network_avg_mb_s`**, `network_peak_mb_s`, **`network_stall_ms`**, **`process_assets_pending_at_end`**, `process_assets_pending_peak`, `assets_errored`, `frames` |
| `asset_failure` | with `completed`, if any group failed | `loading_id`, `realm_bucket`, `reason`, `count` |
| `realm_change_failed` | on realm resolution failure | `loading_id`, `realm_bucket`, `reason` |

- `dismissed_by` ∈ `completion` / `wall_clock_timeout` / `error` / `superseded` / `cancelled`. **`completion` ⟺ the pipeline reached the `done` phase.**
- `realm_bucket` ∈ `world` / `genesis` / `other` / `unknown`. No PII (no coords, URLs, addresses).
- `when` — `on_explorer_ready` is the one cold **lobby→world** entry (**pre-world**); everything else is **in-world** (`on_teleport`, `on_world`, `on_moveto`, `on_goto_realm`/`on_changerealm`, `auto`). So **pre-world ⟺ `when == 'on_explorer_ready'`**. **Filter `when != 'auto'`** to drop background parcel streaming from user-visible loads.
- Milestones are ms-since-begin; **`-1` = never happened**. `phase_breakdown` = entry ms of each phase (`metadata`, `spawning`, `assets`, `ready`, `floating_islands`, `done`), same rule.

### 2. What to analyze — four signals

| Signal | Query | What it tells us |
|---|---|---|
| **Infinite loading** | `completed` where `dismissed_by = wall_clock_timeout` | The stuck-screen bug (#1640 / #2450), now countable per release |
| **Slow connection** | bucket on `network_avg_mb_s`; `network_stall_ms` = time owed content with ~nothing arriving | Throttled loads **hang rather than fail**, so they never show in `assets_errored` — this is their only signal |
| **Completed but incomplete** | `process_assets_pending_at_end > 0` **and** `dismissed_by = completion` | Screen dismissed while the client still owed content — the half-loaded-world symptom |
| **Abandoned load** | a `started` / `progress` `loading_id` with **no** `completed` | User quit / app killed mid-load, before the wall-clock timeout — the highest-signal failure, now visible because each event is flushed immediately |

### 3. Pilot dashboard

Scope every panel to **`when != 'auto'`** (unless noted) and window by release. Segment → your BI tool of choice; these are the panels to stand up first.

| # | Panel | Chart | Source field(s) | Reads |
|---|---|---|---|---|
| 1 | Loads this release | big number | count distinct `loading_id` (`type=started`) | volume / denominator |
| 2 | **Infinite-loading rate** | big number + trend | `% completed where dismissed_by=wall_clock_timeout` | the headline reliability KPI |
| 3 | **Abandoned rate** | big number + trend | `% loading_id with started & no completed` | users giving up mid-load |
| 4 | Duration p50 / p95 | big numbers | `duration_ms` where `dismissed_by=completion` | perceived speed |
| 5 | Load funnel | funnel / bar | share of loads reaching `first_spawn_ms` → `complete_ms` → `scene_rendered_ms` → `completion` (each `≠ -1`) | where loads drop off |
| 6 | Dismissal mix | stacked bar over releases | `completed` grouped by `dismissed_by` | is the timeout slice shrinking? |
| 7 | Duration distribution | histogram, split by `realm_bucket`, `is_world`, `when=='on_explorer_ready'` | `duration_ms` | cold-entry vs in-world, world vs genesis |
| 8 | Connection quality | histogram + scatter | `network_avg_mb_s`; `network_stall_ms` vs `duration_ms` | slow-connection cohort |
| 9 | **25–30% band** | histogram, split by `dismissed_by` | `time_in_25_30_band_ms` | #2450 directly: is it real timeout or a load that only *looks* frozen |
| 10 | Assets owed at end | bar | `process_assets_pending_at_end` (\|`dismissed_by=completion`) + `assets_errored` | completed-but-incomplete + failures |
| 11 | Phase breakdown | stacked bar | per-phase deltas derived from `phase_breakdown` entry ms | which phase eats the time |

Funnel stages (panel 5), each stage = a milestone field being set (`≠ -1`):

```mermaid
flowchart LR
  A["started<br/>(loading_id)"] --> B["first_spawn_ms"]
  B --> C["complete_ms<br/>(reached done)"]
  C --> D["scene_rendered_ms<br/>(best-effort*)"]
  C --> E["completed<br/>dismissed_by=completion"]
```

\* `scene_rendered_ms` is best-effort (`-1` on a share of loads — the session is torn down the instant it completes, so a milestone on a later tick races the teardown). Treat panel 5's last stage as evidence-when-present, not a hard rate.

### 4. Next steps

- **Let one release with #2540 bake first.** Read the `dismissed_by=wall_clock_timeout` rate and the `network_avg_mb_s` distribution **before** touching **#2536 C / I / L / O** — the whole point of shipping the funnel first is that their priority comes from real frequency, not a guess.
- **Two issues to open** (both blocked on precise per-load asset attribution):
  1. **ContentProvider: attribute each requested resource to its requester.** Today `process_assets_pending_*` count the whole process (two cumulative `fetch_add` counters, no requester record), so they can't be turned into a real per-load figure. Unblocks making C and I conclusive.
  2. **The loading session completes while the process still owes content.** Measured here, not fixed: on device the session declared itself complete owing ~72–83 resources and the screen went away (`scene_rendered_ms = -1` on the same load). The half-loaded-world symptom. Blocked on (1) for attribution.

Tracking for everything else stays on **#2536**.

PR #2546, opened 2026-07-16
fix(camera): stop camera clipping through walls — collision + proximity fade (#1814)

### LINKED ISSUE #1814: Implement see-through walls when camera gets too close to surfaces
### Feature Description
When the camera gets very close to walls, floors, or other solid surfaces, the player should be able to see through them. This prevents the view from being completely blocked when navigating tight spaces or when the camera clips into geometry.

### Current Behavior
When the camera intersects with solid geometry (walls, floors, objects), the view becomes blocked or shows clipping artifacts.

### Proposed Behavior
Implement a system that makes surfaces transparent or see-through when the camera gets within a certain distance threshold. This could be achieved by:
- Adjusting camera near plane dynamically
- Making materials transparent when camera proximity is detected
- Implementing a shader-based solution that fades out surfaces near the camera

### Use Cases
- Navigating tight indoor spaces where the camera might clip into walls
- First-person view when standing very close to objects or other avatars
- Preventing disorientation when the camera gets pushed into geometry

### Related Issues
- #1446 — Avatar displays multicolor glitch lines when standing up close (this feature would address similar camera proximity issues)

### Platform
All platforms (mobile, desktop, VR)

### Technical Considerations
This may require shader modifications or render layer adjustments to handle transparency when surfaces are within the camera near plane distance.

---
Requested by Rodri via Slack

manuelmaceira:
## What

Fixes the third-person camera rendering **through solid geometry** in tight indoor spaces (issue #1814), with a **hybrid approach**:

1. **Collision** — the camera respects physically solid geometry (the SpringArm actually shortens).
2. **Proximity fade** — visual safety net for everything collision can't catch (mismatched/missing colliders, the lateral over-shoulder offset, wall corners).

## Root causes fixed (commit 1)

Two distinct root causes in the player camera rig (`godot/src/logic/player/player.tscn` + `player.gd`):

### 1. SpringArm collision mask (primary)
The camera `SpringArm3D` (`Mount`) set no `collision_mask`, falling back to the engine default `1` (`CL_POINTER`). Solid world geometry is on `CL_PHYSICS` (layer 2 — what the player body collides with), and many DCL scene colliders are **physics-only**, so the arm never shortened → camera passed straight through those walls.

- Set `Mount.collision_mask = 2` (`CL_PHYSICS`), matching the player body.
- Add `Mount.margin` so the near plane stops short of the wall face.

### 2. Over-shoulder offset moved the pivot outside the collision capsule
The over-shoulder offset (`0.75`) was applied by translating the **SpringArm pivot itself**, placing the cast origin `0.75` lateral — outside the `0.25`-radius capsule. That lateral segment isn't collision-swept, so hugging a wall the cast could start behind it.

- Insert a `CameraArm` (`Node3D`) below the arm and reparent `Camera3D` under it; apply the over-shoulder X offset on the **camera**, keeping the arm's pivot centered so the collision cast starts from a protected point. The camera keeps the exact same world position/framing as before when fully extended.
- Third-person `spring_length` now uses the pure back distance (`.z = 3.0`) instead of the diagonal `.length()` (~3.09), since the lateral part is handled by the camera offset.

## Proximity fade safety net (commit 2)

Even with perfect collision, scene geometry whose visuals don't match its collider — or that has **no collider at all** (GLTF visible meshes default to `cmask=0`!) — can still end up around the camera. Instead of black inside-out faces, geometry very close to the camera now dissolves out via `BaseMaterial3D.distance_fade` (`PIXEL_DITHER`):

- **Per-fragment** camera distance → works for meshes of any size (unlike `visibility_range`, which is evaluated from the node origin and fails for large walls).
- Stays in the **opaque pipeline** (no transparency sorting side-effects).
- Fully transparent at **0.1m**, fading in up to **0.3m**.

Applied at the two scene-content material chokepoints:
- `gltf_container.gd` → `fix_material` (all GLTF scene materials)
- `material.rs` → `apply_dcl_material_properties` (SDK `Material` on primitives; values mirrored with a cross-reference comment)

Tuning invariant: SpringArm `margin` (bumped `0.2 → 0.3`) ≥ fade start, so a wall the camera is **legitimately** pressed against stays fully visible — the fade band is only ever entered when collision already failed. Values centralized in `proximity_fade.gd` (`FADE_GONE_DISTANCE` / `FADE_START_DISTANCE`).

Deliberate gaps: custom `ShaderMaterial`s (SDK unlit with a separate alpha texture — rare) and the engine fallback material on primitives with no `Material` component.

## How it's structured

Rig config/decision extracted into a pure, `Global`-free `CameraRigHelpers` so the invariants are unit-testable headless without the full player scene or the Rust extension.

## Automated tests

**`godot/src/test/player/test_player_camera_collision.gd`** (standalone `extends SceneTree`, headless):
- **SpringArm mask semantics** against synthetic walls: physics-only wall shortens the arm with `mask=2`; the old default `mask=1` misses it (the bug); pointer-only colliders are ignored; the typical DCL scene collider (layer 1|2) is detected.
- **Rig targets**: third/first-person `spring_length` + camera offset; asserts third-person uses `.z`, not the diagonal.
- **Scene guard**: parses `player.tscn` to pin `Mount.collision_mask`, the centered pivot, and the `CameraArm` structure.

**`godot/src/test/decentraland_components/test_proximity_fade.gd`** (same pattern):
- Fade properties applied correctly; band sanity (`0 < gone < start`); both hooks guarded (incl. reading the Rust file via absolute path); `margin ≥ fade start` invariant against the real `player.tscn`.

Red→green verified on the collision suite (reverting the mask to `1` turns the physics tests red). `gdformat`/`gdlint`/`cargo fmt`/`clippy` clean.

---

## QA test steps

**Builds:** CI artifacts from this PR's build comment (desktop / Android APK / iOS TestFlight).
**Where:** any scene with interiors — Genesis Plaza (0,0) is the easiest spot.

> Heads up: when the camera gets really close to a wall or object (~30 cm), it now shows a soft dithered fade. That's intended, not a bug.

### Third-person camera (what used to break)

1. Go to Genesis Plaza. Walk backwards into an interior wall (try to shove the camera inside the wall).
   **Expected:** the camera shortens and doesn't go through the wall.
2. While hugging a wall, rotate the camera a full 360°.
   **Expected:** no angle shows through the wall.
3. Strafe so the wall is on the avatar's right side (the side the camera leans toward) and rotate.
   **Expected:** still no see-through.
4. Stand in an interior corner, rotate the camera and pitch it up/down.
   **Expected:** the camera doesn't clip into the corner.
5. Walk through doorways and narrow corridors.
   **Expected:** the camera pulls in and recovers smoothly, without clipping through the frames.

### Objects without collision (safety net)

6. Find decorative objects / furniture with no collision and push the camera inside them.
   **Expected:** the object fades out softly around the camera instead of showing black inside faces.
7. Press the camera against a wall that does have collision (the camera stops on its own).
   **Expected:** the wall stays fully visible — it must NOT fade.

### Regression (old stuff still works)

8. Compare with the production client: third-person framing should look the same (avatar shifted to the left, same distance).
9. Switch between 1st/3rd person a few times, also while pressed against a wall (mobile: camera button in the HUD).
   **Expected:** smooth transitions; 1st person centered in the head, 3rd person back to normal.
10. In first person: look straight up/down and walk into walls.
    **Expected:** nothing weird, same as before.
11. Click clickable scene items and hover over avatars.
    **Expected:** clicks register and the avatar highlight outline still works.
12. Enter and exit areas where the scene takes over the camera (cutscenes, special zones).
    **Expected:** the camera returns to the normal mode afterwards.
13. Walk around a busy scene (lots of models).
    **Expected:** nothing fades far from the camera; performance feels normal.

**Platforms:** desktop (at least one of Windows / macOS / Linux) + Android or iOS.


regenesis-claw (approved):
Thorough, well-structured fix — both root causes identified and addressed cleanly.

## What I checked (following REVIEW.md)

**Pre-flight:**
- ✅ Branch is up to date with `main` (0 commits behind)
- ⏳ iOS build still in progress — change touches `player.tscn`, `material.rs`, and `gltf_container.gd` (all platform-visible), but nothing iOS-specific (no native plugins, deeplinks, or `OS.get_name() == "iOS"` branches). The proximity fade uses standard `BaseMaterial3D` distance fade which is engine-level — no platform-specific codepath.
- ✅ No submodule pointer drift

**Collision fix:**
- SpringArm `collision_mask = 2` (CL_PHYSICS) — correct, matches the player body mask.
- Pivot stays centered; over-shoulder offset moved to camera node below the arm via `CameraArm` — collision cast now originates from inside the capsule. Clean separation.
- `spring_length` uses `.z` (back distance) not `.length()` (diagonal) — correct since the lateral is handled by camera offset.

**Proximity fade:**
- Applied at both scene-content material chokepoints (GLTF + SDK Material). PIXEL_DITHER stays in the opaque pipeline — no transparency sorting issues.
- Tuning invariant (margin ≥ fade start) is documented and tested. A wall the camera is legitimately pressed against never enters the fade band.
- Rust hardcodes 0.1/0.3 with a sync comment + the test guards the cross-reference — pragmatic approach.

**Tests:**
- Excellent coverage: synthetic physics collisions (red→green verified), rig targets, scene guard parsing, proximity fade properties, cross-reference guards, and the margin invariant. All headless-compatible.

**Minor note — test plan format:**
The PR has a verification checklist but not the QA-executable format (numbered steps from app open, observable expected result). For a camera behavior change that QA will test on-device, consider adding a case like:

> **Case — Third-person camera against walls**
> 1. Open the app and enter a scene with tight indoor spaces.
> 2. Switch to third-person view.
> 3. Walk into a corner and orbit the camera so it presses against walls.
> **Expected:** Camera pushes away from walls; no see-through-geometry. Clipped geometry dissolves softly instead of black inside-out faces.

Not blocking — the automated tests are strong and the manual verification section covers the intent.

LGTM 🚀

rodrigo685 (changes requested):
Checked this in Checked in Flagtag and Tower of Madness

Standing right next to walls and performing a camera rotation in the wall direction makes the camera clip-through

Expected Behavior: Camera bounces off to maintain view of the character possition, without cliping through objects. Camera rotation towards the colliding object does not perform or bounces off.

Repro Steps
- Go to Flagtag
- Stand next to a building wall mesh
- Rotate the camera towards the mesh
- Be able to see through walls

image


https://github.com/user-attachments/assets/d2492572-2cf9-481e-a4d5-caa61a75c902

rodrigo685 (approved):

kuruk-mm (approved):
Great work!

PR #2547, opened 2026-07-16
feat: username picker with minted names support

sdilauro:
## Summary
- Adds `UsernamePicker` component that shows a dropdown of NFT-minted names (with verified icon) or a free-text input depending on whether the user owns any names
- Extends `DropdownList` with `are_names` mode (verified icon next to selected name, coloured label states per item) and fixes touch/mouse close behaviour for all dropdowns
- Updates `ProfileEditor` to use `UsernamePicker`; saving now correctly sets `has_claimed_name` based on whether a minted name was selected

## Test plan
- [ ] Open profile editor with an account that has **no minted names** → only free-text input is shown, tag `#XXXX` visible
- [ ] Open profile editor with an account that has **one or more minted names** → tabs (Unique / Non-unique) appear; Unique tab shows dropdown with verified icon on selected name
- [ ] Switch between Unique and Non-unique tabs → input switches correctly, `has_claimed_name` reflects the right value on save
- [ ] Select a minted name from the dropdown → verified icon appears on the button, name is populated in text field
- [ ] Save with a minted name selected → profile saved with `has_claimed_name = true`
- [ ] Save with free-text name → profile saved with `has_claimed_name = false`
- [ ] Open and close the profile editor rapidly (race condition) → dropdown never shows duplicated names
- [ ] Smoke-test all other dropdowns (gender, sexual orientation, relationship status) on mobile → open/close correctly on tap


regenesis-claw (changes requested):
Nice work on the `UsernamePicker` component — clean extraction from `ProfileEditor`, good async race guard with `_load_generation`, and the touch/mouse close behavior fix for dropdowns is solid. Followed REVIEW.md.

## Findings

### 🔴 `has_claimed_name` change not tracked by dirty check or save logic

`_check_dirty()` only compares `username_picker.get_name_value()` against `_original_values["username"]`. The `_async_save_profile()` gate is:

```gdscript
if current_username != _original_values.get("username", ""):
    mutable_profile.set_name(current_username)
    mutable_profile.set_has_claimed_name(username_picker.get_is_claimed())
```

If a user owns a minted name that matches their current non-unique display name (or toggles between Unique/Non-unique tabs without changing the text), the claim status changes but:
1. The Save button stays disabled (dirty check misses it)
2. Even if forced, `set_has_claimed_name` is never called (guarded by name-text diff)

**Fix:** Store `"has_claimed_name"` in `_original_values`, add it to `_check_dirty()`, and decouple the `set_has_claimed_name` call from the name-text gate in `_async_save_profile`.

### 🟡 `add_item` has unused `is_name` parameter

`add_item(label, id, is_name)` stores `is_name` in the item dictionary, but `_sync_popup_items` ignores it and uses `are_names and i > 0` instead. The parameter is dead code — either use it or remove it.

### 🟡 `Label_Selected` lost clip/overrun behavior

The `.tscn` diff removes `clip_text = true` and `text_overrun_behavior = 3` from `Label_Selected`, and `size_flags_horizontal = 3` (expand+fill) is gone too. The comment in the script says "all current dropdown options fit" — but minted names can be long. On a narrow mobile screen this could overflow the button bounds. Consider keeping `text_overrun_behavior` as a safety net, or at least `clip_text`.

## Positives

- `_load_generation` pattern for async race prevention is exactly right
- `_updating_from_dropdown` guard avoids signal loops cleanly
- Touch input fix (`InputEventScreenTouch` + `InputEventScreenDrag`) and close-on-release (not press) for the popup layer is a real mobile UX improvement that benefits all dropdowns
- `mouse_filter = 2` added to separators — good hygiene
- Component placement in `molecules/username_picker/` is correct per Atomic Design

## CI / iOS

- Static checks ✅
- Linux, Android, Clippy, iOS builds still in progress — will need to verify before merge
- This PR is pure GDScript/UI, no platform-specific branches — iOS build not strictly required but it's already running

Requesting changes for the dirty-check/save bug (finding #1). The other two are suggestions.

regenesis-claw (approved):
Re-review after 5b14bfeb — delta addresses previous findings:

| # | Finding | Status |
|---|---------|--------|
| 🔴 `has_claimed_name` not tracked in dirty check / save | ✅ Fixed — stored in `_original_values`, added to `_check_dirty()`, and `set_has_claimed_name` is now its own independent `if` block in `_async_save_profile` |
| 🟡 Dead `is_name` param in `add_item` | ✅ Removed |
| 🟡 `Label_Selected` clip/overrun | ⚪ Not addressed (minor, acceptable as-is given the wider button) |

The fix in `_async_save_profile` correctly decouples the name-text change from the claim-status change — both are now independent conditions. Clean and minimal diff.

Static checks ✅. Builds still running but no code changes that would affect them differently from the previous green pass.

Approving.

rodrigo685 (changes requested):
- [x] Based on Eli's comment, the "Select Name" placeholder and also having it selectable in the dropdown is super weird as it's not a valid option, let's remove if possible.

Image

rodrigo685 (approved):
Approved ! Looking great

kuruk-mm (approved):
looks great!

PR #2560, opened 2026-07-20
fix: Animator loop parameter not respected on mobile (#2497)

### LINKED ISSUE #2497: [Bug] Animator component: Loop parameter not respected on mobile
### Platform
Mobile (Android/iOS)

### Version
v1.10.0

### Device
Pixel 6A

### Issue description
The `Loop` parameter of the Animator component does not work correctly on mobile. Animations keep looping even when the `Loop` parameter is disabled.

### Expected behavior
When `Loop` is unchecked, the animation should play once and stop (matching desktop/PC behavior), instead of looping indefinitely.

### Screenshots/Media

Image

### Steps to reproduce
1. Add an Animator component to an entity with `Start playing` enabled and `Loop` disabled.
2. Play the scene on mobile.
3. Observe that the animation keeps looping instead of playing once and stopping.

### Occurrence
Consistent (100% of the time on mobile)

---
Requested by Rodri via Slack

EibrielInv:
## Issue

Fixes #2497

On mobile (Android/iOS) an `Animator` state with `loop` disabled and `playing` enabled keeps looping indefinitely instead of playing once and stopping. Desktop/PC behaves correctly. Reported as consistent (100%) on device.

## Root cause

In the single-animation path of `apply_anims` (`lib/src/godot_classes/animator_controller.rs`), the animation loop mode was only ever set to `LINEAR` when `loop` was `true` — it was **never reset to `NONE`** when `loop` was `false`:

```rust
if state.r#loop.unwrap_or(true) {
    anim.set_loop_mode(LoopMode::LINEAR);
}
```

So the clip kept whatever loop mode it was imported with. On desktop the GLTF import yields a non-looping default, but the **mobile GLTF import path can leave the clip with a looping mode**, so it loops forever regardless of the `loop` flag — hence the platform-specific bug.

## Fix

Set the loop mode explicitly in both cases (`LINEAR` when looping, `NONE` otherwise), so behavior no longer depends on the imported default:

```rust
if let Some(mut anim) = anim_player.get_animation(&anim_name) {
    let loop_mode = if state.r#loop.unwrap_or(true) {
        LoopMode::LINEAR
    } else {
        LoopMode::NONE
    };
    anim.set_loop_mode(loop_mode);
}
```

The multi-animation `AnimationTree` path is unaffected (it stops non-looping clips via `_animation_finished`), which is why this only surfaced for simple single-clip animators.

## QA — how to reproduce & verify

A dedicated test block was added to the test scene, deployed at coordinates **`-31,19`**.

1. Open the test scene on a **mobile device** (Android or iOS) — this is where the bug reproduces.
2. Walk to parcel **`-31,19`**.
3. In the test switcher panel (right side), tap the **"Issue 2497"** tab.
4. Two confetti bursts spawn side by side:
   - **Left** — labelled `loop = false (play once)`
   - **Right** — labelled `loop = true (control)`
5. **Expected (fixed):** the **left** burst plays its animation **once and stops**; the **right** burst keeps looping.
   **Before the fix:** the left burst also looped forever.
6. Tap **"Replay (loop=false)"** in the on-screen panel to re-fire the left (non-looping) clip and confirm it again plays a single burst and stops.
7. Confirm on **desktop** that behavior is unchanged (no regression): left plays once, right loops.

### DIFF
--- a/lib/src/godot_classes/animator_controller.rs
+++ b/lib/src/godot_classes/animator_controller.rs
@@ -671,10 +671,17 @@ pub fn apply_anims(gltf_container_node: Gd<Node3D>, value: &PbAnimator) {
             if let Some(clip_name) = resolved_clip {
                 let anim_name = StringName::from(&clip_name);
                 anim_player.set_speed_scale(state.speed.unwrap_or(1.0));
-                if state.r#loop.unwrap_or(true) {
-                    if let Some(mut anim) = anim_player.get_animation(&anim_name) {
-                        anim.set_loop_mode(godot::classes::animation::LoopMode::LINEAR);
-                    }
+                // Always set the loop mode explicitly. Only setting it when looping
+                // leaves the animation carrying whatever loop mode it was imported
+                // with, which on mobile can be a non-NONE mode, causing it to keep
+                // looping even when Loop is disabled.
+                if let Some(mut anim) = anim_player.get_animation(&anim_name) {
+                    let loop_mode = if state.r#loop.unwrap_or(true) {
+                        godot::classes::animation::LoopMode::LINEAR
+                    } else {
+                        godot::classes::animation::LoopMode::NONE
+                    };
+                    anim.set_loop_mode(loop_mode);
                 }
                 anim_player.play_ex().name(&anim_name).done();
             }
--- a/lib/src/godot_classes/animator_controller.rs
+++ b/lib/src/godot_classes/animator_controller.rs
@@ -641,8 +641,16 @@ pub fn apply_anims(gltf_container_node: Gd<Node3D>, value: &PbAnimator) {
         // Duplicate the AnimationLibrary so each instance has independent animation
         // state. PackedScene.instantiate() shares resources, so without this all
         // instances of the same GLTF would share the same animation objects.
+        //
+        // deep(true) (Godot's duplicate(subresources = true)) is required: a shallow
+        // duplicate copies the library container but keeps pointing at the SAME
+        // Animation sub-resources. Two GltfContainers referencing the same GLB content
+        // hash would then share one Animation, so set_loop_mode / set_speed_scale below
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
Clean, minimal fix. The root cause analysis is spot-on: the old code only set `LINEAR` when `loop=true` but never reset to `NONE` when `loop=false`, leaving the clip with whatever mode the GLTF importer gave it — which differs between desktop and mobile import paths.

The fix correctly restructures the conditional to always set the loop mode explicitly (`LINEAR` or `NONE`), making behavior deterministic regardless of the imported default. The comment explains the *why* clearly.

No concerns:
- No log-level changes (no Sentry impact)
- No platform-specific branches (no iOS build needed)
- Single-animation path only; multi-animation `AnimationTree` path unaffected (as noted in the PR)
- Test scene at `-31,19` with a dedicated tab — good QA coverage

LGTM ✅

Requested by Mateo via Slack

Ludmilafantaniella (approved):
Tested on iOS, Android and Windows. ✅

**Mobile (iOS & Android) — main fix:**
Verified at `-31,19` (Issue 2497 tab):
- `loop = false` plays once and stops correctly ✅
- `loop = true` keeps looping as expected ✅
- "Replay (loop=false)" re-fires the non-looping clip without affecting the looping one ✅
Compared against previous builds where the bug reproduced — behavior is now correct.

**Desktop (Windows) — regression:**
- Animation behavior unchanged ✅
- Scene emotes and actions in Genesis Plaza working correctly ✅

LGTM 🚀

https://github.com/user-attachments/assets/64e938a1-9899-449f-9d8a-e18f175130ed

https://github.com/user-attachments/assets/ce7278e3-3e31-4af3-af83-3bf498f18774

https://github.com/user-attachments/assets/0d324619-3eed-4685-a73f-47c4c78b4f83

kuruk-mm (approved):
code looks great!

PR #2564, opened 2026-07-21
fix: scene-inspector command timeout on corrupted WebSocket frames + wire in-game Pause Scene toggle

### LINKED ISSUE #2563: Bug: Scene inspector timeouts on some commands
Using some commands in the inspector results in timeout

EibrielInv:
## Summary

Fixes two issues surfaced while using the scene-inspector debug channel:

1. **Inspector command timeouts from malformed inbound JSON.** `SceneInspectorWebsocket._handle_message` previously did a single `JSON.parse_string()` and bailed on any failure. Two things broke this:
   - Packets can carry several JSON commands concatenated (the relay coalesces consumer→device frames).
   - Godot's `WebSocketPeer` inbound path can hand us a byte-corrupted frame — a brace duplicated at an object boundary (e.g. `...{"depth":1}}},"id":...`) — which closes the object early and drops the trailing `id`. With no `id`, the ACK is never sent and the caller times out.

2. **In-game "Scene Paused" toggle did nothing.** `menu.gd` only wired `request_pause_scenes` for the pre-explorer path, so the toggle was inert once in the explorer.

## Changes

**`godot/src/logic/scene_inspector_websocket.gd`**
- Fast path unchanged: a single clean frame still parses and dispatches as before.
- On failure, split the packet into balanced top-level `{...}` objects via `_next_balanced_object`, which respects string literals and escapes so payload braces don't miscount.
- Detect the corruption signature (a leftover tail beginning with `,`), splice the tail back on minus the spurious closing brace, and re-validate. Repaired frames dispatch silently — no log noise.
- Parse via `JSON.new().parse()` (`_parse_silent`) instead of `JSON.parse_string()` to avoid the "Expected 'EOF'" error-console spam on malformed input.
- Refactor dispatch into `_dispatch_frame`, which now rejects frames with an empty `id` so the repair path can kick in.

  > The proper fix belongs in `WebSocketPeer`; this keeps the debug channel usable meanwhile.

**`godot/src/ui/explorer.gd`**
- Connect `settings_node.request_pause_scenes` to the existing `_on_control_menu_request_pause_scenes` handler so the in-game toggle works.

## Testing
- Manual: exercised the scene-inspector debug channel; corrupted frames that previously timed out now dispatch and ACK.
- Manual: verified the in-game "Scene Paused" toggle pauses/resumes scenes.

Closes #2563


kuruk-mm (approved):

regenesis-claw (approved):
Clean, well-documented fix. Two focused changes with clear rationale:

**WebSocket corruption repair** — the brace-balanced splitter in `_next_balanced_object` correctly respects string literals and escapes, and the repair heuristic (tail beginning with `,` → splice minus spurious brace) is a pragmatic workaround for the upstream `WebSocketPeer` issue. Using `JSON.new().parse()` instead of `JSON.parse_string()` is the right call to avoid error-console noise on expected malformed input — no Sentry quota impact.

**Pause Scene toggle** — straightforward signal wiring that was simply missing for the explorer path. Good inline comment explaining why.

No platform-specific code touched → iOS build not required. All CI checks pass.

Approving — no issues found.

PR #2566, opened 2026-07-21
fix(sentry): report release as clean semver so build/version filtering works

kuruk-mm:
Makes Sentry able to filter by build/version. Today the release string (`...@1.11.0.1205-01da818-prod`) isn't valid semver, so `release.build:>=N` and `release.version:>=X` return nothing and adoption/regression tracking is disabled. This changes it to clean semver `...@1.11.0+1205`, moving the build number into the numeric `+build` slot. The commit hash and env aren't lost — they're already sent as Sentry `dist` and `environment`.

**Changes**
- `lib/build.rs` — emit `GODOT_EXPLORER_SENTRY_RELEASE = {version}+{build}`.
- `lib/src/godot_classes/dcl_global.rs` — add `DclGlobal.get_sentry_release()`.
- `godot/src/project_main_loop.gd` — use it for `options.release`.

`GODOT_EXPLORER_VERSION` (used by `is_production`, `version_gate`, UI, logs) is untouched.

**QA:** none needed — no behavior change, only the Sentry release string. Optional: after a build ships, check Sentry Releases shows `@<version>+<build>` and `release.build:>=<build>` returns results.

### DIFF
--- a/godot/src/project_main_loop.gd
+++ b/godot/src/project_main_loop.gd
@@ -46,7 +46,9 @@ func _initialize() -> void:
 	if DclGlobal.is_telemetry_disabled():
 		return
 
-	var release_string = "org.decentraland.godotexplorer@" + DclGlobal.get_version()
+	# Clean semver (`{version}+{build}`) so Sentry parses release.version / release.build.
+	# Commit hash and env are carried separately via `dist` / `environment` below.
+	var release_string = "org.decentraland.godotexplorer@" + DclGlobal.get_sentry_release()
 
 	# Detect environment from version string
 	self.is_dev_version = DclGlobal.is_dev()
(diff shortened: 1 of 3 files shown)


leanmendoza (changes requested):
Check if we can still sending the real version string in some custom parameter so we can distinguish between non-prod build from prod ones

kuruk-mm:
> Check if we can still sending the real version string in some custom parameter so we can distinguish between non-prod build from prod ones

we are already doing that with tag environment

check main tags:
image

(all tags [here](https://dcl-regenesis-labs.sentry.io/issues/7469688105/distributions/?environment=staging&project=4510187688361984&query=is%3Aunresolved&referrer=issue-stream&sort=recommended))

leanmendoza (approved):

Issue #672, opened 2025-09-29
Discover v1


rodrigo685:
Redesign Current HUB and behavior so it uses DCL Library Components, has a consistent UX with and acts as the gateway to get players in-world.

- Add Events
- Review All Buttons
- Review News Banner
- Review Search UX
- Review Sorting
- Fail States
- Empty States
- Review "Last Visited" so we bring player counters (not fetching now)

Issue #339, opened 2024-02-28
(80, -8) validate-player-authenticity


sdilauro:
UI Says: "Your dark schemes are not welcome here, go done" even when I log in with Google or Metamask.

![image](https://github.com/decentraland/godot-explorer/assets/22157752/dcf7347b-9328-416e-b334-5439e8fbfe76)

PR #2567, opened 2026-07-22
feat: implement sdk component asset preload

leanmendoza:
## What

Implements the SDK `PBAssetLoad` / `PBAssetLoadLoadingState` components (ids `1213` / `1214`) so scenes can **pre-load assets** ahead of time and avoid download / processing / GPU-upload stalls (pop-in / flickering) when the matching `GltfContainer` later appears. Mirrors the desktop `bevy-explorer` implementation (protocol PR #339, bevy PR #672) on godot/mobile.

A scene sets `PBAssetLoad { assets: [<path>, ...] }` on an entity; the client resolves each path, downloads + processes it, and reports progress back per-asset via `PBAssetLoadLoadingState` (`LOADING` / `FINISHED` / errors).

## How it works

- New handler `lib/src/scene_runner/components/asset_load.rs`, wired into the scene tick as two new `SceneUpdateState` steps (`AssetLoad` reacts to the dirty component; `SyncAssetLoad` polls promises and writes back the loading state).
- Preloads are **refcounted per content-hash**, so duplicate paths across entities load once, and reuse the exact same optimized-vs-runtime decision a real `GltfContainer` makes (so the later load is a genuine cache hit, not a near-miss).
- For GLTF it does the main-thread `ResourceLoader::load` and **retains the `PackedScene`** while `PBAssetLoad` stays on the entity — in this engine the GPU upload is paid at `load`, not `add_child`, so meshes/textures stay GPU-resident and the later real load is instant. Dropped (frees RAM/GPU) on component removal, entity deletion, or scene unload.
- Textures are decoded + GPU-uploaded via the texture cache; other non-GLTF paths warm the download cache on disk.

Also fixes component `1214` registration: the SDK exposes `AssetLoadLoadingState` as a **grow-only value set**, but `build.rs` defaulted it to LWW — added it to the `GROW_ONLY_SET_COMPONENTS` whitelist and write back with `.append()`.

## Testing

A demo scene lives at **MannakiaWorld** (`mannakia.dcl.eth`). On **each** run:

1. **Clean the cache** (fresh download every time — otherwise a warm cache hides the difference).
2. Go to `mannakia.dcl.eth` — the AssetLoad Preload demo loads by default.
3. The scene flips a coin per run (random 50%): **PRELOAD ALL** or **NO PRELOAD**, shown in the UI banner.
4. Hit **Instantiate**:
   - When the run **did** preload → instantiation is noticeably **faster** (assets already downloaded / uploaded).
   - When it **did not** preload → you see the full download + upload stall.

Repeat (clean cache each time) to compare both cases.


lib/src/scene_runner/components/asset_load.rs
```diff
@@ -0,0 +1,476 @@
...
+            reported_state: LoadingState::Unknown as i32,
+        });
+    }
+
+    // If the component was deleted (no paths), the entity is simply dropped.
+    if !tracked.is_empty() {
+        scene.asset_load.entities.insert(*entity, tracked);
+    }
+}
+
```
regenesis-claw, line 176: This `release_preload` takes `&mut Scene` which makes it uncallable from `deleted_entities.rs` (where you only have partial borrows). Consider refactoring the core logic to operate on `&mut AssetLoadState` so it can be reused for entity-deletion cleanup — see the top-level review comment.

lib/build.rs
```diff
@@ -15,11 +15,14 @@ struct Component {
 
 const PROTO_FILES_BASE_DIR: &str = "src/dcl/components/proto/";
 const COMPONENT_BASE_DIR: &str = "src/dcl/components/proto/decentraland/sdk/components/";
-const GROW_ONLY_SET_COMPONENTS: [&str; 4] = [
+const GROW_ONLY_SET_COMPONENTS: [&str; 5] = [
     "PointerEventsResult",
     "VideoEvent",
```
regenesis-claw, line 20: Nit: `per per-asset` → `per-asset`.

regenesis-claw (commented):
Clean, well-documented implementation — the doc comments, refcounting design, and budget-paced `advance_preloads` are all solid. One real issue and a couple of minor notes.

## Findings

### 🔴 Missing cleanup in `deleted_entities.rs`

`update_deleted_entities` cleans up `audio_sources`, `video_players`, `trigger_areas`, `tweens`, `gltf_loading`, etc. for every died entity — but it does **not** touch `scene.asset_load.entities`. When a scene deletes an entity that had `PBAssetLoad`:

1. The `TrackedAsset` entries stay in `asset_load.entities` forever (until scene unload).
2. `release_preload` is never called, so refcounts are never decremented.
3. The `PreloadEntry` (retained `PackedScene` + optional ghost node) is never freed.
4. `sync_asset_load_loading_state` keeps iterating over the dead entity's assets every tick.

On scene unload the whole `Scene` drops so it doesn't leak across scenes, but within a long-lived scene that creates/destroys entities with `PBAssetLoad`, GPU memory and tick cost accumulate.

Suggested fix in `deleted_entities.rs` (after the existing cleanups):

```rust
// Release preload refs for deleted entities
if let Some(tracked) = scene.asset_load.entities.remove(deleted_entity) {
    for asset in &tracked {
        if !asset.hash.is_empty() {
            super::components::asset_load::release_preload_by_hash(
                &mut scene.asset_load, &asset.hash,
            );
        }
    }
}
```

You'd need to extract the refcount-decrement + drop logic from `release_preload` into a public helper that takes `&mut AssetLoadState` (since `deleted_entities` doesn't have a full `&mut Scene` borrow to call the current `release_preload` which borrows `scene` broadly). Alternatively, just inline the `saturating_sub` + `remove` + `ghost.queue_free()` logic.

### 🟡 `release_preload` visibility

`release_preload` is currently `fn release_preload(scene: &mut Scene, hash: &str)` which borrows all of `Scene`. For the fix above (and for general testability), consider making the core logic take `&mut AssetLoadState` instead, so it can be called from `deleted_entities.rs` without needing a full `&mut Scene`.

### 🟡 `build.rs` grow-only comment

Minor: the comment says `per per-asset` (double "per").

## Pre-flight

- ✅ **Branch**: CI all green (static checks, clippy, Linux, Android, iOS export).
- ✅ **iOS build**: present and green (Build iOS export → R2 succeeded).
- ✅ **No submodule drift**.
- ✅ No debug prints, no `.claude/` files, no non-English text.

## Positives

- Excellent doc header explaining the GPU-upload semantics and the `PRELOAD_GHOST_TO_GPU` trade-off.
- Refcounting by content-hash is the right call — dedupes across entities and guarantees the later `GltfContainer` load is a genuine cache hit.
- Budget-pacing in `advance_preloads` prevents frame spikes when many preloads finish simultaneously.
- Correct grow-only-set registration for component 1214 — without this the SDK would miss state transitions.

Requested by Lean via Slack

lib/src/scene_runner/components/asset_load.rs
```diff
@@ -0,0 +1,476 @@
...
+        }
+        root.add_child(&node);
+        if let Some(entry) = scene.asset_load.preloads.get_mut(&hash) {
+            entry.ghost = Some(node);
+        } else {
+            // Preload was released before its ghost landed — don't leak it.
+            let mut node = node;
+            node.queue_free();
+        }
+    }
```
kuruk-mm: idea:
for avoiding to use set_visible per node, we can use a different parent as ghost

`scene.godot_dcl_scene.root_node_3d.ghost_node` (also .preloaded_nodes will work)

also will be clearear when we are going to debug, wdyt?

lib/src/scene_runner/components/asset_load.rs
```diff
@@ -0,0 +1,476 @@
...
+    godot_classes::{dcl_global::DclGlobal, promise::Promise},
+    scene_runner::scene::Scene,
+};
+
+/// Also instantiate a hidden ghost node when a GLTF preload finishes. Warms
+/// skinned-mesh / `AnimationPlayer` setup at the cost of a full extra instance
+/// in RAM. Default OFF — retaining the `PackedScene` already gets meshes and
+/// textures to the GPU; the ghost only adds pipeline/skeleton warmup that is a
+/// no-op on iOS (Metal PSO cache disabled). Flip to experiment / measure.
+const PRELOAD_GHOST_TO_GPU: bool = false;
```
kuruk-mm: do you think a good idea to leave this for experiments?

lib/src/scene_runner/components/asset_load.rs
leanmendoza: ghost mode dismissed for now

lib/src/scene_runner/components/asset_load.rs
leanmendoza: ghost mode dismissed for now

regenesis-claw (commented):
Re-review at `4982ec1` — adversarial pass focused on the requested areas.

## Previous findings: all addressed ✅

| Finding | Status |
|---|---|
| Missing `deleted_entities.rs` cleanup | ✅ `release_entity_preloads()` added, borrows `&mut AssetLoadState` cleanly |
| `release_preload` takes `&mut Scene` | ✅ Refactored to `release_preload_in(&mut AssetLoadState, …)` |
| Ghost node code | ✅ Removed entirely (commit `d752209c`) |
| `per per-asset` typo | ✅ Fixed |

---

## Adversarial analysis — ranked by severity

### 1. 🟡 Stuck preload on never-resolving promise — no per-asset timeout (Medium)

**Repro:** Scene sets `PBAssetLoad { assets: ["building.glb"] }` where the hash is valid in the content mapping but the content server stalls (TCP half-open, slow CDN, offline). The content provider's `fetch_optimized_asset_with_dependencies` / `load_scene_gltf` returns a promise that never resolves or rejects.

**What happens:** `advance_preloads` polls `promise.bind().is_resolved()` each tick → false → skips → entry stays `Loading` forever. The `Gd<Promise>` is retained in `PreloadEntry::promise` for the lifetime of the scene. The SDK sees `LoadingState::Loading` permanently — no error, no escalation, no timeout.

**Impact:** Per-preload, the cost is small (one `Gd<Promise>` + one HashMap entry + one poll per tick). But a malicious or buggy scene can set 100 entities × N assets = hundreds of stuck entries, all polling every tick. Bounded by max entities (512) so it's not unbounded, but it's still unnecessary work.

**Why it matters vs GltfContainer:** GltfContainer has the *scene-level* loading funnel timeout (`async_show_scene_timeout_modal`) that eventually prompts the user. Preloads don't participate in the loading funnel (no `report_asset_loading_started` / `report_asset_loaded` calls), so a stuck preload is invisible — no UX signal, no escape hatch.

**Suggested mitigation (follow-up, not blocking):** Add a wall-clock deadline to `PreloadEntry` (e.g., `created_at: Instant`). In `advance_preloads`, if `Instant::now() - created_at > PRELOAD_TIMEOUT` (say 60s) and the promise hasn't resolved, treat it as `FinishedWithError` and drop the promise. This matches the scene-level timeout behavior without over-engineering.

---

### 2. 🟢 Path-casing change destroys and recreates preload (Low)

**Repro:** Scene sets `PBAssetLoad { assets: ["Models/Foo.glb"] }` (hash X acquired, `ResourceLoader::load` done, `PackedScene` retained). Next tick the scene writes `PBAssetLoad { assets: ["models/foo.glb"] }` — same hash X, different path string.

**What happens:** `reconcile_entity_assets` compares paths by exact string (`p == &old.path`). `"models/foo.glb" != "Models/Foo.glb"`, so the old path is released (refcount → 0, `PackedScene` dropped) and the new path acquires a fresh entry (re-download, re-`ResourceLoader::load`). The entire preload is wasted and restarted.

**Impact:** Wasteful (bandwidth + GPU re-upload), not a leak. Refcounts are balanced. Unlikely in practice — scenes don't normally change path casing mid-run.

**Fix (if desired):** Compare by resolved hash instead of path string when deciding "unchanged". But path identity is also the key for `reported_state`, so this would need a bit of restructuring. Not worth it unless a real scene hits this.

---

### 3. 🟢 Duplicate paths in `assets` array double-count refcount and double-report state (Low)

**Repro:** Scene sets `PBAssetLoad { assets: ["a.glb", "a.glb"] }`.

**What happens:** Both entries resolve to the same hash → `acquire_preload` called twice → refcount = 2. Two `TrackedAsset` entries created. When the preload finishes, `sync_asset_load_loading_state` appends **two** `PbAssetLoadLoadingState` for the same entity+asset with different timestamps. Deletion releases both → refcount 2 → 0 → entry removed. Refcount is balanced ✅ but the SDK sees duplicate state events.

**Impact:** Confusing to scene logic that counts completions. Arguably a scene bug (the scene shouldn't send duplicates), but a `HashSet`-dedup on `new_paths` in `reconcile_entity_assets` would be a one-liner defensive fix.

---

### 4. 🟢 `Scene::cleanup()` doesn't explicitly clear `asset_load` (Informational)

Other per-entity state (`audio_sources`, `video_players`, `tweens`) is explicitly drained in `cleanup()`. `asset_load` relies on Rust's `Drop` when the `Scene` struct is dropped. This is *correct* — `AssetLoadState` contains only Rust-managed data and `Gd<PackedScene>` (ref-counted, not scene-tree nodes). But it's inconsistent with the cleanup pattern. A one-liner `self.asset_load = AssetLoadState::default();` in `cleanup()` would match the convention.

---

## Areas verified clean

| Area | Verdict |
|---|---|
| **Refcount across entity-delete** | ✅ `release_entity_preloads` in `deleted_entities.rs` correctly decrements and drops at zero |
| **Refcount across component-remove** | ✅ LWW `value: None` → `paths = []` → `reconcile_entity_assets` releases all refs |
| **Refcount same hash from N entities** | ✅ Each entity's `acquire` bumps refcount; each entity's death/removal decrements independently |
| **CRDT LWW (1213) vs GOS (1214)** | ✅ `PBAssetLoad` read as LWW, `PBAssetLoadLoadingState` registered in `GROW_ONLY_SET_COMPONENTS` and written via `.append()` |
| **Delete + PBAssetLoad on same JS tick** | ✅ `take_dirty()` strips died entities from LWW dirty set before scene update loop — `update_asset_load` never sees the dead entity |
| **Entity version recycling** | ✅ `(number, version+1)` is a distinct `SceneEntityId` — old preloads cleaned, new entity starts fresh |
| **Tick budget early-return** | ✅ `advance_preloads` returns `false` → state machine re-enters `SyncAssetLoad` next tick; emit loop (cheap) always completes; no starvation |
| **Promise reject** | ✅ `is_resolved()` covers both fulfilled and rejected; `is_rejected()` distinguishes → `FinishedWithError` |
| **Scene unload** | ✅ Rust drop of `AssetLoadState` releases all `Gd<PackedScene>` refs; no scene-tree nodes held (ghosts removed) |
| **Content-mapping case sensitivity** | ✅ `get_hash()` lowercases internally; `path.to_lowercase()` at call site matches |
| **Timestamp wrapping** | ✅ `wrapping_add(1)` on u32; grow-only set entries are unique by value, timestamp is for SDK ordering only |

---

## Verdict

The entity-deletion leak from the first review is fixed. The remaining findings are all low-severity — #1 (stuck preloads) is the only one worth a follow-up issue, and even that is inherited from the content provider's lack of per-promise timeouts (same as GltfContainer). The refcounting, CRDT wiring, and budget pacing are all solid.

No blockers — code looks good to merge.

Requested by Lean via Slack

lib/src/scene_runner/components/asset_load.rs
```diff
@@ -0,0 +1,446 @@
...
+/// A shared preload keyed by content hash.
+pub struct PreloadEntry {
+    /// How many tracked (entity, asset) pairs in this scene reference this hash.
+    pub refcount: u32,
+    /// Whether this is a GLTF asset (taken all the way to a retained `PackedScene`)
+    /// vs a generic file (download-cache warmup only).
+    pub is_gltf: bool,
+    /// In-flight download promise. `None` once resolved (or never started).
+    pub promise: Option<Gd<Promise>>,
+    /// Resource path the GLTF loads from (`res://glbs/<hash>.scn` for optimized,
```
regenesis-claw, line 81: Consider adding a `created_at: Instant` field here. In `advance_preloads`, if `Instant::now() - created_at > PRELOAD_TIMEOUT` and the promise is still pending, transition to `FinishedWithError` and drop the promise. This would prevent stuck preloads from polling forever when the content server is unreachable. Follow-up material — not blocking.

kuruk-mm (approved):
looking great

PR #2569, opened 2026-07-22
feat: block entry to private worlds and show the "is private" modal

### LINKED ISSUE #1725: Modals: Prevent Scene/World Load for Private Worlds - Implementation
## **Description**
The client should read the user permissions to enter a given world and and prevent loading when the user is not allowed to join it.

[FIGMA](https://www.figma.com/design/IG9xWnAveE63CrLNSIIJ7V/Modals?node-id=272-296&t=N6n7PIo5fyjTfRy5-1)

**Acceptance Criteria**
- [ ] Client reads Private World before world load
- [ ] If the user is not allowed, private world modal shows

Image

leanmendoza:
Closes #1725

Worlds can restrict access to an allow-list configured in Builder / Creator Hub, but the client never read it. worlds-content-server only enforces it for comms, so a non-allowed user still loaded and walked the world with comms silently dead.

This reads `GET /world/<name>/permissions` **before** the navigation starts and, when the user isn't allowed, blocks it and shows a simple modal instead of ever entering the world.

### How it works

- **Prefetch at the entry points.** The block runs in `async_join_world` / `async_teleport_to` *before* any UI — before the loading screen, the menu close, or booting the explorer scene. On a refusal the modal appears with no transition and the user stays exactly where they were; nothing of the private world loads. This also covers the cold-start-from-lobby case (no explorer yet), which previously booted the explorer straight into the world before blocking.
- **Safety-net gate.** `Realm.async_set_realm()` keeps a gate before `realm_changing.emit()` / the `/about` fetch, so entry points that don't prefetch (deeplink cold start, SDK `changeRealm`, resume, settings) are still covered. It emits a dedicated `realm_access_denied` signal, separate from `realm_change_failed`, so it raises the modal instead of the generic "World unavailable" toast.
- **Cached + warmed.** The access answer is cached briefly (30s), so the prefetch and the gate share a single fetch. The cache is warmed when a place's jump-in card opens, so by the time JUMP IN is tapped the decision is already cached and resolves synchronously — an allowed world goes straight to the loading screen with no visible gap, a private one shows the modal instantly.
- **The modal** is a blocker with the design's block icon and copy (`<name> is private` / `Only invited people can enter.`) and a single `OK` button that just dismisses it, leaving the user where they were.
- Only `permissions.access` is evaluated, mirroring the server's own access checker.
- A refused world is pointed back at the main realm (`_clear_boot_realm_if_denied`) so it doesn't reopen the modal on every cold start.

### Fail-open by design

Denies **only** a plain wallet allow-list the address is absent from (and not the owner). `unrestricted`, community-based allow-lists, `shared-secret`, `nft-ownership`, unknown types, network errors, malformed responses and a missing address all resolve to "allowed". Comms remains the real enforcement, so a false negative costs nothing while a false positive would lock users out.

### Test plan

Fixtures (confirm current server state with `curl -s https://worlds-content-server.decentraland.org/world/<name>/permissions`):
- `welcomeguides.dcl.eth` — wallet allow-list, denies any wallet not on it.
- `aesironline.dcl.eth` — `access.type: unrestricted`.
- `ryumi.dcl.eth` — community-based allow-list (empty `wallets`, non-empty `communities`) → intentionally allowed.

Sign in with any wallet **not** on the allow-list (a guest works).

1. **Denied, in session (main path)** — already in Genesis or a world, open `welcomeguides`'s card in Discover and press JUMP IN. The modal must appear **over Discover, with no loading flash and without entering the world**: block icon, `welcomeguides is private` / `Only invited people can enter.` / a single `OK`. Tapping `OK` dismisses it and leaves you exactly where you were (no forced Discover, no trap).
2. **Denied, cold start from the lobby** — from the lobby/Discover with no world loaded yet, jump into `welcomeguides`. It must **not** boot the explorer or land you "in-world" first; the modal shows over Discover and `OK` returns you to Discover.
3. **Denied, deeplink cold start** — launch straight into it (`?realm=welcomeguides.dcl.eth`). The modal must appear and the world must never load. Kill and relaunch: it must start on the main realm, not reopen the modal (boot realm was reset).
4. **Denied, other entry points** — `/goto welcomeguides`, a chat `world:` link, and an SDK `changeRealm` each show the modal with no transition and leave the user where they were.
5. **Allowed world, no gap** — open `aesironline`'s card, wait a moment, then JUMP IN. It must go straight to the loading screen with **no visible "card closes → gap → loading" step** (the cache is warmed on card open). The world loads normally.
6. **No regression** — a normal Genesis teleport is unaffected (no extra HTTP for Genesis; one `GET /permissions` per world, warmed at card open). `ryumi` (community-gated) still lets you in — no modal.
7. **Allowed user** — with a wallet that *is* on the allow-list (or the world's owner), `welcomeguides` loads normally with comms working.

`gdformat` / `gdlint` clean; all touched files pass `check-gdscript`. Validated on desktop with a guest wallet, driving the live client over the debug hub.

### Known limitations (intentional)

- Community-based allow-lists aren't evaluated (needs social-service membership resolution) → those users are let through. Single condition to flip later.
- `shared-secret` / `nft-ownership` worlds aren't blocked — unchanged behaviour, no password UI in this client.
- The 30s cache means a user removed from an allow-list can still enter for up to ~30s; comms remains the real enforcement.
- No in-world revocation guard: permissions revoked while the user is inside don't evict them.


godot/src/logic/realm.gd
```diff
@@ -148,6 +152,21 @@ func async_set_realm(new_realm_string: String, search_new_pos: bool = false) ->
...
 	)
+
+	# Private worlds (#1725): refuse before emitting, fetching or mutating anything, so a
+	# world the user can't enter never starts loading.
+	var world_name := WorldPermissionsHelper.world_name_from_realm(
+		new_realm_string, candidate_realm_url
+	)
+	if not world_name.is_empty() and not await WorldPermissionsHelper.async_is_allowed(world_name):
+		printerr("[REALM] Access denied for private world: ", world_name)
+		# Close the analytics loading episode without raising the generic failure toast.
```
regenesis-claw, line 164: 🟡 **Sentry quota** — `printerr` routes to Sentry (`push_error`/`printerr` → Godot error → Sentry issue). A user being denied entry to a private world is an **expected, handled condition** — the feature is working as designed. This will fire a Sentry event every time someone taps a private world they're not on the list for.

Downgrade to `print` (or `push_warning` if you want it in warnings):
```gdscript
print("[REALM] Access denied for private world: ", world_name)
```

godot/src/ui/components/organisms/modal/modal_manager.gd
```diff
@@ -981,6 +1013,26 @@ func _on_ban_pre_check_go_to_discover() -> void:
 	# Case 3: Already in discover — modal closed, discover is already behind
 
 
+func _on_private_world_go_to_discover() -> void:
+	close_current_modal()
+
+	if Global.is_orientation_portrait():
```
regenesis-claw: 🔴 **Cold-start denial may bypass the "not stranded" guard.** On a cold start via deeplink into a private world, the app is already in portrait (Discover) when the denial fires. This early return exits without ever setting `private_world_block_active = true` (that's gated behind the portrait check on line 1029).

Consequence: user presses "BACK TO DISCOVER" → modal closes → user tries to close Discover → `reshow_blocked_entry_modal()` returns `false` → they drop into landscape with no realm loaded. This is exactly the scenario the test plan's **"Not stranded"** case describes should NOT happen.

The ban pre-check (`_on_ban_pre_check_go_to_discover`) doesn't have this early return — it always sets `ban_pre_check_active = true`. Consider the same approach here:
```gdscript
func _on_private_world_go_to_discover() -> void:
	close_current_modal()
	_force_hide_loading_screen()

	if not Global.is_orientation_portrait():
		Global.set_orientation_portrait()
		Global.open_discover.emit()

	if Global.realm.get_realm_string().is_empty():
		private_world_block_active = true
```
This way, when already in portrait it skips the orientation/emit calls (no-op, as you intended) but still evaluates the empty-realm guard.

regenesis-claw (changes requested):
Clean, well-structured PR — the single-choke-point gate in `async_set_realm`, the fail-open design, and the boot-loop prevention via `_clear_boot_realm_if_denied` are all solid. Two issues below, one of which may break the "Not stranded" test case.

**Branch note:** 3 commits behind `main` — not blocking review, but please merge before final approval to ensure CI signals are trustworthy.

**iOS build:** not required — no platform-specific code touched.

Followed repo REVIEW.md.

Ludmilafantaniella:
@leanmendoza reviewed this, PR and modal look correct, but I think the flow needs a tweak.

Repro: I'm in a scene, type `/goto [private world]` in chat. The loading bar starts, then cancels and returns me to the scene I was already in, showing the "is private" modal on top. The only button is "BACK TO DISCOVER", which sends me to Discover instead of just closing the modal and letting me stay where I already am.

Could we add a "Close" option, or have the button just dismiss the modal instead of forcing navigation to Discover?

Screenshot_20260727_100402_Decentraland

Also found: I was able to access ryumi.dcl.eth (has permissions set) with a guest account on mobile. Desktop explorer works fine, this only happens on mobile.

regenesis-claw (approved):
Re-reviewed after commits ab44809..71f25bb. Both issues from my first review are resolved:

1. **Cold-start stranding (was 🔴)** — completely reworked. The old `_on_private_world_go_to_discover` (with its portrait early-return bug) is gone. The modal now uses a simple "OK" dismiss, and the cold-boot fallback is handled cleanly in `_on_realm_access_denied`: when the explorer is valid but has no realm, it falls back to main realm via `call_deferred`. `private_world_block_active` state removed entirely — much simpler design, no more trapping in Discover.

2. **`printerr` → Sentry noise (was 🟡)** — downgraded to `print` with a comment explaining why. ✅

Additional improvements in these commits:
- Request dedup via `_inflight` dictionary — card-open warm + fast JUMP IN share a single GET.
- Naming cleanup: `reshow_ban_modal` / `_on_menu_close_ban_recheck` — private-world modal intentionally excluded since its OK button just dismisses.
- UX updated per design feedback: "OK" button that leaves user where they were, instead of forcing navigation to Discover.

Branch is merged with main. CI all green (Linux, Android, static checks, clippy, iOS export). No platform-specific code touched — iOS build not required but present and passing anyway.

Followed repo REVIEW.md.

Ludmilafantaniella:
### 🐛 Issues found:

@leanmendoza Deeplink-specific issue, behaves differently per platform (Discover search and `/goto` both work fine on both):

Setup: app in background or close → open `https://mobile.dclexplorer.com/?realm=welcomeguides.dcl.eth`

- **Android:** skips Discover entirely — loads directly into a random scene with the error modal on top.
- **iOS:** Discover + modal show correctly at first, but teleporting from there to a different (non-private) scene incorrectly re-triggers the welcomeguides modal while the target scene loads underneath.

https://github.com/user-attachments/assets/c3723d10-d098-4987-842a-aaf075d63ed5

Ludmilafantaniella (approved):
**Approved** ✅ — verified on iOS and Android, everything works correctly.

- Deeplink, app fully closed (cold start): modal shows correctly, world never loads.
- Deeplink, app in background: same, modal shows correctly.
- From Discover: modal appears over Discover, OK dismisses and leaves me where I was.
- `/goto world` and `/goto world.dcl.eth` from a scene: modal shows correctly in both forms, no transition, left where I was.

No issues found. LGTM.

https://github.com/user-attachments/assets/2b806507-71c3-427b-95c8-6ee58d7d96d1

https://github.com/user-attachments/assets/2f33bf47-c527-4765-a1f0-7d63f8c17378

🟢 Regression performed:
- Log In/Out
- Teleport to scenes/worlds from Discover and also using chat commands 
- Backpack  - chnaging wearables
- Emote functionality

kuruk-mm (approved):

PR #2571, opened 2026-07-22
fix: reduce spurious scene-room reconnects (onSceneEnter fires repeatedly under poor connectivity)

EibrielInv:
## What

  Under poor connectivity the scene-room reconnect loop fired **overlapping** reconnect
  attempts: `reconnect_scene_room()` only spawns an async gatekeeper request and returns,
  leaving `scene_room = None` for the whole round-trip, so the 5s retry fired again every
  tick while one was still in flight. Each successful attempt re-announces the player, which
  the authoritative server counts as a fresh join → repeated `onSceneEnter`, each rejoin with
  a null position.

  This matches the reference (Unity) explorer behavior: **one `onSceneEnter` per genuine
  reconnect is expected**; the bug was the Godot explorer generating *extra* reconnects.

  ## Changes (`lib/src/comms/communication_manager.rs`)

  - **In-flight guard** — a `scene_room_connect_in_flight` marker + the pure, unit-tested
    `should_start_scene_room_reconnect(...)`, so at most one reconnect attempt is in flight at
    a time (20s lost-signal safety net). Adds a `SceneRoomConnectionRequest::Failed` terminal
    signal so a failed attempt clears the marker.
  - **Exponential backoff** on the scene-room retry (5→10→20→40→60s, reset on disconnect and on
    successful connect) so a persistently-failing reconnect stops hammering the gatekeeper.

  ## Result
  
  A single scene-room disconnect now produces **one** reconnect attempt and **one** server-side
  re-join, instead of the 3× burst captured before the fix — matching Unity.

**Before (3 reconnects)**
imagen

**Now (only one reconnect)**
imagen

  ## Tests

  - 6 unit tests for the guard: `cd lib && cargo test --features use_livekit should_start_scene_room_reconnect`
  - `cargo clippy -- -D warnings` and `cargo fmt --check` clean.

  ## QA — how to reproduce / verify

  1. Launch the explorer into the affected World:
     `cargo run -- run -- --skip-lobby --fake-deeplink "decentraland://jump?realm=skychaser.dcl.eth"`
  2. Once loaded and standing still, **degrade the network connection** (e.g. macOS Network Link
     Conditioner with a lossy/high-latency profile, or briefly cut connectivity) to force a
     comms reconnect. Note the wall-clock timestamp of the reconnect.
  3. Ask the scene admin for connection logs around that timestamp.
  4. **Expected:** each genuine reconnect yields **one** connection/`onSceneEnter` on the server
     (Unity-equivalent), not a burst of several within a few seconds.

### DIFF
--- a/lib/src/comms/communication_manager.rs
+++ b/lib/src/comms/communication_manager.rs
@@ -45,12 +45,59 @@ pub enum SceneRoomConnectionRequest {
     Banned {
         scene_id: String,
     },
+    /// A reconnect attempt finished without producing a connection (the
+    /// gatekeeper request failed). Carries no room change; it only clears the
+    /// in-flight marker so the reconnect loop can try again.
+    Failed {
+        scene_id: String,
+    },
 }
 
 #[derive(Debug)]
 enum SceneAdapterError {
     Banned,
     Other(String),
+}
+
+/// Safety net for the scene-room reconnect in-flight guard: if an attempt's
+/// terminal signal is ever lost, treat it as stale after this long and allow a
+/// fresh attempt rather than wedging reconnection forever.
+#[cfg(feature = "use_livekit")]
+const SCENE_ROOM_CONNECT_TIMEOUT: Duration = Duration::from_secs(20);
+
+/// Base delay before the first scene-room reconnect attempt after a disconnect.
+#[cfg(feature = "use_livekit")]
+const SCENE_ROOM_RECONNECT_BASE_SECS: u64 = 5;
+/// Cap for the exponential backoff between failing scene-room reconnect attempts.
+/// Backoff (base → 2× per failure, capped here) stops a persistently-failing
+/// reconnect from hammering the gatekeeper every 5s while connectivity is poor —
+/// each successful attempt is a server-side re-join (issue #2382).
+#[cfg(feature = "use_livekit")]
+const SCENE_ROOM_RECONNECT_MAX_SECS: u64 = 60;
+
+/// Decides whether the scene-room reconnect loop should spawn a new attempt now.
+/// Pure (no Godot/Tokio deps) so the in-flight guard is unit-testable. The guard:
+/// don't start a reconnect while one is already in flight, unless it has been
+/// outstanding longer than `connect_timeout` (a lost-signal safety net).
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
Clean, well-structured fix for a real production issue. Reviewed against `REVIEW.md`.

**Pre-flight:**
- ⚠️ Branch is 8 commits behind `main`, but none of those commits touch `communication_manager.rs` — no conflict risk. Merge/rebase before merging.
- ✅ iOS build present and green (required: PR touches `lib/src/comms/`).
- ✅ No submodule drift.

**What's good:**
- The pure `should_start_scene_room_reconnect()` function is an excellent design choice — decouples the guard logic from Godot/Tokio for clean unit testing.
- The `strip_prefix("livekit:")` refactor replaces the old `starts_with` + `unwrap_or` pattern — safer and more idiomatic.
- Exponential backoff (5→10→20→…→60s cap) with reset on both disconnect and successful connect is textbook.
- The 20s safety-net timeout on the in-flight marker prevents permanent wedging if a terminal signal is lost — good defensive design.
- All `scene_room_connect_in_flight` clearing paths are comprehensive: checked `_on_change_scene_id`, `clean()`, `release_comms()`, and all three terminal outcomes (`Connect`/`Failed`/`Banned`). ✅
- Log levels are correct: `warn!` for the initial disconnect (actionable, already existed), `debug!` for routine attempt/failure (not Sentry). No quota risk.
- 6 unit tests cover the key invariants well.

**One minor observation (not blocking):**
- The `Failed` handler clears `scene_room_connect_in_flight` without checking whether `scene_id` matches `current_scene_id`. If a scene change happens while a reconnect for the *old* scene is still in flight, the stale `Failed` could clear the in-flight marker for the new scene's attempt. In practice this is benign — scene changes go through `_on_change_scene_id` which already clears the marker and doesn't use the reconnect timer — but storing the scene_id alongside the `Instant` in the marker (e.g. `Option<(String, Instant)>`) would make the guard airtight. Fine as a follow-up.

Approving — please merge `main` before merging.

Requested by Mateo via Slack

kuruk-mm (approved):
looks great!

PR #2575, opened 2026-07-23
fix: match interaction-hint icon to SDK-replaced control icons

### LINKED ISSUE #2557: Controls Customization: Icons in the over highlight should be the same as the replaced icons in the gamepad

EibrielInv:
## What

  When the crosshair points at an interactable 3D entity, the interaction hint ("over highlight") now shows the same icon a scene set for that action via `PBTouchScreenControls` — matching the replaced glyph on
  the on-screen gamepad button — instead of the default keyboard letter / built-in glyph.

  ## Changes

  - `sdk_touch_controls_applier.gd`: add shared static `get_custom_icon_for_action()` — single source of truth for an action's scene-replaced icon (`{hash, url}`), reading `Global.touch_controls_*`.
  - `joypad.gd`: build its per-action icon map via the shared helper (no behavior change).
  - `tooltip_label.gd`: after computing the default glyph, async-load the scene-replaced icon (mirrors the joypad's content-provider fetch) and swap it in on both mobile and desktop. The default stays visible 
  until the texture loads (no flicker) and if it never does. `_physics_process` re-resolves the glyph when the action's icon changes live (e.g. from the Controls Builder) while the tooltip is on screen.

imagen

 ## QA testing steps

  Test scene is deployed at **`-31,19`**.

  1. Enter the test scene and select the **Controls Builder** block (the "Control Creator" dashboard — it's the default block).
  2. A row of colored interactive spheres appears, one per button (JUMP, POINTER, E, F, 1–4).
  3. Point the crosshair at a sphere → the interaction hint appears with that button's **default** glyph (keyboard letter / built-in icon).
  4. In the Controls Builder panel, assign a custom icon to that button (cycle its icon to fishing / binocular).
  5. Verify the hint icon now matches the icon shown on the on-screen gamepad button for that action — on both desktop and mobile.
  6. **Live update:** with the crosshair still on the sphere, change that button's icon in the panel → the hint icon updates in place (no need to look away and back).
  7. Remove the custom icon (set it back to none) → the hint reverts to the default glyph.
  8. Regression: hover an interactable in a scene with no `PBTouchScreenControls` → the hint behaves exactly as before (keyboard letters on desktop, mapped glyphs on mobile).

  Closes #2557

### DIFF
--- a/godot/src/ui/components/molecules/pointer_tooltip/tooltip_label.gd
+++ b/godot/src/ui/components/molecules/pointer_tooltip/tooltip_label.gd
@@ -22,6 +22,9 @@ var text_down := ""
 var text_up := ""
 var last_state_pressed := false
 var stylebox: StyleBox
+# Content hash of the scene-replaced (PBTouchScreenControls) icon we're currently loading,
+# so a stale async fetch can bail if the tooltip is reused for another action meanwhile.
+var _custom_icon_hash: String = ""
 
 @onready var label_action = %Label_Action
 @onready var texture_rect_action_icon = %TextureRect_ActionIcon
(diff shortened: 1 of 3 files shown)


kuruk-mm (approved):
looks great!

Issue #1538, opened 2026-02-25
SDK Particle System Support 


rodrigo685:
## Summary

Add support for a **Particle System SDK component** that allows scene creators to spawn and configure particle effects in their scenes. This is a long-requested community feature (tracked in [decentraland/sdk#319](https://github.com/decentraland/sdk/issues/319) under "Other > Particle systems") and currently has **no protocol definition** — it needs to be designed from scratch.

## Current State

- **Protocol**: No `particle_system.proto` exists in `decentraland/protocol`. A new SDK component must be defined.
- **Unity Renderer**: Particles exist only for **internal renderer effects** (emotes like Clap/Heart/Confetti, avatar spawn/teleport VFX, skybox clouds) under `unity-renderer/Assets/Particles/`. These are hardcoded prefabs, **not** exposed to the SDK. Related issues:
  - [unity-renderer#7](https://github.com/decentraland/unity-renderer/issues/7) — Particles for avatar locomotion
  - [unity-renderer#9](https://github.com/decentraland/unity-renderer/issues/9) — Particle-based emotes (StickersSystem)
  - [unity-renderer#1208](https://github.com/decentraland/unity-renderer/issues/1208) — Particle effects for emotes
- **Godot Explorer**: No particle system component implemented. Godot natively provides `GPUParticles3D` and `CPUParticles3D` nodes which can be leveraged.

### Protocol Definition
A new protobuf component (e.g. `particle_system.proto`) needs to be added to `decentraland/protocol` with fields covering at minimum:

- **Emitter shape**: Box, Sphere, Cone, Point
- **Emission**: Rate over time, burst count
- **Particle lifetime**: Min/max duration
- **Particle properties**: Start size (min/max), start color, start speed, gravity modifier
- **Velocity**: Direction, speed over lifetime
- **Rendering**: Billboard mode, texture (content URL), blend mode
- **Size over lifetime**: Curve or start/end values
- **Color over lifetime**: Gradient or start/end colors
- **Max particles**: Cap for performance/scene limits
- **World vs local space**: Simulation space

### Implementation in Godot
- Map the SDK component to Godot's `GPUParticles3D` node (preferred for performance) with `CPUParticles3D` as a fallback on lower-end devices
- Process material properties map to `ParticleProcessMaterial` in Godot
- Texture/material properties map to the particle's draw pass material
- Scene resource limits should cap `max_particles` to prevent abuse

### Performance & Safety
[...]

PR #2588, opened 2026-07-27
feat(sdk): implement ParticleSystem component (#1538)

manuelmaceira:
# SDK Particle System Support (Draft)

Implements the `ParticleSystem` SDK component (proto id **1217**, already published in `@dcl/protocol`) in the Godot Explorer, closing the renderer side of #1538.

The mapping follows the **Unity Explorer reference implementation** ([`SDKComponents/ParticleSystem`](https://github.com/decentraland/unity-explorer/tree/main/Explorer/Assets/DCL/SDKComponents/ParticleSystem)) so both clients behave the same way.

## How it works

`PBParticleSystem` CRDT updates map onto a `GpuParticles3D` node (with `ParticleProcessMaterial` + a quad draw pass) per entity, following the same handler pattern as the other SDK components (`lib/src/scene_runner/components/particle_system.rs`).

| Area | Implementation |
|---|---|
| Emission shapes | point/sphere/box native; **cone** approximated with a flat ring + spread (Godot has no cone shape). Axis matches Unity's local +Z, adjusted for the DCL→Godot Z flip |
| Emission | rate (`amount = rate × lifetime`, Godot's cycle model), lifetime, loop, prewarm, `active`, playback state (play/pause/stop) |
| Bursts | per-frame driver with `emit_particle()` — Unity's fixed **5s cycle** semantics, `cycles`/`interval`/`probability`, manual shape/direction/speed/color sampling |
| Motion | initial velocity range, gravity × DCL gravity, `additional_force` (world-space converted), `limit_velocity` via Godot 4.6's native `velocity_limit_curve` (exact for `dampen = 1`, the default) |
| Size/rotation/color | `scale_min/max` + float-baked curve textures (CurveTexture would clamp to [0,1]), `color_initial_ramp` (random-between) + `color_ramp` (over lifetime), sRGB→linear conversion, face-travel-direction, initial rotation baked into the draw mesh for non-billboard quads |
| Rendering | `BILLBOARD_PARTICLES` + **`BILLBOARD_KEEP_SCALE`** (without it the draw shader drops particle scale), back-face culling like Unity's material, blend modes (alpha/add/multiply), sprite-sheet animation, textures via the standard content pipeline (path→hash resolved through the content mapping) |
| Limits | 5k particles/emitter, 50k/scene budget, clamped sprite-atlas tiles, sanitized quaternions (scenes can send zero/unnormalized ones) |
| Presence policy | user leaves the scene → stop + clear every system; re-enters → restart from scratch (matches Unity's `ParticleSystemPlaybackSystem`). Event-driven via `on_current_parcel_scene_changed` because idle scenes only tick every ~120s |

## QA Steps

You'll need a build of this branch and the Unity client for reference. No code access needed — navigation is via deeplink.

1. **Open the test scene** in this branch's build by clicking/opening this deeplink on the device (or pasting it in the browser on desktop):
   `decentraland://?dclenv=zone&realm=sdk7testscenes.dcl.eth&position=0,7`
   You should land in the "Particle System Showcase" scene (parcel `0,7` of the sdk7 test world).
2. **Open the same deeplink in the Unity client** for side-by-side reference.
3. **Walk through the emitters and compare** with Unity: each labeled effect (Fire Ember, Magic Aura, Snowfall, Vortex Spiral, Bat/Bee Swarm, Campfire, Fireworks, etc.) should look roughly the same — shape, direction, size, color, sprite-sheet animation.
4. **Presence policy**: walk out of the scene's parcels → every particle effect should disappear instantly. Walk back in → effects restart from the beginning. Should match Unity's behavior exactly.
5. **Edge cases (optional)**: open `kuruk.dcl.eth` (mainnet world) in both clients — it has ~86 labeled test-case emitters (bursts, weird values, additional force rows) to compare.

Expected result: effects are visually comparable to Unity. Known delta: semi-transparent particles may still read slightly brighter than Unity (under investigation, see below).

## Testing

- [`kuruk.dcl.eth`](https://decentraland.org/play/?realm=kuruk.dcl.eth) — dedicated particle test world (~86 emitters)
- `sdk7testscenes.dcl.eth` parcel `0,7` ([scene source](https://github.com/decentraland/sdk7-test-scenes/tree/main/scenes/0%2C7-particle-system)) — side-by-side comparison with the Unity client
- Verified: shapes, directions, sizes, colors, sprite sheets, bursts, presence policy (walk out → all die instantly; walk in → restart)

## Known deltas / follow-ups

- **Brightness**: semi-transparent stacks still read slightly brighter than Unity. Ruled out so far: glow (tested with bloom off), culling (now single-sided like Unity), blend factors (identical SrcAlpha/OneMinusSrcAlpha), color space (both linear). Under investigation — likely global render pipeline (tonemap/HDR) rather than the component.
- `limit_velocity.dampen < 1` is approximated as a hard clamp (a custom `shader_type particles` ShaderMaterial would be needed for exact per-frame excess removal).
- Rotation on billboards reduces to the view-axis (Z) component, matching Unity's billboard behavior; full 3D rotation only applies to non-billboard quads (baked into the mesh).
- Sphere shape emits in random directions (Godot) instead of strictly radial (Unity) — visually close for small radii.
- One-time ~25ms spike in the `ParticleSystem` state when a scene with many emitters first applies (material/texture creation).
- Mobile/LOW-end: budget constants are fixed; hook into the dynamic graphics manager for quality scaling.

## Out of scope (other repos)

- `decentraland/js-sdk-toolchain`: expose the component in `@dcl/sdk` (the test scenes already use it via `defineParticleSystemComponent`).


lib/src/scene_runner/components/particle_system.rs
```diff
@@ -0,0 +1,1006 @@
...
+}
+
+/// Poll pending particle textures and apply them once loaded (mirrors the
+/// `dirty_materials` flow in material.rs).
+fn poll_particle_textures(scene: &mut Scene) {
+    let mut keep_dirty = false;
+    let mut content_provider = DclGlobal::singleton().bind().get_content_provider();
+
+    for item in scene.particle_systems.values_mut() {
+        if !item.waiting_texture {
```
regenesis-claw, line 391: 🟡 The AABB doesn't include the displacement from `gravity_vec` or `additional_force`. A system with `gravity = 1.0` and `lifetime = 5s` drops particles ~122 m (`½ × 9.81 × 25`), well beyond this bound. Consider:
```rust
let accel_reach = 0.5 * gravity_vec.length() as f64 * lifetime * lifetime;
let reach = (speed_max * lifetime as f32 + shape_extent + accel_reach as f32 + 2.0).max(4.0);
```
(Requires computing `gravity_vec` before this point or passing it in.)

lib/src/scene_runner/components/particle_system.rs
```diff
@@ -0,0 +1,1006 @@
...
+                    playing,
+                    looping: new_value.r#loop.unwrap_or(true),
+                    lifetime: new_value.lifetime.unwrap_or(5.0).max(0.01),
+                    user_in_scene,
+                    last_value: new_value,
+                },
+            );
+
+            scene.dirty_particle_systems = true;
+        }
```
regenesis-claw, line 287: 🟡 For infinite-cycle bursts (`cycles = 0`), this `while` loop iterates `cycle_time / interval` times — up to ~5 000 with the 0.001 minimum clamp. Each entry then calls `emit_burst_particles(count.min(5000))`, so a single malicious burst spec can trigger millions of `emit_particle` API calls in one frame. Consider capping total burst emissions per frame to `MAX_AMOUNT_PER_EMITTER`.

regenesis-claw (commented):
Excellent feature PR — well-structured, thorough documentation, proper scene lifecycle (create/update/delete/teardown), and the Unity-parity mapping table in the PR body is genuinely helpful for reviewers and future maintainers. Followed `REVIEW.md`.

## Pre-flight

- ✅ Branch is up to date with `main` (0 behind).
- ℹ️ **iOS build**: mobile distribution was just triggered (pending). This is a pure Rust component (no iOS platform branches, no native plugins, no GDScript). Not blocking.
- ✅ No submodule changes.

## Findings

### 🟡 Visibility AABB ignores gravity and additional_force

`particle_system.rs` ~L385 — the AABB is derived from `speed_max × lifetime + shape_extent + 2.0`, but doesn't account for acceleration-based displacement. A system with `gravity = 1.0` and `lifetime = 5.0` drops particles ~122 m (`½ × 9.81 × 25`), well beyond a typical ~12 m AABB derived from speed alone. Particles that leave the AABB are frustum-culled — they vanish when the camera aims at them.

Suggested fix: include the acceleration displacement in `reach`:
```rust
let accel_reach = 0.5 * gravity_vec.length() as f64 * lifetime * lifetime;
let reach = (speed_max * lifetime as f32 + shape_extent + accel_reach as f32 + 2.0).max(4.0);
```
Not a correctness blocker (no crash, no data corruption), but can cause noticeable visual popping on any scene that uses gravity.

### 🟡 Burst CPU cost is unbounded per frame

`drive_bursts` (L268) — for infinite-cycle bursts (`cycles = 0` → `remaining_cycles = None`), the inner `while burst.next_fire <= cycle_time` loop iterates `cycle_time / interval` times. With `interval = 0.001` (the minimum clamp) and `cycle_time ≈ 5.0`, that's ~5 000 entries pushed to `pending`. Each entry then calls `emit_burst_particles` which loops up to `count.min(5000)` times calling `emit_particle` — a Godot API call.

A malicious scene sending `{ interval: 0.001, cycles: 0, count: 5000 }` can trigger **~25 M** `emit_particle` calls in a single frame. The GPU silently drops excess particles (capped by `amount`), but the CPU cost of the API calls is real.

Consider capping total burst emissions per frame per emitter (e.g. `MAX_AMOUNT_PER_EMITTER`) to bound the CPU cost:
```rust
let mut total_emitted = 0u32;
for (count, probability) in pending {
    if total_emitted >= MAX_AMOUNT_PER_EMITTER as u32 { break; }
    if rng.gen::<f32>() <= probability {
        let capped = count.min((MAX_AMOUNT_PER_EMITTER as u32).saturating_sub(total_emitted));
        emit_burst_particles(&mut item.node, &item.last_value, capped, &mut rng);
        total_emitted += capped;
    }
}
```

### Nit: `serde_json::to_string` runs even when debug logging is off

L190 — the serialization inside the `if let Ok(json) = serde_json::to_string(…)` block runs unconditionally; only the `tracing::debug!` output is gated by log level. Not a hot path (only fires on dirty components), but easily avoidable:
```rust
if tracing::enabled!(tracing::Level::DEBUG) {
    if let Ok(json) = serde_json::to_string(&new_value) {
        tracing::debug!("ParticleSystem apply entity {:?}: {}", entity, json);
    }
}
```

## What looks good

- Clean separation: process material (physics/motion) vs draw material (rendering) vs burst driver vs presence policy — each a focused function.
- Proper resource cleanup in `deleted_entities.rs`, `Scene::dispose_all()`, and the remove-node path.
- Presence policy matches Unity's `ParticleSystemPlaybackSystem`: stop+clear on leave, restart on enter, event-driven from `scene_manager`.
- Log levels are well-calibrated — all `warn!`, no `error!`, debug serialization behind `tracing::debug!`.
- Defensive quaternion sanitization (`dcl_quat_to_godot`) handles zero/unnormalized input scenes can send.
- Budget capping per-emitter and per-scene with a clear path to hook into the dynamic graphics manager.

No blockers — the AABB and burst findings are correctness and performance improvements that could land in a follow-up. Solid work on a long-awaited feature (#1538).

rodrigo685 (approved):

kuruk-mm (approved):
Great work!

PR #2591, opened 2026-07-27
fix(social): stop the friends 2-second reconnect loop (Sentry noise)

### LINKED ISSUE #2590: [Bug] Friends connection drops and reconnects every 2 seconds (floods Sentry)
## What happens

After being logged in for about 30 seconds, the friends connection starts dropping and reconnecting every 2 seconds, and it never stops. In the logs it looks like this, repeating forever:

```
[FriendsPanel.SubscriptionState] subscription dropped — reconnecting in 2s
```

Because of this:

- **Sentry gets flooded.** Most of the "social" reconnect noise we see in Sentry comes from this loop — the same errors reported over and over, from every logged-in player, every few seconds.
- Friend online/offline status stops updating.
- The friends panel keeps refreshing for no reason, and we hit the server with useless requests every 2 seconds.

## Why it happens

Every 30 seconds the client re-subscribes to the friends streams (friendship, connectivity, blocks) on the same connection, "just in case". The problem: the RPC protocol has **no way to close a stream that is already open**, so the server still sees the old subscription as active. When the new subscription arrives, the server says "you already have one" and closes it right away. The client thinks the connection dropped, waits 2 seconds, subscribes again, gets closed again... forever.

The server even has a metric for this exact situation (`subscription_duplicates_total`) with a comment saying *"a sustained high rate on the metric indicates a client stuck re-subscribing"*. That client is us.

The Unity explorer doesn't have this problem because it follows one simple rule: **never re-subscribe while a stream is still alive.**

Bonus finding: the "empty BlockUpdate" events we were seeing are part of the same story. Before closing a duplicate stream, the server sends a small "stream closed because: duplicate subscription" note. Our protos are older and don't know that field, so it decodes as an empty update.

## How to reproduce

1. Log in with a real account (not guest).
2. Wait ~30 seconds.
3. Watch the logs: the drop/reconnect cycle starts and never stops.

kuruk-mm:
Fixes #2590

## What was wrong

Every 30 seconds we re-subscribed to the friends streams on the same connection, "to be safe". But the RPC protocol can't close a stream that is already open, so the server still had the old subscription and rejected each new one as a duplicate, closing it instantly. The client saw that as a dropped connection and retried every 2 seconds — forever. This loop is where most of the social reconnect noise in Sentry comes from. Full investigation in #2590.

## What this PR does

It applies the same rule the Unity explorer already uses: **never replace a live stream.**

- If a subscription is already running, calling subscribe again simply does nothing (and reports success). So the 30s safety timer and the reconnect handler can keep calling subscribe without breaking anything.
- If subscribing fails or times out, we drop the whole connection and reconnect fresh. That's the only safe recovery: a failed subscribe can leave a stream half-open on the server, and we can never close it from our side.
- `disconnect()` now also stops the stream listeners, so logging out and back in starts clean.
- (First commit) Ignore the empty `BlockUpdate` the server sends right before closing a stream — it's a "stream closed" notice that our older protos decode as an empty update.

No GDScript changes needed: the existing timers and reconnect logic in `explorer.gd` become harmless once subscribe is idempotent.

## How I tested it

- **Before:** logged in with a real account, the drop/reconnect loop started ~30s after login and repeated every 2s forever (`stream CLOSED by remote` on every new subscription, confirmed with an instrumented rpc client).
- **After:** ran the client for several minutes — zero stream drops, subscriptions stayed alive through many 30s timer cycles, and the server no longer closed anything.
- `cargo fmt` and `cargo clippy` clean.

## Test plan (QA)

1. Log in with a real account (not a guest) and open the friends panel.
   - [x] Friends list and requests load fine, online/offline status of friends is correct.
2. Stay in-world for at least 10 minutes, then use friends again.
   - [x] The panel still works: status updates arrive, and sending/accepting a friend request works.
3. Minimize the app (on mobile: send it to the background) for a few minutes, then bring it back.
   - [ ] Friends recovers on its own: the list loads and online/offline status updates again, without restarting the app.


regenesis-claw (approved):
Excellent fix for a high-impact Sentry noise source. The approach — making subscribe idempotent via an atomic active-flag instead of tearing down live streams — directly mirrors the Unity explorer's strategy and is the correct fix for the RPC protocol constraint.

**What I verified:**

1. **`SubscriptionActiveGuard` RAII pattern** — clean. The `Drop` impl guarantees the flag clears on every exit path (success, cancellation, timeout, panic). The guard ownership chain is correct: created after CAS → moved into `async_subscribe_to_*` → moved into the detached `tokio::spawn` listener task. If the outer 15s timeout fires *before* the inner spawn, the future drops and the guard drops with it (flag clears). If it fires *after*, the listener task is already independent and holds the guard until stream end.

2. **`compare_exchange` gate** — correct use of CAS for idempotency. The `SeqCst` ordering is stronger than strictly needed (`Acquire`/`Release` would suffice) but this isn't a hot path, so clarity > micro-optimization.

3. **`disconnect()` cancellation** — the new code cancels all three stream handler tokens on disconnect, which unblocks the handler tasks → guards drop → flags clear → re-login subscribe calls succeed. There's a brief race window where a subscribe call between disconnect and handler exit would no-op, but the 30s timer self-heals this within one cycle. Fine.

4. **`SUBSCRIBE_TIMEOUT_SECS = 10` (inner) < 15s (outer)** — smart layering. The inner timeout lets the manager layer clean up the connection (`try_clear_connection`) before the outer Godot-layer timeout drops the future mid-RPC. This prevents the half-open stream problem the PR description identifies.

5. **Empty `BlockUpdate` filter** — the guard against the server's stream-closed notice is minimal and correct. `address.is_empty()` is a safe discriminator since real block events always have an address.

6. **Connectivity updates timeout** — bonus consistency fix: `subscribe_to_connectivity_updates` now has the same timeout+error-handling wrapper that friendship and block subscriptions already had.

7. **Log levels** — all new logs are appropriate: `tracing::debug!` for no-op/empty-update (no Sentry), `tracing::error!` for genuine subscribe failures and timeouts (actionable faults). Consistent with REVIEW.md logging discipline.

8. **No GDScript changes needed** — confirmed. The existing 30s timer and reconnect logic in `explorer.gd` become harmless once subscribe is idempotent.

All CI checks pass (Linux, Android, iOS, Docker, static checks, clippy). Branch is only 3 commits behind main — no concerning drift.

No issues found. Approving.

Followed repo REVIEW.md for this review.

Requested by Mateo via Slack

Ludmilafantaniella (commented):
QA results on this PR (Windows & iOS):

✅ Friend request notification — arrives correctly
✅ Accepting a friend request — works correctly
❌ Online status after accepting — friend doesn't show up under 'Online' friends (reproduced with two mutual friends online at the same time, neither shows for the other — screenshot attached)

Image

Ludmilafantaniella (commented):
## ❌ Request Changes - retested on Android + iOS.
Retested after the latest commits

**Online status still broken**
Two mutual friends, both online at the same time (Android + iOS), still don't show up as "Online" for each other. This is the same issue previously flagged - doesn't look resolved yet.

**Nearby tab - broken profile thumbnails**
In the Nearby tab, both avatars do show up for each other, but with a default/generic thumbnail instead of their actual profile picture (see attached screenshot with Ryumi highlighted).

**New issue - Friends system fully breaks after a while (iOS)**
After some minutes in-world on iOS, the Friends panel stops loading entirely and shows:
> "Whoops... Friends System is Down! Something went wrong and we couldn't retrieve your friends."

Screenshots attached for all three cases. Given the online-status issue persists and a new full-failure state showed up on Mac, I don't think this is ready to approve yet.

Screenshot_20260810_170707_Decentraland
Screenshot_20260810_170551_Decentraland
Screenshot_20260810_171118_Decentraland


cc @kuruk-mm

kuruk-mm:
Follow up https://github.com/decentraland/godot-explorer/issues/2590

PR #2599, opened 2026-07-28
Test APK against Adreno ubershader crash-fix engine branch

gonpombo8:
## Summary

Pins `GODOT_USE_BRANCH` to `fix/adreno-ubershader-crash` so CI builds the APK against the engine backport in decentraland/godotengine#19 (fix for the Adreno 6xx ubershader startup crash).

Draft — not meant to merge. After the engine PR merges: bump `GODOT_BUILD_SHA` to its merge commit and reset the pin to `None`.

## What could break

If merged as-is, every dev/CI pulls an unmerged, mutable engine build. The Android job here fails if it runs before the engine branch finishes publishing artifacts — re-run it in that case.

## How to test

Install the APK from the build-report comment on a POCO X3 Pro: it must reach the lobby instead of crashing ~6 s after launch. Sanity-check a known-good device.

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -28,10 +28,9 @@ pub const GODOT_BUILD_SHA: &str = "aed178f10";
 /// branch's CI build. An explicit `--branch` on the CLI still takes precedence over this.
 ///
 /// Reset to `None` once the branch is merged and `GODOT_BUILD_SHA` is bumped to the merge commit —
-/// leaving a branch pinned here makes every dev/CI pull an unmerged engine build. Currently `None`:
-/// the issue #2002 iOS main-thread-freeze fix (decentraland/godotengine#17) is merged and
-/// `GODOT_BUILD_SHA` above points at its merge commit.
-pub const GODOT_USE_BRANCH: Option<&str> = None;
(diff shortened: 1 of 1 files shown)


regenesis-claw (commented):
Small, self-contained change (`src/consts.rs` only) that pins `GODOT_USE_BRANCH` to the engine backport branch for on-device testing. Correctly labeled draft/do-not-merge with a clear checklist for the follow-up (bump `GODOT_BUILD_SHA`, reset pin to `None` once decentraland/godotengine#19 merges).

CI: Android build succeeded and produced an APK. iOS build failed with `invalid Zip archive` fetching the branch's iOS template — this matches the race condition you called out in the PR body (job ran before the engine branch's own CI finished publishing artifacts); the engine PR (#19) has since gone green on all platforms except a still-pending macOS Editor job, so a re-run of this iOS job should succeed now.

Nothing to block here given the explicit draft/do-not-merge intent. Just flagging: don't forget to remove this pin before it's accidentally merged as-is (it would leave CI/devs pulling a mutable, unmerged engine build).

Requested by Gon via Slack.

kuruk-mm:
Not required. Godotengine was merged to main

PR #2606, opened 2026-07-30
fix: populate PBPrimaryPointerInfo on the scene RootEntity (closes #2411)

### LINKED ISSUE #2411: [Bug] PBPrimaryPointerInfo (worldRayDirection) not being populated
## Summary

The renderer **never populates `PBPrimaryPointerInfo`** (SDK component id **1209**) on the scene's `RootEntity`. The proto is generated and wired into the CRDT layer (`get_primary_pointer_info` / `get_primary_pointer_info_mut` exist), but nothing ever writes a value — so `world_ray_direction`, `screen_coordinates` and `screen_delta` are always **`undefined`** to scenes.

For most scenes this is silently degraded behavior, but any SDK scene that takes one of these optional vectors and feeds it into a **serialized component** (e.g. a `Raycast` direction) hits an SDK serialization crash that **kills the entire scene** (black world).

## Impact

- High: a single scene reading `PrimaryPointerInfo.worldRayDirection` without a guard takes the whole scene down on our client.
- Mobile-dominant: the crash path runs whenever the cursor is **not** pointer-locked, which is **always the case on Android/iOS** (and on desktop with an unlocked cursor).
- Reproduced live on **Genesis Plaza** (`dclexperience2.dcl.eth`) under `--emulate-android`: the scene black-screens at load.

## Root cause

`PrimaryPointerInfo.getOrCreateMutable(engine.RootEntity)` returns a component whose optional `world_ray_direction` (`Vector3`) is `undefined` because the renderer never sets it. A scene that does:

```ts
Raycast.getMutable(engine.CameraEntity).direction = {
  $case: 'globalDirection',
  globalDirection: pointerInfo.worldRayDirection!  // undefined on our client
}
```

writes a `Raycast` (component **1067**) whose direction vector is `undefined`. On the next tick the SDK serializes the dirty component and the protobuf `Vector3` encoder does `value.x`:

```
TypeError: Cannot read properties of undefined (reading 'x')
  at Object.t [as encode]       (<loader>) // Vector3 encode -> reads .x
  at Object.t [as encode]       (<loader>) // Raycast.direction oneof encode
  at Object.serialize           (<loader>) // component 1067 serialize
  at Object.getCrdtUpdates      (<loader>)
  at Object.i [as update]       (<loader>)
```

Because the throw happens **inside** `getCrdtUpdates → serialize`, the scene never emits a valid CRDT frame, `CommunicatedWithRenderer` is never set, and our own watchdog kills the scene after 10 consecutive errors (`lib/src/dcl/js/mod.rs:577` — "too many errors without renderer interaction: shutting down"). End result: the world never renders.

## Where to fix

The component is already defined and CRDT-registered:
- Proto: `lib/src/dcl/components/proto/decentraland/sdk/components/primary_pointer_info.proto` (id 1209)
  - `optional PointerType pointer_type = 1`
  - `optional Vector2 screen_coordinates = 2`
  - `optional Vector2 screen_delta = 3`
  - `optional Vector3 world_ray_direction = 4`

It just needs to be **populated each frame on `SceneEntityId::ROOT`**, the same way `PbPointerLock` is written today on `SceneEntityId::CAMERA` in `lib/src/scene_runner/update_scene.rs:597-613`. The accessors `SceneCrdtStateProtoComponents::get_primary_pointer_info_mut(...)` already exist.

Suggested population:
- `world_ray_direction`: the world-space ray from the camera through the cursor (desktop) or screen center (mobile/locked). We already compute camera ray directions for pointer/raycast hit-testing — reuse that.
- `screen_coordinates` / `screen_delta`: cursor pixel position and per-frame delta (desktop mouse). On touch, set from the active touch point; `POT_NONE`/center is acceptable until touch drag is implemented.
- `pointer_type`: `POT_MOUSE` on desktop, `POT_TOUCH` on mobile.

At minimum, populating `world_ray_direction` resolves the scene-killing crash class.

## Workaround already shipped (scene side)

Genesis Plaza's pond raycast (`waterPond.ts`) was patched to guard the undefined and fall back to a camera-forward `localDirection`, marked as a temporary mobile workaround referencing this issue. That unblocks Genesis Plaza, but every other SDK scene that reads `PrimaryPointerInfo` without guarding remains exposed until the client implements it.

## Acceptance criteria

- [ ] `PBPrimaryPointerInfo` is written on `RootEntity` each frame with a valid `world_ray_direction`.
- [ ] `screen_coordinates` / `screen_delta` / `pointer_type` populated (desktop mouse + mobile touch).
- [ ] Genesis Plaza fishing minigame raycast works on mobile without the scene-side workaround.
- [ ] (Nice to have) renderer hardening so one scene's bad component can't black-screen the world.

EibrielInv:
## What & why

The renderer never wrote **PBPrimaryPointerInfo** (SDK component **1209**) onto a scene's `RootEntity`, so `worldRayDirection`, `screenCoordinates` and `screenDelta` were always `undefined` to scenes. Any SDK scene that feeds `worldRayDirection` into a serialized component (e.g. a `Raycast` direction) hit a protobuf `Vector3` encode crash (`Cannot read properties of undefined (reading 'x')`). That throw happens inside `getCrdtUpdates → serialize`, so the scene never emits a valid CRDT frame and our watchdog kills it after 10 errors → **black world**. This is the default state whenever the cursor is not pointer-locked — i.e. **always on Android/iOS**. Reproduced on Genesis Plaza's fishing minigame under `--emulate-android`.

Fixes #2411.

## What was done

- **`lib/src/scene_runner/scene_manager.rs`**
  - Added a `last_cursor_position` field (tracks the previous screen point for `screenDelta`).
  - In `physics_process`, compute an `Option<PbPrimaryPointerInfo>` once per frame using the same screen-point selection as `get_current_mouse_entity` (`cursor_position` when `raycast_use_cursor_position`, else `viewport_center`) and `camera_node.project_ray_normal(...)` for the world ray. Applies the Godot→DCL Z flip. Produces `None` when there is no camera node.
  - Threaded the value into `_process_scene`.
- **`lib/src/scene_runner/update_scene.rs`**
  - Added the `primary_pointer_info` parameter.
  - In `ComputeCrdtState`, after the PointerLock block, writes it to `SceneEntityId::ROOT` with a `PartialEq` changed-guard (mirrors the existing `camera_mode` / `pointer_lock` writes).

**Populated fields:** `pointerType = POT_MOUSE` (when a camera exists), `screenCoordinates` (pixels), `screenDelta` (per-frame pixel delta), `worldRayDirection` (world-space ray, Z-flipped to DCL space).

## QA steps

Use the **Test Scene** (available at coordinates **-31,19**), which has a dedicated **PrimaryPointerInfo** test block.

1. Open the world and walk to the scene at **-31,19**.
2. In the right-hand test switcher, select the **PrimaryPointerInfo** tab.
3. **Verify the component is populated** — a debug panel appears top-left showing, updated live every frame:
   - `component present: yes` (it must NOT say `NO — not written!`)
   - `pointerType`, `screenCoordinates`, `screenDelta`, and a non-`undefined` `worldRayDirection`.
4. **Normal view interaction** — hold the pointer: the on-pad sphere rolls toward where you look (uses `worldRayDirection`). It stays on the ground and can't leave the pad.
5. **Top View interaction** — press the **Top View** button (parks a top-down virtual camera and hides the joystick). Hold and swipe/drag: the sphere rolls in the swipe direction (uses `screenDelta`). Press **Reset Ball** to recenter it, and **Follow Player** to return.


lib/src/scene_runner/scene_manager.rs
```diff
@@ -1436,6 +1438,43 @@ impl SceneManager {
...
+            } else {
+                self.viewport_center
+            };
+            let dir = camera_node.project_ray_normal(screen_point);
+            let screen_delta = screen_point - self.last_cursor_position;
+            self.last_cursor_position = screen_point;
+
+            PbPrimaryPointerInfo {
+                pointer_type: Some(PointerType::PotMouse as i32),
+                screen_coordinates: Some(
```
regenesis-claw, line 1525: Non-blocking: `PotMouse` is hardcoded here, but on Android/iOS the input source is touch. If a scene ever branches on `pointerType` to distinguish input methods, it would get the wrong answer on mobile. Consider selecting `PotMouse` vs `PotTouch` based on the platform or `raycast_use_cursor_position`. Fine as a follow-up.

regenesis-claw (approved):
Clean, well-scoped fix for a crash-causing gap in the SDK contract. The approach is solid — compute once per frame in `physics_process`, write with a `PartialEq` changed-guard matching the existing `camera_mode`/`pointer_lock` pattern. Z-flip is correct per codebase convention.

One minor observation (non-blocking):

**`PointerType::PotMouse` hardcoded on all platforms** — on Android/iOS, the actual input is touch, not mouse. If any SDK scene branches on `pointerType` to distinguish mouse vs touch input, it would get the wrong value on mobile. Consider checking `OS.get_name()` or `raycast_use_cursor_position` to select `PotMouse` vs `PotTouch`. Not a blocker since scenes today primarily consume `worldRayDirection` / `screenDelta`, but worth a follow-up.

Ludmilafantaniella (approved):
## ✅ Approve
Tested on iOS + Android (primary focus), plus a quick check on Windows/macOS.

**Core fix**
- [x] Followed QA steps 1-5 on the Test Scene (-31,19): component populates correctly (`component present: yes`), `pointerType`/`screenCoordinates`/`screenDelta`/`worldRayDirection` all show valid values
- [x] Normal view: sphere follows `worldRayDirection`, stays on pad
- [x] Top View: sphere follows `screenDelta`, Reset Ball / Follow Player work
- [x] Regression: played the Genesis Plaza fishing minigame on iOS and Android - no crash.

No blockers on the core fix - iOS/Android look solid.

A few non-blocking observations from desktop testing (flagging in case useful, not sure if in scope since focus was mobile):
1. On Windows, Top View vertical drag seems inverted - dragging up/down moves the ball the opposite way
2. Follow Player behaves differently per platform - Windows needs held click while moving camera, mobile just needs camera movement
3. Multi-touch (mobile): second finger drag while holding with the first doesn't move the ball - not sure if multi-touch is in scope

Happy to file these separately if worth tracking.

https://github.com/user-attachments/assets/fe5e4354-81b8-4991-88ce-6fa0da377629

Image

kuruk-mm (approved):
lookign great

Issue #2583, opened 2026-07-27
IAP - Update Credits Packs icons 


regenesis-claw:
## Description
A new Credits icon has been designed by Elizabeth (Figma: https://www.figma.com/design/jjNmP9Z9pdw5fgg10Fg5W7/Product?node-id=5237-11195). We need to update the Credits and IAP Packs icons used in the native mobile app to match the new design.

## Context
Rollout should be coordinated with the Foundation team's marketplace release timeline so the icon update ships at the same time across mobile and the embedded marketplace (see companion issue in decentraland/marketplace).

<img width="500" height="550" alt="Image" src="https://github.com/user-attachments/assets/8c95604f-554d-4600-8ba9-78d95ab2edc5" />

**Change MARKETPLACE for SHOP.** 

## Notes
- Foundation team may still be iterating on the packages icons — confirm final assets before implementation.
- Coordinate release timing with marketplace rollout (tracked separately).

---
Requested by Rodri via Slack

ElzathRusso:
[Figma](https://www.figma.com/design/3QwlsmPqaS5DkqquUXHH7b/IAP?node-id=775-20682&t=XpLf0kz1dDj4kuIG-1)

**Change Credits Image:**

- [ ] Discover, topbar
- [ ] Credits, top bar
- [ ] Credits, each pack
- [ ] Backpack, my credits
- [ ] Backpack, cards with credits
- [ ] Modal Success

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/4a4311d1-9289-4e1e-bbe2-c6d9c30768c0" />

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/e0e97e94-016a-460d-9c2b-04fc567f7860" />

---

**Change text MARKETPLACE for SHOP:**
- [ ] Backpack, button
<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/0ee33841-9ce0-4d69-acee-b87a81e41f1d" />



- [ ] Backpack, category empty state with link to marketplace
<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/337ff726-f590-492f-8a82-528da92f5e16" />

PR #2608, opened 2026-07-31
feat(iap): mobile Shop — credits packs, new art, and the signing fixes behind them

### LINKED ISSUE #2609: IAP - New Mobile Shop and Credits IAP Setup + Icons Change
## Scope

Configure the mobile client to to the new Shop URL and process the new credit packs, plus the visual asset changes for the new credit design. This involves updating the endpoint/URL used by the mobile client when opening or embedding the Shop experience.

## Acceptance Criteria

# New Credits IAP & Icons-->[SEE FIGMA DESIGN](https://github.com/decentraland/godot-explorer/issues/2583) and [Prices Definitions](https://github.com/decentraland/godot-explorer/issues/2611)
- [ ] Create and configure new IAP Purchase and Prices
- [ ] Adjust Credits Quantities in UI
- [ ] Adjust Credits and Credit Packs ICON Assets
- [ ] Adjust wording "GO TO MARKETPLACE" for "GO TO SHOP"
- [x] Verify new Credits Purchases can be done
- [ ] Verify credit balance updates after purchase

# New Shop
- [ ] Mobile client opens the new Shop URL instead of the old one
- [ ] Verify the Shop loads correctly in the embedded webview
- [ ] Confirm navigation and interactions work as expected with the new URL
- [ ] Confirm purchases are successfully done and wearables accesible in-game

Image

---
Requested by Rodri via Slack

leanmendoza:
Groups the mobile Shop work: the storefront routes on zone, the `credits_tier_b` packs, the new Credits art, and the signed-fetch fixes that were blocking purchases.

Closes #2609. Also applies the assets from #2583.

## What changed

**Shop**
- `urls::marketplace()` resolves per environment — zone serves the storefront at `/shop`
- Storefront deep links are built from the env's route table
- Copy: "GO TO SHOP", and Settings → "Shop"

**Credits**
- `credits_tier_b1/b2/b3` packs registered and offered (40 / 100 / 260)
- Balance read from the USD rail that IAP actually credits
- Suggestions priced in the credits that buy them; the legacy MANA rail is gone from the client, so the balance is always in that same unit
- New pack art, inline icon, and purchase-success modal art
- Purchase history reads `/credit-orders`
- Design pass on the credits surfaces: widget/icon sizing and padding to spec, the whole pack card is a tap target, and the price button gets its pressed and disabled states back

**Auth**
- A sign-in callback no longer switches the environment mid-login
- The signed-fetch metadata fold is scoped to credits-server, so comms-gatekeeper and SDK7 `signedFetch` keep their ADR-44 bytes

**Tooling**
- `copy_if_modified` publishes through a rename, which avoids the macOS dylib overwrite hang

## QA

iPhone only, except for the Android step.

**Setup**

1. Kill the app from the app switcher.
2. Open this deep link on the phone:
   `decentraland://open?dclenv=catalyst::org,comms::org,events::org,mobilebff::org,profile::zone,zone`
   Easiest way is to send it to yourself on Slack and tap it from there.

**Credits**

3. Buy a credits pack. The whole card is tappable, not just the price button.
4. Open Purchases and check the purchase is listed.
5. Go back from Purchases and open it again — the button has to stay usable.
6. Keep buying until you hit the daily limit — it should stop you with a clear message.

**Accounts**

7. As a guest, buying should not be allowed.
8. Signed in with OTP, Google or Apple, buying should work.

**Shop**

9. Heads up: you log in twice — once in the app, once again in the Shop.
10. Buy wearables from more than one category, and an emote.
11. Go back to the app: the items you bought should be there.

**Back on org**

Every link above is built from a per-environment route table, and the setup step pins you to zone — so the org table needs its own pass.

12. Kill the app and open `decentraland://open?dclenv=org`.
13. Backpack: suggestions appear, with prices in a believable range (single digits to a few hundred).
14. Tap a suggestion, then DETAIL: the item page opens in the browser — not a "Page not found".
15. Pick a category with no suggestions: the "go to shop" link in the empty state opens the marketplace.
16. Settings → Shop opens, and so does the claim-a-NAME link.

**Android**

17. Nothing IAP should be visible: no credits in the Discover navigation, no suggestions in the Backpack.

## Known

- **Purchase history on org lists only Stripe orders.** credits-server [#590](https://github.com/decentraland/credits-server/pull/590) unions the Apple rail into `/credit-orders`; it is deployed on zone but not yet on org.
- **The history shows credits, not the amount charged.** No source has it: `/credit-orders` reports `usdCents`, which is the charge on the Stripe rail but the credits' $0.10 face value on the Apple one (40 credits → $4.00 against an $8.59 charge), and `iap_credits` stores no price or currency. Showing the money needs credits-server to persist `price`/`currency` off Apple's signed transaction.
- **Switching `dclenv` twice in one session needs a restart.** The first explicit env (a deep link or `--dclenv`) now outranks whatever a later deep link carries — that is what stops a sign-in callback from restarting the session mid-login. Kill the app between env switches.


godot/src/ui/components/molecules/marketplace_recommended_section/marketplace_recommended_section.gd
```diff
@@ -191,19 +214,36 @@ func _populate_cards(items: Array):
...
 	_set_rarity_background(card, item_data.get("rarity", "common"))
 
-	var price = _parse_price(item_data.get("minPrice", item_data.get("price", "0")))
-	var item_url = item_data.get("url", "")
-	var full_url = str(DclUrls.marketplace()) + item_url if not item_url.is_empty() else ""
+	# `priceCredits` is the price in the credits an In-App Purchase buys — the same
+	# unit as the balance the card compares it against to choose between DETAIL and
+	# GET CREDITS. The entry also carries `manaWei`, a different currency: 1 MANA is
+	# worth ~0.63 credits, so its figure reads about 1.6x too expensive.
+	var price := int(item_data.get("priceCredits", 0))
```
regenesis-claw: [P1] This makes the card price *always* credits, but `Iap.get_balance()` is only credits when the server answered with a `usd` block — `_async_fetch_balance` still has a MANA fallback (`totals.nonExpiring / _WEI_PER_MANA`). On any environment that hasn't switched the USD rail on, the card then compares credits against MANA: `wearable_item.gd:155` picks DETAIL vs GET CREDITS off that comparison, and `marketplace_cta_button.gd` picks its label the same way, so both are wrong by the ~1.6x you documented right here.

Two consistent options:
- If every environment that can run IAP now credits USD, the MANA branch is dead — drop it (or keep it and `push_warning` once) so the unit can't diverge.
- If org can still answer MANA, the price has to follow the same rail: read `manaWei` when the balance came from `totals`, and `priceCredits` when it came from `usd`.

Either way please state which one org returns today — that's the env real users are on, and it's the one the QA plan never exercises.

godot/src/iap/iap_manager.gd
```diff
@@ -750,12 +807,18 @@ func _async_signed_iap(path: String, method: int, body: String) -> Variant:
...
 	# Resolve the base URL per call so a runtime environment switch is picked up.
+	# Signed with folded metadata: credits-server runs crypto-middleware >=6.0.0, which
+	# rebuilds the payload with the metadata verbatim, so a folded signature only matches a
+	# folded header. Safe here and nowhere else on this client — credits-server reads
+	# `productId` off the body and never touches `authMetadata`.
 	var url := DclUrls.credits_server() + path
-	var response = await Global.async_signed_fetch(url, method, body)
+	var response = await Global.async_signed_fetch(url, method, body, true)
 	if response is PromiseError:
+		printerr("[IAP] ", path, " transport/auth error: ", response.get_error())
```
regenesis-claw: [P1] `_async_signed_iap` is called from `_async_poll_balance_after_purchase`, which retries `_POST_PURCHASE_POLL_ATTEMPTS` (12) times at 5s intervals. A phone that loses connectivity right after a purchase — the single most likely moment for it — now emits 12 error-level events per purchase, all identical. Per REVIEW.md §5 that's a quota burst, and a transport error here is expected and already recovered by the retry loop.

Suggested: downgrade to `print`, or keep it error-level but emit it only on the last attempt.

```suggestion
		print("[IAP] ", path, " transport/auth error: ", response.get_error())
```

godot/src/iap/iap_manager.gd
```diff
@@ -750,12 +807,18 @@ func _async_signed_iap(path: String, method: int, body: String) -> Variant:
...
 		return null
 	var json = response.get_string_response_as_json()
 	if not (json is Dictionary):
+		printerr("[IAP] ", path, " unparseable response: ", response.get_string_response())
```
regenesis-claw: [P2] This ships the whole credits-server response body into the Sentry event. That body is wallet-scoped purchase data (orders, amounts, ids), so an unparseable-response incident becomes a PII payload in the issue — and, when the failure is an HTML error page from an edge proxy, a multi-KB one. Log the status/length and truncate the body, e.g. `response.get_string_response().substr(0, 200)`, or drop the body entirely and keep `path`.

godot/src/iap/iap_manager.gd
```diff
@@ -408,6 +440,7 @@ func _async_begin_purchase(product_id: String, wallet: String) -> void:
 	if not envelope.get("ok", false):
 		var code := str(envelope.get("code", ""))
 		var reason := str(envelope.get("reason", ""))
+		printerr("[IAP] quote denied for ", product_id, ": code=", code, " reason=", reason)
```
regenesis-claw: [P2] A denied quote is an expected outcome with a defined `else` branch — you call `_show_quote_denied_modal(code, reason)` on the next line, and the QA plan lists hitting the daily limit and buying as a guest as *pass* conditions. Per REVIEW.md §5 that's a `push_warning`/`print`, not a Sentry issue: as written, every guest tap and every rate-limited buyer opens one.

```suggestion
		print("[IAP] quote denied for ", product_id, ": code=", code, " reason=", reason)
```

src/copy_files.rs
```diff
@@ -38,8 +38,31 @@ pub fn copy_if_modified<P: AsRef<Path>, Q: AsRef<Path>>(
...
-            .map(|_| println!("Copying {}", dest_path.to_string_lossy()))?;
+        // Copy to a sibling temp file and rename it into place instead of
+        // writing over the destination directly. `fs::copy` truncates and
+        // rewrites the existing file, so the inode stays the same -- and on
+        // macOS, rewriting a code-signed dylib that another process still has
+        // mapped wedges the vnode: every later `dlopen` of that path blocks
+        // forever inside dyld's slice validation, and the stuck processes
+        // survive SIGKILL. Renaming publishes a fresh inode, so new loads get a
+        // clean file and whoever still has the old one keeps it.
+        let tmp_path = dest_path.with_file_name(format!(
```
regenesis-claw, line 54: [P2] The temp name is derived only from the destination file, so two concurrent `xtask` invocations targeting the same output (two builds, or a build racing an export) both write `<name>.tmp`: one `fs::copy` interleaves into the other's file, and the loser's `remove_file` on the error path can delete the winner's temp before its `rename`. Since the whole point is to publish atomically, make the temp unique — appending `std::process::id()` (and/or a counter) to the suffix is enough.

godot/src/ui/components/molecules/marketplace_recommended_section/marketplace_recommended_section.gd
```diff
@@ -252,14 +292,38 @@ func _set_rarity_background(card: WearableItem, rarity: String):
...
+		return true
+	var gender := _string_field(item_data, "gender")
+	# Null for emotes, which have no body shape to mismatch — those are kept.
+	return gender.is_empty() or gender == "unisex" or gender == player_gender
+
+
+## Reads a string field, treating a JSON null as absent. `str(null)` is "<null>",
+## a non-empty string that would pass an is_empty() guard and reach a URL.
+func _string_field(item_data: Dictionary, key: String) -> String:
+	var value = item_data.get(key)
```
regenesis-claw, line 329: [P2] `_string_field` exists because this feed sends JSON nulls, but the numeric reads next to it aren't guarded the same way: `int(item_data.get("priceCredits", 0))` (line 237) and, in `iap_manager.gd`, `int(entry.get("credits", 0))` / `float(entry.get("createdAt", 0))` all take the *present-but-null* path, where the default never applies and the conversion pushes a Godot error (→ Sentry) before yielding 0. A `_int_field`/`_float_field` sibling to this helper closes it in three lines.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`.

**Pre-flight**
- Branch is up to date with `main` (ahead 18, behind 0).
- This touches auth/comms/IAP, so an iOS build is required — present and green (`🍏 iOS → R2 (unsigned export)` 13m41s). Static checks, Clippy, Linux, Android and Docker all pass; `🤖 Android → APK ready` still pending at review time.
- No submodule pointer drift.

The signed-fetch part is the right shape: `SignedMetadata::Verbatim` is the default, the fold is opt-in, and both gatekeeper call sites plus SDK7 `op_signed_fetch_headers` are pinned to `Verbatim` with a stated reason. I verified the blast radius of the two signature changes: `sign_request` has exactly 4 call sites (all updated in this diff), `async_get_identity_headers` is only called from `global.gd` (updated), and `Global.async_signed_fetch`'s new arg is defaulted so the other 13 GDScript callers keep ADR-44 bytes. The `Storefront` route table with unit tests over both tables is a good call.

Holding on the following.

| # | Sev | Finding |
|---|-----|---------|
| 1 | P1 | Suggestion prices are now in credits on every env, but the balance falls back to MANA when the server sends no `usd` block |
| 2 | P1 | `printerr` in `_async_signed_iap` fires once per poll iteration (12x per purchase) → Sentry quota |
| 3 | P2 | The unparseable-response `printerr` ships the full credits-server body to Sentry |
| 4 | P2 | `printerr` for a quote denial, which is an expected and already user-handled outcome |
| 5 | P2 | `_async_fetch_total` still asks `first=0` — unverified against the new `/v3/catalog/unified` |
| 6 | P2 | Numeric fields aren't null-guarded even though `_string_field` exists because this feed carries nulls |
| 7 | P2 | `copy_if_modified` publishes through a fixed `<name>.tmp` sibling |

**5 — `_async_fetch_total` on the new endpoint (unverified).** `_async_fetch_total()` still calls `_build_catalog_url(category, 0, 0)` (unchanged line, `marketplace_recommended_section.gd:200`) and reads `total` off the response. That contract was verified against `/v2/catalog`; this PR swings the whole section over to `/v3/catalog/unified`, which the PR itself says does *not* share a response shape. If the unified handler rejects `first=0` or names the count differently, `total <= 0` hides the entire suggestions section — silently, on the happy path. I can't check the endpoint from here; please confirm on zone *and* org, and add it as a QA case (the current plan never asserts that suggestions render at all).

**Test plan.** The QA block is executable and starts from a cold app, which is what we want, but it's missing the regression case for the surfaces this PR changed outside IAP: Backpack suggestions render with sane prices, an item card opens a real page (not a 404), the empty-state "go to shop" link, Settings → Shop, and the NAME-claim link — all of which changed route builders. Please add at least one case covering those on **org**, since every route/unit change here is env-conditional and the whole setup block pins the tester to a zone-composed env.

**Not blocking, worth a line in the description:** warm-start `dclenv` switching now needs an app restart once an env was set explicitly (the `dcl_env_explicit` guard). That's a deliberate trade and it's commented, but it changes a dev/QA workflow people use daily, so it should be visible in the PR body rather than only in the code.

No security issues found: no secrets in the diff, the metadata fold is correctly scoped to the one service that never reads metadata back, and no auth gate is loosened. The one data-exposure note is finding 3 below.

Requesting changes on 1 and 2; the rest are fine as follow-ups if you disagree.

ElzathRusso:
**Credits widget all screens** (discover, credits, and backpack)

- [x] Seems bigger and has poor padding
Screenshot 2026-08-20 at 4 25 57 PM

[Credits size 40x40 ](https://www.figma.com/design/3QwlsmPqaS5DkqquUXHH7b/IAP?node-id=775-21417&t=GZ7ty1vHLR7iwVmW-1)Widget height 60, width responsive

Screenshot 2026-08-20 at 4 33 22 PM

---

**Credits screen:**

- [x] Bundles: make the entire card interactive. 
- [x] Bundles: Increase the button size (back and font)
Screenshot 2026-08-20 at 4 25 07 PM

- [x] Check button states (the pressed state is now gray)
Screenshot 2026-08-20 at 4 27 30 PM

- Bundles: 100 image, seems a little small, increase (image size 182x182) [Link](https://www.figma.com/design/3QwlsmPqaS5DkqquUXHH7b/IAP?node-id=785-12719&t=GZ7ty1vHLR7iwVmW-1)

Screenshot 2026-08-20 at 4 51 11 PM


---
**Purchases screen**

- [x] Balance: check icon size and padding:
Screenshot 2026-08-20 at 4 42 41 PM

- [x] Purchase history:
Screenshot 2026-08-20 at 4 37 03 PM

Screenshot 2026-08-20 at 4 44 48 PM

- [ ] Extra if we can:
Change the amount of credit purchased for the price:
Screenshot 2026-08-20 at 4 46 19 PM


___
**Bug**

- [x] The purchase history doesn't appear.

- [x] After opening Purchases and going back, the Purchases icon remains disabled, so you can't access the section again.


https://github.com/user-attachments/assets/4920e6e3-40ef-4c94-8116-f83d560df6cd

regenesis-claw (approved):
Re-review of the delta since my last pass (`674de70` → `dc6dfd5`, 7 commits, 11 files). All six findings from the previous review are resolved, and Elizabeth's design/bug list is covered except for the two items she left unchecked. No new blockers — approving.

## Previous findings

| # | Finding | Status |
|---|---------|--------|
| P1 | Credits vs MANA unit mismatch | **Fixed.** The MANA rail is gone from the client (`_WEI_PER_MANA` deleted); a missing `usd` block now reports 0 with a one-shot `print`, so every surface that compares a balance against `priceCredits` compares the same unit. |
| P1 | Sentry burst from the post-purchase poll | **Fixed.** The transport/auth branch is `print` now, with the poll-count rationale in the comment. |
| P2 | Whole credits-server body dumped into the log | **Fixed.** Truncated to `_ERROR_BODY_EXCERPT_CHARS` (200). |
| P2 | `printerr` on a denied quote | **Fixed.** `print`, since a guest tap and the daily cap both land there with a modal already handling it. |
| P2 | `first=0` / `total` contract on `/v3/catalog/unified` | **Answered** — documented as verified on org and zone, for wearables and emotes, and called out as a QA case. I can't confirm the endpoint's behaviour from the diff, so this stays author-verified. |
| P2 | Missing null guards on numeric fields | **Fixed.** `_int_field`/`_float_field` added in both `iap_manager.gd` and `marketplace_recommended_section.gd`, applied to `credits`, `createdAt`, `total`, `priceCredits`, `chainId`. |
| P2 | Fixed `.tmp` sibling races concurrent xtask runs | **Fixed.** The temp name carries the pid. |

## Delta beyond the review

- **Purchases button stuck disabled** (Elizabeth's second bug) — real fix, and the diagnosis is right: the fetch callback enabled unconditionally while the shop view disabled on an empty list, and nothing re-fires `transaction_history_updated` from the history view. `_update_history_button()` is now the single rule for both.
- **Whole card is a tap target** — the layering holds up: `Button_Card` is the first child of the PanelContainer (drawn behind, hit-tested last), every node in front is `mouse_filter = 2`, and `Button_Price` is back to the default STOP so it can't double-fire. `_set_purchase_enabled()` moves both together, so the card body can't buy while the terms modal is up.
- **Header clone in the queue requester** — correct: `take()` left `None` on the response the caller still holds, and this only runs when the network inspector is attached.
- The `else` branch in `_async_show_terms_then_purchase()` also triggers when a modal exists whose `tree_exited` is already wired, which would re-enable early and drop the accept handler. Not reachable today (the card is disabled for the duration), so noting it rather than asking for a change.

## Two things worth confirming before this ships to production

1. **[P2] `usd` rail on org.** Dropping the MANA fallback is the right call, but it means an environment without the `usd` block reports 0 — the widget shows 0 and every suggestion falls to GET CREDITS. Worth confirming the flag is on in org before the release build, not after.
2. **[P2] Empty history has no empty state.** With `disabled = history.is_empty()` the Purchases button is greyed out and unexplained. On org that is the expected state until credits-server #590 lands there, so an Apple buyer sees a dead button right after a successful purchase. Enabling it with an empty-state message would separate "nothing to show" from "broken" — this is also Elizabeth's still-unchecked "purchase history doesn't appear" item.

## CI

All required checks green on `dc6dfd5`: Static checks, Clippy, Linux, Docker, Android, and `🍏 iOS → R2 (unsigned export)` — the iOS build this PR needs, since it touches auth. The only non-green entry is a `Build iOS export → R2` job still pending in the mobile-distribution run.

Security: no new issues. The signed-fetch scoping was verified in the previous pass and is untouched here; the response-body dump that was the one exposure note is now truncated.

---
Requested by Lean via Slack

kuruk-mm (approved):
looks great from my side!

PR #2616, opened 2026-08-03
fix: HUD overlays no longer swallow touch input meant for scene UI

### LINKED ISSUE #2578: [iOS] Debug panel causes invisible touch-blocking overlay, breaks UI close buttons + settings tabs
### Platform
iOS (local preview)

### App Version
v1.11.0

### Device Information
iPhone 15

### Issue Description
When the debug panel is enabled in local preview on iOS, an invisible/transparent overlay appears over part of the screen that blocks touches/clicks from being registered. This overlay area moves/shifts when the debug panel is opened or closed.

This is significantly affecting UI development/testing, since many "close" buttons for scene UIs end up positioned inside the blocked area, making it very difficult to dismiss UI panels once opened (reproduced with both Dead Surge and Cozy Farm UIs).

Additionally, in the app Settings screen, only the "Graphics" tab can be selected — no other settings tabs respond to taps. This may be caused by the same underlying issue (possibly the same invisible blocking overlay affecting tab selection), but this hasn't been confirmed.

Reported by Agustin Carriso; confirmed independently by Manuel Maceira, who saw the same symptom (could not modify Settings while the debug panel was enabled) the day before.

A minimal reproduction scene was created to visualize the affected areas (red squares mark zones where taps/clicks are not registered):
https://github.com/agustincarriso/dcl-mobile-screen-touch-test

Production (non-local-preview) builds do not exhibit this issue — only local preview.

### Expected Behavior
Enabling/disabling the debug panel should not create any invisible overlay blocking touch input elsewhere on screen, and all Settings tabs should remain selectable regardless of debug panel state.

### Screenshots / Media
Screenshots were shared in the reporting Slack thread showing red-square repro zones where taps are not registered, at different debug panel states. See the Slack thread for the original attachments (not re-attached here since issue creation via API can't attach files directly).

### Steps to Reproduce
1. Run local preview on iOS (v1.11.0)
2. Open the debug panel
3. Load the repro scene (https://github.com/agustincarriso/dcl-mobile-screen-touch-test) or a scene with UI close buttons (e.g. Dead Surge, Cozy Farm)
4. Attempt to tap UI elements/close buttons located in the area indicated by the repro scene's red squares — taps do not register
5. Toggle the debug panel open/closed and observe the blocked area shifts position
6. Separately, open in-app Settings and try tapping tabs other than "Graphics" — they don't respond

### Occurrence
100% while debug panel is active in local preview (per reporter); not observed in production builds

---

kuruk-mm:
Closes #2578

## Problem

On local preview (most visible on iOS), an invisible overlay blocked touch input over large screen regions — scene UI under it never received taps, close buttons died, and only the Graphics settings tab responded. The repro scene's red cells map exactly to a 500×280 rect that shifts when the stats panel expands/collapses.

## Root cause

The whole HUD and the SDK scene's UI (`SceneUIContainer`) live in one Control tree ordered by sibling index, and in Godot only `MOUSE_FILTER_IGNORE` is click-through against **lower siblings** — the `*Container` default `PASS` still claims the hit and only bubbles to ancestors. Three HUD widgets above `SceneUIContainer` relied on defaults:

1. **Debug panel body** (`debug_panel.tscn`): the right-aligned `VBoxContainer` defaulted to PASS while its `Control_Panel` child reserves 500×280 *even when the console is collapsed* (`Button_ShowHide` only hides the inner `TabContainer`). This is the shifting invisible rect from the issue — it moves because the stats panel shares the same right-aligned HBox. It also sat on top of the settings panel, which is why the non-Graphics tabs were dead.
2. **Navbar profile corner** (`navbar.tscn`): the 120×136 corner `Control` defaulted to STOP and its toggle `Button` was full-rect — ~2.5× the visible 80×80 bubble.
3. **Version/FPS strip** (`explorer.tscn`): bottom-center `HBoxContainer` defaulted to PASS.

## Fix

- `debug_panel.tscn`: `mouse_filter = 2` on `VBoxContainer` and `HBoxContainer_TopButtons`. The Console/Reload buttons and the open console still receive input (Buttons are STOP, `TabContainer` is PASS).
- `navbar.tscn`: `mouse_filter = 2` on the corner `Control`; the toggle `Button` hitbox now mirrors `Panel_Profile`'s 80×80 centered rect.
- `explorer.tscn`: `mouse_filter = 2` on `HBoxContainer_VersionFPS` (its Labels already ignore input).
- New regression lint `lib/src/hud_input_lint.rs`: plain `cargo test` unit tests (no Godot runtime) that parse the three `.tscn` files and pin these invariants — the same PASS-default trap already bitten before (`notifications.gd`), so this guards the bug class, not just these nodes.

Not touched here: `HBoxContainer_RightPanels` full-rect STOP while Friends/Settings/Notifications is open (intentional modal barrier), and the Console/Reload buttons being docked below scene UI via the virtual-joystick parent (a scene with pointer events over the top-right can still swallow *their* taps — inverse of this bug, worth a follow-up).

## Test plan

- [x] TDD: lint tests written first and failing (RED), pass after the tscn changes (GREEN)
- [x] `cargo test -- --skip auth` in `lib/` — 198 passed, 0 failed
- [x] `cargo clippy --tests -- -D warnings` clean; `cargo fmt` applied
- [ ] On-device verification with the [touch-test scene](https://github.com/agustincarriso/dcl-mobile-screen-touch-test): all cells reachable in preview with the debug panel enabled, both stats-panel states
- [ ] Sanity: Console/Reload buttons, stats toggle, profile navbar open/close, settings tabs all still tappable


kuruk-mm:
Follow-up pushed: `SafeMarginContainerDebug` (debug console + scene stats) moved between `SceneUIContainer` and `LeftRightSafeContainer`, so the debug tools still win input over scene UI but now render **below** the Friends/Notifications/Settings panels instead of on top of them. Ordering is pinned by a new `hud_input_lint` test. Re-added the `build` label for a fresh mobile build.

EibrielInv (approved):
Great work!

PR #2625, opened 2026-08-04
fix: subscribe Cast v2 LiveKit video tracks and port Unity stream selection

### LINKED ISSUE #2624: LiveKit video transmission not working: Cast v2 streams are never subscribed, and there is no stream selection
**Priority: high — scheduled for the current sprint (August v2).**

## Symptom

In-world screens using `livekit-video://` VideoPlayer sources stay black while a DCL Cast (browser) stream is live. Reproduced in `gather.dcl.eth` ("Gathering Stage" scene).

## Root cause

LiveKit rooms are joined with `auto_subscribe: false`, and the client only calls `publication.set_subscribed(true)` for participants whose identity ends in `-streamer` — the **legacy RTMP ingress** naming. Current comms-gatekeeper Cast v2 streamers join the scene room with identity `stream:{placeId}:{uuid}` (`role: 'streamer'` in metadata) — see `decentraland/comms-gatekeeper` `src/logic/cast/cast.ts`. So the video track is never subscribed and no frames ever arrive.

Two further gates in `lib/src/comms/adapter/livekit.rs` would drop the track even if subscribed:
- video branch requires `is_streamer || address.is_some()` — both false for `stream:...` identities
- streamer-audio branch requires the `-streamer` suffix, so Cast v2 audio is dropped too

## Secondary gaps (same area)

1. **No stream selection**: all video frames from all tracks are multiplexed into a single synthetic-address slot and broadcast to every `livekit-video://` player — two simultaneous streams interleave/flicker. unity-explorer keys tracks by `(identity, sid)` and selects one stream by priority (presentation bot → screen share → active speaker with 1.5s hold — see unity-explorer `docs/cast.md`).
2. `presentation-bot:` participants (authoritative video sources in Unity) are unknown to the Godot client.
3. The livekit connection string is rebuilt with `Uri::host()`, silently dropping explicit ports (breaks self-hosted/local LiveKit servers).

## Fix

- Subscribe by streamer identity (`-streamer`, `stream:`, `presentation-bot:`) **and** subscribe any video-kind publication (player tokens are mic-only, so every video track is an authorized stream — matches Unity).
- Port unity-explorer's stream selection: sid-keyed track registry + priority selection (presentation bot → unmuted screenshare → dominant active speaker with 1.5s hold → keep current → first available); forward only the selected track's frames to scenes.
- Preserve the URL authority (host + port) when parsing the connection string.

Verified end-to-end against a local livekit-server with simulated Cast v2 streamers (camera + screenshare preemption + fallback on track end).

kuruk-mm:
Fixes #2624

## Problem

`livekit-video://` screens stayed black during DCL Cast streams. Scene rooms connect with `auto_subscribe: false` and the client only subscribed publications from `-streamer` identities (legacy RTMP ingress). Cast v2 browser streamers join as `stream:{placeId}:{uuid}` (comms-gatekeeper `src/logic/cast/cast.ts`), so their video track was never subscribed; two further identity gates would have dropped the frames anyway. Additionally, the client had no stream selection: frames from **all** video tracks were multiplexed into a single synthetic-address slot and broadcast to every player, so concurrent streams interleaved.

## Changes

**`lib/src/comms/adapter/livekit.rs`**
- `is_streamer_identity()` — matches all three streamer identity shapes: `{id}-streamer` (RTMP ingress), `stream:{placeId}:{uuid}` (Cast v2), `presentation-bot:…` (bots)
- Subscribe streamer publications by identity **and** any video-kind publication from any participant — player tokens are mic-only (gatekeeper), so every video track is an authorized stream (matches unity-explorer)
- Video frames now carry the track `sid`; `InitVideo` carries publisher identity, source kind, and muted state; new events forwarded: `TrackMuted`/`TrackUnmuted`, `TrackUnsubscribed`, `ActiveSpeakersChanged`, `VideoTrackEnded` on stream drain
- `parse_conn_string()` — preserves the URL authority (host **and** port; `Uri::host()` silently dropped explicit ports) and fails gracefully on invalid input

**`lib/src/comms/adapter/message_processor.rs`**
- Sid-keyed video track registry + `best_video_track()`, a port of unity-explorer `LivekitPlayer.BestFollowCandidate`/`BestInitialVideoKey` (see unity-explorer `docs/cast.md`): presentation bot → unmuted screen share (muted share falls through) → dominant active speaker with a video track (1.5 s hold; the on-screen speaker keeps the stream) → keep current → first available
- Only the selected track's frames are forwarded to `scene.process_livekit_video_frame`
- Block-list filtering by publisher identity (wallet-identity publishers) instead of the synthetic address

**Deliberate follow-ups (not in this PR):** `livekit-video://{identity}/{sid}` pinning, playing all audio tracks simultaneously (Unity behavior; we keep the single streamer-audio stream), camera-off placeholder texture.

## Test plan

- [x] 16 new unit tests: streamer identity matching (4), conn-string parsing (5), stream selection (7 — bot/share/camera priority, muted-share fall-through, speaker-follow with hold, dominant-speaker-keeps-stream, next-speaker fall-through, dead-track fallback)
- [x] Full lib suite: 250 passed / 0 failed; `cargo fmt` + `cargo clippy -D warnings` clean
- [x] Live end-to-end against local `livekit-server` (docker) with simulated streamers driven via the debug hub: `stream:…` identity subscribed → `InitVideo` 640x360 → frames forwarded; camera selected → screenshare preempts → falls back to camera on track end
- [ ] Real DCL Cast stream in a world (needs a streaming key / scene admin)
- [ ] Mobile smoke test (Android/iOS build)


leanmendoza (approved):
LGTM!

lib/src/comms/adapter/message_processor.rs
```diff
@@ -153,11 +173,89 @@ struct ProfileFetchFailure {
...
+    let first_by_order = |mut candidates: Vec<(&String, &VideoTrackInfo)>| -> Option<String> {
+        candidates.sort_by_key(|(_, info)| info.order);
+        candidates.first().map(|(sid, _)| (*sid).clone())
+    };
+
+    let bots: Vec<_> = tracks
+        .iter()
+        .filter(|(_, info)| info.identity.starts_with(PRESENTATION_BOT_IDENTITY_PREFIX))
+        .collect();
+    if let Some(sid) = first_by_order(bots) {
```
regenesis-claw, line 221: This selects a bot track even when `muted == true`, unlike the screenshare tier (line 230) which explicitly checks `!info.muted`. If a presentation bot pauses its track, the viewer sees a frozen/blank frame instead of falling through to an active speaker or screenshare.

Either add `&& !info.muted` here, or document that Unity also pins muted bots intentionally.

lib/src/comms/adapter/livekit.rs
```diff
@@ -835,6 +888,46 @@ fn spawn_livekit_task(
                                 }
                             }
                         }
+                        livekit::RoomEvent::TrackMuted { participant: _, publication } => {
+                            let _ = sender.send(IncomingMessage {
```
regenesis-claw, line 892: `TrackMuted` fires for **all** track kinds (audio + video). This sends a `VideoTrackMuted` message even for voice tracks. The processor silently drops it (sid not in `active_video_tracks`), so no functional bug, but it generates unnecessary message traffic. Add `if publication.kind() == livekit::track::TrackKind::Video` to avoid the noise.

lib/src/comms/adapter/livekit.rs
```diff
@@ -835,6 +888,46 @@ fn spawn_livekit_task(
...
+                                    muted: true,
+                                },
+                                address: H160::zero(),
+                                room_id: room_id.clone(),
+                            }).await;
+                        }
+                        livekit::RoomEvent::TrackUnmuted { participant: _, publication } => {
+                            let _ = sender.send(IncomingMessage {
+                                message: MessageType::VideoTrackMuted {
+                                    sid: publication.sid().to_string(),
```
regenesis-claw, line 904: Same as `TrackMuted` above — fires for all track kinds, sends `VideoTrackMuted` for audio unmutes too.

regenesis-claw (commented):
## Adversarial review — concrete issues ranked by severity

Followed `REVIEW.md`. Pre-flight: branch is up to date (`CLEAN`), CI all green, iOS build passed. No prior reviews.

### 1. (Medium) `best_video_track` selects a muted presentation bot over everything

The screenshare tier has an explicit `!info.muted` guard, but the bot tier does not. If a presentation bot mutes its track (e.g. between slides, or paused playback), the selection sticks on the bot's frozen/blank frames instead of falling through to an active screenshare or camera.

Either the bot tier should also check `!info.muted`, or this needs a comment explaining why Unity intentionally keeps the bot pinned even when muted.

### 2. (Medium-Low) `active_speakers` is not room-scoped

`ActiveSpeakersChanged` overwrites `self.active_speakers` unconditionally regardless of `room_id`. If the client is connected to multiple LiveKit rooms simultaneously (e.g. comms room + scene room), a speakers update from a non-streaming room clobbers the streaming room's speaker list. This breaks the dominant-speaker-follow tier of stream selection.

Practical risk is low today (voice + cast are typically in the same scene room), but it's architecturally fragile. A room-scoped `HashMap<String, Vec<String>>` for `active_speakers` would fix it.

### 3. (Low) `TrackMuted`/`TrackUnmuted` fire for audio tracks too

`livekit.rs` sends `VideoTrackMuted` for *every* track kind — audio mute/unmute events included — because neither handler checks `publication.kind()`. The processor silently drops them (sid not in `active_video_tracks`), so it's functionally harmless but generates unnecessary message traffic proportional to voice activity.

A `publication.kind() == TrackKind::Video` guard in both handlers would keep it clean.

### 4. (Unverified) PeerLeft cleanup doesn't reach streamer-identity tracks

`handle_peer_left` retains tracks where `identity_h160 != Some(address)`. For streamer identities (`stream:...`, `presentation-bot:...`), `identity_h160` is `None`, so the retain predicate is always `true` — those tracks survive PeerLeft. This relies on `TrackUnsubscribed` firing *before* `ParticipantDisconnected` in LiveKit's event order to clean them first. LiveKit's docs suggest this is the normal order, but I can't verify the guarantee from this diff alone. If that ordering assumption ever breaks, streamer tracks leak in the registry.

### 5. (Unverified) iOS smoke test

PR description has `[ ] Mobile smoke test (Android/iOS build)` unchecked. The PR changes `lib/src/comms/adapter/` — per `REVIEW.md` this is iOS-sensitive. CI iOS build passed, but no manual confirmation that Cast video renders correctly on iOS.

### Claims verified against the diff

- `is_streamer_identity` matches all three identity shapes — **verified**
- `parse_conn_string` preserves host + port via `authority()` — **verified** (old code used `host()` which dropped ports)
- sid-keyed registry replaces H160-keyed map — **verified** (fixes collision when multiple non-wallet streamers existed)
- `best_video_track` priority cascade matches described algorithm — **verified** (modulo issue #1 above)
- block-list uses `identity_h160` not `message.address` — **verified**
- only selected track's frames forwarded — **verified** (`selected_video_sid` gate in VideoFrame handler)
- 16 unit tests — **verified** (5 parse + 4 identity + 7 selection)
- graceful invalid-conn-string handling — **verified** (`Option` return + early exit with `LK_STATE_DISCONNECTED`)

Requested by Lean via Slack

PR #2626, opened 2026-08-04
fix: emote wheel shows empty slots instead of ghost default icon

### LINKED ISSUE #2458: Emote Wheel: Show Empty Slots When No Emote Is Equipped
## Description

While fixing the emote wheel rarity backgrounds (#1743), Seba noticed a UX inconsistency: when a user unequips an emote from a slot in the backpack, that slot shows up empty in the backpack (as expected), but the Emote Wheel still displays the *default* emote icon in that slot — even though it can no longer be used/triggered.

This is confusing: the wheel visually implies an emote is available in that slot, but nothing happens when selected.

## Expected Behavior

The Emote Wheel should visually reflect empty slots (no emote equipped) instead of falling back to showing a default/ghost icon that isn't actually functional.

This requires design input to define what an "empty slot" should look like in the wheel (e.g. dimmed/placeholder state, "+" to equip, hidden slot, etc.) before implementation.

## Context

- Found while working on the emote wheel rarity background fix (#1743)
- A quick fix was attempted but needs more investigation + a defined design for empty slot states
- Related: #1743 (Emote wheel missing rarity background color/outline)

Requested by Rodri via Slack

sdilauro:
Closes #2458

## What

When a user unequips an emote from a backpack slot, the Emote Wheel used to keep showing the *default* emote icon in that slot — a ghost that looks usable but triggers nothing. The wheel now reflects empty slots explicitly.

`avatar_data.get_emotes()` returns an empty string for an unequipped slot; the wheel was calling `async_load_from_urn("")`, which errored out and left the previous icon on screen. It now renders a proper empty state instead.

## Also

As part of this fix I also reworked the emote wheel item to make future redesigns easier. The slot is now built from a single, vertically-laid-out template that's cloned and rotated into place at runtime, instead of ten hand-positioned instances — so tweaking the geometry only means editing one node. Its visual states were simplified from several stacked, visibility-toggled outline nodes down to a single outline whose color is driven by state, with the state colors exposed as @exports for quick iteration from the inspector. The result is fewer moving parts, a cleaner layout, and a slot that's much simpler to restyle when the design evolves.

image

## Test plan

1. Open the backpack → Emotes tab, and **unequip** an emote from one of the 10 slots.
2. Open the **Emote Wheel** (emotes button). The just-unequipped slot shows an **empty slot** (dimmed purple background + `+` icon + purple outline), not the old default emote icon.
3. **Tap / click the empty slot** → the wheel closes and the backpack opens on the **Emotes** tab.
4. **Equip an emote** back into that slot, reopen the wheel → the slot now shows the emote thumbnail with its correct rarity color/outline.
5. Trigger a non-empty emote from the wheel → it plays as before.
6. **Regression – backpack grid**: in the Emotes tab, equip/unequip and select emotes; selection highlight, equipped mark and empty thumbnail behave exactly as before.


regenesis-claw (approved):
Good fix and a solid cleanup of the wheel item architecture. Approving.

**What I checked:**
- Empty-slot lifecycle: `set_empty()` sets `_is_empty = true` + dirty flag; `async_load_from_urn()` clears it back to `false` — no stale ghost state.
- All node references (`%Pressed`, `%Equiped`, `%Outline`, `%TextureRect_Add`) use `get_node_or_null` + `is_instance_valid` guards — the shared script works safely in both the wheel item (has `%Outline`/`%TextureRect_Add`, no `%Pressed`/`%Equiped`) and the square backpack item (vice versa). No nil-access risk.
- `duplicate()` clone in `_ready()` correctly clears `unique_name_in_owner` on clones to avoid ambiguous `%` lookups.
- `_on_item_rect_changed` (runs with `@tool`) uses inline `get_node_or_null("%TextureRect_Add")` instead of the `@onready` cached var — correct, since `@onready` may not have fired in editor context.
- Outline state precedence (pressed > equipped > empty > normal) is sensible and covers all combinations.
- Empty-slot tap → opens backpack emotes tab via `Global.open_backpack.emit(true)` + haptic — good UX, returns early before the emote-disabled check (correct, since there's nothing to play).
- Hover label shows "Equip Emote" for empty slots instead of a blank name.
- Near-zero `rotation = 5.1e-05` on `Control_Wheel` is a harmless editor float artifact.
- No debug prints, no non-English comments, no `.claude/` files.
- CI all green (Linux, Android, iOS, static checks, clippy).
- Test plan is thorough and QA-executable.

Followed repo REVIEW.md.

ElzathRusso (approved):

Ludmilafantaniella (approved):
## ✅ Approve
Tested on iOS and Android.

- [x] Unequipped an emote from backpack - Emote Wheel shows proper empty slot (dimmed purple + `+` icon + purple outline), no more ghost default icon
- [x] Tapping the empty slot closes the wheel and opens backpack on Emotes tab
- [x] Re-equipping an emote shows correct thumbnail with proper rarity color/outline
- [x] Non-empty emotes still trigger and play correctly from the wheel
- [x] Regression: backpack grid equip/unequip/select behaves as before - selection highlight, equipped mark, empty thumbnail all unaffected

https://github.com/user-attachments/assets/551dab19-900e-4579-824b-8aa8ca161d46

PR #2640, opened 2026-08-05
fix: re-parented GLTF container pins the loading screen for 2 minutes

leanmendoza:
## Problem

A scene can sit on the loading screen for **~2 minutes** while it is already fully playable behind it — the "RUN ANYWAY" path works instantly and everything is there. Observed on an Android device against a local SDK preview, but neither defect is preview-specific.

The hang is not infinite: it ends at **125s**, terminated by the 120s per-container timeout in `gltf_container.gd`, not by anything recovering.

## Root cause

**1. A re-parented `GltfContainer` never resumes its load.**

Re-parenting an entity is a normal SDK operation, and it takes the container out of the scene tree and back in. `_exit_tree` detached the container from its shared `LoadGroup` but left `dcl_gltf_loading_state` at `LOADING` with `_requested_hash` still set — and `_ready`, the only caller of `async_load_gltf`, runs **once per node**, so re-entering the tree re-requested nothing.

The container was then in `LOADING` with no group behind it and no way back: Rust kept counting it in `scene.gltf_loading`, so the session's `expected_assets` never balanced and the Assets phase never cleared.

On-device trace that pinned it (instrumented build, since reverted). One `coin_01.glb` group:

```
12:18:00.829  CREATE   coin_01.glb
              ATTACH   x30
12:18:05.34–.43  UNREGISTER x30   group state=STATE_REALIZING(3), every waiter still LOADING
              ERASE-retire       waiters=0
              (zero further events for this group)
```

Entities 2341–2346 each show `ATTACH=1, UNREGISTER=1` and never re-register. Six containers, stuck forever.

**2. The Assets phase had no escape hatch, which turned 6 stuck assets into a full-screen stall.**

`check_loading_timeouts` already force-marks a scene ready after `SCENE_TIMEOUT_SECS` without progress — but it only fed `ready_scenes`, and the `all_loaded` gate never consulted that set. So the one existing recovery path could not actually release the phase, and **6 unresolved assets out of 231** held the entire loading screen.

## Fix

**`godot/src/decentraland_components/gltf_container.gd`**
- `_exit_tree` now clears `_requested_hash`, which is what marks the container as "detached from any group". The loading state deliberately **stays** `LOADING` across the detach, so a re-parent resumes rather than reporting a false `FINISHED`.
- New `_enter_tree` re-requests the load when it finds itself mid-load and detached. The guard is `state == LOADING and _requested_hash.is_empty()`, so a finished or errored container is never reloaded. On the very first tree entry `_enter_tree` runs before `_ready` while the state is still `UNKNOWN`, so the guard fails and no double request happens; `GltfLoadingCoordinator.request()` is idempotent anyway (`if group.waiters.has(container): return`).

**`lib/src/scene_runner/loading_session.rs`**
- A scene in `ready_scenes` no longer gates the Assets phase — either it genuinely finished, or the timeout sweep force-marked it ready. This connects the existing timeout to the gate it was always meant to release.

## Testing

- New `test_stalled_asset_does_not_pin_assets_phase` reproduces the scenario end to end: 6 assets started, 5 loaded, discovery settled, scene goes quiet past `SCENE_TIMEOUT_SECS` → the sweep marks it ready → the session must reach `Done` at 100%. Verified non-vacuous (fails on the pre-fix gate).
- `cargo test --lib loading_session` → 5/5 pass. `cargo clippy -- -D warnings` clean, `gdformat`/`gdlint` clean over `godot/`.
- **Not yet verified on device:** the GDScript side has no automated coverage. Repro is: start a local preview, enter the scene, kill the preview server mid-load — the loading screen should now dismiss instead of waiting out the 120s timeout. Adding the `build` label to get an APK for that check.

### DIFF
--- a/godot/src/decentraland_components/gltf_container.gd
+++ b/godot/src/decentraland_components/gltf_container.gd
@@ -26,14 +26,30 @@ func _ready():
 
 
 func _exit_tree():
-	# Detach from the shared load group
+	# Detach from the shared load group. Leaving the tree is NOT the same as
+	# finishing the load: `dcl_gltf_loading_state` deliberately stays LOADING so a
+	# re-parented container resumes rather than reporting a false FINISHED. Clearing
+	# `_requested_hash` is what marks it as "no longer attached to any group" —
+	# `_enter_tree` uses exactly that to re-request.
 	if not _requested_hash.is_empty():
 		GltfLoadingCoordinator.unregister(self, _requested_hash)
+		_requested_hash = ""
 
 	# Free pending orphan node (legacy path)
 	if dcl_pending_node != null:
 		dcl_pending_node.queue_free()
 		dcl_pending_node = null
+
+
+func _enter_tree():
+	# Re-parenting an entity (a normal SDK operation) takes its GltfContainer out of
+	# the tree and back in. `_ready` only ever runs once, so without this the load is
+	# never re-requested: the container sits in LOADING with no group behind it, Rust
+	# keeps counting it in `scene.gltf_loading`, and the loading screen stays pinned
+	# until the 120s per-container timeout. Only re-request when we are genuinely
+	# mid-load and detached — a FINISHED or errored container must not reload.
+	if dcl_gltf_loading_state == GltfContainerLoadingState.LOADING and _requested_hash.is_empty():
+		async_load_gltf.call_deferred()
 
 
 #region Loading Flow
--- a/lib/src/scene_runner/loading_session.rs
+++ b/lib/src/scene_runner/loading_session.rs
@@ -360,8 +360,15 @@ impl LoadingSession {
                 // Check if all registered assets are loaded
                 let total_expected: u32 = self.expected_assets.values().sum();
                 let total_loaded: u32 = self.loaded_assets.values().sum();
+                // A scene already in `ready_scenes` no longer gates this phase: either it
+                // genuinely finished, or `check_loading_timeouts` force-marked it ready after
+                // SCENE_TIMEOUT_SECS without progress. That timeout was the only escape from a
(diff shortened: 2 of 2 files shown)


regenesis-claw (approved):
LGTM — both halves of the fix are well-reasoned and the new test is solid.

**Pre-flight:** branch CI all green, iOS build present, no submodule drift.

**Review notes (following REVIEW.md):**

**GDScript (`gltf_container.gd`):**
- The `_exit_tree` → clear `_requested_hash` + new `_enter_tree` re-request pattern is correct. The guard (`state == LOADING and _requested_hash.is_empty()`) correctly distinguishes: first tree entry (state UNKNOWN → skip), re-parent mid-load (LOADING + empty hash → re-request), finished/errored container (state != LOADING → skip).
- `call_deferred` is appropriate for tree-entry context and also coalesces multiple re-parents within a single frame.
- `GltfLoadingCoordinator.request()` is confirmed idempotent (`if group.waiters.has(container): return` at line 97), so even if somehow called twice, no harm.

**Rust (`loading_session.rs`):**
- Adding `self.ready_scenes.contains(scene_id)` to the `all_loaded` gate is the correct fix — it connects the existing `check_loading_timeouts` → `mark_timed_out_scenes_ready` escape hatch to the gate it was always meant to release. Previously the timeout marked scenes ready but the asset gate never checked that set.
- The new test (`test_stalled_asset_does_not_pin_assets_phase`) properly exercises the full chain: 5/6 assets loaded → timeout sweep → phase should advance. Non-vacuous (verified to fail on pre-fix code).

No concerns — both changes are minimal, well-scoped, and independently correct.

Requested by Lean via Slack

kuruk-mm (approved):
looks great

PR #2641, opened 2026-08-05
fix: loading screen pinned forever when no island parcel is in view

### LINKED ISSUE #2634: [Bug] Loading Stuck when entering Pixel Arcade (Trivia) scene
### Platform
Mobile (RC build / Release Candidate)

### App Version
Release Candidate (RC) — exact version not specified

### Device Information
Not specified

### Issue description
The "Pixel Arcade" trivia scene shows a blocking modal on entry that prevents normal play. Reported by Gon: it worked the first time entering the scene, but now happens consistently on every entry. Pressing "Continue" on the modal anyway lets the scene work despite the warning. Rodri confirmed reproducing the same issue.

Likely a loading/scene-related issue rather than a generic client bug — flagged for investigation with per-scene metrics.

Scene repo: https://github.com/TtLMaV/ThePixelArcade/tree/main/ThePixelArcade-TriviaTest

### Expected behavior
Scene should load normally without a blocking modal (or the modal should only appear when genuinely relevant, e.g. actual loading/compat issue), consistent with the first successful load.

### Screenshots/Media

Image

### Steps to reproduce
1. Enter "The Pixel Arcade" (Trivia scene) on a mobile RC build
2. Observe a modal appears on entry
3. Press "Continue" anyway
4. Scene loads/works despite the modal

### Occurrence
Started happening consistently after working once on first entry.

---
Requested by Rodri via Slack

leanmendoza:
Fixes #2634 

## Problem

The loading screen can stay up **indefinitely** over a scene that is already fully loaded and rendered. The user only gets out via the "Loading is taking longer than expected → START ANYWAY" modal, and everything works instantly once they do.

Captured on a Samsung A54, instrumented build (instrumentation not included here). The session sat like this for over 8 minutes:

```
t=  4.1s  phase=Assets  assets=40/40  ready=1
          waiting_islands=true  islands=0/176
          grace_ms=4029 (passed=true)  stable=true  all_loaded=true
...
t=470.0s  phase=Assets  assets=40/40  ready=1
          waiting_islands=true  islands=0/176   ← unchanged
```

Every real gate condition is `true` from 4.1s on. The phase is held by one flag.

## Root cause

`tick_culling` only emits `generation_complete` when `in_view_candidates > 0`:

```rust
if self.generating
    && in_view_candidates > 0
    && self.in_view_all_materialized(player, view, &camera)
```

A candidate counts toward that number when it is within the `dist <= 1` ring **or** passes `parcel_in_camera_view`. When every nearby candidate fails the camera test and none sits in the inner ring, the counter stays at zero and the condition becomes unsatisfiable. The same pass enqueues nothing in that state, so `generation_progress` never fires either and `generated_so_far` stays at 0 — generation is stuck with no way out.

The observed run makes the distinction concrete. The candidates are **not** far away:

```
islands candidates set: n=176 player=(0, 0) view=7
                        sample=[(-1, -3), (14, 20), (-3, 13), (2, -2), ...]
islands tick: candidates=176 in_view=0 missing_enqueued=0 active=0 pending=0
              generated=0/176 player=(0, 0) materialized=false
```

`(-1,-3)` and `(2,-2)` are well inside `player ± 7`. They are excluded by the visibility filter, not by distance. Across 64 consecutive samples `in_view` was 0 every time, with zero early-return aborts — the tick ran fine, it just had nothing it considered visible.

This reaches the loading screen because `generation_complete` is the only thing that clears `waiting_for_floating_islands` on the loading session (`scene_fetcher.gd``_on_islands_generation_complete`), and `LoadingPhase::Assets` returns early while that flag is set:

```rust
LoadingPhase::Assets => {
    if self.waiting_for_floating_islands {
        return false;
    }
```

So the phase is held regardless of asset state — which is why the dump shows `all_loaded=true` going nowhere.

## Fix

Treat "no candidates in view" as vacuously complete, debounced by `EMPTY_VIEW_COMPLETE_MS` (1s) so a startup frame where the camera is not yet positioned cannot cut generation short. The debounce field resets in `set_candidate_parcels` and `clear_all`.

Terrain building is unaffected: enqueueing is not gated on `generating`, so parcels are still built as they come into view later. The fix changes when generation reports *done*, not what it builds.

## Testing

Verified on device (Samsung A54) against the World that reproduced the stall, same build otherwise:

```
islands START count=176 has_session=true
islands candidates set: n=176 player=(0, 0) view=7
islands tick: candidates=176 in_view=0 ... generated=0/176
islands COMPLETE via empty-view escape (candidates=176, player=(0, 0), view=7)
islands FINISH has_session=true
t=2.1s phase=Assets ... waiting_islands=false islands=176/176
```

Generation completes ~1s in, the flag clears, and the loading session proceeds and finishes on its own. Without the fix the identical path held for minutes.

No unit test: `DclFloatingIslandsManager` is a `#[class(base=Node)]` whose tick needs a live camera, scenario RID and physics space, so the stuck state is not reachable from a headless test.

## Notes

- This is **independent** of #2640, which fixes two other causes of a pinned loading screen. That PR's Assets-gate change does not help here, because the `waiting_for_floating_islands` early return happens *before* the gate it relaxes is evaluated.
- Not preview-specific — reproduced on a deployed World, no local preview involved.
- The upstream question of *why* `parcel_in_camera_view` rejects parcels that are a couple of tiles from the player is not addressed here and is worth its own look. This PR removes the unrecoverable state; it does not change which parcels are considered visible.

### DIFF
--- a/lib/src/godot_classes/dcl_floating_islands_manager.rs
+++ b/lib/src/godot_classes/dcl_floating_islands_manager.rs
@@ -26,6 +26,11 @@ use crate::godot_classes::floating_islands::{
     PARCEL_HALF_SIZE, PARCEL_HEIGHT_BOUND, PARCEL_SIZE, TERRAIN_MATERIAL_PATH,
 };
 
+/// How long `tick_culling` must keep finding zero candidate parcels in view before it
+/// treats generation as finished. Debounced rather than immediate so a startup frame
+/// where the camera is not yet positioned cannot cut generation short.
+const EMPTY_VIEW_COMPLETE_MS: u64 = 1000;
+
 #[derive(GodotClass)]
 #[class(base=Node)]
 pub struct DclFloatingIslandsManager {
@@ -43,6 +48,9 @@ pub struct DclFloatingIslandsManager {
     generating: bool,
     generated_so_far: i32,
     generation_total: i32,
+    /// Ticks-msec of the first tick that found zero candidates in view while generating;
+    /// 0 whenever the last tick did find some. See `EMPTY_VIEW_COMPLETE_MS`.
+    empty_view_since_msec: u64,
 
     scenario: Rid,
     physics_space: Rid,
@@ -100,6 +108,7 @@ impl INode for DclFloatingIslandsManager {
             generating: false,
             generated_so_far: 0,
             generation_total: 0,
+            empty_view_since_msec: 0,
             scenario: Rid::Invalid,
             physics_space: Rid::Invalid,
             terrain_material: None,
@@ -214,6 +223,7 @@ impl DclFloatingIslandsManager {
         self.generating = true;
         self.generated_so_far = 0;
         self.generation_total = self.candidates.len() as i32;
+        self.empty_view_since_msec = 0;
     }
 
     #[func]
@@ -248,6 +258,7 @@ impl DclFloatingIslandsManager {
         self.generating = false;
         self.generated_so_far = 0;
         self.generation_total = 0;
+        self.empty_view_since_msec = 0;
         if let Some(worker) = &self.worker {
             while worker.rx.try_recv().is_ok() {}
         }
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
LGTM — clean fix for a real loading-screen hang.

**Pre-flight:** branch CI all green, iOS build present, no submodule drift.

**Review notes (following REVIEW.md):**

- The debounce approach (`EMPTY_VIEW_COMPLETE_MS = 1s`) is the right call — it prevents the startup-camera-not-yet-positioned false positive while still catching the real stuck state promptly. The sentinel `empty_view_since_msec = 0` is safe since `get_ticks_msec()` is always > 0 at runtime.
- Resets in `set_candidate_parcels` and `clear_all` are correct — the debounce timer doesn't survive across candidate-set changes.
- `tracing::debug!` for the escape path is the right log level (not an error/warning, just diagnostic info for loading telemetry).
- The comment block explaining the historical bug is valuable for future readers.
- No unit test expected here — `DclFloatingIslandsManager` needs a live camera + scenario RID + physics space, so the stuck state isn't reachable from a headless test. The on-device verification in the PR description is sufficient.

Requested by Lean via Slack

kuruk-mm (approved):
great!

PR #2642, opened 2026-08-06
feat: HUD + Social Panel Redesign — Iteration #1

### LINKED ISSUE #2343: Controls: Align HUD positioning between Android and iOS
## Description

HUD buttons are positioned differently between Android and iOS due to how each platform reports safe area margins. This inconsistency makes it difficult for scene creators to build predictable mobile UIs, as elements that appear correctly placed on Android end up misaligned on iOS.

### Observed Behavior

- **Android:** HUD is correctly positioned within safe margins.
- **iOS:** HUD is offset (shifted to the left relative to Android) and can render outside of safe margins, causing misalignment with scene UI.

Because scene creators using `ScreenInsetArea` inherit the same positioning logic as the client HUD, fixing the HUD consistency will also fix scene UI alignment across platforms.

### Context

- iOS reports safe margins on both sides (left and right), while Android only reports on the correct side. This causes the HUD to land in different positions.
- Scene tested: `Adventureland.dcl.eth` — scene UI that looks correctly aligned on Android is off-set and outside safe margins on iOS.
- Scene creators can use the [`ScreenInsetArea` component](https://docs.decentraland.org/creator/scenes-sdk7/building-for-mobile/safe-area#reserved-margins) to avoid overlap, but the underlying HUD inconsistency still makes cross-platform UI development unpredictable.

### Expected Behavior

HUD buttons should be positioned consistently between Android and iOS so that scene creators can rely on a single layout working across both platforms.

---
Requested by Rodri via Slack

### LINKED ISSUE #2352: Controls Customization: HUD + Social Panel Redesign - Iteration #1
## Problem:
Current HUD and Social Panel layouts need to be redesigned to accommodate the full controls customization feature set and provide a more flexible, user-friendly interface for mobile players.

## Goal:
Implement the redesign for the new HUD and Social Panel to support controls customization while improving overall usability and visual consistency.

## Requirements:
- Redesign HUD layout to integrate with controls customization system
- Redesign Social Panel to work within the new HUD framework
- Ensure compatibility with different control presets and customization options
- Maintain accessibility and usability across different device sizes
- Provide clear visual hierarchy and intuitive interactions

TBD: Detailed design specifications and implementation approach

---
Requested by Rodri via Slack

### LINKED ISSUE #2638: Define InteractableArea + DeviceSafeInset and expose them to scenes at runtime
##  **⚠️ Problem:**
Creators building mobile UI have no runtime way to ask the client where scene UI can safely live — the HUD-reserved margins are published as hardcoded pixel constants (480px left, 400px right, 58px top/bottom) that every scene copies by hand, and the HUD redesign in #2352 will silently invalidate all of them. Scene UI then lands under the joystick and action buttons where it competes for taps, which drives the mobile "feels broken" perception that suppresses `Playtime` and featuring.

## **🏁 Scope:**
Formally define two distinct concepts — **`InteractableArea`** (the region free of the Decentraland system HUD, client-owned and HUD-layout dependent) and **`DeviceSafeInset`** (the hardware notch / status bar / home indicator / rounded-corner inset) — and expose both to scenes as live values the mobile client reports, so layouts adapt instead of hardcoding. Out of scope: changing HUD layout itself (#2352), the rendering behavior of `ScreenInsetArea`, and any desktop-client equivalent beyond reporting zeroed values.

## **📝 Deliverables:**
- [ ] Written definition of `InteractableArea` and `DeviceSafeInset` — what each covers, what owns it, and how they compose — agreed with SDK and docs before implementation
- [ ] Mobile client reports current `DeviceSafeInset` values (top/right/bottom/left) to the scene runtime, updating on rotation and on system-bar show/hide
- [ ] Mobile client reports current `InteractableArea` derived from the live HUD layout, not from constants — it must change automatically when the HUD changes
- [ ] Values normalized `[0.0–1.0]` against the 1600×720 reference resolution, matching the convention already published in the creator docs
- [ ] Desktop client reports `DeviceSafeInset` as `(0,0,0,0)` and `InteractableArea` as the full viewport, so cross-platform scene code needs no branching
- [ ] Values change reactively — scenes re-layout on rotation, HUD show/hide, and the granular Hide UI toggles from #2055
- [ ] `CLICK_BUTTON`-style instrumentation is **not** required here; instead, confirm no regression in the safe-area behavior shipped in #2148 and #1898

## Metrics
1. **Target metric:** `Playtime`.
2. **Expected direction and size:** reduce scene-UI-obscured-by-HUD reports and raise the share of featured scenes that lay out correctly on mobile. No baseline exists — this is measured qualitatively via creator reports and QA passes until a scene-UI telemetry event exists.
3. **How it's measured:** QA regression pass on the safe-area cases from #1898 and #2148, plus creator-reported issues in the mobile channels. No Segment event is introduced by this issue.

## Dependency
- #2352 — Controls Customization: HUD + Social Panel Redesign - Iteration #1. `InteractableArea` is derived from HUD layout, so the two must agree on a single source of truth or the values go stale again the moment the redesign ships.
- #2351 — HUD + Social Panel Redesign (design). The design defines which screen regions the HUD claims.

## Notes
- Naming is not settled — `InteractableArea` vs `SafeUiArea` vs extending the existing `ScreenInsetArea`/`UiCanvasInformation` surface. Needs an SDK-side decision before implementation.
- Open question: is `InteractableArea` a single rectangle, or does it need to express the "right-side gap" between the profile avatar and the action cluster that the docs currently document separately?
- Open question: does exposing this require a protocol change in `decentraland/protocol`, or can it ride on an existing component?

## **🔗 References:**
- [decentraland/docs `creator/build-for-mobile/develop/safe-area.md`](https://github.com/decentraland/docs/blob/main/creator/build-for-mobile/develop/safe-area.md) — the page that currently hardcodes the reserved margins (480/400/58 px) and documents `ScreenInsetArea` for hardware insets. It is the clearest existing statement of both concepts, and the thing this issue makes dynamic.
- [decentraland/godot-explorer#1898](https://github.com/decentraland/godot-explorer/issues/1898) — full-screen scene overlays not covering the safe area; its "Suggested Fix" option 3 explicitly asks for a scene-accessible API for safe-area drawing. Closed by shipping `ScreenInsetArea`, which covers hardware insets only — the HUD half is still unaddressed.
- [decentraland/godot-explorer#2148](https://github.com/decentraland/godot-explorer/pull/2148) — safe-area-aware scene UI; the existing implementation path any new reporting must extend rather than duplicate.
- [decentraland/godot-explorer#2112](https://github.com/decentraland/godot-explorer/pull/2112) — HUD safe-area regression caused by a scene-tree restructure. Concrete evidence that HUD-derived margins drift when nobody owns them; the reason `InteractableArea` must be computed, not constant.
- [decentraland/godot-explorer#1100](https://github.com/decentraland/godot-explorer/pull/1100) — `--emulate-ios` / `--emulate-android` CLI flags with `safe_area_presets.gd`. The existing test harness for verifying inset values without a physical device.

sdilauro:
Closes #2352 
Closes #2343
Closes #2638 

## What's in here
- **The whole HUD is framed by one shared safe-area rectangle.** Navbar, chat, the navbar's side panels, emotes and the joypad now line up inside the same rectangle, driven by a single resource (`hud_margins.tres`) so the margins are defined once instead of per-widget — and it's orientation-aware (landscape vs portrait).
- **Camera mode moved to Settings › Gameplay**, and it's now a **dropdown** (First / Third person) instead of the old joystick button. When a scene locks the camera the dropdown greys out with a "set by the creator" note, and the whole section is hidden until you're actually in a world.
- **The virtual joystick travels the whole screen** (it's no longer boxed into the rectangle), while the joypad stays aligned to it.
- **Hide UI reworked**: the "Show UI" button now lives with the HUD and survives hiding; restoring always brings the chat back.
- Version/FPS strip pulled out of the rectangle, plus cleanup (removed the old camera button and its dead code, orphaned assets, stale comments).
- **Symmetric side margins**: a one-sided device inset (a landscape notch) widens the opposite side to match, so the HUD stays centered.
- **Tap any empty area to dismiss** the open panel or the chat — a full-screen catcher removes the dead zones the inset layout had introduced.
- **Scene interactable area** reported to creators (the HUD rectangle minus the fixed chat column, reported by the chat itself), with a dev-only overlay + `%` readout behind a Settings › Developer toggle.

## Test plan
Everything below is meant for a **phone (iOS or Android)**, in **both portrait and landscape** unless a step says otherwise. Desktop is fine for a quick sanity pass but it isn't the target.

### 🎥 Camera mode (Settings › Gameplay)
- [x] Open Settings › Gameplay — you should see a **Camera Mode dropdown** (not a toggle).
- [x] Pick **First person** → the camera actually goes first-person. Pick **Third person** → back to third.
- [x] Enter a scene that locks/forces the camera → the dropdown greys out and shows **"The camera mode is set by the creator. It can't be changed."** Leave the scene → it's editable again.
- [x] Before entering any world (lobby / discover), the Camera section shouldn't be there at all.
- [x] Nice-to-check: with Settings open in two spots, changing the mode in one updates the other.

### 🕹️ On-screen controls
- [x] The **joystick graphic reaches every edge** — no clipping near the bottom.
- [x] When the joystick first shows up (before you've touched it), its **resting spot doesn't overlap the emote button**.
- [x] You can start the joystick **anywhere** on the screen and move around.
- [x] The **joypad** (action buttons) sits inside the HUD rectangle, not jammed into the corner.
- [x] Rotate to **portrait** → joystick and joypad go away (chat takes over). Rotate back → they return.
- [ ] Tap a scene's own on-screen button → it presses that button instead of moving the avatar.

### 🙈 Hide UI
- [x] Turn on **Hide Interface** (Settings › Gameplay) → the HUD hides and a small **Show UI** button appears top-right.
- [x] You can still **walk around** while the UI is hidden.
- [x] Tap **Show UI** → the whole HUD comes back, **including the chatbar**, and anything that was closed stays closed.

### 💬 Social panels & chat
- [x] Open the navbar → **Notifications / Friends / Settings** panels open and line up on the left.
- [x] Tapping **any empty area** (including the side margins and the left/joystick side) closes the open panel or the chat — no dead zones.
- [x] In **portrait**, open the chat from the chatbar → it fills almost the whole screen (small margins, not a skinny column).
- [x] Type a message → the input sits right above the keyboard with no odd gap, and sending works.
- [x] Back in landscape → chat / navbar / panels line up with the same rectangle as everything else.

### 🧭 General layout & regressions
- [x] Nothing in the HUD overlaps, and nothing hugs the notch / home indicator / camera cutout.
- [x] On a phone with a **notch/cutout**, in landscape both sides of the HUD sit the **same** distance from the edge — the clean side matches the notch side, so the HUD stays centered (not lopsided).
- [x] Rotate the phone **the other way in landscape** (the notch jumps to the opposite side) → the HUD stays symmetric, matching the new notch side.
- [x] Rotate back and forth a few times → margins recompute cleanly, nothing gets stuck.
- [x] Quick poke at the **backpack** and **emote wheel** to make sure they still open and close fine.
- [x] Settings › Developer → **Show Interactable Area** (off by default) draws a blue rect + `%` readout starting to the right of the chat; it doesn't change when the chat/notifications show or hide.


regenesis-claw (changes requested):
Followed `REVIEW.md`. Pre-flight: branch is up to date with `main` (mergeable, just blocked by the failing check below). Not purely backend — this touches the HUD/mobile input surface, and there IS a green iOS-adjacent signal here: `🤖 Android` build passed and posted an APK, but no `build`/`build-ios` label has been added yet, so there's no iOS artifact. Per REVIEW.md Section 0: **holding approval until a maintainer adds the `build` label and an iOS build goes green**, given this PR touches safe-area/virtual-keyboard-adjacent layout on mobile.

## 🔴 Blocker — CI is red (Linux build/test failure)
`lib/src/hud_input_lint.rs::debug_panel_body_does_not_block_scene_ui` fails:
```
node HBoxContainer_TopButtons with parent MarginContainer/VBoxContainer not found — was it renamed?
```
`debug_panel.tscn` now wraps `HBoxContainer_TopButtons` in a new `header` `MarginContainer` (`MarginContainer/VBoxContainer/header/HBoxContainer_TopButtons` instead of `MarginContainer/VBoxContainer/HBoxContainer_TopButtons`) — good, `mouse_filter = 2` is still correctly set on both the new `header` and `HBoxContainer_TopButtons`, so the actual click-through behavior looks right. But the lint test's `find()` call at `lib/src/hud_input_lint.rs:85-91` (`debug_panel_body_does_not_block_scene_ui`) still points at the old parent path and wasn't updated. Note the PR *did* correctly update the sibling tests in the same file (`debug_tools_sit_between_scene_ui_and_right_panels`, `version_fps_strip_does_not_block_scene_ui` for the `LeftRightSafeContainer` → `SafeAreaHud` rename) — this one just got missed. Fix:
```rust
find(&nodes, "HBoxContainer_TopButtons", "MarginContainer/VBoxContainer/header"),
```

## Notes on Lean's earlier feedback (thread, 18:09 UTC)
Looks addressed in this diff: `scene_manager.rs` now tracks `interactable_area` and a separate `safe_area`, with `explorer.gd::_push_scene_interactable_area()` calling both `Global.scene_runner.set_interactable_area(...)` and the new `set_safe_area(...)`, and `screen_inset_area` in `create_ui_canvas_information()` is built from `safe_area` (not `interactable_area`) — so `screen_inset_area` should no longer include the chat column, matching the protocol doc Lean linked. Worth a quick on-device sanity check against the two rects since I can't run the client here.

## Other findings
- No debug-print / mis-leveled Sentry logs, no `.claude/` files, no non-English strings, no `DclGlobal.is_ios/is_android/is_mobile` usage, no stray `TODO`/`FIXME`, no unguarded `.unwrap()`/`.expect()` in the diff — clean on all the usual red flags.
- `hud_margins.tres` single source of truth for margins + orientation-aware `portrait_left/right` is a nice simplification vs. per-widget margins.
- Comments in `explorer.gd` (e.g. the `_set_explorer_hud_elements_visible` invariant note about Hide-UI/chat/write-mode) are the good "why" kind REVIEW.md asks for.

## Test plan
Good shape overall (starts from opening the app, phone-first, portrait/landscape called out). One gap: the "Scene interactable area" dev overlay change is a new feature but I don't see an explicit QA case for toggling `Settings › Developer` → interactable-area overlay and checking the `%` readout — worth a one-line case per REVIEW.md Section 4 format.

**Recommendation: Requesting changes** — fix the stale `hud_input_lint.rs` path so CI goes green, and get an iOS build (`build` label) before this merges given the mobile-surface changes. Everything else here looks solid; happy to re-review once CI is green.

regenesis-claw (commented):
**Re-review** (delta: `0b041ec` → `658103a`, 1 commit)

| Previous finding | Status |
|---|---|
| 🔴 `hud_input_lint.rs::debug_panel_body_does_not_block_scene_ui` stale path | ✅ Fixed — node renamed `header` → `MarginContainer_Header` (follows PascalCase convention), lint test updated to `MarginContainer/VBoxContainer/MarginContainer_Header`. `mouse_filter = 2` still correctly set. |

Clean fix, exactly what was needed.

**Remaining before I can approve:**
1. **CI** — Linux build (which runs the lint test) is still in progress on this commit. Waiting for green.
2. **iOS build** — no `build` label has been added yet. Per REVIEW.md pre-flight, this PR touches mobile HUD/safe-area layout and needs a green iOS artifact before approval. A maintainer can add the `build` label to trigger one.

Once Linux passes and an iOS build goes green, ping me and I'll approve immediately — no other code concerns.

regenesis-claw (commented):
## Hostile re-review — concrete bugs only

Reviewed the full 4028-line diff (37 files), every GDScript handler, the Rust scene_manager changes, and the explorer.tscn node tree. Traced every state-machine path I could identify (hide-UI, write-mode, portrait/landscape, panel open/close, backpack, dismiss-catcher, orientation change). **CI is all green on `333e05e2` including Linux + iOS.**

### Bugs found: 0

I wanted to find something. I couldn't. Every path I traced resolves correctly.

### Items I tried to break (and failed)

| Hypothesis | Verdict |
|---|---|
| `_refresh_hud_dismiss` races with `close_chat` in the dismiss-catcher tap handler | **Not a bug.** `_close_all_panels()` calls `_refresh_hud_dismiss()` (dismiss stays STOP because chat is still visible), then `Global.close_chat.emit()` fires → `chat.exit_chat()` → `visibility_changed` signal → `_refresh_hud_dismiss()` fires again and correctly sets IGNORE. Trailing-edge signal resolves the state. |
| Navbar stays hidden after portrait→landscape rotation | **Not a bug.** `_on_size_changed()` sets `visible = window_size.x > window_size.y` and only early-returns when `_manually_hidden` is true (only set by `_enter_menu_screen`). Normal HUD state doesn't set it. |
| Chat permanently hidden after opening/closing friends panel | **Not a bug.** `_open_friends_panel()` hides `chat_panel`; `_close_all_panels()` → `chat_panel.show()` (guarded by `not _session_hide_main_hud`). |
| `_on_backpack_open` ignores the `on_emotes` flag (behavioral regression) | **Intentional.** Old code only acted on `on_emotes=true` (emote wheel). New code runs for all opens, treating Backpack as a fullscreen menu screen (same as Discover). PR description explicitly documents this. |
| Camera mode dropdown `index` vs enum value mismatch | **Happens to work.** Items added in enum order (FIRST_PERSON=0, THIRD_PERSON=1), so list index equals enum value. Fragile but not broken. |
| `InteractableAreaDebug` ColorRect eats clicks | **Not a bug.** `mouse_filter = 2` (IGNORE) in the tscn. |
| `HudDismissCatcher` blocks gameplay when nothing is open | **Not a bug.** Starts IGNORE in tscn; `_refresh_hud_dismiss()` only sets STOP when a panel or chat is visible. |
| `Notifications` lost `custom_minimum_size` override in `chat_panel.tscn` | **Not a bug.** Same value (328,0) is baked into `notifications.tscn` itself. The override was redundant. |
| `chat_notifications_width` read before set | **Not a bug.** Set in `chat_panel._ready()` from `notifications.custom_minimum_size.x`; `explorer.gd` reads it in `get_interactable_area()` which runs in `_process`, well after `_ready`. |
| `debug_panel` null access (was dynamically instantiated, now `@onready`) | **Not a bug.** Now a static child of `InteractableHUD` in explorer.tscn; always valid. Visibility toggled instead of instantiate/free. |
| `scene_stats_panel.toggle_host` null when `_create_toggle_button` runs | **Not a bug.** Set by `explorer.gd` before `add_child(scene_stats_panel)`, so it's populated before `_ready()` fires. Fallback to `get_parent()` exists anyway. |

### Unverified (can't run the client)

1. **Symmetric margins on one-sided-notch devices** — the `symmetric_horizontal` logic in `safe_margin_container.gd` maxes left/right and applies both. Correct in code, but I can't verify visual centering on a real device with an asymmetric notch (e.g. iPhone landscape left-notch). Need on-device QA.
2. **Portrait chat bottom compensation** — `_apply_default_margins` subtracts the SafeMarginContainer's bottom from `PORTRAIT_BOTTOM` to avoid double-counting. The math looks right but I can't verify the keyboard doesn't overlap the input on a real phone.
3. **Scene stats panel positioning** — `_reposition_below_toggle()` pins via `_toggle_btn.get_global_rect()` every frame. On orientation change the toggle moves, so the panel should follow, but I can't verify it doesn't flicker or land off-screen on small devices.
4. **Joystick travels the whole screen** — the tscn changes remove the camera button and its touch handling; `_active_area` is full-screen. PR claims this, but I can't verify touch targets don't conflict with the dismiss catcher or joypad on-device.

regenesis-claw (approved):
All previous findings resolved. CI fully green (Linux, Android, iOS). Hostile re-review found zero concrete bugs across 37 files — traced every state-machine path (hide-UI, write-mode, portrait/landscape, panel lifecycle, dismiss-catcher, backpack, orientation). 4 items flagged as unverified (need on-device QA). Approving.

leanmendoza (approved):

rodrigo685:
- [ ] Text Chat Focus Mode Overflows device 

image

- [x] Backpack loading state (spinner) sits below an empty state that it shouldn't be there


Now
Screenshot 2026-08-10 at 11 32 13 AM

https://dclregenesislabs.slack.com/files/U09DYGTRXDH/F0BP1AB8F0T/screenrecording_08-10-2026_11-24-18_1.mp4?origin_team=T091V1U5GP6

Previously
image

https://dclregenesislabs.slack.com/files/U09DYGTRXDH/F0BPBBDQSNQ/screenrecording_08-10-2026_11-36-34_1.mp4?origin_team=T091V1U5GP6


- [x] Emote Wheel needs to be auto-closed when Menu Opens to avoid overlap, and button hidden (Check with @ElzathRusso )

image

rodrigo685 (changes requested):
See requested changes above.

kuruk-mm (approved):
looks great!

ElzathRusso (approved):

regenesis-claw:
**Checked build 1506** (commit `e2c8300`, delta since my last approval on `333e05e2`).

This lands `feat(hud): mutually exclusive navbar/chat/emote overlays; joypad combo & hide-ui fixes` plus a couple of QA-driven fixes — all three items @rodrigo685 flagged in the PR comments look directly addressed:

| Rodrigo's finding | Fix |
|---|---|
| Text Chat Focus Mode overflows device | `chat_message.gd`: `autowrap_mode = TextServer.AUTOWRAP_WORD_SMART` — long space-less tokens (URLs, wallet addresses) now break instead of overflowing the panel. |
| Backpack loading spinner sits below an empty state | `backpack.gd`/`backpack_landscape.tscn`: `RightEditorContainer` is now explicitly hidden while loading and shown on data-ready, and the loading-spinner nodes were reordered in the tscn draw order. |
| Emote Wheel needs to auto-close when Menu opens, button hidden | `explorer.gd` rework: navbar, chat-focus, and the emote wheel are now mutually exclusive overlays via new `_hide_movement_controls()`/`_show_movement_controls()` + `_on_chat_focus_entered/_exited()` handlers. Opening the navbar now explicitly closes the emote wheel and hides its button; the emote wheel closes when a side panel/navbar is opened, hides the joypad/joystick while open, and restores them on close. `joypad.gd` also collapses its "+" combo overflow whenever the joypad is hidden by any flow (via `_notification(NOTIFICATION_VISIBILITY_CHANGED)`), so it can't silently reappear open next time.

I checked the new interlocking overlay logic for the usual class of bug here (state left stuck after an overlay closes mid-transition, or two "restore" paths fighting each other) — the code consistently follows a documented "hides run last so they win over the restores" pattern across all the new handlers, and I didn't find a path where movement controls or the chat get stuck hidden. `_close_all_panels()` restores things directly instead of reusing the new `_show_movement_controls()` helper (minor duplication, not a bug).

CI is fully green on this commit (Linux, Android, iOS, Clippy, static checks, Docker). The rest of the delta (billboard `targetEntity`, `PBPrimaryPointerInfo`, Sentry sample-rate flag) came in via merge-from-main, already reviewed/merged separately — not new work on this branch.

Still approved from my side. Given the scope of the overlay rework, worth a quick on-device pass specifically on: opening the navbar while the emote wheel is open (and vice versa), opening chat write-mode while the emote wheel is open, and the joypad "+" combo after a Hide-UI toggle.

Ludmilafantaniella (approved):
## ✅ Approve
Tested on Android and iPhone 16 Pro Max, portrait and landscape.

- [x] Camera Mode dropdown works correctly (First/Third person), greys out with correct message in camera-locked scenes (tested with Metadynelabs), hidden pre-world
- [x] Joystick/joypad positioning, edges, and rotation behavior all correct; scene buttons still work as expected
- [x] Hide UI / Show UI works correctly, closed panels stay closed
- [x] Chat and social panels align properly in both orientations, tap-to-dismiss works with no dead zones
- [x] No overlap with notch/Dynamic Island; HUD stays symmetric across rotation on iPhone 16 Pro Max
- [x] Backpack, emote wheel, and Developer interactable area overlay all working fine

https://github.com/user-attachments/assets/1f926584-88a7-440b-a3f1-cc64db8868b2

regenesis-claw (approved):
Re-approving after commit 45e77bd (build 1510).

**Fix verified**: emote wheel open/close now only hides/shows the `virtual_joystick` (movement stick) instead of calling `_hide_movement_controls()`/`_show_movement_controls()` — so the joypad action buttons (E/F/+/jump) stay visible while the emote wheel is open, which is exactly what was requested.

Also: `scene_stats_panel` now parents into `hbox_debug_tools` directly instead of floating on `ui_root`, keeping it properly docked with the Console/Reload buttons.

CI all green (Linux, Android, iOS, Clippy, static checks, Docker). No issues found.

rodrigo685 (approved):

PR #2647, opened 2026-08-06
fix: emote props missing on optimized (mobile) builds

### LINKED ISSUE #2617: [Bug] Some emotes with animated GLTFs don't show up in Mobile
### Platform
Mobile (iOS & Android)

### App Version
1.11-prod

### Device Information
Not specified

### Issue description
Some emotes that use animated `.glb`/`.gltf` prop assets don't show up / render on mobile. Confirmed this is *not* Pulse-related — Lean confirmed the same behavior happens in prod today, so it's a pre-existing issue unrelated to the Pulse migration. Likely something specific to how the animated asset is loaded/rendered rather than a networking/comms issue.

### Expected behavior
The animated GLTF prop attached to the emote should render normally on mobile, matching desktop/Unity behavior.

### Screenshots/Media

Image

Image

### Steps to reproduce
1. Equip/play an emote that includes an animated GLTF prop asset
2. Observe the prop/asset does not appear on mobile

### Occurrence
Not specified

---
Requested by Rodri via Slack

kuruk-mm:
Emote props (surfboard, TV, etc.) were invisible on mobile because some optimized `-mobile.zip` bakes on the CDN are defective in two different ways, and the client had no way to cope with either. The original "find the prop by name" fix wasn't enough — the prop was found but nothing ever made it visible (Case A), or the prop mesh wasn't in the zip at all (Case B). This PR makes the client handle both defective bake shapes, and "worked on macOS" turned out to be a stale local cache, not a platform difference (Case C). Verified live on Android (Motorola Edge 60 Pro) with a fresh install downloading the real CDN zips.

Closes #2617

> **QA note:** verified on 30+ emotes (props + base emotes). 4 emotes still play without their prop (Unmonday Dreams, Swing this way, Planting Seeds, Retro Gamer) — they hit a third bake shape not covered here and are tracked separately in #2755.

## The cases

**Case A — raw re-baked zips (e.g. Surf, `bafybeihx7km…-mobile.zip`).** The zip is a plain GLTF→scn conversion that skipped emote processing: no `EmoteAnimations` player, prop unrenamed/unflipped, and **no `visible` track**. Prop visibility is 100% animation-driven (the controller `hide()`s props at load; only a baked `Armature_Prop_{hash}:visible` track shows them), so the emote played with the prop attached but invisible. Fix: `extract_emote_from_scene` detects the raw shape and runs the same `process_emote_animations` the runtime pipeline uses (renames + flips the prop, remaps tracks, bakes visible/audio tracks). Animations are duplicated first — they're shared Resources across `PackedScene.instantiate()`, so processing them in place twice (second avatar loading the same emote) would corrupt them. Likely bake-side cause worth an infra follow-up: emote and scene/wearable bakes share the CDN key `{hash}-mobile.zip`, so a bake of the same GLB hash under another asset type clobbers the emote-shaped zip.

**Case B — prop dropped at bake time (e.g. Watching TV: prop armature authored as `Armature_prop`, lowercase p).** The pipeline's `find_child("Armature_Prop")` is case-sensitive, so the bake dropped the prop **geometry** from the zip while keeping the prop **animation** (animation-name matching lowercases). These props never worked anywhere in this client, desktop included. Fix: the pipeline now finds the prop case-insensitively (`find_prop_node`) and remaps its tracks by the authored name — fixing runtime processing and all future bakes. Since existing zips physically lack the mesh, `EmoteLoader` also self-heals: when a bake has a prop animation but no prop node, it re-processes the original GLB at runtime once and caches the result.

**Case C — stale caches.** Desktop "worked" only because a July processed-shape zip was still in `user://content/`, short-circuiting the CDN download of the broken re-bake. Also, runtime-processed emote scenes saved by the old case-sensitive pipeline are missing their props on existing installs — `LOCAL_ASSETS_CACHE_VERSION` bumped 4→5, so installs clear the content cache once and re-process.

## Changes

- `lib/src/content/content_provider.rs` — detect raw (unprocessed) bakes in `extract_emote_from_scene`; new `extract_raw_emote_from_root` duplicates the shared animations and runs `process_emote_animations`.
- `lib/src/content/gltf/emote.rs` — `find_prop_node()`: case-insensitive prop armature lookup; track remapping keyed off the prop's authored name instead of the `"Armature_Prop"` literal.
- `godot/src/decentraland_components/avatar/wearables/emote_loader.gd` — self-heal fallback (re-process the original GLB when an optimized bake has a prop clip but no prop node); extraction refactored into `_async_extract_from_path`.
- `godot/src/decentraland_components/avatar/avatar.gd`, `avatar_emote_controller.gd` — pass file name + content mapping through so the loader can self-heal.
- `godot/src/global.gd` — `LOCAL_ASSETS_CACHE_VERSION` 4→5.

## Test plan

- [ ] **Case A (raw zip):** On a fresh install, open the app, sign in, equip and play the Surf emote (`0x6cd6546c…:0:7`). Expected: the surfboard appears under the avatar, animates with the emote, and disappears when the emote ends.
- [ ] **Case B (dropped prop):** Play the Watching TV emote (`0x701d9cad…:0:387`). Expected: the prop appears and animates (first play may take a moment longer — one-time re-process of the original GLB); subsequent plays and app restarts show it immediately.
- [ ] **Regression (working prop emotes):** Play a prop emote that already worked (e.g. any emote whose prop showed before). Expected: prop still shows/hides correctly, no duplicated or leftover props after switching emotes.
- [ ] **Regression (prop-less emotes):** Play a base emote (wave/clap). Expected: plays normally, no errors in logs.
- [ ] **Remote avatars:** Have a second account play the Surf and TV emotes nearby. Expected: props visible on the other player too.
- [ ] **Cache bump:** Upgrading from a previous build clears the content cache once (first load re-downloads assets), then behaves normally.


Ludmilafantaniella (commented):
## 🐛 Issues found

 Tested the build from this PR and there are still emotes with missing props:
- [Unmonday Dreams #86](https://decentraland.org/marketplace/contracts/0x5ed2f9e19b473858f67911f4971fdfb0971c69d2/tokens/86)
- [Swing this way #78](https://decentraland.org/marketplace/contracts/0x447f3cc893d55897c791dd5bbe01bbbc7c89015e/tokens/78) 
- [Planting Seeds #163](https://decentraland.org/marketplace/contracts/0xca53b9436be1d663e050eb9ce523decbc656365c/tokens/105312291668557186697918027683670432318895095400549111254310977699)
- [Retro Gamer #14](https://decentraland.org/marketplace/contracts/0x0ae365f8acc27f2c95fc7d60cf49a74f3af21573/tokens/315936875005671560093754083051011296956685286201647333762932932622) — TV prop still not showing

cc @kuruk-mm

Image

## QA ✅ 

In addition to the props reported as missing above, I also tested 30+ other emotes (including base emotes with no props) to verify the fix didn't break anything existing. All of them worked correctly; props visible, animating properly, and hiding when the emote ends.

Leaving this as evidence that the fix covers the vast majority of cases, with the issue narrowed down to the 4 emotes reported above.

https://github.com/user-attachments/assets/3c3bd9d8-3c0b-4035-8fcc-7291bedf9882

https://github.com/user-attachments/assets/49b5706a-3c61-43b4-86cc-80a2afc1833e

https://github.com/user-attachments/assets/8de94a0e-5c4f-4f86-a377-ec03ebc41bdb

kuruk-mm:
Thanks @Ludmilafantaniella 🙏 — the 4 emotes still missing their prop (Unmonday Dreams, Swing this way, Planting Seeds, Retro Gamer) are now tracked in #2755. Since the fix covers the other 30+ emotes verified, merging this one and handling those 4 separately.

leanmendoza (approved):

PR #2650, opened 2026-08-06
fix(content): skip texture pipeline for impostor-only body snapshots

manuelmaceira:
## Problem

On mobile, avatar body snapshots go through the generic content pipeline, which ETC2-compresses the shared `Gd<Image>` in-place to build the entry's texture. The impostor capturer then reads `TextureEntry.image` and `set_impostor_texture` calls `resize`/`convert`/`generate_mipmaps` on it — all three reject compressed formats, spamming Sentry (~6.4k events/day: "Cannot resize in compressed image formats", "Cannot generate mipmaps from compressed image formats", "Image format must match texture's image format") and leaving impostor billboards untextured.

But the body snapshot's `texture` is **never rendered**: the impostor uploads the raw pixels into its own TextureArray. The compression was pure waste for this path — compress in the loader, decompress in the consumer.

## Fix

- `load_image_texture` gains a `compress` flag; `fetch_avatar_body_texture` and `fetch_default_avatar_body_texture` pass `false`
- `compress=false` keeps the image raw (RGBA8) and ships a 2x2 placeholder texture — the entry's texture is never rendered, so no real texture is built at all (saves VRAM + a compress)
- `set_impostor_texture` compresses the entry image to ETC2 **after** consuming it (mobile only), so session-cached promise entries reclaim the same RAM as before; a recapture enters through the `decompress()` guard and drops the inherited mip chain before re-uploading

Net: no compress→decompress cycle on first capture, same steady-state RAM, no error spam.

## Test plan

- [ ] Crowded spot (Genesis Plaza), walk away until avatars switch to impostor billboards → billboards show body textures, no compressed-image errors in logs
- [ ] Sentry: the three compressed-image errors stop for this build
- [ ] Regression: near↔far LOD transition still fades normally; impostor recapture after LRU eviction still textures correctly

### DIFF
--- a/lib/src/avatars/avatar_scene.rs
+++ b/lib/src/avatars/avatar_scene.rs
@@ -587,6 +587,28 @@ impl AvatarScene {
         };
 
         let mut img = image;
+        // The entry image outlives this call (it stays in the content
+        // provider's promise cache): on mobile we compress it to ETC2 at the
+        // end of this function to reclaim RAM, so a recapture re-reads it
+        // compressed. resize/convert/mipmaps all reject compressed formats —
+        // decompress first.
+        if img.is_compressed() {
+            let err = img.decompress();
+            if err != godot::global::Error::OK {
+                tracing::warn!(
+                    "set_impostor_texture: failed to decompress source image for slot {}: {:?}",
+                    impostor_id,
+                    err
+                );
+                return;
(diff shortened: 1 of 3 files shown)


kuruk-mm (approved):
Looks great!

Issue #616, opened 2025-08-12
Feature: Scene Dynamic Lights


kuruk-mm:
Protocol spec: https://github.com/decentraland/protocol/pull/296

Scene for testing:
[Lights.Testing.zip](https://github.com/user-attachments/files/21744797/Lights.Testing.zip)

Bevy Reference:

https://github.com/user-attachments/assets/7a1ce2d1-8c6d-4dcb-9530-a4e5eb13f23f

Unity Reference:

https://github.com/user-attachments/assets/f0572e45-19ce-4d58-a7d1-6c3f7bb2f8ce

PR #2653, opened 2026-08-06
fix(scene): VisibilityComponent affects only entity's own visuals, not child entities (#1888)

### LINKED ISSUE #1888: [Bug]: Scene objects invisible until interaction + selective GLTF animation failures
### Platform
Mobile (iOS and Android)

### App Version
Current release

### Device Information
Not specified

### Issue Description
Two related rendering/animation issues observed in specific scenes on mobile:

**Issue 1: bloomgarden.dcl.eth - Objects invisible until interaction**
- Some plants in the scene don't display/render initially
- Objects become visible only after user interacts with them
- This suggests a visibility or culling issue triggered by interaction events

**Issue 2: pepino.dcl.eth - Selective GLTF animation failures**
- Some seed animations play correctly (e.g., onion seeds ✓)
- Other seed animations don't play (e.g., potato seeds ✗)
- Inconsistent behavior across similar objects in the same scene
- All seeds likely use similar GLTF animation setups, so selective failure is unexpected

### Expected Behavior
- All scene objects should be visible from initial load (unless intentionally hidden by scene logic)
- All GLTF animations should play consistently if the objects are set up similarly

### Steps to Reproduce

**For bloomgarden.dcl.eth:**
1. Jump to bloomgarden.dcl.eth on mobile
2. Observe that some plants are invisible on initial load
3. Interact with the invisible plant locations
4. Notice plants become visible after interaction

**For pepino.dcl.eth:**
1. Jump to pepino.dcl.eth on mobile
2. Observe seed animations
3. Notice onion seeds animate correctly
4. Notice potato seeds (and others) don't animate
5. All seeds should have similar animation setups

### Occurrence
Consistent in these specific scenes

### Additional Context

**Potentially related issues:**
- #1755 - Objects with low scale values culled/invisible (might explain bloomgarden if plants use small scale)
- #1793 - Animation sync issues between clients
- #1784 - Animation playback issues in Try On feature
- #1189 - Visibility range culling implementation

**Hypothesis:**
- bloomgarden issue could be initial visibility/culling bug that's reset on interaction
- pepino issue suggests selective GLTF animation loading/playback failure - possibly related to animation clip names, file structure, or model-specific attributes

**Impact:**
- Breaks immersion in farming/gardening scenes
- Inconsistent behavior confuses content creators
- May indicate broader GLTF animation or visibility system issues

---
Requested by A Daniel Garcia Aranda via Slack

manuelmaceira:
Closes #1888 

## Problem

Issue #1888 (part 1): scene objects invisible until interaction (bloomgarden.dcl.eth — plants invisible until watered).

**Root cause:** SDK7 semantics say `VisibilityComponent` affects ONLY the entity's own content, not child entities. Our `update_visibility` called `hide()`/`show()` on the entity `Node3D`, hiding the **entire subtree** — including nested child entities that have their own `VisibilityComponent`.

In bloomgarden, plant slot entities are authored `visible=false` while their rose **child entities** are `visible=true`. The reference client shows the roses (per-entity visibility); we hid them with the parent. Watering flips the slot's visibility, which is why plants "appeared" on interaction.

Diagnosis confirmed live via Scene Inspector bridge: rose meshes reported `is_visible_in_tree=false` with a hidden entity ancestor, while their own visibility and animation state were healthy.

## Fix

- Tag each entity's own visual children (`GltfContainer`, `MeshRenderer`, `TextShape`, `VideoPlayer`) at add time (`add_own_visual_child`).
- `update_visibility` toggles only those tagged children; child entities (`DclNodeEntity3d`) keep their own visibility.

Verified in bloomgarden: wilted roses and water drops now render from spawn (guest profile), matching the reference client.

## Also in this PR

- `node_query` command for the Scene Inspector bridge: live node probing over WebSocket (find nodes by name, dump arbitrary props, call methods). Used to diagnose this issue; generally useful for scene debugging on device.

## Not covered (can't reproduce)

Issue #1888 part 2 (pepino.dcl.eth selective GLTF animation failures): seeds animate the same as the reference client in the current build. Left for a follow-up if it resurfaces with a clearer repro.

## Tests

Pre-commit full suite PASS (fmt, gdlint, clippy, Rust unit tests, integration tests).

## QA Steps

1. Launch the explorer (any platform) and go to **bloomgarden.dcl.eth** (e.g. `--realm bloomgarden.dcl.eth` or jump from the world menu).
2. Walk to the garden beds.
3. **Verify wilted (dark/drooping) roses are visible in the beds right after spawn, without watering or interacting with anything.**
4. Verify water drops floating above the garden are visible.
5. Water a plant — it should transition/open as before (no regression on interaction).
6. Compare against the reference client: the garden should look the same.


Ludmilafantaniella (commented):
🟡 🐛 With Issues

**Android - iOS (iPhone 16 Pro Max:** Roses and water drops render correctly on spawn (main fix confirmed working). However, I couldn't find any way to actually water the plants on mobile - neither tapping nor the E/F keys trigger the watering action (works fine with a click on Explorer). This blocks fully verifying step 5 of the QA steps (watering transitions the plant) on mobile.

Flagging both for visibility.

https://github.com/user-attachments/assets/4d8cac49-68d9-49de-95a8-8a2d32dc64c1

@manuelmaceira

manuelmaceira:
@Ludmilafantaniella The way to water the plants is pressing the click button, not E/F.
Works correctly on my side 
Screenshot 2026-08-17 at 11 33 56 AM

manuelmaceira:
> @Ludmilafantaniella The way to water the plants is pressing the click button, not E/F. Works correctly on my side Screenshot 2026-08-17 at 11 33 56 AM

still reviewing on android/ios

manuelmaceira:
> > @Ludmilafantaniella The way to water the plants is pressing the click button, not E/F. Works correctly on my side Screenshot 2026-08-17 at 11 33 56 AM
> 
> still reviewing on android/ios

Already tested. Working correctly on both android and ios, could you re-test @Ludmilafantaniella ?

lib/src/scene_runner/godot_dcl_scene.rs
```diff
@@ -388,3 +388,36 @@ impl GodotDclScene {
...
+    }
+}
+
+/// Add a direct child that is part of the entity's own visuals, tagging it so
+/// `set_own_visuals_visible` can toggle it and applying the entity's current
+/// own-visibility. Nested entity nodes (DclNodeEntity3d) must NOT be added through
+/// here — they manage their own visibility.
+pub fn add_own_visual_child(node_3d: &mut Gd<Node3D>, child: &Gd<Node>) {
+    let visible = node_3d
+        .get_meta("dcl_own_visible")
```
regenesis-claw: [P1] `get_meta` with no default pushes a Godot error when the entry is missing, and `dcl_own_visible` is only ever set by `set_own_visuals_visible` — so every entity that never got a VisibilityComponent (the vast majority) logs an error here for each visual child it adds. That lands in the Godot error stream → Sentry, on the scene-load hot path.

```suggestion
    let visible = node_3d
        .get_meta_ex("dcl_own_visible")
        .default(&true.to_variant())
        .done()
        .try_to::<bool>()
        .unwrap_or(true);
```

(a `has_meta` guard is equally fine and matches `raycast.rs` / `trigger_area.rs` / `scene_manager.rs`.)

godot/src/tool/scene_inspector_bridge.gd
```diff
@@ -138,6 +138,29 @@ func _on_command(cmd: String, args: Dictionary, request_id: String) -> void:
...
+				func(n): return qname.to_lower() in n.name.to_lower()
+			)
+			for n in found.slice(0, qmax):
+				var entry := {
+					"name": str(n.name), "path": str(n.get_path()), "class": n.get_class()
+				}
+				for p in qprops:
+					entry[p] = var_to_str(n.get(p))
+				for c in args.get("calls", []):
+					if n.has_method(c[0]):
```
regenesis-claw: [P2] `c[0]` / `c.slice(1)` assume every entry in `calls` is an Array with at least one element. A malformed payload (a bare string, or `[]`) raises a GDScript runtime error inside `_on_command`, which aborts before `_send_ack` — the caller then hangs waiting for an ack it never gets. Worth a cheap guard:

```suggestion
					if c is Array and not c.is_empty() and n.has_method(c[0]):
```

regenesis-claw (changes requested):
Solid diagnosis and the right fix direction — per-entity visibility semantics (own visuals only, child entities untouched) match SDK7 and the reference client, and the tag-at-add-time + apply-current-state pattern handles the async GLTF/mesh arrival ordering correctly. Reviewed against `REVIEW.md`. Two things block merge.

## Findings

| # | Sev | Where | Issue |
|---|-----|-------|-------|
| 1 | P1 | `nft_shape.rs:57`, `avatar_shape.rs:187`, `particle_system.rs:203`, `light_source.rs:265` | Own-visual children not tagged → VisibilityComponent no longer hides them |
| 2 | P1 | `godot_dcl_scene.rs:414` | Unguarded `get_meta` → Godot error per visual child → Sentry quota burst |
| 3 | P2 | `scene_inspector_bridge.gd:159` | `node_query` assumes each `calls` entry is a non-empty Array |

### [P1] The own-visual tag list is incomplete — this trades one visibility bug for another

Before this PR, `node_3d.hide()` hid the entire subtree, so *every* visual attached to the entity went dark. Now only children tagged via `add_own_visual_child` do. Four components still add their visuals with a plain `add_child` and are therefore no longer affected by `VisibilityComponent` at all:

- `nft_shape.rs:57` — `NftShape` is an explicitly documented VisibilityComponent target in SDK7; this is a straight regression.
- `avatar_shape.rs:187` — `AvatarShape` likewise.
- `particle_system.rs:203` (#1538) and `light_source.rs:265` (#616) — neither has its own visibility handling; both currently rely on the parent hide.

I checked: none of those four files contain any `visib` handling of their own, so `update_visibility` was their only path. Please route them through `add_own_visual_child` too (or, for particles/lights, state explicitly why they're intentionally exempt — a light with `visible=false` on its entity emitting light is a behavior change worth calling out either way).

### [P1] `get_meta` without a guard fires a Godot error for (almost) every entity

See inline comment. Godot's `Object.get_meta(name)` with no default pushes an error when the entry is missing, and `dcl_own_visible` only exists on entities that actually received a VisibilityComponent — i.e. a small minority. Every GLTF/mesh/text/video child added on all other entities will emit `The object does not have any 'metadata' entry with name "dcl_own_visible"` on the scene-load path. That routes to the Godot error stream and into Sentry (REVIEW.md §5, log discipline) — a per-entity error during scene load is exactly the quota-burst class the `NOISE_PATTERNS` denylist exists for. The repo convention is a `has_meta` guard everywhere else (`raycast.rs`, `trigger_area.rs`, `scene_manager.rs`).

### Verified / not blocking

- Colliders (`mesh_collider.rs`) are deliberately untagged — correct, hiding never disabled collision and shouldn't now.
- Branch is up to date with `main` (0 behind), CI green (Static checks, Clippy, Linux, Android, Docker, iOS export). iOS artifact present, so the REVIEW.md pre-flight iOS gate is satisfied even though this diff is platform-agnostic Rust.
- No security issues found. `node_query` adds arbitrary property reads / method calls over the inspector WebSocket, but the bridge already exposes an `eval` verb through the same channel, so it is not a new capability class.
- Unverified: the "verified in bloomgarden, matches the reference client" claim — I can't run the client. QA (@Ludmilafantaniella) confirmed the spawn-visibility fix on Android/iOS; step 5 (watering) is still unconfirmed on mobile per her comment.

Requesting changes on 1 and 2.

lib/src/scene_runner/components/light_source.rs
```diff
@@ -262,7 +262,10 @@ pub fn update_light_source(
                             // Important:
                             // Add the node to the tree first so Godot runs _ready()
                             // before setting texture or debug display data.
-                            node_3d.add_child(&new_light_node.clone().upcast::<Node>());
+                            add_own_visual_child(
```
regenesis-claw, line 265: [P1] This node's `visible` already has an owner: `light_source_component.gd:397` (`set_light_enabled`) sets `visible = enabled and dcl_lights_system_enabled` on the root, and `create_or_update_light` calls it on every LightSource update — including the one right after this `add_own_visual_child`. So the own-visibility applied here is overwritten immediately, and a `VisibilityComponent{visible:false}` light stays lit; conversely `set_own_visuals_visible(true)` can re-enable a light the scene turned off with `active:false`.

Suggest dropping the `visible = …` assignment from `set_light_enabled` so the root's visibility belongs to VisibilityComponent alone — `_update_light_rendering_by_range()` already gates on `runtime_light_enabled` and fades `current_light.visible` off via `_apply_fade()`, and `set_light_enabled` already calls `request_recompute()`.

regenesis-claw (changes requested):
Delta re-review of `a22a100` (vs `7f4e66e`) — all three previous findings are addressed.

|---|------------------|--------|
| 1 | [P1] `nft_shape` / `avatar_shape` / `particle_system` / `light_source` not tagged as own visuals | ✅ all four routed through `add_own_visual_child` |
| 2 | [P1] Unguarded `get_meta("dcl_own_visible")` → Godot error per visual child | ✅ `has_meta` guard added, matches repo convention |
| 3 | [P2] `node_query` assumed each `calls` entry is a non-empty Array | ✅ `c is Array and not c.is_empty()` guard |

I checked the four newly-tagged nodes for an existing owner of their `visible` property. Three are clean: `avatar.gd` only hides its own children (nickname / wearable nodes), never the `DclAvatar` root; `GpuParticles3D` and `NFTShape` nodes are never toggled outside this path. `LightSource` is not.

### [P1 — new] `LightSource` root node's `visible` is now double-owned

`light_source_component.gd:397` (`set_light_enabled`) does:

```gdscript
func set_light_enabled(enabled: bool) -> void:
	runtime_light_enabled = enabled
	visible = enabled and dcl_lights_system_enabled
```

That writes the **root** node's `visible` — the exact node `add_own_visual_child` now tags and `set_own_visuals_visible` now drives. `create_or_update_light` calls `set_light_enabled(active)` on every LightSource CRDT update, including the one immediately after the node is created (`light_source.rs:265` → `create_or_update_light` → `set_light_enabled` at `light_source.rs:59`). Net effect, both directions:

- Entity with `VisibilityComponent{visible:false}` + `LightSource`: `add_own_visual_child` applies `visible=false`, then `set_light_enabled(true)` overwrites it back to `true` in the same update. The light stays lit — and any later LightSource property change re-clobbers it. Before this PR the parent `hide()` won, so this is still a regression on the pre-PR behavior, just narrowed to lights.
- Reverse: a scene that sets `LightSource.active=false` (root `visible=false`), then flips `VisibilityComponent` to `true`, gets the deactivated light switched back on by `set_own_visuals_visible`.

Minimal fix: drop the `visible = …` line from `set_light_enabled` and let the root's `visible` be owned solely by VisibilityComponent. The off-state is already fully handled downstream — `_update_light_rendering_by_range()` gates on `runtime_light_enabled` and calls `_set_light_render_state(false)`, which fades `_energy_scale` to 0 and `_apply_fade()` sets `current_light.visible = false` (`light_source_component.gd:484`). `set_light_enabled` already calls `request_recompute()`, so the recompute path covers it.

(If you'd rather not touch the lights component in this PR, the alternative is leaving `light_source.rs` on plain `add_child` and filing a follow-up — but then `VisibilityComponent` silently doesn't apply to lights, which is worth stating in the PR body.)

### CI

Static checks and Clippy green on `a22a100`; Linux / Android / iOS still running at review time.

Requesting changes on the one new item only — everything else is resolved and I'm happy with the rest of the change.

Ludmilafantaniella:
The option to water it isn't showing up on either iOS or Android.
Doesn't show the widget to tapping

https://github.com/user-attachments/assets/b86834bb-0e53-45c7-865a-562353d586ac

manuelmaceira:
Solved, it was a problem with the widgets hiding by distance to camera (third person had the camera too far away) 
Screenshot 2026-08-26 at 10 02 40 AM

lib/src/scene_runner/components/pointer_events.rs
```diff
@@ -182,12 +182,21 @@ impl PbPointerEvents {
...
+// third person even with the crosshair on the entity. Use whichever of
+// camera/player is closer: first-person is unchanged (they coincide), third
+// person becomes playable.
 pub fn event_info_in_range(
     max_distance: Option<f32>,
     max_player_distance: Option<f32>,
     camera_distance: f32,
     player_distance: f32,
 ) -> bool {
+    let camera_distance = camera_distance.min(player_distance);
```
regenesis-claw, line 199: [P2] This mixes two reference frames. `camera_distance` is `raycast_hit.length` (camera → hit point), but `player_distance` is `entity_player_distance` = distance to the entity's `base_3d` **origin**, not the hit point. For an entity whose pivot sits far from its geometry (scene-scale GLTF terrain/building with the origin near the parcel origin), the pivot can be metres closer to the player than anything aimable, so a `maxDistance: 3` entry becomes effectively unbounded.

Suggest keeping the `max_distance` leg hit-based by passing `player_position.distance_to(raycast_hit.position)` for this comparison, and leaving `entity_player_distance` (pivot) to the `max_player_distance` leg where it's specced. That preserves the third-person fix without loosening large-pivot entities.

Also worth noting: `entity_player_distance` returns `f32::MAX` when the entity is gone, which `min()` correctly turns into a no-op here — good failure mode, just make sure any refactor keeps it.

While you're here, the bullet list above (lines 180-185) still describes the default case as "camera distance only"; it's now the closer of camera/player.

regenesis-claw (approved):
Delta re-review of `8c2519d` (vs `a22a100`). Two PR-authored commits since the last pass (`64a8755`, `8c2519d`) plus a merge from main.

|---|------------------|--------|
| 1 | [P1] nft_shape / avatar_shape / particle_system / light_source not tagged as own visuals | ✅ fixed in `a22a100` |
| 2 | [P1] Unguarded `get_meta("dcl_own_visible")` | ✅ `has_meta` guard, still present at head |
| 3 | [P2] `node_query` assumed each `calls` entry is a non-empty Array | ✅ guarded |
| 4 | [P1] `LightSource` root `visible` double-owned by `set_light_enabled` | ✅ fixed in `64a8755` |

**Finding 4 verified end-to-end, not just taken on trust.** With the `visible = enabled and dcl_lights_system_enabled` line gone, I traced every off-path:
- `set_light_enabled(false)` → `request_recompute()` → manager tick → `_update_global_light_budget()` → `_update_light_rendering_by_range()` → `_set_light_render_state(false)` → fade → `_apply_fade()` sets `current_light.visible = false`. The `use_global_light_budget == false` branch also calls `_update_light_rendering_by_range()` on every registered light, so it is covered too.
- Global disable (`apply_light_settings` / `apply_graphic_profile_settings`) goes through the same `request_recompute()` path and `_update_light_rendering_by_range()` short-circuits on `not dcl_lights_system_enabled`.
- Cold start is safe: `_energy_scale` initialises to `0.0` and both `set_spot`/`set_point` call `_apply_fade()` immediately after `add_child(current_light)`, so a light created while the system is disabled is born with `current_light.visible == false` — it can no longer rely on the removed root-`visible` write.

The only behavioural side effect is that `set_light_enabled(false)` is now deferred by one manager tick + `FADE_DURATION` instead of hard-cutting the root. That reads as an improvement (matches the fade comment at the top of the file), not a regression.

### The new commit `8c2519d` (pointer-events range) — approving, with three P2 notes

The change itself is correct for the reported symptom and the added unit test covers the four branches. It is also a semantics change to a protocol-specified field, so it deserves the notes below rather than a silent pass.

**[P2] `max_distance` is now satisfiable by a pivot distance, while the camera leg is a hit-point distance.** `camera_distance` is `raycast_hit.length` (camera → actual hit point); `player_distance` comes from `entity_player_distance`, which is `player_position.distance_to(node_3d.get_global_position())` — the entity's *origin*, not the hit point. Taking `min()` of the two mixes reference frames. For a large GLTF whose pivot sits far from its visible geometry (a scene-scale terrain/building mesh with pointer events and its origin near the parcel origin), the pivot can be metres closer to the player than anything you can actually aim at, so a `maxDistance: 3` entry becomes effectively unbounded. Using the player → *hit point* distance for the `max_distance` leg keeps it hit-based and preserves the third-person fix:

```rust
// at the call sites: player_position.distance_to(raycast_hit.position)
```

(`max_player_distance` should keep using the pivot distance — that one is specced against the entity, and `entity_player_distance`'s `f32::MAX` fail-closed default is correct there. Note the `min()` also makes `f32::MAX` a no-op for the camera leg, which is the right failure mode.)

**[P2] Divergence from the reference client is now real and undocumented.** unity-explorer's `InteractionInputUtils` gates strictly on `raycastResultForSceneEntities.GetDistance() <= info.MaxDistance` — camera-origin distance, no player fallback — and its MCP docs state the same rule ("the entity's `maxDistance` applies ... get close first"). After this commit a mobile/third-person player can trigger an interaction that Desktop refuses at the same position. That may well be the right call for mobile (third person is the norm there), but it's a cross-client behaviour difference a scene author can hit; worth a line in the PR body and ideally a heads-up to the protocol/Unity side so the two clients converge deliberately rather than by accident.

**[P2] Scope.** This is an interaction-range semantics change riding in a VisibilityComponent PR. It's related in origin (same bloomgarden repro), but it's independently revertable and independently risky — a separate PR/issue would make it easier to bisect if a scene reports "clicks register from too far away" later. Also, the bullet list at lines 180-185 still says "camera distance only" for the default case; it's now the closer of camera/player and should be folded into the new comment.

### CI

All five checks green at `8c2519d` (Static checks, Clippy, Linux, Android, Docker). The new `#[cfg(test)]` module runs in the Linux `cargo test` job. No iOS build on this PR — no `build`/`build-ios` label and no artifact comment; the diff is Rust + GDScript with no iOS-specific surface, so I'm not treating that as blocking, but flagging it per REVIEW.md pre-flight.

### Security

No security issues found. No new external input, no auth surface, no secrets; the pointer-range change only widens a client-side interaction gate, and scene-authored `max_distance` remains an upper bound in both legs.

Approving — all four prior findings resolved and verified, remaining items are P2. Note that a bot approval does not satisfy godot-explorer branch protection; a human codeowner is still needed to unblock merge.

Ludmilafantaniella (approved):
## ✅ QA Approved

Platforms tested: iOS & Android
Scenes covered:
- bloomgarden.dcl.eth — Wilted roses and water drops render correctly from spawn, without needing to interact first (main fix from this PR confirmed). Watering a plant works correctly on mobile in third person.
- cozy farm - No visibility regressions found; objects render as expected from spawn.
- Genesis Plaza (regression) - Verified scene emotes and backpack (avatar) emotes continue to work correctly, no regressions introduced by this change.

Result: No blockers found. Approving from QA side.

https://github.com/user-attachments/assets/b441808e-3185-4f7a-98d3-b599c871cc3f

cc @manuelmaceira

kuruk-mm (approved):
great work!

PR #2660, opened 2026-08-07
fix(content): send User-Agent on HTTP clients to stop CDN 403s (#1766)

### LINKED ISSUE #1766: [Bug]: Arweave-hosted images not loading on mobile (remain white)
### Platform
Android, iOS

### App Version
0.55

### Device Information
Not specified

### Issue Description
Images hosted on Arweave at the Riviera second floor (-150,-54) do not load on mobile — they remain white/blank. The same images load correctly on the Unity client.

### Expected Behavior
Arweave-hosted images should load and display correctly on mobile, matching Unity client behavior.

### Steps to Reproduce
1. Visit Riviera second floor at coordinates -150,-54 on mobile (Android or iOS v0.55)
2. Observe the image displays on the second floor
3. Note that the images remain white/blank and do not load
4. Compare with Unity client where images load correctly

### Occurrence
100% (confirmed on both Android and iOS)

---

Reported by Roustan via Discord, confirmed by ToxicWaifu

Mobile deeplink: https://mobile.dclexplorer.com/open?position=-150,-54

manuelmaceira:
Closes #1766 

## Problem

Arweave-hosted images (Riviera 2nd floor `-150,-54`, casa Roustan `37,-114`) never loaded — frames stayed gray/blank. Reported on mobile (#1766) but reproduced on desktop too: **not platform-specific**.

## Root cause

`arweave.net` sits behind CDN77 bot protection that rate-limits clients sending **no `User-Agent` header**: the first request passes, everything after gets **403 Forbidden** for minutes (each attempt resets the window). `reqwest` sends no UA by default, so our downloads were 403ed.

The error then died silently inside the texture promise, leaving the material permanently untextured — which is why this went unnoticed for months.

The Unity client has no Arweave-specific code; it passes simply because UnityWebRequest always sends a UA.

## Evidence (same IP, same moment)

| client | result |
|---|---|
| curl (even concurrent bursts) | 200 always |
| reqwest, no UA (rustls, native-tls, h1 — all) | first OK → 403 |
| reqwest + UA (ours / Chrome / Unity's) | **6/6 × 200, even in bursts** |

## Fix

- Set `User-Agent: DecentralandGodotExplorer UnityWebRequest/1.0 (UnityWebRequest)` on the `ResourceProvider` and `HttpQueueRequester` reqwest clients. Identifiable to our own catalyst/WAF/Segment traffic (Godot vs Unity) while keeping the Unity token for CDN allowlists tuned to it.
- Log a `tracing::warn!` when `fetch_texture_by_url` fails (previously invisible).
- Re-baselined the 8 avatar snapshot images: with a UA, the avatar renderer's optimized-bundle fetch now succeeds on the CI runner (it went UA-less before), so wearables render with baked textures instead of runtime-decoded ones — a small but deterministic pixel shift (byte-identical across 5 CI runs; main stayed at 100% in the same window). The old snapshots encoded the UA-less fallback; the new ones match what the client renders now. Compare step back to 100.00000%.

## QA steps

1. Open the client (any platform) and go to the Riviera 2nd floor: deeplink `decentraland://open?position=-150,-54` or `/goto -150,-54` in chat.
2. Verify the second-floor frames display their images (previously: gray/blank).
3. Repeat at casa Roustan `37,-114`.
4. Sanity check: other images/textures must keep loading normally (wearables, scenes, places).

## Test plan

- `test_arweave_download_with_user_agent` (network; `#[ignore]`d so default CI never hits arweave.net — manual repro via `cargo test --release -- --ignored`): fetches a real Riviera image through `fetch_resource_with_data`, asserts JPEG magic bytes.
- E2E desktop at `-150,-54` with cleared cache: all 6/6 Arweave JPEGs re-downloaded and rendered.
- Full pre-commit suite passes (fmt, gdlint, clippy, unit + integration tests).

## Follow-up (not in this PR)

- Report upstream: arweave.net's CDN77 rules penalize legitimate API clients that omit a UA.
- Other reqwest clients still send no UA (`decentraland_auth_server.rs`, `thirdweb_guest.rs`, xtask downloader) — they only hit first-party endpoints today, but the trap is identical.


lib/src/lib.rs
```diff
@@ -6,6 +6,11 @@
 
 use godot::prelude::*;
 
+/// User-Agent for all outgoing reqwest HTTP clients. reqwest sends none by
+/// default, and CDN bot protection (e.g. arweave.net's CDN77) rate-limits
+/// UA-less clients into 403s (#1766).
+pub const USER_AGENT: &str = concat!("DecentralandGodotExplorer/", env!("CARGO_PKG_VERSION"));
```
kuruk-mm: can we fake a chrome tab or use the same as unity-explorer is using?

manuelmaceira: Yes, we can, unity has `UnityWebRequest/1.0 (UnityWebRequest)` and chrome uses `Mozilla/5.0 ... Chrome/126.0.0.0 Safari/537.36` which one would be the correct one to use? using unity feels off to me but in the case in the future someone blocks unity's UA (low chance of happening) we would have the same behavior as the reference client.

.github/workflows/docker_build_test.yml
```diff
@@ -67,8 +67,13 @@ jobs:
...
         run: |
           mkdir -p avatars-output
-          podman run --rm -v \
-            $(pwd)/tests/avatars-test-input.json:/app/avatars.json \
+          # --only-no-optimized: the optimized-bundle fetch races local decoding
+          # and ETC2-baked textures render slightly differently, making the
+          # snapshot comparison flip between ~100% and ~88% depending on runner
+          # network timing. Force runtime decoding for deterministic output.
+          podman run --rm \
+            -e PRESET_ARGS="--avatar-renderer --avatars avatars.json --only-no-optimized" \
```
regenesis-claw: [P1] This drops the optimized-bundle/ETC2 path from the only avatar snapshot test, and it isn't mentioned in the PR description.

The timeline on this branch (Docker green on `main`, failed at 88.4% on `18f9076e` — the first commit carrying the Unity UA — then this commit) suggests the divergence is a consequence of the UA change rather than pre-existing runner flakiness. Could you confirm whether you saw the same flip on `main` without the UA?

If it is pre-existing: link the evidence here and file a follow-up issue for the flaky optimized snapshot. If it isn't: please re-baseline the snapshots (or add a second job that keeps the optimized path covered) instead of narrowing the test — mobile ships the optimized path, so losing that coverage is exactly backwards.

lib/src/lib.rs
```diff
@@ -6,6 +6,12 @@
 
 use godot::prelude::*;
 
+/// User-Agent for all outgoing reqwest HTTP clients. reqwest sends none by
+/// default, and CDN bot protection (e.g. arweave.net's CDN77) rate-limits
+/// UA-less clients into 403s (#1766). We reuse Unity Explorer's UA: if a CDN
+/// ever blocks it, both clients break together and the issue gets flagged fast.
+pub const USER_AGENT: &str = "UnityWebRequest/1.0 (UnityWebRequest)";
```
regenesis-claw: [P2] Following up on @kuruk-mm's question, since it's still open: this client is not only used for arweave. `HttpQueueRequester` also carries profile deploys, `signed_login`, the realm/scene-entity coordinator and the Segment analytics uploads — all first-party. Sending Unity's UA there means our own catalyst logs, WAF rules and Segment traffic can no longer distinguish Godot clients from Unity clients, which is a real observability cost for a header whose only job was to be non-empty.

A hybrid keeps both properties — identifiable to us, still carrying the Unity token for any CDN allowlist that has been tuned for it:

```suggestion
pub const USER_AGENT: &str = "DecentralandGodotExplorer UnityWebRequest/1.0 (UnityWebRequest)";
```

Happy either way, but this needs @kuruk-mm's explicit sign-off before merge rather than an unanswered question.

lib/src/content/resource_provider.rs
```diff
@@ -918,6 +923,43 @@ mod tests {
         let _ = tokio::fs::remove_dir_all(path).await;
     }
 
+    /// Arweave images must download through arweave.net's CDN bot protection
+    /// (which 403s clients without a User-Agent — issue #1766).
+    /// Network-dependent, unlike the localhost-server tests above.
+    #[tokio::test]
+    async fn test_arweave_download_with_user_agent() {
```
regenesis-claw, line 936: [P1] Linux CI runs `cargo test --release -- --skip auth`, so this test reaches out to `arweave.net` on every build. The test right below it carries the comment "Serve the image bytes from localhost instead of the public httpbin.org service" — the repo already moved away from network-dependent tests for this reason. A gateway outage, a rate-limit on the shared runner egress IPs, or this transaction going cold will turn unrelated PRs red, and what's being asserted is CDN77's bot policy, not our code.

Suggest `#[ignore]` with a doc line saying it's the manual repro for #1766 (run via `cargo test -- --ignored`), or gate it behind an env var.

lib/src/content/resource_provider.rs
```diff
@@ -918,6 +923,43 @@ mod tests {
...
+        let url = "https://2ijy24lhsmfwf3sjzeza7dz45xq5n27wtwgcs4xiruge4nxxayha.arweave.net/0hONcWeTC2LuSckyD4887eHW6_adjCly6I0MTjb3Bg4";
+        let file_hash = "arweave_test_tx".to_string();
+        let absolute_file_path = format!("{}/{}", path, file_hash);
+        let data = provider
+            .fetch_resource_with_data(url, &file_hash, &absolute_file_path)
+            .await
+            .expect("arweave fetch failed");
+
+        // JPEG magic bytes
+        assert_eq!(&data[..3], &[0xFF, 0xD8, 0xFF]);
```
regenesis-claw: [P2] `&data[..3]` panics with an index-out-of-bounds if the gateway ever returns a short body (an error page, a truncated response) — which masks the real failure behind a slice panic. `starts_with` fails cleanly:

```suggestion
        assert!(data.starts_with(&[0xFF, 0xD8, 0xFF]), "expected JPEG magic bytes");
```

lib/src/content/content_provider.rs
```diff
@@ -1610,8 +1615,12 @@ impl ContentProvider {
 
             loading_resources.fetch_add(1, Ordering::Relaxed);
 
+            let url_for_log = url.clone();
             let result =
                 load_image_texture(url, sent_file_hash, content_provider_context, true).await;
+            if let Err(error) = &result {
+                tracing::warn!("fetch_texture_by_url failed for {}: {}", url_for_log, error);
```
regenesis-claw, line 1622: [P2] Good addition — this is the log whose absence hid the bug. One calibration note per REVIEW.md §5: `tracing::warn!` routes to Godot's warning stream and is captured by Sentry, and this fires once per failed texture URL. A scene with a batch of dead image URLs (or a user on a flaky connection) can emit this dozens of times per load, and the texture path already has a fallback, so it's "expected-and-handled" by the guide's definition.

Not blocking, but consider `debug!` here (the `debug!` you added above is already the right level), or keep `warn!` and confirm it can't fire per-retry.

regenesis-claw (changes requested):
Solid root-cause work — the evidence table (curl vs reqwest-no-UA vs reqwest+UA) makes the diagnosis convincing, and adding the UA at the two client construction sites is the right minimal fix. The `warn!` on `fetch_texture_by_url` failure is a genuine improvement: this bug hid for months precisely because that error was swallowed.

**Pre-flight (REVIEW.md §0):** branch is up to date with `main` (0 behind, 11 ahead) ✅. All checks green on head `92375f9b`, including a real iOS export and Android APK ✅. No submodule drift ✅. Reviewed against `REVIEW.md`.

Two things hold approval.

---

### 🔴 P1 — the CI workflow change looks like it is hiding a rendering delta caused by this PR

`docker_build_test.yml` now forces `--only-no-optimized` on the avatar snapshot test. This is undocumented in the PR description and it permanently removes the optimized-bundle/ETC2 path — the path mobile actually ships — from the only avatar snapshot test we have.

What the history shows:
- On `main`, the last 10 orchestrator runs have a green Docker job with the optimized path enabled.
- On this branch, the Docker job failed on `18f9076e` (the first commit that contained the Unity UA) with `mati_avatar_test_face.png ... 88.44273% similar`.
- The next commit, `9adfcf79`, is the workflow change that makes the test skip the optimized path.

That sequence reads as: adding a User-Agent changed which content path wins (optimized bundle fetch now succeeds / wins the race where it previously didn't), and the resulting avatar render diverges ~12% from the snapshots. If that's what happened, it's a real behavior change of this PR — desktop and mobile users start receiving optimized assets on requests that previously failed — and it deserves to be in the description, not silenced in CI.

Please answer explicitly: was this snapshot flakiness reproducible on `main` before the UA change? If yes, link the evidence and file a follow-up issue for the flaky optimized-path snapshot (and mention the workflow change in the PR body). If no, the coverage loss isn't the right fix — re-baseline the snapshots for the optimized path, or gate the flag to a separate job so both paths stay covered. I flagged this rather than assuming the worst, but I can't approve while the only avatar snapshot test silently stops exercising the path that changed. (Unverified from here: whether the optimized-assets bucket was in fact 403ing UA-less clients in production too.)

### 🔴 P1 — `test_arweave_download_with_user_agent` puts a third-party CDN in the default CI test set

Linux CI runs `cargo test --release -- --skip auth`, so this test hits `arweave.net` on every build. The sibling test in the same module carries the comment *"Serve the image bytes from localhost instead of the public httpbin.org service"* — the repo deliberately moved away from exactly this pattern. An arweave gateway outage, a rate-limit on the shared GitHub runner egress IPs, or that one transaction going cold turns unrelated PRs red, and the assertion is really testing CDN77's bot policy rather than our code. Mark it `#[ignore]` (documented as a manual repro for #1766) or gate it behind an env var / feature. Detail on the assertion in the inline comment.

---

### 🟡 P2

1. **The description contradicts the code.** It says `User-Agent: DecentralandGodotExplorer/<version>`; the code sends `UnityWebRequest/1.0 (UnityWebRequest)`. Please update the body.
2. **The UA choice is still an open thread** — @kuruk-mm asked and the reply was a question, so nobody has signed off. See the inline note on the constant; it affects our own first-party observability, not just arweave.
3. **Fix is partial.** Other reqwest clients still send no UA: `decentraland_auth_server.rs` (3 builders), `thirdweb_guest.rs` (7), and the xtask downloader. They only hit first-party endpoints today, so no 403 now — but the trap is identical. A follow-up issue (or a tiny shared constructor) would close it.
4. **Test plan shape.** The QA steps are reproducible, which is more than most — but they don't follow the Case/Setup/Steps/Expected checklist from REVIEW.md §4, and the cleared-cache precondition (which matters: a cached texture will render fine either way) is only mentioned in the *Test plan* section, not in the QA steps QA actually runs.

### Security

No security issues found. No secrets, no user-controlled input reaching the new code paths, and the added logs print a URL only. Worth noting the UA is static, so it leaks no client version — which is also why it removes version segmentation on our side.

**Requesting changes** on the two P1s.

.github/workflows/docker_build_test.yml
manuelmaceira: Answer to the explicit question: no, this was NOT reproducible on main — it is a deterministic consequence of the UA change, not pre-existing flakiness or runner timing.

Evidence (all from CI logs):
- 4 branch runs spanning Aug 10–12 (d9c6afdd, 83a927a6, 0d951e99, 18f9076e, 2a22996e) produced **byte-identical** similarity values for all 8 images: 98.50346, 96.57324, 99.63964, 97.78515, 99.08166, 96.05916, 96.29728, 88.44273. A timing race would vary per run; this is a deterministic render difference.
- main in the same window (e.g. 00d61910, 0ef326e7, 54a9b342, 69a8567a): **100.00000%** on every image, every run.
- The only functional delta between those two sets is the User-Agent. reqwest 0.11.27 sends **no UA at all** by default (verified empirically against a local echo server), so pre-fix the avatar renderer's optimized-bundle HEAD/GET (`check_remote_file_exists` → `resource_provider.client`) went UA-less and deterministically fell back to runtime decoding — which is what the old snapshots encoded. With a UA the fetch succeeds, the render shifts slightly (baked textures), and CI reproduced that shift identically every time.

Resolution, per your instruction ("re-baseline the snapshots for the optimized path, or gate the flag to a separate job"): reverted the `--only-no-optimized` change (workflow back to main's exact state) and re-baselined all 8 snapshots from the CI-produced render of the fixed client. Verified locally: `compare-image-folders` returns 100.00000% for all 8. The default optimized-capable path — the one mobile ships — stays fully covered.

lib/src/content/resource_provider.rs
manuelmaceira: Done. Marked `#[ignore]` with a doc comment documenting it as the manual repro for #1766:

    cargo test --release -- --ignored test_arweave_download_with_user_agent

Default CI no longer touches arweave.net. Also applied the starts_with fix on the same test (see the other thread).

lib/src/content/resource_provider.rs
manuelmaceira: Done — switched to `data.starts_with(&[0xFF, 0xD8, 0xFF])` with a length message; fails cleanly on short bodies.

lib/src/content/content_provider.rs
manuelmaceira: Kept `warn!` deliberately — surfacing this error is the point of the PR (it was invisible for months). It fires once per failed texture URL, not per retry: `fetch_texture_by_url` → `load_image_texture` is a single attempt per URL with no internal retry loop. A scene with dozens of dead URLs is exactly the case worth seeing in Sentry. The `debug!` on the success path stays at debug as you suggested.

lib/src/lib.rs
manuelmaceira: Valid point and the hybrid keeps both properties. Left the string as Unity's for now since the thread with @kuruk-mm is still open — happy to switch to `DecentralandGodotExplorer UnityWebRequest/1.0 (UnityWebRequest)` the moment that's signed off. PR description updated to match the actual string sent (was already describing the old value).

.github/workflows/docker_build_test.yml
manuelmaceira: Update: per direction, removed the snapshot re-baseline as well. The branch now leaves the Docker workflow and avatar snapshots identical to `main` (no `--only-no-optimized`, no snapshot updates). If the next CI run still shows the same 88.44% delta, that confirms the render shift is deterministic and we will need either this flag or a snapshot re-baseline. Leaving it this way to let the CI run speak for itself.

Ludmilafantaniella (approved):
## ✅ Approve by QA

Tested on Android (Galaxy A35) and iPhone (16Pro).

- [x] Riviera 2nd floor (-150,-54) - frames display images correctly (previously gray/blank)
- [x] Casa Roustan (37,-114) - same fix confirmed, images load correctly
- [x] Sanity check: other textures/images still load normally (wearables, scenes, places) - no regressions

No blockers on mobile. Note: there's still a failing Linux build check on CI (unrelated to this QA pass) - worth confirming that's resolved before merge.

https://github.com/user-attachments/assets/88cc09fc-5819-4859-bdf1-78a7df572a90

regenesis-claw (changes requested):
Delta re-review at `d1ca0542` (previous review was at `92375f9b`).

| # | Finding | Status |
|---|---|---|
| P1 | CI workflow forced `--only-no-optimized`, masking a possible PR-caused render delta | ⚠️ Root cause confirmed, resolution still open |
| P1 | Live-network arweave test in the default `cargo test` set | ✅ Fixed — `#[ignore]` + doc comment naming the manual repro |
| P2 | `&data[..3]` panics on a short body | ✅ Fixed — `starts_with` with a length message |
| P2 | `warn!` volume / Sentry quota | ✅ Answered — single attempt per URL, no retry loop; keeping `warn!` is the right call given the bug it hid |
| P2 | Description contradicted the code | ✅ Fixed — body now states the actual UA |
| P2 | UA choice unresolved | 🟡 Still open — needs @kuruk-mm's call |
| P2 | Other reqwest clients still UA-less | ❌ Not addressed, no follow-up issue linked |
| — | Branch state | ❌ New — branch is 8 commits behind `main` and the diff is now ~35 files |

---

### ⚠️ P1 — the render shift is now confirmed, and it's bigger news than the PR title

Thank you for the evidence — 5 branch runs with byte-identical similarity values against 100.00000% on `main` in the same window settles it: this is a deterministic render change caused by the User-Agent, not runner flakiness. That's exactly what needed to be established, and it materially changes what this PR *is*.

The implication deserves to be spelled out in the description rather than living in a review thread: if UA-less requests were losing/failing the optimized-bundle fetch (`check_remote_file_exists` → `resource_provider.client`), then **every user was silently falling back to runtime decoding for optimized content, and this PR switches them onto the optimized/baked path**. That is a much larger behavioral change than "arweave images load" — it touches avatar/wearable texture quality and load performance for everyone, on the path mobile ships. Please add it to the PR body, and add a QA case for it: side-by-side avatar + wearable appearance on a real device before/after, since the changed snapshot is a *face* at 88.44%.

What's still open is the resolution. The branch currently has neither the flag nor a re-baseline, so the Docker avatar snapshot job — queued behind the in-progress Linux build, it hasn't run on this head yet — is expected to reproduce the same 88.44%. Letting CI speak is a reasonable step, but the PR can't merge in this state. My read: the re-baseline you had is the right end state (it keeps the optimized path covered, which forcing the flag does not), **provided a human looks at the new face snapshot and confirms the delta is the expected baked-texture difference and not a quality regression**. 88.44% is a large enough diff on a face that it deserves eyes, not just a green check. @kuruk-mm — this is your call, and it's the only thing blocking from my side.

### ❌ New — branch is behind `main` and the diff picked up unrelated commits

REVIEW.md §0.1: the branch is 8 commits behind `main`, and the rebase re-applied commits that are already merged there (LightSource #616, the avatar state-machine/glider fixes, the xtask download retries, the engine bump). The PR now shows ~35 changed files instead of 4. I verified the blobs for the duplicated files are byte-identical to `main`, so there's no functional divergence hiding in there — it's noise, not drift. Merging `main` into the branch should collapse the diff back to the 4 files this PR actually owns, and it makes the CI signal trustworthy again.

### 🟡 Remaining P2s

- The UA string is still Unity's, pending sign-off. @kuruk-mm, the tradeoff in one line: reusing Unity's UA means our own catalyst logs, WAF rules and Segment traffic can't tell Godot clients from Unity ones (`HttpQueueRequester` carries profile deploys, `signed_login`, the realm coordinator and analytics). `DecentralandGodotExplorer UnityWebRequest/1.0 (UnityWebRequest)` keeps the Unity token for any CDN tuned to it while staying identifiable.
- No follow-up issue yet for the UA-less clients in `decentraland_auth_server.rs`, `thirdweb_guest.rs` and the xtask downloader. Given we now know a missing UA can silently change which content path wins, that's worth a tracked issue rather than a mental note.

Everything I flagged as fixable got fixed properly, and the CI archaeology you did was the right answer to the question. **Requesting changes** only on the open snapshot decision and the branch update.

kuruk-mm (approved):
looks great

regenesis-claw (dismissed):
Delta re-review at `32becb42` (previous: `d1ca0542`).

| # | Finding | Status |
|---|---|---|
| ❌→✅ | Branch behind `main`, diff inflated to ~35 files | ✅ Fixed — merged `main`, 0 behind, diff is back to the 4 files this PR owns (+73/−1) |
| ⚠️ | Avatar snapshot render shift | ⚠️ Reproduced by CI a 4th time; **safe to re-baseline** — analysis below |
| — | Linux CI red on this head | ❌ Not caused by this PR — teardown segfault, details below |
| 🟡 | UA string (Unity's vs identifiable hybrid) | 🟡 Still waiting on @kuruk-mm |
| ❌ | Follow-up issue for the other UA-less reqwest clients | ❌ Still not filed |

---

### The snapshot question — I compared the images, and re-baselining is safe

The Docker job on `d1ca0542` (before the `main` merge) failed exactly as predicted: `Error: 7 files are too different!` with the same value set for the 4th time in a row (98.50346 / 96.57324 / 99.63964 / 97.78515 / 99.08166 / 96.05916 / 96.29728 / **88.44273**). Deterministic, not flaky — settled.

So I pulled the `docker-snapshots` artifact and diffed the renders against the committed baselines pixel by pixel:

| pair | mean abs channel diff | max | pixels differing >16/255 | pixels identical |
|---|---|---|---|---|
| `mati_avatar_test_face.png` (256×256) | 2.65 | 72 | 1 117 (1.70%) | 40.6% |
| `mati_avatar_test.png` (256×512) | 0.80 | 55 | 195 (0.15%) | 79.0% |

Visually the two renders are indistinguishable — same pose, same silhouette, same materials, same colours. The difference is low-amplitude noise spread across textured surfaces (skin, hood, hat), which is exactly the signature of lossy baked/compressed textures replacing runtime-decoded ones. No geometry, shading or colour regression. The reason `88.44273%` looks alarming is that `compare-image-folders` appears to count any non-identical pixel, so uniform compression dither tanks the score while the image is visually the same.

Conclusion: the re-baseline you originally had was correct, and the coverage-preserving option. Recommend restoring it (all 8 snapshots, from the CI-produced render), leaving the Docker workflow at `main`'s state. That was the human eyeball the last review asked for — @kuruk-mm, flagging it for you since it changes the committed baselines.

### Linux red on this head is not yours

`🐧 Linux` failed on the `test-avatar` step, but the test itself passed — the log shows `[test_avatar_state_machine_graph] PASS`, then the engine crashed during shutdown (`ERROR: Condition "!configured" is true.` at `core/string/string_name.cpp:264`, followed by `File must be opened before use`, exit `signal 11`). That's a teardown-time crash in the headless harness, with no path to the HTTP client changes in this diff, and the same suite passed on `d1ca0542`. A re-run should clear it; it also looks worth a separate issue for the flaky shutdown segfault in the avatar test runner. Docker was skipped as a downstream consequence, so the snapshot job has not re-run since the `main` merge.

### Still open

- The UA string is unchanged pending @kuruk-mm's call (Unity's UA vs `DecentralandGodotExplorer UnityWebRequest/1.0 (UnityWebRequest)`).
- No follow-up issue yet for the UA-less clients in `decentraland_auth_server.rs`, `thirdweb_guest.rs` and the xtask downloader.

The code in this PR is in good shape — the remaining work is the snapshot re-baseline and a CI re-run. **Requesting changes** only so the PR doesn't merge with a known-failing snapshot job.

.github/workflows/docker_build_test.yml
manuelmaceira: Snapshots re-added. The rerun on 32becb42 confirmed the diagnosis: Linux avatar regression passed (the segfault was flaky), while Docker failed with the exact same byte-identical values (88.44273) — so the render delta is deterministic and tied to the UA. Re-baselining stays. The new CI run on 91c3e5e715 should be green.

PR #2686, opened 2026-08-12
feat: Preview Mode HUD layout

### LINKED ISSUE #2679: Preview Mode: New HUD Layout Implementation
## Context

Implement the Design for the Preview Mode HUD specified below per issue #2668 

## Scope

Implement the approved preview-mode HUD layout: placement of Console/Debug, Scene Performance & Limits, and Reload controls/

## ⚠️ Dependency / Merge Order

Per Mateo's guidance in Slack, this work should NOT start until both of the following are merged to `main` to avoid conflicts:
1. PR #2642 — feat: HUD + Social Panel Redesign — Iteration #1 (Seba)
2. PR #2663 — feat: Implementing stats panel design (Eibriel)

## Design

[Figma](https://www.figma.com/design/Y5cdeJvgF5M5Ye1pdIea61/Scene-Performance-Plugin?node-id=81-2282&t=0z5S5XGxQPQ8ybFW-1)

- [ ] Hide the Discover and Chat buttons.
- [ ] Place the three buttons next to the Menu button.
- [ ] When one panel is opened, the other should close. Both panels should appear in the same position.
- [ ] Add color state, selected is red, default is white.
- [ ] Use an icon for the Reload button.

---
**Video

https://github.com/user-attachments/assets/d0fc5927-9ff1-46d3-bcaf-9ca5128f99be

---
Screenshots**
Image

Image

Image

## References
- #2668 — Mobile Creators Preview Tools HUD Layout Design (design spec this issue implements)
- #2351 — parent HUD + Social Panel Redesign
- #2638 — defines `InteractableArea` + `DeviceSafeInset`
- #2493 — Scene Performance UX-refactor (panel styling reference)
- #2451 — Debug/Reload buttons regression in preview (v1138) — do not repeat

---
Requested by Rodri via Slack

sdilauro:
Closes #2679 

## Summary

Implements the **Preview Mode HUD layout** (#2679). A new `preview_hud_panel` organism hosts the **Console**, **Scene Status** and **Reload** controls in a single top-left toolbar that replaces the chat while active — in the settings **Dev Tools → Scene Logs** entry, in a **preview realm**, or via a `scene-stats=true` **deep link**.

- **New `preview_hud_panel` organism** (static, hidden `InteractableHUD` child): a 3-button header (Console / Scene Status / Reload) over a body that shows the console or the scene-stats overlay — mutually exclusive, same position (opening one closes the other).
- **Behaves like the chat vs. the navbar**: the whole toolbar (header included) is hidden while a navbar side panel is open, and restored header-only (toggles off) when the navbar collapses. It shares the bottom-left slot with the chat and never coexists with it.
- **Reuse, not rewrite**: `debug_panel` (console) and `scene_stats_panel` keep their own logic; their old headers / floating toggles were removed. The static `DebugPanel` moved out of `explorer.tscn` into the new organism. The console is always present so it keeps capturing scene logs; the scene-stats overlay is created on demand.
- **Availability**: Console + Reload show whenever the toolbar is active; the **Scene Status** button + panel appear only in preview / deep-link (settings Scene Logs exposes Console + Reload only).
- **Button states**: Console/Scene Status show a border + icon in `#FF2D55` (red) when open and `#FCFCFC` (white) when closed; Reload is borderless with a white icon that turns `#DF9CFF` (lilac) while held.
- Zero-cost in a normal production run: no preview, no settings toggle, no deep link → toolbar stays hidden and the chat behaves normally.

## Out of scope but required to make the toolbar usable

The relocated Reload button (and the console) were effectively unusable without two changes outside the pure-HUD scope of #2679, so they ride along in this PR:

- **Reload in Worlds** — `lib/src/realm/scene_entity_coordinator.rs`: `reload_scene_data` relied on `update_position`, which never re-fetches fixed/global scenes (Worlds are not re-discovered by parcel position, unlike Genesis City). Reloading a World scene therefore erased it and left an empty parcel. Added a `reloadable_entity_bases` map so a fixed/global scene re-issues its entity fetch on reload; city-mode keeps the position re-discovery path.
- **Loading screen on reload** — `godot/src/logic/scene_fetcher.gd`: show the loading screen up front on reload so killing the scene doesn't flash an empty parcel before it re-loads (skipped for preview hot-reload, which stays seamless).

**Design note (for reviewers):** a `scene-stats=true` deep link now shows the toolbar and hides the chat in *any* realm (production included), because the toolbar is the shared container for the Scene Status button. Flagging in case we'd rather have that deep link expose *only* the Scene Status button without disturbing the chat.

## Test plan

Verified on desktop against the deployed World `sebadl.dcl.eth` in preview (`--fake-deeplink "decentraland:///?realm=sebadl.dcl.eth" --preview`). Formal QA / on-device (iOS/Android) pass still pending.

- [ ] **Settings → Dev Tools → Scene Logs** (non-preview realm): toggling ON hides the chat and shows the toolbar with **Console + Reload only** (no Scene Status); Console toggles the console panel; toggling OFF restores the chat.
- [ ] **Preview realm / `--preview`**: toolbar shows **Console + Scene Status + Reload**; Console and Scene Status are mutually exclusive in the same spot.
- [ ] **`scene-stats=true` deep link**: the Scene Status button + panel are available.
- [ ] **Navbar interaction**: opening a navbar side panel (friends/notifications/settings) hides the entire toolbar; collapsing the navbar restores it header-only with all toggles off.
- [ ] **Reload**: reloads the current/previewed scene in **Genesis City** and **Worlds** (the World scene comes back rather than an empty parcel); the loading screen appears immediately (no empty-parcel flash); the console shows the reloaded scene's `console.log` output.
- [ ] **Colors**: Console/Scene Status border + icon are white when closed and red when open; Reload is borderless with a white icon that turns lilac while held.
- [ ] **Zero-cost / no regression**: a normal run (no preview, no settings toggle, no deep link) does not show the toolbar and the chat, hide-UI, and navbar flows behave as before.
- [ ] `cargo run -- check-gdscript`, `gdformat`/`gdlint godot/`, and `cd lib && cargo fmt --all && cargo clippy -- -D warnings` all pass; `hud_input_lint` tests pass.

### DIFF
--- a/godot/src/logic/scene_fetcher.gd
+++ b/godot/src/logic/scene_fetcher.gd
@@ -1192,6 +1192,12 @@ func _on_try_spawn_scene(
 func reload_scene(scene_id: String) -> void:
 	var scene = loaded_scenes.get(scene_id)
 	if scene != null:
+		# Show the loading screen up front so killing the scene doesn't flash an empty parcel
+		# before it re-loads. Skipped for preview hot-reload, which is meant to be seamless.
+		if not _is_hot_reloading:
+			var explorer = Global.get_explorer()
+			if is_instance_valid(explorer):
+				explorer.loading_ui.enable_loading_screen("", "on_reload")
 		var scene_number_id: int = scene.scene_number_id
 		if scene_number_id != -1:
 			Global.scene_runner.kill_scene(scene_number_id)
--- a/godot/src/logic/scene_fetcher.gd
+++ b/godot/src/logic/scene_fetcher.gd
@@ -1194,6 +1194,10 @@ func reload_scene(scene_id: String) -> void:
 	if scene != null:
 		# Show the loading screen up front so killing the scene doesn't flash an empty parcel
 		# before it re-loads. Skipped for preview hot-reload, which is meant to be seamless.
+		# If the re-fetch fails it's still taken back down: a null definition completes the
+		# loading session (report_scene_fetched), and any deeper failure is caught by the
+		# loading screen's own watchdog (Timer_CheckProgressTimeout: inactivity / max-time
+		# -> timeout modal + hide), which starts when enable_loading_screen is shown.
 		if not _is_hot_reloading:
 			var explorer = Global.get_explorer()
 			if is_instance_valid(explorer):
--- a/godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.gd
+++ b/godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.gd
@@ -66,8 +66,6 @@ func _ready() -> void:
 ## realm or a `scene-stats=true` deep link offers it; the settings Scene Logs entry exposes
 ## Console + Reload only. The console/debug panel is never torn down — it stays static.
 func set_scene_status_available(available: bool) -> void:
-	if not is_node_ready():
-		await ready
 	button_scene_status.visible = available
 	if available:
 		if not is_instance_valid(scene_stats_panel):
@@ -86,8 +84,6 @@ func set_scene_status_available(available: bool) -> void:
 ## Point the scene-stats overlay at the scene being previewed (forwarded from explorer.gd
 ## on preview / scene changes). No-op until the overlay exists.
 func set_scene(scene_id: int) -> void:
-	if not is_node_ready():
-		await ready
 	if is_instance_valid(scene_stats_panel):
 		scene_stats_panel.set_scene(scene_id)
 
--- a/godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.gd
+++ b/godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.gd
@@ -1,0 +1,190 @@
+class_name PreviewHudPanel
+extends MarginContainer
+
+## Preview-mode HUD toolbar (issue #2679). Lives in the same bottom-left slot as the
+## ChatPanel and replaces it while active (settings Scene Logs, preview realm, or a
+## `scene-stats=true` deep link). Placed statically in explorer.tscn — its console
+## (debug_panel) is always present so it keeps capturing scene logs; only the scene-stats
+## overlay (and its header button) is created on demand, here inside the hub.
+##
+## Header buttons: Console (toggles the console), Scene Status (toggles the scene-stats
+## overlay; shown only in preview/deep-link), Reload (reloads the current scene). Console
+## and Scene Status share the same body position and are mutually exclusive — opening one
+## closes the other. Selected buttons tint red, default is white.
+
+const COLOR_ACTIVE: Color = Color("ff2d55")
+const COLOR_DEFAULT: Color = Color("fcfcfc")
+# Reload icon while the button is held down (it never toggles nor shows a border).
+const COLOR_RELOAD_PRESSED: Color = Color("df9cff")
+const BUTTON_STATES: Array[String] = [
+	"normal",
+	"pressed",
+	"pressed_mirrored",
+	"hover",
+	"hover_mirrored",
+	"hover_pressed",
+	"hover_pressed_mirrored",
+	"focus",
+]
+const SCENE_STATS_SCENE: PackedScene = preload(
+	"res://src/ui/components/organisms/scene_stats_panel/scene_stats_panel.tscn"
+)
+
+var scene_stats_panel: SceneStatsPanel = null
+# Each toggle button gets its own copy of the scene's border stylebox so its border color can
+# be set independently (the scene shares one stylebox across both) — recolored per open/close.
+var _console_box: StyleBoxFlat = null
+var _stats_box: StyleBoxFlat = null
+
+@onready var button_console: Button = %Button_Console
+@onready var button_scene_status: Button = %Button_SceneStatus
+@onready var button_reload: Button = %Button_Reload
+@onready var body: Control = %Body
+# Untyped: the DebugPanel script has no class_name, so its custom methods
+# (set_console_visible/reload_current_scene) dispatch cleanly.
+@onready var debug_panel = %DebugPanel
+
+
+func _ready() -> void:
+	button_console.toggle_mode = true
+	button_scene_status.toggle_mode = true
+	button_console.button_pressed = false
+	button_scene_status.button_pressed = false
+	button_console.pressed.connect(_on_console_pressed)
+	button_scene_status.pressed.connect(_on_scene_status_pressed)
+	button_reload.pressed.connect(_on_reload_pressed)
+	# Console starts collapsed; scene-stats is created on demand via set_scene_status_available.
+	debug_panel.hide()
+	button_scene_status.hide()
+	_console_box = _own_border_stylebox(button_console)
+	_stats_box = _own_border_stylebox(button_scene_status)
+	_setup_reload_style()
+	_refresh_toggle_colors()
+
+
+## Create or free the scene-stats overlay (and show/hide its header button). Only a preview
+## realm or a `scene-stats=true` deep link offers it; the settings Scene Logs entry exposes
+## Console + Reload only. The console/debug panel is never torn down — it stays static.
+func set_scene_status_available(available: bool) -> void:
+	if not is_node_ready():
+		await ready
+	button_scene_status.visible = available
+	if available:
+		if not is_instance_valid(scene_stats_panel):
+			scene_stats_panel = SCENE_STATS_SCENE.instantiate()
+			scene_stats_panel.hide()
+			body.add_child(scene_stats_panel)
+	else:
+		if button_scene_status.button_pressed:
+			button_scene_status.set_pressed_no_signal(false)
+		if is_instance_valid(scene_stats_panel):
+			scene_stats_panel.queue_free()
+			scene_stats_panel = null
+		_refresh_toggle_colors()
+
+
+## Point the scene-stats overlay at the scene being previewed (forwarded from explorer.gd
+## on preview / scene changes). No-op until the overlay exists.
+func set_scene(scene_id: int) -> void:
+	if not is_node_ready():
+		await ready
+	if is_instance_valid(scene_stats_panel):
+		scene_stats_panel.set_scene(scene_id)
+
+
+## Forward a scene console line to the embedded (always-present) debug panel.
+func on_console_add(scene_title: String, level: int, timestamp: float, text: String) -> void:
+	if not is_node_ready():
+		return
+	debug_panel.on_console_add(scene_title, level, timestamp, text)
+
+
+## Collapse back to header-only: both overlays closed, every toggle off. Called when the
+## toolbar is restored (e.g. after the navbar collapses), mirroring the chat returning to
+## its closed state.
+func reset() -> void:
+	if not is_node_ready():
+		return
+	button_console.set_pressed_no_signal(false)
+	button_scene_status.set_pressed_no_signal(false)
+	debug_panel.hide()
+	if is_instance_valid(scene_stats_panel):
+		scene_stats_panel.hide()
+	_refresh_toggle_colors()
+
+
+func _on_console_pressed() -> void:
+	Global.send_haptic_feedback()
+	if button_console.button_pressed:
+		button_scene_status.set_pressed_no_signal(false)
+		if is_instance_valid(scene_stats_panel):
+			scene_stats_panel.hide()
+		debug_panel.show()
+		debug_panel.set_console_visible(true)
+	else:
+		debug_panel.hide()
+	_refresh_toggle_colors()
+
+
+func _on_scene_status_pressed() -> void:
+	Global.send_haptic_feedback()
+	if button_scene_status.button_pressed:
+		button_console.set_pressed_no_signal(false)
+		debug_panel.hide()
+		if is_instance_valid(scene_stats_panel):
+			scene_stats_panel.show()
+	elif is_instance_valid(scene_stats_panel):
+		scene_stats_panel.hide()
+	_refresh_toggle_colors()
+
+
+func _on_reload_pressed() -> void:
+	Global.send_haptic_feedback()
+	debug_panel.reload_current_scene()
+
+
+## Give the button its own copy of the scene's border stylebox for every visual state, so its
+## border color can be set independently of the other toggle (they share one in the scene).
+## Preserves the scene's shape/width — we only recolor the border later.
+func _own_border_stylebox(button: Button) -> StyleBoxFlat:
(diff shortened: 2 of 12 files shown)


godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.gd
```diff
@@ -0,0 +1,190 @@
...
+	_console_box = _own_border_stylebox(button_console)
+	_stats_box = _own_border_stylebox(button_scene_status)
+	_setup_reload_style()
+	_refresh_toggle_colors()
+
+
+## Create or free the scene-stats overlay (and show/hide its header button). Only a preview
+## realm or a `scene-stats=true` deep link offers it; the settings Scene Logs entry exposes
+## Console + Reload only. The console/debug panel is never torn down — it stays static.
+func set_scene_status_available(available: bool) -> void:
```
regenesis-claw, line 68: [P0] `gdlint` fails here and on line 88, which turns `📊 Static checks` red and cascades into *every* build job being skipped:

```
preview_hud_panel.gd:68: Error: Async Function name "set_scene_status_available" is not valid (async-function-name)
preview_hud_panel.gd:88: Error: Async Function name "set_scene" is not valid (async-function-name)
```

The `await ready` makes both coroutines, and the repo's async naming rule wants an `async_` prefix (cf. `async_teleport_to`, `async_show_scene_timeout_modal`).

My suggestion is to drop the `await` rather than rename: this panel is an `@onready` child of `explorer.tscn`, so Godot readies children before the parent — by the time `explorer._ready()` → `_update_preview_hud()` runs, `is_node_ready()` is already true, and every other caller (`_update_preview_ui`, `_on_control_menu_request_debug_panel`) is post-`_ready` too. The guard is unreachable in practice, and removing it keeps both methods synchronous (no coroutine returned to the non-awaiting callers in `explorer.gd`) and fixes the lint without renaming call sites.

If you'd rather keep the defensive guard, rename to `async_set_scene_status_available` / `async_set_scene` and update the two call sites in `explorer.gd`.

godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.tscn
```diff
@@ -0,0 +1,143 @@
...
+script = ExtResource("1_script")
+metadata/mobile_preview_orientation = "landscape"
+
+[node name="VBoxContainer" type="VBoxContainer" parent="." unique_id=900190433]
+custom_minimum_size = Vector2(0, 450)
+layout_mode = 2
+mouse_filter = 2
+theme_override_constants/separation = 0
+
+[node name="MarginContainer" type="MarginContainer" parent="VBoxContainer" unique_id=806649657]
```
regenesis-claw, line 44: [P1] This `MarginContainer` has no `mouse_filter`, so it defaults to `MOUSE_FILTER_STOP`. Every other node in this scene is correctly `IGNORE` (root, VBox, `HBoxContainer_Header`, `Body`), but this one isn't — and it's the widest node in the header chain: it's the child of a VBox inside a full-rect `MarginContainer`, so it stretches to the **full HUD width** while only ~250px of it is actual buttons. The result is a transparent ~100px-tall, full-width strip that swallows taps/clicks meant for scene UI underneath whenever the toolbar is visible.

That's the same class of bug as #2578 / the `debug_panel` body fix, and it's precisely what pure layout containers are supposed to avoid (REVIEW.md §5: "pure layout containers with no own visuals should be `MOUSE_FILTER_IGNORE`").

```suggestion
[node name="MarginContainer" type="MarginContainer" parent="VBoxContainer" unique_id=806649657]
layout_mode = 2
mouse_filter = 2
```

Worth a quick check in preview against a scene with UI in the top-left band to confirm it's gone.

lib/src/hud_input_lint.rs
```diff
@@ -78,25 +78,14 @@ fn assert_click_through(node: &TscnNode, scene: &str) {
...
-    );
-    assert_click_through(
-        find(
-            &nodes,
-            "HBoxContainer_TopButtons",
-            "MarginContainer/VBoxContainer/MarginContainer_Header",
-        ),
-        "debug_panel",
-    );
+    assert_click_through(find(&nodes, "VBoxContainer", "."), "debug_panel");
```
regenesis-claw, line 88: [P2] `debug_tools_sit_between_scene_ui_and_right_panels` is removed here (correctly — `SafeMarginContainerDebug` no longer exists), but nothing replaces it for the new toolbar, and the suite now has no coverage of `preview_hud_panel.tscn` at all. That gap is why the header `MarginContainer` `MOUSE_FILTER_STOP` above slipped through.

Suggest adding a test in the same style as `debug_panel_body_does_not_block_scene_ui`:

```rust
/// The preview HUD toolbar spans the full HUD rect but only its buttons and the
/// console body should claim input; every layout container in between must be
/// IGNORE so the reserved-but-empty area doesn't eat taps meant for scene UI.
#[test]
fn preview_hud_toolbar_does_not_block_scene_ui() {
    let nodes = parse_tscn(
        "../godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.tscn",
    );
    for (name, parent) in [
        ("VBoxContainer", "."),
        ("MarginContainer", "VBoxContainer"),
        ("HBoxContainer_Header", "VBoxContainer/MarginContainer"),
        ("Body", "VBoxContainer"),
    ] {
        assert_click_through(find(&nodes, name, parent), "preview_hud_panel");
    }
}
```

godot/src/ui/components/organisms/preview_hud_panel/preview_hud_panel.tscn
```diff
@@ -0,0 +1,143 @@
...
+anchor_bottom = 1.0
+grow_horizontal = 2
+grow_vertical = 2
+mouse_filter = 2
+theme_override_constants/margin_right = 0
+script = ExtResource("1_script")
+metadata/mobile_preview_orientation = "landscape"
+
+[node name="VBoxContainer" type="VBoxContainer" parent="." unique_id=900190433]
+custom_minimum_size = Vector2(0, 450)
```
regenesis-claw: [P2] Two bits of inert config worth cleaning up while you're here:

1. `custom_minimum_size = Vector2(0, 450)` — this VBox is the only child of a full-rect `MarginContainer`, which fits its child to the full rect, so the 450px minimum never applies. Either it's leftover from an earlier layout, or the intent was a fixed-height toolbar that isn't actually taking effect.
2. `SubResource("StyleBoxFlat_pressed")` (bordered) is assigned to Reload's `pressed`/`hover_pressed` states, but `_setup_reload_style()` overrides **all** states with the borderless `normal` box at runtime — so that subresource is dead and the `.tscn` reads as if Reload gets a border when held, which contradicts the code and the PR description. Drop the `pressed`/`hover*` overrides from the `.tscn` (or the runtime override), so the two agree.

lib/src/realm/scene_entity_coordinator.rs
```diff
@@ -508,26 +515,34 @@ impl SceneEntityCoordinator {
             };
 
             self.fixed_desired_entities.insert(entity_base.hash.clone());
+            // Remember how to re-fetch this scene (used by reload_scene_data).
+            self.reloadable_entity_bases
```
regenesis-claw, line 519: [P2] `reloadable_entity_bases` only ever grows within a realm — entries are inserted here and in `_set_fixed_desired_entities_global_urns`, but `_set_fixed_desired_entities` deliberately doesn't clear it (per the comment above), so a hash stays in the map after the scene drops out of `fixed_desired_entities`. Two consequences, both minor but easy to avoid:

- Reloading a scene that is no longer desired re-issues a fetch for it instead of falling through to the position-rediscovery path.
- The map grows across every fixed-scene set for the life of the realm (bounded by distinct scenes, so not a leak in practice, but unbounded in principle).

A cheap fix that keeps the global-scene entries intact: retain only hashes present in `fixed_desired_entities ∪ global_desired_entities` at the end of `_set_fixed_desired_entities`, rather than skipping the cleanup entirely. Alternatively gate the lookup in `reload_scene_data` on `fixed_desired_entities.contains(&scene_id) || global_desired_entities.iter().any(...)`.

godot/src/logic/scene_fetcher.gd
```diff
@@ -1192,6 +1192,12 @@ func _on_try_spawn_scene(
 func reload_scene(scene_id: String) -> void:
 	var scene = loaded_scenes.get(scene_id)
 	if scene != null:
+		# Show the loading screen up front so killing the scene doesn't flash an empty parcel
+		# before it re-loads. Skipped for preview hot-reload, which is meant to be seamless.
+		if not _is_hot_reloading:
+			var explorer = Global.get_explorer()
```
regenesis-claw, line 1202: [P2] Question rather than a blocker: showing the loading screen up front means the dismissal now depends entirely on the reload path completing. For a fixed/World scene the new `reload_scene_data` route re-issues an entity fetch — if that request 404s or fails (server hiccup, scene un-deployed between load and reload), does anything take the loading screen back down, or is the user stuck on it until they teleport?

If the scene-timeout modal path covers it, worth a one-line comment here saying so. If not, consider clearing the loading screen on the failure branch of the reload fetch.

regenesis-claw (changes requested):
Solid consolidation — the toolbar hub is a clean home for console/scene-stats/reload, the `_restore_bottom_left_hud` / `_hide_bottom_left_hud` pair replaces the scattered `chat_panel.show()/hide()` calls nicely, and the fixed/global reload fix in `scene_entity_coordinator.rs` is a genuine bug fix with a clear rationale. Two things block merge.

**Pre-flight (REVIEW.md §0)**
- Branch is up to date with `main` (merge commit `1140425`). ✅
- **CI is red**: `📊 Static checks` fails on `gdlint` (2 errors, see below); every downstream job (Linux / Clippy / Android / Docker) is skipped as a result, so there is currently *no* build signal on this PR at all.
- iOS build: none present, and no `build` label. This PR touches no native iOS plugin, `OS.get_name()` branch, deeplink, keyboard/safe-area, or `av`/`comms` code, so per §0.2 an iOS build is **not required** — but it is a HUD layout change that QA will exercise on device, so adding the `build` label once static checks are green is worth it.

## Findings

| # | Sev | Where | Issue |
|---|-----|-------|-------|
| 1 | P0 | `preview_hud_panel.gd:68,88` | `gdlint` fails — CI red, no builds produced |
| 2 | P1 | `preview_hud_panel.tscn:57` | Header `MarginContainer` defaults to `MOUSE_FILTER_STOP` → full-width invisible strip eats scene-UI input |
| 3 | P2 | `hud_input_lint.rs` | Deleted ordering test not replaced with a click-through test for the new toolbar |
| 4 | P2 | `preview_hud_panel.tscn` | Dead `StyleBoxFlat_pressed` on Reload + inert `custom_minimum_size` on the VBox |
| 5 | P2 | `scene_entity_coordinator.rs` | `reloadable_entity_bases` entries are never removed when a scene stops being desired |
| 6 | P2 | `scene_fetcher.gd:1195` | Loading screen shown up front on reload — confirm the failure path still dismisses it |

Details inline. (2) is the one I'd most want fixed before this reaches a preview build: a ~100px full-width transparent band over the top of the HUD swallowing taps meant for scene UI is exactly the class of regression the `hud_input_lint` suite exists to catch, and the test that would have caught the old version of it was removed in this PR.

### Security
No security issues found — no network/auth/user-input surface touched; the Rust change re-issues an already-derived content-server URL from an existing `EntityBase`.

Requesting changes on (1) and (2); the P2s are fine as follow-ups or quick fixes in the same push.

_Requested by Seba via Slack_

regenesis-claw (approved):
## Re-review — delta on `114042569` → `aa3ab648` (4 commits)

All blockers addressed, and the fixes are the right ones rather than the minimal ones.

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| 1 | P0 | `gdlint` async-function-name on `set_scene_status_available` / `set_scene` | ✅ `await ready` guards dropped (`13f90af`) — both setters synchronous again, no call-site churn in `explorer.gd`. `📊 Static checks` is **green**. |
| 2 | P1 | Header `MarginContainer` defaulting to `MOUSE_FILTER_STOP` | ✅ `mouse_filter = 2` added (`f8b01e9`). Every container in the scene is now IGNORE; only the three buttons and the console/scene-stats bodies claim input. |
| 3 | P2 | No `hud_input_lint` coverage for the new toolbar | ✅ `preview_hud_panel_containers_do_not_block_scene_ui` added (`f8b01e9`), asserting click-through on all four containers — this now regression-guards (2). |
| 4 | P2 | Dead `StyleBoxFlat_pressed` + inert `custom_minimum_size(0, 450)` | ✅ Both removed (`aa3ab64`), along with the redundant per-state overrides on the two toggle buttons that `_own_border_stylebox()` replaces at runtime. The `.tscn` and the code now agree on what Reload looks like when held. |
| 5 | P2 | `reloadable_entity_bases` never pruned | ✅ `prune_reloadable_entity_bases()` added and called from both desired-set setters (`b255c0c`). |
| 6 | P2 | Loading screen dismissal on a failed reload fetch | ⚠️ Unchanged, unanswered — see below. |

### Notes on the new code

- I checked the ordering hazard in the prune: `_set_fixed_desired_entities_global_urns` pushes into `global_desired_entities` on **both** the cached and the not-yet-cached branch (lines 576 and 591), so a global scene whose fetch is still in flight is still in the `desired` union when `retain` runs and its reloadable base survives. That was the one way this prune could have silently re-broken the Worlds reload fix — it doesn't. Same for `_set_fixed_desired_entities`: it clears only `fixed_desired_entities`, and the union preserves the global entries.
- The lint test's node paths (`VBoxContainer` → `.`, `MarginContainer` → `VBoxContainer`, `HBoxContainer_Header` → `VBoxContainer/MarginContainer`, `Body` → `VBoxContainer`) all match the current `.tscn`.

### Still open

- **(6)** remains a genuine question rather than a blocker: on a fixed/World reload whose entity re-fetch fails (404, un-deployed scene, network hiccup), does anything take the up-front loading screen back down? If the scene-timeout modal covers it, a one-line comment in `reload_scene` saying so would close the loop; if not, it's a decent follow-up issue.
- **CI**: `📊 Static checks` and the gate pass; `🐧 Linux`, `🔧 Clippy` and `🤖 Android` are still running at the time of this review. The Rust delta is small and self-contained (a `HashMap::retain` over an already-imported `HashSet` plus one new `#[test]`), so I don't expect trouble — but this approval assumes they land green. Worth adding the `build` label afterwards so QA gets a device build for a HUD layout change of this size.

Approving.

_Requested by Seba via Slack_

kuruk-mm (approved):
great job! code looks clean! tested and working perfectly! congrats

Image

PR #2689, opened 2026-08-12
fix(auth): complete the signin deep link on a cold start

### LINKED ISSUE #2644: [Bug] Google Sign-In loop prevents login on Android (Xiaomi Redmi Note, v1.11.0)
### 🕹️Platform

Android

### 🔢App Version

v1.11.0

### 📱Device Information

Xiaomi Redmi Note (Android)

### 📄Issue description

User (MiguelAmargo) reports Google Sign-In gets stuck in a loop and they are unable to log in at all. Tapping "Continue with Google" opens the browser, they sign in successfully with their email, the browser hands control back to the app, but instead of completing login the app shows the sign-in screen again. This repeats every attempt, so the user is completely locked out.

User previously signed in once with this account and stayed logged in without issue; the loop only started recently.

This matches a cluster of Google Play Store reviews reported since 2026-07-01 with the same symptom (browser auth succeeds, app returns to sign-in screen instead of completing login). Working theory: the OS reclaims memory / kills the app process during the browser hop on lower-spec devices, wiping the in-memory `pending_mobile_auth` state before the `decentraland://open?signin=...` deep link can complete the login handshake.

### ✅Expected behavior

After successfully signing in with Google in the browser, the app should complete login and take the user into Decentraland, not loop back to the sign-in screen.

### 🖼️Screenshots / Media

[N/A (reported via Discord, no video/screenshot provided)](https://discord.com/channels/417796904760639509/1534873024858427433)

### ▶️Steps to reproduce

1. Open Decentraland mobile app (Android, Xiaomi Redmi Note, v1.11.0)
2. Tap "Continue with Google"
3. Browser opens, sign in with Google account successfully
4. Get returned to the DCL app
5. App shows sign-in screen again instead of completing login (repeats indefinitely)

### 👁️Occurrence

100% currently (user is fully locked out, cannot log in at all)

---
Reported via Discord by MiguelAmargo. Requested by ToxicWaifu via Slack.

leanmendoza:
closes #2644

Mobile sign-ins are lost when the OS kills the app while the user is in the browser. `pending_mobile_auth` lives only in that process's memory, so the returning `decentraland://open?signin=<identityId>` deep link hits a fresh boot with the flag gone and `deep_link_router` drops the token with a `printerr` — the user lands back on ACCOUNT_HOME after a sign-in they already completed.

Over 14 days, joining the auth site's `Login Success` to the in-app `AUTH_SUCCESS` on the shared session id: **92.0%** of sign-ins reach the app when the process survives, **26.5%** when it cold-starts. Cold starts hit 32% of confirmed sign-ins, so ~244 logins were lost in that window (~17/day), and it has been getting worse per version (0.70: 80.6% → 1.12: 69.7%).

The guard protected nothing: the field is `Option<()>`, a unit type with no payload, and `complete_mobile_auth(identity_id)` is a self-contained `fetch_identity_by_id` that only needs the id from the link. An invalid or expired id already fails there with a real error.

## What changed

- Drop the Rust gate in `complete_mobile_connect_account`.
- On the cold path, park the token on `deep_link_router` instead of completing it inline: there, `process_deep_link` runs from `Global._notification`, before the first scene exists, so `wallet_connected` / `profile_changed` would fire into a lobby that has not connected them yet.
- `lobby.gd` redeems it — from `_ready` when the link arrived first, from the new `deep_link_signin_parked` signal when it arrives later (Android can deliver the launch intent asynchronously). The token is single-shot.
- Redeeming takes precedence over session recovery; if the token is expired (a re-tapped old link), `_on_auth_error` runs the recovery that was skipped, so it cannot cost an existing session.
- The warm path — app alive through the browser hop — is unchanged.

Resumed sign-ins report `AUTH_BROWSER_OPEN` with `method: "deeplink_cold_start"`, so the funnel can measure the fix separately from ordinary in-process logins.

Validated on an Android device.

## How to reproduce the cold start (Android)

Waiting for Android to kill the app on its own is unreliable — it depends on OEM policy. Handing the auth flow to external Chrome makes it deterministic, because the app can then be killed without touching the browser:

1. Signed out, tap **Sign in → Google**. The in-app browser opens.
2. **Right away**, before the auth page redirects, open the ⋮ menu and pick **Open in Chrome**. Timing matters here — once it redirects it is too late.
3. With the flow now in external Chrome, kill the app from the recents switcher.
4. Finish signing in in Chrome and let it redirect back. The app opens fresh.
5. **Expected:** a brief "Finishing sign in…", then signed in — Welcome Back for an existing profile, avatar creation for a new one. Before this PR it landed on ACCOUNT_HOME with the login discarded.

Verified on device with this build.

## QA

The new code only runs when the app was killed mid-login, so nothing below should look different from today. That is the point: the risk here is breaking the ordinary sign-in everyone uses, not the rare one. Please run these on real devices — the simulator/emulator does not kill apps the way a phone does.

Test 1 and 2 are the important ones. Test 3 is quick and worth doing. Test 4 is optional and needs adb.

### 1. Normal sign-in — Android

1. Install the build and open it. If it opens you straight into the app, go to **Settings → Sign out** first so you land on the start screen.
2. Tap **Sign in**, then **Google** (any provider is fine).
3. The browser opens. Pick your account and complete the login there. Don't touch anything else — just let it finish.
4. The browser says it is sending you back, and the app comes to the front on its own.

**Expected:** you end up signed in. An account that already has an avatar shows the "You're signed in as …" welcome screen; a brand-new account goes to avatar creation. Carry on to the end and check you reach the main menu / world like always.

**Something is wrong if:** you land back on the start screen, you see "Authentication failed", or the spinner just sits there.

### 2. Normal sign-in — iOS

Exactly the same steps as test 1, on an iPhone.

iOS never showed the original bug, but it runs the same sign-in code, so this is purely checking we did not break it.

### 3. Reopening the app while already signed in — Android and iOS

1. Starting from the account you signed into above, close the app completely (swipe it away from the app switcher).
2. Open it again.

**Expected:** it goes straight in, no sign-in screen, same account as before.

### 4. A broken sign-in link (optional, Android, needs adb)

Checks that a stale or malformed link cannot cost someone their session.

1. Sign in normally and confirm you are in. Close the app completely.
2. Run:
   ```
   adb shell am start -a android.intent.action.VIEW -d "decentraland://open?signin=not-a-real-id"
   ```

**Expected:** the app opens, briefly shows "Finishing sign in…", cannot use that made-up id, and quietly falls back to the session you already had — you end up signed in as before.

**Something is wrong if:** you are left on an "Authentication failed" screen, or signed out entirely.


lib/src/auth/dcl_player_identity.rs
```diff
@@ -797,16 +802,17 @@ impl DclPlayerIdentity {
...
-        if self.pending_mobile_auth.take().is_none() {
-            tracing::error!("No pending mobile auth to complete");
-            self.base_mut().call_deferred(
-                "_error_getting_wallet",
-                &["No pending mobile auth".to_variant()],
-            );
-            return;
-        };
+        // Consume the flag when this process did start the flow; its absence is not an error.
+        self.pending_mobile_auth.take();
```
regenesis-claw, line 842: [P1] Dropping the gate here also drops the only thing that made **Cancel** stick.

`abort_try_connect_account()` clears `pending_mobile_auth`. Pre-PR that meant a deep link arriving after a cancel hit the gate and was refused. Post-PR the same link is indistinguishable from a cold start: `has_pending_mobile_auth()` is false, the router parks it, the still-alive lobby redeems it, and the user is signed in against an explicit cancel.

Repro (warm, no OS kill involved):
1. Signed out, tap Sign in -> Google. In-app browser opens.
2. Back out to the app and tap **Cancel** on the spinner.
3. Return to the browser and finish the login (or let an already-issued redirect fire).
4. The link comes back, the app shows "Finishing sign in…" and signs you in.

That also makes "the guard protected nothing" and "the warm path is unchanged" both false as written. The fix does not need the gate back — it needs the cancel to be distinguishable from a cold start. Suggest an explicit `cancelled` marker: have `abort_try_connect_account()` record that this process cancelled a mobile auth, and have the lobby refuse to redeem a parked token in that state (a real cold start has an empty marker because the process is new).

godot/src/deep_link_router.gd
```diff
@@ -208,11 +224,46 @@ func _route_teleport() -> void:
...
+	# to `printerr` and drop the token, stranding the user on ACCOUNT_HOME after a sign-in
+	# they had already completed — ~26% of logins that hit a cold start survived it.
+	#
+	# Don't complete it here: on a cold start this runs from Global._notification, before the
+	# first scene exists, so wallet_connected/profile_changed could fire into a lobby that
+	# hasn't connected them yet. Park the token and let the lobby redeem it on its own terms.
+	print("[DEEPLINK] signin token arrived with no pending auth (cold start) — parking it")
+	_pending_signin_identity_id = identity_id
+	_pending_signin_parked_at_ms = Time.get_ticks_msec()
+	deep_link_signin_parked.emit.call_deferred()
```
regenesis-claw, line 263: [P1] The parked token outlives its only consumer. `deep_link_signin_parked` is connected in the lobby and nowhere else, so when the deep link arrives while the app is warm and *past* the lobby — explorer or menu, i.e. the user is already signed in — nothing redeems it, and it stays live for `PENDING_SIGNIN_MAX_AGE_MS` (5 minutes).

`Global.sign_out()` swaps to a fresh lobby scene, whose `_ready` calls `take_pending_signin_identity_id()` first thing.

Failure scenario:
1. Signed in and in-world, the user taps an old/forwarded `?signin=` link (mail, chat, browser history). Nothing visibly happens; the token is parked.
2. Within 5 minutes they go Settings -> Sign out, intending to switch accounts.
3. The new lobby redeems the parked token and signs them straight back in as whoever that link belongs to, skipping ACCOUNT_HOME. `sign_out()` has already wiped `session_account`, so the `auth_error` fallback has nothing to fall back to either.

The account that gets signed in is chosen by a link the user may have received from someone else, which makes this more than a UX wart. Two options: park only when this is genuinely a fresh boot (no wallet ever connected in this process, or `Time.get_ticks_msec()` under a small boot window), and/or refuse to redeem a token that predates the current lobby's `_ready`.

godot/src/ui/pages/auth/lobby.gd
```diff
@@ -898,6 +916,36 @@ func _on_button_try_again_pressed():
...
+
+	print("[DEEPLINK] Resuming the interrupted sign-in from the parked token")
+	_resumed_signin_from_deep_link = true
+	# Put the lobby in the state the browser flow would have left it in, so the
+	# wallet_connected -> profile_changed chain lands on Welcome Back / avatar creation
+	# instead of leaving the user parked on ACCOUNT_HOME with a wallet already connected.
+	waiting_for_new_wallet = true
+	# Distinct auth_method: AUTH_BROWSER_OPEN is the same screen, but this one is a resumed
+	# cold start, and the funnel needs to tell the two apart to measure the fix.
+	show_auth_browser_open_screen("Finishing sign in...", "deeplink_cold_start")
```
regenesis-claw, line 952: [P1] This screen shows `button_cancel`, and Cancel is not handled on the resume path.

`_on_button_cancel_pressed()` calls `abort_try_connect_account()` + `show_auth_home_screen()`. But (a) `abort_try_connect_account()` only aborts `try_connect_account_handle` — the completion started here runs on an untracked `handle.spawn`, so it keeps going; (b) `_ready` already returned early, so the session recovery it skipped never runs; (c) `_resumed_signin_from_deep_link` is left `true`.

Failure scenario: a user with a stored session cold-starts on a stale link, the fetch is slow, they tap Cancel. They now sit on ACCOUNT_HOME **signed out**, with a session on disk that will not be restored until the next launch. Then, seconds later, one of two things yanks the UI out from under them: `wallet_connected` (the completion succeeded -> `waiting_for_new_wallet` is still true -> Welcome Back for the sign-in they cancelled) or `auth_error` (-> the fallback fires the splash screen + `try_recover_account`).

At minimum, clear `_resumed_signin_from_deep_link` and run the skipped recovery in `_on_button_cancel_pressed()` when it is set. Worth adding a timeout too — otherwise Cancel is the only exit from "Finishing sign in…".

godot/src/ui/pages/auth/lobby.gd
```diff
@@ -898,6 +916,36 @@ func _on_button_try_again_pressed():
...
+	# Same reason the session-recovery path below awaits it: on a sandbox StoreKit build the
+	# hybrid env has to be settled before the profile fetch this kicks off, or the profile
+	# loads from the wrong backend. No-op off iOS, and capped at 5s.
+	await Iap.async_await_env_resolved()
+	Global.player_identity.complete_mobile_connect_account(identity_id)
+	return true
+
+
+func _on_deep_link_signin_parked() -> void:
```
regenesis-claw, line 961: [P2] This entry point can fire long after `_ready` finished, and it does not check what the lobby is already doing.

Cold start with an existing session where Android delivers the intent late: `_ready` takes no token, awaits `Iap`, calls `try_recover_account()`, sets `loading_first_profile = true`. The link then lands, this handler redeems it, sets `waiting_for_new_wallet = true` and starts a second identity flow concurrently with the recovery. Whichever `profile_changed` arrives first hits the `loading_first_profile` branch in `_async_on_profile_changed`, which `return`s after `async_close_sign_in()` — so the `waiting_for_new_wallet` branch never runs, no Welcome Back, and the user enters the app while the other identity is still being applied.

Related coverage gap: if the recovery already navigated to explorer/menu, the lobby is gone and this signal has no receiver — the sign-in this PR exists to save is still lost in that window, and the token then sits parked (see the router comment).

Also, the coroutine's result is discarded here (no `await`), so a redemption that fails inside is invisible to the caller. Guarding on `waiting_for_new_wallet or loading_first_profile` before redeeming would cover both.

regenesis-claw (changes requested):
Reviewed as a hostile skeptic: every claim in the description checked against the diff, findings only.

**Verified as described:** `pending_mobile_auth` really is `Option<()>` with no payload; `complete_mobile_auth(identity_id)` really is self-contained (`fetch_identity_by_id`, errors route to `_error_getting_wallet` -> `auth_error`); the parked token really is single-shot (cleared on read before the staleness check); the resumed screen really does emit `AUTH_BROWSER_OPEN` with `method: "deeplink_cold_start"` via `track_screen_viewed`'s extra payload.

**Not verified (no access):** the 92.0% / 26.5% / 32% funnel figures and the ~244-lost-logins estimate; "validated on an Android device"; whether the backend rejects a *replayed* `identity_id` (relevant now that a token can be redeemed minutes after it was parked, see P1-3).

**Two description claims are falsified by the diff:**

1. *"The guard protected nothing."* It protected Cancel. `abort_try_connect_account()` sets `pending_mobile_auth = None`, so pre-PR a deep link arriving after the user tapped Cancel hit the gate and was refused. Post-PR that same link takes the new cold path, parks, emits, and the live lobby redeems it — the cancelled sign-in completes anyway.

2. *"The warm path — app alive through the browser hop — is unchanged."* It is changed for exactly the case above: warm process, flag cleared by Cancel, deep link returns. That is the warm path, and its behaviour is now the opposite of what it was.

3. *"It cannot cost an existing session."* Only the `auth_error` route is covered. `_ready` returns early on a successful redeem, so session recovery never runs; tapping Cancel on the "Finishing sign in…" spinner (`button_cancel` is shown by `show_auth_browser_open_screen`) drops the user on ACCOUNT_HOME with the on-disk session never restored, and `abort_try_connect_account()` does not abort the completion task (`complete_mobile_connect_account` spawns an untracked handle; only `try_connect_account_handle` is aborted), so the cancelled flow can still land afterwards.

Findings are on the diff lines below. P1 x3, P2 x2.

**Pre-flight (REVIEW.md):** branch is MERGEABLE against `main`. This touches deeplinks, so an iOS build is required — the `Build iOS export → R2` check is present but still *pending*, along with Linux and Android. Static checks and Clippy are green. I am not treating a pending iOS job as a green iOS artifact.

Ludmilafantaniella (approved):
## ✅ Approve by QA

Tested on Android (Galaxy A35) and iPhone(16 Pro).

- [x] Cold start repro (Android): signed out → Sign in → Google, switched to external Chrome mid-flow, killed the app, finished sign-in in Chrome and returned - app resumes and completes sign-in correctly (previously landed on ACCOUNT_HOME with login discarded)
- [x] Normal sign-in via the in-app browser screen (Android) - completes correctly, reaches welcome screen / avatar creation as expected
- [x] Normal sign-in (iOS) - same flow works correctly, no regressions
- [x] Reopening the app while already signed in (Android + iOS) - goes straight in, no sign-in screen, same account

https://github.com/user-attachments/assets/f438c57c-f2c6-458d-89db-713287d3020b

regenesis-claw (approved):
## Re-review of the review follow-ups (head `4cfce7bc`) — Approved

All four findings from my 2026-08-12 review (`837dc97`) and the five from the stacked follow-up (now folded in here) are addressed. Verified against the delta `398ea0b3...4cfce7bc`:

| Prior finding | Status |
|---|---|
| [P1] Dropping the Rust gate also drops what made **Cancel** stick | Fixed — `mobile_auth_cancelled` (gated by `mobile_auth_started` so the native WalletConnect Cancel doesn't set it) is checked first in `_handle_signin_deep_link`, plus a monotonic `mobile_auth_attempt` re-checked on the main thread in `_mobile_auth_completed` / `_mobile_auth_failed`. That closes the post-`await` window `abort()` can't reach. |
| [P1] Parked token outlives its only consumer when the app is warm and past the lobby | Fixed — non-empty `get_address_str()` now refuses the token, so `sign_out()` can't hand a fresh lobby someone else's sign-in. |
| [P1] Cancel unhandled on the resume path | Fixed — `_on_button_cancel_pressed` clears `waiting_for_new_wallet` unconditionally (correct: `login.gd` arms it for every browser sign-in) and runs `_fall_back_to_stored_session()` so a cancelled resume doesn't leave a stored session unrestored. |
| [P2] `_on_deep_link_signin_parked` doesn't check what the lobby is already doing | Fixed — guarded on `waiting_for_new_wallet` / `loading_first_profile` / the new `_recovering_session` (which covers exactly the `Iap.async_await_env_resolved()` suspension the old flags missed), and the token is consumed rather than left parked. |
| [P1] `EphemeralAuthChain::new` panics on an unvalidated key/expiration | Fixed — `LocalWallet::from_bytes` pre-validated and the expiration goes through `u64::try_from` + `checked_add`, both returning errors that reach `auth_error`. |
| [P2] Splash overlay dead-end on the error screen | Fixed — `_show_auth_error` calls `SplashOverlay.fade_out()`. |

`try_connect_account_handle.replace(complete_handle)` with `abort()` on the previous handle is the right call (dropping a tokio `JoinHandle` detaches), and the attempt token is what actually makes the cancel authoritative — the handle is belt-and-braces.

### Remaining, non-blocking

[P2] On the **cold-start resume** path `mobile_auth_started` is false (fresh process), so `abort_try_connect_account` sets `mobile_auth_cancelled = false`. If the user cancels the "Finishing sign in…" screen and Android re-delivers the same launch intent — the asynchronous delivery this PR explicitly handles — the token is parked again and, with no stored session to trigger `loading_first_profile`, redeemed a second time, signing in the user who just cancelled. Setting `mobile_auth_started = true` inside `complete_mobile_connect_account` too would make the marker cover both entry points.

[P2] `mobile_auth_cancelled` is process-wide and cleared by `start_mobile_connect_account`: cancel → tap Sign in again → finish the *first*, still-open browser tab, and token #1 is honoured. The user is mid-sign-in in that scenario, so the outcome is defensible; noting it only because the cancel marker reads as stronger than it is.

### CI / QA

All checks green on `4cfce7bc` (Linux, Android, iOS export, Docker, Clippy, static checks). iOS export built successfully; no `build`/`build-ios` label on the PR, so no TestFlight artifact.

Worth re-running the QA sign-off: @Ludmilafantaniella approved on `398ea0b3`, and these commits change the **warm** sign-in and Cancel paths, not just the cold-start one — QA steps 1 (normal sign-in) and the Cancel case are the ones that regressed most cheaply here.

No security issues found.

Requested by Lean via Slack

leanmendoza:
new build 1585  @Ludmilafantaniella  we need to retest since the codebase you tested changed

kuruk-mm (approved):
great work!

PR #2690, opened 2026-08-12
feat: avatar masks (upper-body emotes) + EmoteState playback signals

### LINKED ISSUE #2621: Implement Avatar Masks (upper-body-only emotes) support
## Summary

The mobile client does **not** support Avatar Masks — the `avatarMask` parameter on the `triggerEmote` / `triggerSceneEmote` restricted actions, which lets a scene play an emote on the upper body only (e.g. waving while still walking) instead of overriding the full-body animation.

## Verification (godot-explorer, 2026-08-03)

- `op_trigger_emote` in `lib/dcl/js/restricted_actions.rs` only accepts `emote_id: String` — no mask param.
- `op_trigger_scene_emote` only accepts `emote_src: String` and `looping: bool` — no mask param.
- `RpcCall::TriggerEmote` / `TriggerSceneEmote` (`lib/dcl/scene_apis.rs`) carry no mask field either.
- No `AvatarMask` / `AM_UPPER_BODY` / bone-masking logic exists anywhere in the repo (checked `lib/` and `godot/` trees).
- Conclusion: any scene calling `triggerEmote(emote, { avatarMask: AvatarMask.AM_UPPER_BODY })` silently ignores the mask on mobile and always plays the full-body clip.

## Protocol Reference

- `decentraland/protocol#368` — feat: avatar masks (adds `decentraland/sdk/components/common/avatar_mask.proto`, `AvatarMask` enum with `AM_UPPER_BODY`)
- `decentraland/protocol#373` — feat: avatar masks (experimental follow-up)
- `decentraland/protocol#442` — feat: avatar emote command mask field (adds mask to `AvatarEmoteCommand`)

## Unity Reference Implementation

- `decentraland/unity-explorer#7462` — feat: avatar masks (original implementation)
- `decentraland/unity-explorer#8384` — fix: revert "feat: avatar masks (#7462)" (had to be rolled back once)
- `decentraland/unity-explorer#8480` — feat: avatar masks v2 (re-implementation, merged)
- `decentraland/unity-explorer#9532` — fix: scene emotes masked loop (open, follow-up fix)

Key files:
- `RestrictedActionsAPIWrapper.cs` (Infrastructure/SceneRuntime/Apis/Modules/RestrictedActionsApi) — `TriggerEmote(string predefinedEmote, uint? mask)` / `TriggerSceneEmote(string src, bool loop, uint? mask)`; converts the optional wire `AvatarMask` (`AM_UPPER_BODY`) into a runtime `AvatarEmoteMask` (absent = full body).
- `MaskedLegacyEmoteBlender.cs` (AvatarRendering/AvatarShape/UnityInterface) — resolves non-masked bones for a given mask and blends the emote animation layer only onto the masked bone subset.
- `EmoteMaskCatalog.cs` (AvatarRendering/AvatarShape/UnityInterface) — maps `AvatarEmoteMask` → Unity's native `AvatarMask` asset (bone selection).
- `EmotePlayer.cs` (AvatarRendering/Emotes/Systems/Play) — applies the resolved mask when playing an emote.
- `AvatarBase.cs` (AvatarRendering/AvatarShape/UnityInterface) — `StartMaskedLegacyEmote(AnimationClip clip, AvatarMask avatarMask, bool loop)`.
- Proto: `AvatarMask.gen.cs` (Protocol/DecentralandProtocol).

## Implementation Guideline (mobile)

1. Extend `RpcCall::TriggerEmote` / `RpcCall::TriggerSceneEmote` (`lib/dcl/scene_apis.rs`) with an optional mask field.
2. Update `op_trigger_emote` / `op_trigger_scene_emote` (`lib/dcl/js/restricted_actions.rs`) to read the optional `mask: u32` param from JS and forward it.
3. Wire the mask through `handle_restricted_actions.rs::trigger_emote` down to the avatar animation system (`decentraland_components/avatar/avatar.gd`).
4. On the Godot side, implement upper-body-only blending — likely via an `AnimationTree` blend mask / bone mask (Godot supports `AnimationNodeBlendTree` filters per-bone) so the emote plays on upper body bones while the lower body keeps the locomotion animation, mirroring `MaskedLegacyEmoteBlender.cs`.
5. Currently only one mask value exists in the protocol (`AM_UPPER_BODY`), so the initial implementation only needs to support that case (absent mask = current full-body behavior, unchanged).

---
Requested by Rodrigo via Slack

### LINKED ISSUE #2655: Support EmoteState in AvatarEmoteCommand (emote playback completion signals)
## Summary

`decentraland/protocol#459` adds an `EmoteState` field to `PBAvatarEmoteCommand` so scenes can tell **when an emote stopped**, not just when it started. Today a scene that plays an emote has no way to know it ended — the common workaround is a hardcoded `setTimeout(clipLength)`, which drifts and is simply wrong when the emote gets cancelled early.

The Godot explorer currently only ever reports the "started" case, so scenes written against the new API will never see a completion signal here.

## What the protocol adds

```protobuf
enum EmoteState {
  ES_STARTED     = 0;  // zero value — entries from older explorers read as "started"
  ES_FINISHED    = 1;  // non-looping emote completed naturally
  ES_INTERRUPTED = 2;  // cancelled: movement, teleport, superseded by another emote,
                       // explicit stop, or scene change
}

message PBAvatarEmoteCommand {
  string emote_urn = 1;
  bool loop = 2;
  uint32 timestamp = 3;
  optional AvatarMask mask = 4;
  optional EmoteState state = 5;   // <-- new; absent means ES_STARTED
}
```

The field is additive and backwards compatible: existing explorers that never write it are read as `ES_STARTED`, which is what they mean today.

Scenes consume it by reading the grow-only set on the player entity and looking at the `state` of each appended entry.

## Where we stand in this repo

The explorer appends to `AvatarEmoteCommand` (component 1088) in two places, both of them "started"-only:

- `lib/src/scene_runner/scene_manager.rs:1004` — `on_primary_player_trigger_emote`, for the local player, fanned out to every loaded scene.
- `lib/src/avatars/avatar_scene.rs:1032` — `on_avatar_trigger_emote`, for remote avatars, only pushed to the scenes the avatar is actually in.

Both construct the message with `mask: None` and `timestamp: 0`; the timestamp is replaced with a real per-entity monotonic counter later, at flush time, in `lib/src/scene_runner/components/avatar_data.rs:73-90`. That counter is what gives the new state entries their ordering, so it already works in our favour — a `FINISHED` entry appended after its `STARTED` entry will carry a strictly higher timestamp.

Both call sites are driven by the same `emote_triggered` signal that `AvatarEmoteController` emits when playback *begins* (`godot/src/decentraland_components/avatar/avatar_emote_controller.gd:202`), wired up at `godot/src/ui/explorer.gd:289` for the local player and `lib/src/avatars/avatar_scene.rs:838` for remote ones. There is no counterpart signal for playback ending.

The good news is that the controller **already detects** every termination case we need — it just keeps the knowledge to itself:

| Case | Where it already happens |
|---|---|
| Natural end of a non-looping emote | `avatar_emote_controller.gd:882-889` — `process()` notices the playback position stopped advancing and clears `playing_single` / `playing_mixed` |
| Cancelled because the player moved | `avatar_emote_controller.gd:870-875` — same `process()`, outside the teleport grace window |
| Explicit stop (backpack, profile, preview UIs) | `stop_emote()`, `avatar_emote_controller.gd:164` |
| Superseded by another emote | `play_emote()`, `avatar_emote_controller.gd:172` |
| Avatar left the scene | `on_avatar_changed_scene`, `lib/src/avatars/avatar_scene.rs:1073` |

So the work is mostly about surfacing state that is already computed, rather than writing new playback tracking.

## Proposed work

1. **Update the vendored proto.** `lib/src/dcl/components/proto/decentraland/sdk/components/avatar_emote_command.proto` is a local copy and still lacks the enum and field 5. It has to be synced once protocol#459 merges; `prost` regenerates the Rust type from it at build time.

2. **Emit an emote-ended signal from `AvatarEmoteController`.** Add something like `emote_finished(urn, interrupted)` next to the existing `emote_triggered`, declared on `DclAvatar` (`lib/src/godot_classes/dcl_avatar.rs:147` is where `emote_triggered` lives) and fired from the five termination points listed above. `process()` currently just flips the `playing_*` booleans; it needs to remember which URN was playing so it can name it in the signal.

3. **Append the terminal entry.** Connect the new signal for the local player (mirroring `explorer.gd:289`) and for remote avatars (mirroring `avatar_scene.rs:838`), and append a second `PBAvatarEmoteCommand` with the same `emote_urn` and `loop` but `state: ES_FINISHED` or `ES_INTERRUPTED`. Set `state: ES_STARTED` explicitly on the two existing start paths.

4. **Cover scene-triggered emotes too.** When a scene writes `AvatarEmoteCommand` on an `AvatarShape` entity, `update_avatar_shape_emote_command` (`lib/src/scene_runner/components/avatar_shape.rs:206`) plays it on that avatar. Those need terminal entries as well, otherwise the feature only works for the local player and remote players.

5. **Cover keyboard and emote-wheel emotes.** They already go through `play_emote()`, so they come along for free once step 2 is done — but they should be part of the test pass, since the Unity implementation explicitly calls them out.

### One hazard worth knowing about up front

`update_avatar_shape_emote_command` reacts to the component being *dirty* and replays `emotes.back()`. Explorer-side appends do show up as dirty on the next `take_dirty()` — that is exactly how the `onPlayerExpression` event works today (`lib/src/dcl/js/events.rs:421-450`). So once we start appending `FINISHED` / `INTERRUPTED` entries for `AvatarShape` entities, that read path will see them and try to play the emote again, in a loop.

The fix is to make the play path ignore any entry whose state is not `ES_STARTED`, and it should go in as part of this work rather than being discovered later.

## Reference implementations

- `decentraland/protocol#459` — the proto change (and `#458`, the experimental variant)
- `decentraland/unity-explorer#9589` — Unity's implementation. Note its scope is wider than this issue: it also adds completion signals for `AudioSource` and `Animator`, which we are deliberately leaving out here.
- `decentraland/js-sdk-toolchain#1516` — SDK side, exposes `state` to scenes
- `decentraland/docs#157` — docs
- `decentraland/sdk7-test-scenes#83` — test scenes

## Testing

The Unity PR's QA scenes are the natural target. Connect a build to the `sdk7testscenes.dcl.eth` world on Sepolia and check the emote scene:

```
--realm sdk7testscenes.dcl.eth --dclenv zone --position 89,-11 --skip-version-check true
```

Cases to confirm, for the local player, a remote player, and a scene-spawned `AvatarShape`:

- A non-looping emote played to the end produces `STARTED` then `FINISHED`.
- Walking out of an emote produces `STARTED` then `INTERRUPTED`.
- A looping emote produces `STARTED` and nothing else until it is stopped or interrupted.
- Triggering a second emote mid-playback produces `INTERRUPTED` for the first, `STARTED` for the second.
- No emote is ever replayed as a side effect of a terminal entry being appended.

## Out of scope

The `mask` field (field 4, upper-body-only emotes) is tracked separately in #2621.

EibrielInv:
## Summary

  Implements the two emote protocol features the mobile client was missing, plus two CRDT bugs found along the way.

  - **Avatar Masks (#2621)** — `triggerEmote`/`triggerSceneEmote` accept `mask: AM_UPPER_BODY`, so an emote plays on the upper body while locomotion keeps driving the legs (wave while walking). Wired through
  all three entry points: restricted actions, the `AvatarEmoteCommand` component, and rfc4/Pulse comms (Unity-compatible, so masked emotes replay masked on other clients).
  - **EmoteState (#2655)** — the explorer now appends a terminal `PBAvatarEmoteCommand` entry with `state: ES_FINISHED` / `ES_INTERRUPTED` when playback ends, so scenes no longer need `setTimeout(clipLength)` 
  guesswork. Covers the local player, remote avatars, and scene-owned `AvatarShape` NPCs.

  Closes #2621
  Closes #2655

  ## How it works

  **Masked playback.** `avatar.tscn`'s tree root is now an `AnimationNodeBlendTree` wrapping the existing locomotion state machine, with an `AnimationNodeOneShot` layer above it. Per play, the OneShot's filters
  are built from the clip's own tracks minus root motion and the lower body (`Avatar_Hips` + legs) — matching Unity's `UpperBodyAvatarMask` exactly. Prop and audio tracks pass the filter, so prop emotes keep 
  working. Masked emotes are exempt from cancel-on-move by construction (that is the feature), and `set_emoting` was fixed so Pulse doesn't broadcast a stop for a walking masked emote.

  **State reporting.** A single `_emit_emote_finished()` choke point in the controller (mirroring Unity's `StopEmote(reason)`) fires a new `emote_finished` signal from every termination path — natural end, 
  movement cancel, explicit stop, supersede, freeze — guaranteeing exactly one terminal entry per start. The component read path skips non-`ES_STARTED` entries so terminal entries can't re-trigger playback (a 
  hazard Unity structurally avoids by never reading the component back).

  **Proto.** `EmoteState` is added via the established `build.rs` patch idiom rather than a pin bump, since the proto tree is gitignored and rewritten by `cargo run -- install`. It no-ops once the pinned 
  `@dcl/protocol` ships the field.

  ## Bugs fixed along the way
  
  - **CRDT append order (all grow-only components).** `append_gos_component` iterated a *reverse* index forward, so whenever a tick carried more than one entry the scene received them newest-first. Emotes 
  exposed it (supersede logged STARTED before INTERRUPTED), but it also affects `PointerEventsResult`, `TriggerAreaResult`, `VideoEvent` and `AssetLoadLoadingState` — for video it caused silent event *loss*, 
  since the SDK dedups by timestamp and discarded the entries that arrived after a newer one. Possibly the cause of the long-disabled video-player scene test.
  - **`onPlayerExpression` re-emitting old entries.** The handler read the dirty count but indexed from the *front* of the value set, re-sending the oldest entries once the set grew past that count.

  ## Out of scope

  Masked *scene* emotes are not masked for remote observers — rfc4's `SceneEmote` message carries no mask field upstream. Player emotes do carry it.

  ## QA

  Test scene at **-31,19** (Test Scene → **Emote Mask** tab). Steps are also printed in the panel:

  1. **MASKED on** → Wave → move with the joystick — legs walk, arms wave, emote is *not* cancelled.
  2. **MASKED off** → Wave → move — emote cancels at once, log shows `INTERRUPTED`.
  3. **MASKED on** → Dance or Hammer → keep moving — loops until another emote or a stop.
  4. **MASKED on** → Clap or Snowfall — prop appears and animates, audio plays.
  5. Fire Wave and let it finish — log shows `STARTED` then `FINISHED`.
  6. Fire an emote, then fire another mid-playback — log shows `INTERRUPTED` (first) then `STARTED` (second), **in that order**.


Ludmilafantaniella (approved):
## ✅ Approve by QA

Tested on Android (Galaxy A35) and iOS (iPhone 16 Pro), using the Emote Mask test tab at -31,19.

- [x] MASKED on → Wave → move with joystick - legs walk, arms wave, emote not cancelled
- [x] MASKED off → Wave → move - emote cancels immediately, log shows INTERRUPTED
- [x] MASKED on → Dance/Hammer → keep moving - loops correctly until another emote or stop
- [x] MASKED on → Clap/Snowfall - prop appears, animates, and audio plays correctly
- [x] Fire Wave and let it finish - log shows STARTED then FINISHED
- [x] Fire an emote, then another mid-playback - log shows INTERRUPTED (first) then STARTED (second), correct order

Regression:
- Scene emotes Genesis Plaza
- Emotes (backpack, emote wheel)

https://github.com/user-attachments/assets/ce4bb41a-dc5c-4386-a776-393892651cb6

https://github.com/user-attachments/assets/ac4e1297-b393-4e07-80f1-6cbe841588f6

godot/src/decentraland_components/avatar/avatar_emote_controller.gd
```diff
@@ -858,6 +1009,25 @@ func set_teleport_grace() -> void:
 ## Process emote state. Cancel emote if player is moving (not idle).
 ## idle: true if player is not moving
 func process(idle: bool):
+	# Masked (upper-body) emotes coexist with locomotion, so movement never
+	# cancels them; only their natural end is tracked here — independent of idle.
+	if playing_masked:
```
regenesis-claw, line 1065: [P1] This is the only place a natural end is detected, but `Avatar._process()` returns before calling `process()` when `_lod_state == LODState.FAR` (and for `show_only_wearables` avatar shapes), and no teardown path calls `stop_emote()`.

So a remote avatar or NPC that goes far away mid-emote reports its terminal entry only once it comes back in range, and one that disconnects or is deleted mid-emote never reports one at all — a scene awaiting `ES_FINISHED`/`ES_INTERRUPTED` hangs forever, which is worse than the `setTimeout` it replaces.

Emitting `_emit_emote_finished(true)` on avatar teardown (and on entry to `LODState.FAR`, where playback is frozen anyway) would close it. If that's out of scope here, please state the limitation in the PR body and open a follow-up.

lib/src/dcl/crdt/message.rs
```diff
@@ -325,7 +325,12 @@ pub fn append_gos_component(
         return Err("Component not found".into());
     };
 
-    for i in 0..*elements_count {
+    // `to_binary` takes a reverse index (0 = newest entry), so iterate the count
+    // backwards to emit the new entries in chronological (append) order — with
+    // several appends in one tick (e.g. an emote's INTERRUPTED terminal entry
+    // followed by the next emote's STARTED) the scene must see them in the order
+    // they happened.
+    for i in (0..*elements_count).rev() {
```
regenesis-claw, line 333: [P2] Verified this is correct — `to_binary` computes `entry.len() - reverse_element_index - 1`, so reversing the range does emit the newest `elements_count` entries oldest-first.

It is also the widest-reaching change in the PR: `PointerEventsResult`, `TriggerAreaResult`, `VideoEvent` and `AssetLoadLoadingState` all change emission order, and none of them are covered by the QA pass or by a test. Could we add a unit test over `append_gos_component` asserting chronological order for `elements_count > 1`, plus a QA regression case on the video-player scene and a scene that produces several pointer results in one tick?

regenesis-claw (changes requested):
Solid, well-documented piece of work — the mask plumbing is consistent end-to-end (internal `-1/0` sentinel → component `AvatarMask` → wire `AvatarEmoteMask`), and `emote_mask.rs` carries unit tests for exactly the off-by-one that would otherwise bite on Unity interop. I independently verified three of the load-bearing claims:

- `GrowOnlySet::to_binary` really does take a **reverse** index (`entry.len() - reverse_element_index - 1`), so `(0..count).rev()` is the correct chronological order. The bug and the fix are both real.
- `GenericGrowOnlySetComponentOperation::get` returns the raw `VecDeque`, front-indexed, so the `onPlayerExpression` `start..len` fix is also correct.
- Upstream `decentraland/protocol` `avatar_emote_command.proto` already declares `optional EmoteState state = 5` with the same three values — the `build.rs` patch is wire-compatible with what will land on the pin bump.

Holding approval on one correctness gap plus two pre-flight items.

---

## Pre-flight

- **Branch conflicts with `main`** (`mergeStateStatus: DIRTY`). Per REVIEW.md §0 this has to be resolved before merge; CI here is green but was computed on a base from Aug 12.
- **iOS build: present and green** ✅ (`🍏 iOS → R2 (unsigned export)` SUCCESS, Android APK SUCCESS, static checks/clippy/Linux/Docker all green). Platform-sensitive change, requirement satisfied.
- QA sign-off from @Ludmilafantaniella on Android + iOS at this exact SHA — noted and not re-litigated.

---

## P1 — terminal `EmoteState` is not guaranteed for remote avatars / scene NPCs

The PR description says the terminal entry "covers the local player, remote avatars, and scene-owned `AvatarShape` NPCs". That holds only while the avatar stays near and on-screen.

`_emit_emote_finished()` for a natural end is only reachable from `AvatarEmoteController.process()`, and `Avatar._process()` returns early before calling it:

```gdscript
if _lod_state == LODState.FAR:
	return
...
if is_avatar_shape and avatar_data != null and avatar_data.get_show_only_wearables():
	animation_tree.active = false
	return
...
emote_controller.process(self_idle)
```

On top of that, nothing calls `stop_emote()` when an avatar is torn down — `freeze_on_idle()` (which does emit) is only used by the backpack preview, the scene-renderer tool and the client tests, not by the LOD/removal path.

Consequences for a scene that replaces its `setTimeout(clipLength)` with an `EmoteState` await (the stated purpose of #2655):
- A remote avatar or NPC that walks to `LODState.FAR` mid-emote reports its terminal entry only if/when it comes back into range — arbitrarily delayed.
- A remote avatar that disconnects, or an `AvatarShape` entity that is deleted, mid-emote **never** reports a terminal entry. The scene's await never resolves, which is strictly worse than the timeout it replaced.

Suggested fix: emit `_emit_emote_finished(true)` from the avatar teardown path (and on entry to `LODState.FAR`, since playback is frozen there anyway). If that's too invasive for this PR, the limitation needs to be stated in the PR body and in the docs for the field, plus a follow-up issue — silently unresolvable promises are the failure mode scenes will hit first.

---

## P2 — worth addressing, not blocking

1. **`godot/project.godot`: `boot_splash/show_image=true` removed** with no mention in the description. `show_image` defaults to `true` so it's behaviourally a no-op, but it's unrelated to emotes and looks like editor drift (REVIEW.md Tier 1 §5). Please restore the line.

2. **The `append_gos_component` order fix has the widest blast radius in the diff and zero coverage.** It changes emission order for `PointerEventsResult`, `TriggerAreaResult`, `VideoEvent` and `AssetLoadLoadingState` as well as emotes. I believe the fix is right, but the QA pass only exercised emotes. Please add a regression case for the video-player scene (the one the description suspects this broke) and for a multi-click-per-tick pointer-events scene, and/or a unit test over `append_gos_component` asserting chronological order for `elements_count > 1`.

3. **`MASKED_EMOTE_EXCLUDED_TRACKS` matches full track paths as exact strings.** Any clip whose armature/bone naming differs from `Armature/Skeleton3D:Avatar_*` — plausible for a scene-authored emote GLB — will have its hips and legs pass the filter and fight the locomotion state machine, with no diagnostic. Matching on the bone name after `:` (or asserting the clip has at least one excluded track when a mask was requested) would fail loudly instead of silently.

4. **Looping masked emotes re-fire on a polled edge.** The OneShot has to read `active == false` for one frame before `process()` re-fires it, and each re-fire replays the 0.3 s `fadein_time`, so a looping upper-body emote gets a blend seam once per iteration. Driving the loop from the `AnimationNodeAnimation`'s own loop mode would avoid the round-trip.

5. **`build.rs` declares `EmoteState` nested inside `PBAvatarEmoteCommand`; upstream declares it at file scope.** Wire-identical, as the comment says, but the generated Rust path changes from `pb_avatar_emote_command::EmoteState` to `sdk::components::EmoteState` when the `@dcl/protocol` pin bumps, breaking all five import sites. It's a loud compile error rather than a silent break, so this is fine — but declaring it at file scope in the patch would make the eventual pin bump a true no-op.

---

## Verified, no action needed

- `player_avatar_node` is assigned `player.avatar` in `explorer.gd`, so moving the `emote_triggered` wiring into `set_player_avatar_node` connects against a node that actually has both signals; the `is_connected` guard covers Explorer rebuilds.
- All `send_emote` call sites were updated (emote wheel + restricted actions); no stale 1-arg callers remain.
- `body.mask ?? -1` in `RestrictedActions.js` correctly preserves `AM_UPPER_BODY = 0` (a `||` there would have been a bug).
- `on_avatar_trigger_emote`'s `.expect("avatar not found")` replaced with a `let ... else` — good catch, that was reachable via the deferred signal.
- **Security:** no findings. No new user-controlled input reaches a shell, path, query, or deserializer; the mask sentinel is range-checked on both conversion directions and unknown wire values fall back to full body.

Requesting changes on the P1 (or on an explicit, documented decision to ship the gap with a follow-up issue) and the merge conflict.

regenesis-claw (approved):
## Re-review — delta at `27508b39` ("review fixes")

Every finding from the previous review is addressed, and two of the fixes are better than what I asked for. Approving the code.

|---|---|---|
| P1 | Terminal `EmoteState` never emitted when an avatar is torn down mid-emote | ✅ fixed (see below) |
| P2.1 | `project.godot`: `boot_splash/show_image=true` removed | ✅ restored |
| P2.2 | `append_gos_component` order change had no coverage | ✅ unit test added |
| P2.3 | Mask exclusion list matched full track paths as exact strings | ✅ now bone-name keyed + warning |
| P2.4 | Looping masked emote replayed `fadein_time` on every re-fire | ✅ loops inside the node |
| P2.5 | `EmoteState` nested vs upstream file-scope | ✅ now file-scope |

**P1.** `take_unfinished_emote()` is the right shape — handing the urn/mask back synchronously and clearing state in one call means the teardown path can't double-report, and it sidesteps the real problem (a deferred `emit_signal` on a dying node is dropped). Both teardown paths are covered: `AvatarScene::remove_avatar` for a departing peer, and the `new_value.is_none()` branch of `update_avatar_shape` for a deleted NPC. The ordering comment in `remove_avatar` is correct — `append_avatar_emote_command` resolves the active scenes through `avatar_godot_scene`, which is still populated at that point. The `AvatarShape` path pushes straight into `scene.avatar_scene_updates` rather than re-entering the scene runner, which avoids a re-borrow.

**P2.2.** The test asserts the property that matters (emitted offsets increase with append order) without re-implementing the CRDT parser, and it's generic over grow-only components rather than emote-specific. Linux (which runs `cargo test`) is green at this SHA, so it passes.

**P2.3.** Keying on the bone name after `:` is the fix; the `push_warning` when a masked clip excludes nothing is a genuinely better answer than what I suggested — it turns a silent no-op mask into a diagnosable one, and `push_warning` (not `push_error`) is the right level for a recoverable authoring problem, so no Sentry-quota cost. The `Armature_Prop` carve-out for colon-less tracks matches the existing convention in `_hide_all_props` and the prop-loading path.

**P2.4.** `use_custom_timeline` + `timeline_length = animation.length` + `LOOP_LINEAR` is the correct way to do this — and the comment explaining that `loop_mode` is only honoured with a custom timeline (whose length otherwise defaults to 1 s) is exactly the kind of Godot-quirk note that saves the next reader an hour. Dropping the loop branch out of `process()` follows: the OneShot now stays active for the lifetime of a looping masked emote, so only stop/supersede can end it.

**P2.5.** Also fixed a real latent bug I missed: `clear_patched_proto_root()` ran *after* the component loop, so it was deleting the patched `avatar_emote_command.proto` copy that loop had just written. Moving it above the loop is correct.

---

## Remaining — not blocking, but gate the merge on these

1. **Re-QA is required.** @Ludmilafantaniella's approval is against `7ed509b5`; the loop mechanism and the entire mask-filter predicate changed since. Cases 3 (looping masked emote) and 4 (prop + audio) exercise precisely the changed code. Two cases are worth adding while she's in there:
   - a remote peer that disconnects mid-emote → the scene receives `ES_INTERRUPTED` (the new P1 path, currently untested by anything);
   - the video-player scene and a scene producing several pointer results in one tick → confirms the `append_gos_component` reordering in the field, not just in the unit test.
2. **`_is_excluded_from_mask` excludes every colon-less track that isn't `Armature_Prop*`.** That's broader than the old exact-match list, which only excluded `"Armature"`. Emote audio is played through `play_emote_audio()` rather than an animation audio track as far as I can tell, so I don't think anything is affected — but case 4 ("audio plays") is the one that would catch it if a merged clip does carry a node-level audio or method track.
3. **LOD FAR still delays the terminal entry.** `Avatar._process()` returns before `emote_controller.process()` when `_lod_state == LODState.FAR`, so a remote avatar that walks far away mid-emote reports `ES_FINISHED` only once it comes back in range. The *loss* case is now closed by the teardown flush, so this is a delay rather than a hang — fine to ship, worth a sentence in the PR body so scene authors don't design around a tight deadline.
4. **PR body is now slightly out of date**: "the OneShot's filters are built from the clip's own tracks minus root motion and the lower body" describes the old path-matching behaviour, and the QA section doesn't mention the new regression cases.
5. **CI is still in flight** at `27508b39`: static checks, clippy, Linux and Docker are green; 🤖 Android and 🍏 iOS export are still building. Both were green on the previous head and this delta is GDScript + Rust with no platform branches, so I don't expect surprises — but per REVIEW.md §0 this shouldn't merge until the iOS artifact is green.

No security findings in the delta. Note that a `regenesis-claw` approval doesn't satisfy branch protection here — a human codeowner still needs to sign off.

kuruk-mm (approved):
great!

PR #2691, opened 2026-08-12
feat: two-finger pinch-to-zoom for in-world camera

### LINKED ISSUE #2636: Gesture-Based 3p/1p Camera Toggle (Pinch)
##  **⚠️ Problem:**
Pinch-to-zoom is the near-universal convention in mobile 3D games (Roblox, Minecraft mobile), but on mobile the only way to switch POV is the HUD toggle button that replaced pinch in #1336 — players attempt the gesture, get nothing, and stay in one camera. This is the same discoverability failure reported in #1224 and it suppresses `Playtime` by making the world harder to read and navigate.

## **🏁 Scope:**
Implement a two-finger pinch on the in-world camera that drives 3rd-person camera distance between clamped min/max, entering 1st person past the near clamp and returning to 3rd on pinch-out — the Roblox model. The existing camera-mode toggle button stays and remains in sync; out of scope: Backpack/avatar-preview zoom (#1943, #2164), FOV, desktop scroll-wheel, and any new HUD affordance.

## **📝 Deliverables:**
- [ ] Two-finger pinch recognized on the in-world camera and mapped to 3rd-person camera distance, continuous (not stepped)
- [ ] Distance clamped to defined min/max; pinch-in past the near clamp enters 1st person, pinch-out from 1st person exits back to 3rd person at the near clamp distance
- [ ] Camera-mode toggle button state stays in sync with gesture-driven POV changes (and vice versa)
- [ ] Gesture does not conflict with the joystick / camera-drag / scene-UI input arbitration from #2083 — pinching over scene UI must not move the camera
- [ ] Zoom level persists for the session and resets to the default 3rd-person distance on realm/scene change
- [ ] `CLICK_BUTTON` event on every completed zoom gesture is implemented
  - `event`: `CLICK_BUTTON` · `screen_name`: `IN_WORLD` · `button_text`: `ZOOM` · `extra_properties`: `{zoom_direction: 'zoom_in' | 'zoom_out'}`
- [ ] Verify the event lands in Metabase with correct properties before closing

## Metrics
1. **Target metric:** `Playtime`.
2. **Expected direction and size:** increase the share of sessions that change camera POV at least once. No baseline exists — the `CLICK_BUTTON` event above is what produces it, so the target is set after one release of data rather than guessed here.
3. **How it's measured:** Segment/Metabase `CLICK_BUTTON` where `button_text = ZOOM`, split by `zoom_direction`, expressed as % of DAU. Follows the event convention consolidated in #2377 and used for the Hide UI toggle in #2058.

## Dependency
- #2352 — Controls Customization: HUD + Social Panel Redesign - Iteration #1. The gesture shares the in-world camera input surface being reworked there; sequence accordingly.
- #2351 — HUD + Social Panel Redesign (design). Confirm the design does not introduce a HUD element that conflicts with a two-finger gesture zone.

## Notes
- Min/max distance values and the 1st-person entry threshold are not yet defined — needs a values pass during grooming or a short in-editor tuning session.
- Open question: should pinch in 1st person be a no-op, or swallowed entirely?

## **🔗 References:**
- [decentraland/godot-explorer#1336](https://github.com/decentraland/godot-explorer/pull/1336) — the PR that **removed** pinch-zoom camera-mode switching on mobile and replaced it with the HUD toggle button; this issue reintroduces pinch as continuous distance control that crosses into the mode toggle. Its test plan is the regression baseline.
- [decentraland/godot-explorer#1224](https://github.com/decentraland/godot-explorer/issues/1224) — Playtest #1 finding that 1st-person POV is undiscoverable; the original driver for #1336 and the reason a gesture affordance matters.
- [decentraland/godot-explorer#2083](https://github.com/decentraland/godot-explorer/pull/2083) — unified joystick + camera input through Godot's native `gui_input` with tree-order arbitration. Any new pinch handler must go through this path, not a manual hit-test.
- [decentraland/godot-explorer#2377](https://github.com/decentraland/godot-explorer/issues/2377) — the `SCREEN_VIEWED` / `CLICK_BUTTON` event convention (`screen_name` / `button_text` / `extra_properties`) this issue's analytics deliverable follows.
- No prior art found in `decentraland/unity-explorer` or `decentraland/docs` (searched "pinch", "zoom", "camera zoom", "third person camera", 2026-08-05) — desktop has no touch surface, so there is no parity target for this gesture.

sdilauro:
Closes #2636.

Reintroduces two-finger **pinch-to-zoom** on the in-world camera (mobile), replacing the discoverability gap left when #1336 swapped pinch for a HUD toggle. Pinch drives the third-person camera distance continuously and crosses into/out of first person — the Roblox model.

## Zoom model (`player.gd` / `camera_rig_helpers.gd`)
- A single continuous **`_zoom_level`** scalar spans the whole range. Values in `[THIRD_PERSON_MIN_DISTANCE, THIRD_PERSON_MAX_DISTANCE]` map straight to the SpringArm `spring_length`; below `MIN` it drops into first person. Mode crossings route through `Global.set_camera_mode()`, so the settings camera-mode dropdown (and the Rust camera) stay in sync both ways. A hysteresis band below `MIN` (`FIRST_PERSON_ZOOM_LEVEL`) stops the mode flickering at the boundary; re-entering third person lands exactly at the near clamp.
- Staying in third person writes the spring length **live** (no tween — the gesture is already continuous). Because that path kills the mode-change tween, it now **re-pins the over-shoulder `lateral_offset`** so a pinch that crosses into third person doesn't leave the avatar centered.
- Zoom persists within a scene and **resets to the default third-person view on a deliberate transition** — hooked to `Global.loading_finished` (Discover jump / teleport / realm change, i.e. the loading-screen path), never on a per-parcel walk. Skipped while a scene forces the camera mode; scene-forced `CameraModeArea` block/unblock snapshots and restores the zoom.

## Mobile input model (Roblox-style)
Two zones, arbitrated by which control owns each touch:

| Zone | 1st finger | 2nd finger |
|---|---|---|
| **Joystick** (`ActiveArea`, bottom-left **40% × 67%**) | walk | rotate camera |
| **Camera** (everything else — "free" touches) | rotate camera | **pinch-to-zoom** |

- **`MobileCameraInput`** is a full-screen catcher. Its own **`gui_input`** handles the *free* touches (single-finger look, pinch eligibility, chat-close, focus regrab, cinematic) — a touch that reaches it was consumed by no UI, so it is UI-safe by construction. It **also** taps the global **`_input`** for the two things `gui_input` can't see: the pinch tracking/recognition, and the **second finger inside the joystick's area** (which the joystick consumes) driving camera rotation.
- **Pinch candidates are restricted to free touches** (`_free_touches`, populated from `gui_input`). A pinch over the joystick, scene UI, HUD panels or open chat is therefore never recognized and never steals those drags.
- **Recognizer**: with no competing gesture in the free/camera area, a pinch is simply the two-finger spread changing past a threshold (~16 px) — no "both fingers must move" / axial-angle gate, so a thumb-anchored or angled pinch is caught. Candidates are re-evaluated on drag as well as press/release.
- The **joystick movement zone changed** from the old full-height left band to the Roblox bottom-left **40% × 67%** rect (frees the top-left + right for the camera). Note this is a touch-zone change for *all* mobile users, not only the new gesture.

## Robustness / QA fixes
- **Focus-out cleanup** — all touch state is cleared on `NOTIFICATION_APPLICATION_FOCUS_OUT`.
- **Releases processed first, unconditionally** — a touch-up during cinematic/pointer mode (or before the player resolves) can't strand a phantom finger that would later confirm a spurious pinch.
- **Zoom↔mode self-heal** — `begin_pinch_zoom()` reconciles `_zoom_level` with the actual camera mode, so the QA "pinch stops responding until you toggle the POV dropdown" state can't persist across gestures.
- **Joystick render offset** — the drawn thumbstick is converted from `ActiveArea`-local into the shader's space, so it tracks the finger after the zone was re-anchored.
- **Floor guard** — some scene ground meshes have no usable collider (single-sided shell / `cmask=0`), so the collision-sweep casts slip through and the camera dips below the visible floor at far zoom (confirmed via `DCL_CAM_DEBUG` → `NO-COLLIDER on camera path`; the orange Genesis Plaza floor clips, the cobblestone doesn't). The camera is now kept above the **player's own ground contact** (its `CharacterBody3D` origin), independent of scene geometry — additive, only ever raises the camera.

## Analytics
Every completed gesture with a net change emits `CLICK_BUTTON` · `screen_name: IN_WORLD` · `button_text: ZOOM` · `extra_properties: {zoom_direction: 'zoom_in' | 'zoom_out'}` via `Global.metrics.track_click_button`, on release, only when the final zoom differs from the start (net-zero gestures emit nothing).

## Known follow-ups
- No unit test yet for the pinch recognizer (would need extracting the spread math into a pure helper à la `CameraRigHelpers`).
- Analytics event not yet verified in Metabase (needs a build).

## Test plan (device only — iOS/Android)
Setup: open the app, log in, enter a world in Genesis City, landscape.

1. **Zoom out/in** — two fingers in the camera area, spread apart → camera pulls out; pinch together → pulls in. **Pass:** smooth, not stepped; stops at the far limit without running away.
2. **Enter/exit first person** — pinch fully in → drops to first person (avatar hides); pinch back out → returns to third person near the avatar. **Pass:** no flicker pinching back and forth across the boundary.
3. **Walk + camera together** — walk with the left thumb while spreading two fingers in the camera area. **Pass:** the avatar keeps walking *and* the camera zooms (independent zones).
4. **Walk-left + rotate-right at once** — left thumb pushes the joystick left while a right finger drags the camera right. **Pass:** it walks and rotates; it is **not** read as a pinch.
5. **Second finger in the joystick area** — one thumb walking, a second finger in the joystick rect dragging. **Pass:** the camera rotates (no zoom).
6. **UI arbitration** — pinch/drag on top of an interactive scene-UI panel, a HUD panel, or with chat open. **Pass:** the camera does not move; the UI/tap behaves normally.
7. **Scene-change reset** — zoom out (or go first person) → Discover-jump to another scene → after the loading screen you're in the default third-person view.
8. **Persistence** — zoom, then walk around the *same* scene → zoom stays.
9. **Settings sync** — the camera-mode dropdown matches the current POV; changing it there moves the camera.
10. **Recovery** — background the app mid-pinch and return → single-finger look still works; pinch still responds (no stuck state).
11. **Wall + floor collision** — camera doesn't clip through walls at any zoom, and stays above the floor at max zoom-out even on scene ground with no collider (e.g. the orange Genesis Plaza floor).
12. **Floor-guard edge cases (known tradeoff)** — the floor guard references the player's own feet (scene floors often have no collider to measure against), so it can't tell "legitimately below the feet plane over a drop" from "clipping the floor on flat ground". At max zoom over a **raised ledge / top of a staircase**, pitching fully up can pin the vertical framing (feet + 0.15 m) instead of tracking further down — a **framing stall, not a clip**. Stand at the edge of a raised platform at max zoom, pitch up, then walk down a staircase. **Pass:** no floor clip; a slight framing stall / step-wise bump over the drop is acceptable.
13. **Analytics** — with `--rust-log=dclgodot::analytics=debug,warn`, a zoom gesture logs a `ZOOM` / `IN_WORLD` Click Button with `zoom_direction`; a no-change gesture logs nothing.

> Given the input path changed substantially since the last QA pass, a **fresh QA on the current APK** is recommended over a delta review.


godot/src/logic/player/player.gd
```diff
@@ -254,6 +288,13 @@ func _on_player_profile_changed(new_profile: DclUserProfile):
 func _on_scene_changed(_scene_id: int) -> void:
 	_locomotion_settings = Global.scene_runner.get_current_scene_locomotion_settings()
 	_apply_locomotion_settings()
+	# On mobile you can't walk between scenes — every scene entry goes through a
+	# loading screen (Discover jump, teleport, realm change), so a scene change is
+	# a deliberate transition and the natural point to reset the pinch zoom back to
+	# the default third-person view. Desktop crosses parcels by walking and has no
+	# pinch input, so it's left untouched.
+	if Global.is_mobile():
+		_reset_zoom_to_default()
```
regenesis-claw, line 312: [P1] `on_change_scene_id` is not a loading-screen signal. `scene_manager.rs` emits it from `on_current_parcel_scene_changed()`, which fires whenever `current_parcel_scene_id` differs from the previous frame's — and `compute_scene_distance()` recomputes that id every frame from the avatar's world position → parcel (it even resets to `SceneId::INVALID` first, so stepping onto an unloaded parcel counts).

So on mobile this fires while simply **walking across a parcel boundary** in Genesis City. Two effects:

- the zoom the user just set snaps back to the default a few steps later;
- `_reset_zoom_to_default()` calls `Global.set_camera_mode(THIRD_PERSON)`, so a user in first person — including one who chose it from the Settings dropdown, unrelated to this feature — is force-yanked to third person by walking into the next parcel.

Tie the reset to an actual deliberate transition (teleport / Discover jump / realm change — the loading-screen path) rather than the per-frame parcel scene id. If a per-scene reset is still wanted, at least leave the camera *mode* alone and only restore the distance when already in third person.

godot/src/logic/player/player.gd
```diff
@@ -152,8 +182,12 @@ func set_camera_mode(mode: Global.CameraMode, play_sound: bool = true):
 		if play_sound:
 			UiSounds.play_sound("ui_fade_out")
 	elif mode == Global.CameraMode.FIRST_PERSON:
+		# Park the zoom scalar below the min so a pinch-out has to cross the
+		# hysteresis band before it returns to third person.
+		_zoom_level = CameraRigHelpers.FIRST_PERSON_ZOOM_LEVEL
```
regenesis-claw, line 195: [P2] This also runs on the scene-forced path: `_on_camera_mode_area_detector_block_camera_mode()` calls `set_camera_mode(forced_mode, false)`, so entering a first-person `CameraModeArea` parks `_zoom_level` at `FIRST_PERSON_ZOOM_LEVEL`. On unblock, `set_camera_mode(stored_camera_mode_before_block, false)` sees `_zoom_level < THIRD_PERSON_MIN_DISTANCE` and falls back to the default distance — the user's zoom is silently lost after crossing a trigger area.

Consider snapshotting the zoom next to `stored_camera_mode_before_block` and restoring it on unblock.

godot/src/logic/player/mobile_camera_input.gd
```diff
@@ -88,23 +95,66 @@ func _handle_touch(event: InputEventScreenTouch) -> void:
...
+			if not _pinch_active:
+				_begin_pinch()
 	else:
 		_touch_positions.erase(event.index)
 		if event.index == _drag_index:
 			_drag_index = -1
 		if _touch_positions.size() < 2:
 			_two_fingers = false
+		if _pinch_active and (event.index == _pinch_index_a or event.index == _pinch_index_b):
+			_end_pinch()
```
regenesis-claw: [P2] Ending the pinch only on the two captured indices leaves the state stuck if a release event never arrives (app backgrounded mid-gesture, OS touch cancel): `_pinch_active` stays true with a stale entry in `_touch_positions`, and because `_handle_drag()` now early-returns for *any* index while `_pinch_active`, single-finger camera look is dead until a matching index happens to be pressed and released again. A stale position also makes the next `_update_pinch()` distance jump, i.e. a large spurious zoom.

Same guard covers the 3-finger case: lift one pinch finger with three down and the two remaining fingers can never start a new pinch, since `_begin_pinch()` is only reachable from the press branch.

Suggest ending unconditionally on the count instead:

```suggestion
		if _touch_positions.size() < 2:
			_two_fingers = false
			if _pinch_active:
				_end_pinch()
		elif _pinch_active and (event.index == _pinch_index_a or event.index == _pinch_index_b):
			_end_pinch()
```

regenesis-claw (changes requested):
Nice shape overall — a single continuous `_zoom_level` scalar that spans both modes is the right model, routing mode crossings through `Global.set_camera_mode()` keeps the settings dropdown and the Rust camera in sync, the `camera_mode_change_blocked` guard on both `apply_pinch_zoom()` and `_reset_zoom_to_default()` correctly respects scene-forced camera modes, and the constant-ordering unit test is a good touch. Desktop is unaffected (`_zoom_level` initializes to `THIRD_PERSON_CAMERA.z`, which is exactly what `rig_targets(true).spring_length` returned before). CI is fully green including the iOS export job, so REVIEW.md §0 pre-flight is satisfied and the branch is up to date with `main` (0 behind).

One blocker, on the scene-change reset.

---

### 🔴 P1 — the reset premise is wrong: `on_change_scene_id` fires while walking

`godot/src/logic/player/player.gd:296` (inline comment below).

The PR description says *"On mobile you can't walk between scenes — every scene entry goes through a loading screen"*. That isn't what this signal means. `on_change_scene_id` is emitted from `on_current_parcel_scene_changed()` in `lib/src/scene_runner/scene_manager.rs`, which runs whenever `current_parcel_scene_id != last_current_parcel_scene_id`; `current_parcel_scene_id` is recomputed **every frame** in `compute_scene_distance()` straight from the avatar's world position → parcel. So it fires every time the player walks across a parcel boundary into a different deployed scene in Genesis City — no loading screen involved. It also resets to `SceneId::INVALID` at the top of that function, so walking into a parcel whose scene isn't loaded (roads, empty parcels) fires it too.

Two consequences on a phone in Genesis City:

1. The zoom the user just set snaps back to the default a few steps later — which contradicts the "zoom persists within a scene" deliverable in spirit (the user reads "scene" as "the place I'm in", not "this parcel deployment").
2. Worse: `_reset_zoom_to_default()` force-calls `Global.set_camera_mode(THIRD_PERSON)`. A user who is in first person — whether via pinch **or** via the Settings → camera-mode dropdown — gets yanked out to third person just by walking into the next parcel. That's a regression on an existing, unrelated feature (the dropdown), not just on the new gesture.

Hook the reset to an actual deliberate transition instead: teleport / Discover jump / realm change (the loading-screen path), e.g. a realm-change or loading-screen signal, or gate on "the scene entrypoint changed" rather than the per-frame parcel scene id. If a per-scene reset is genuinely wanted, at minimum don't change the camera *mode* — only restore the third-person distance when already in third person, and leave a deliberate first-person choice alone.

---

### 🟡 P2 — findings

**1. Scene-forced camera-mode areas silently destroy the persisted zoom.** `player.gd:187` (inline). Walking into a `CameraModeArea` that forces first person runs `set_camera_mode(forced_mode, false)`, which parks `_zoom_level` at `FIRST_PERSON_ZOOM_LEVEL`. On unblock, `set_camera_mode(stored_camera_mode_before_block, false)` hits the `_zoom_level < THIRD_PERSON_MIN_DISTANCE` branch and resets to the default distance. The user's zoom is gone after crossing a first-person trigger area. Consider snapshotting `_zoom_level` next to `stored_camera_mode_before_block` in `_on_camera_mode_area_detector_block_camera_mode()` and restoring it on unblock.

**2. Stale pinch state if a touch release is lost.** `mobile_camera_input.gd:107` (inline). The end condition is keyed on the two captured indices; a dropped release (app backgrounded mid-gesture, OS touch cancel) leaves `_pinch_active` true with a stale entry in `_touch_positions`, and since `_handle_drag()` now returns early for *any* index while `_pinch_active`, single-finger camera look is dead until a matching index is pressed and released again. Also covers the 3-finger case: lift one pinch finger with three down and the two remaining fingers can't start a new pinch (`_begin_pinch()` is only reachable from the press branch).

**3. Boundary crossings re-fire the mode sound and pop the spring length.** Each crossing goes through `Global.set_camera_mode()` → `set_camera_mode(mode)` with `play_sound` defaulting to `true`, so a pinch that oscillates around the boundary retriggers `ui_fade_in`/`ui_fade_out`. And on exiting first person mid-gesture, the 0.25 s tween is started and then immediately killed by the next `apply_pinch_zoom()` frame, which writes `spring_length` directly — so the camera jumps from `-0.2` to `1.5` in one frame instead of easing. Both are arguably acceptable, but worth a device pass; if the pop reads badly, ease the first exit frame.

**4. `PINCH_ZOOM_SENSITIVITY` is meters-per-raw-pixel**, so the physical gesture feel scales with screen density — a 3x-DPI phone needs ~3x the finger travel of a 1x tablet for the same zoom. `HORIZONTAL_SENS`/`VERTICAL_SENS` in the same file have the same property, so this matches existing precedent; flagging only so the tuning pass happens on both a high- and a low-DPI device rather than one.

**5. Test plan format.** REVIEW.md §4 asks for QA-executable cases (setup only when non-obvious → numbered steps starting from opening the app → an observable expected result). The current list is a good coverage checklist but reads as reviewer notes: e.g. *"Pinch back and forth around the 1st/3rd boundary → no flicker between modes"* has no starting point and no pass/fail line QA can tick without reading the code. Worth reshaping the ~4 highest-risk cases (boundary crossing, scene-change reset, scene-UI arbitration, joystick + pinch concurrency) into that format before it goes to QA.

---

### Unverified

Every box in the test plan is unchecked, and the analytics deliverable is explicitly marked pending a build — so the device behavior (gesture feel, no-flicker at the boundary, the scene-UI arbitration claim, the `ZOOM`/`IN_WORLD` event landing in Metabase) is author-stated but not demonstrated in this PR. The `gui_input` arbitration argument is sound from reading the code, but it's inherited reasoning from #2083, not something this diff proves.

**Security:** no security issues found — no new input parsing, network, auth, or persistence surface; the only new outbound data is a `zoom_direction` enum string in an existing analytics event.

Requesting changes on the P1. The rest are follow-up-friendly.

godot/src/logic/player/player.gd
sdilauro: Fixed in b06d5887. Moved the reset off `on_change_scene_id` onto `Global.loading_finished`, which only fires behind a loading screen (Discover jump / teleport / realm change) — never when walking across a parcel boundary. `_on_scene_changed` keeps only its locomotion work; the reset now lives in `_on_loading_finished` (still mobile-gated).

Verified on device: walking across parcels no longer resets the zoom or changes the camera mode; a Discover jump between scenes in Genesis City still resets to the default third-person view after the loading screen.

godot/src/logic/player/player.gd
sdilauro: Fixed in b06d5887. Snapshotted the zoom into `stored_zoom_level_before_block` next to `stored_camera_mode_before_block` on block, and restore it *before* `set_camera_mode` reads it on unblock — so the user's third-person distance survives a scene-forced first-person `CameraModeArea` instead of falling into the `< THIRD_PERSON_MIN_DISTANCE` default fallback.

Verified on device: entering a game that forces a cinematic/first-person camera and then leaving restores the previous camera state.

godot/src/logic/player/mobile_camera_input.gd
sdilauro: Fixed in b06d5887. The pinch now ends on `_touch_positions.size() < 2` (not on a specific captured index), so a lost release / OS touch-cancel can no longer strand `_pinch_active`.

I didn't apply the suggestion verbatim: instead of `_end_pinch()` in the `elif` branch, it calls `_sync_pinch()`, which re-seats the gesture onto the two surviving fingers — so lifting one finger of a 3-finger touch keeps the pinch alive and lets it restart, rather than ending it. Also added a safety-net in `_handle_drag` that ends the pinch if the count ever drops below two mid-drag, so single-finger look always recovers.

godot/src/logic/player/mobile_camera_input.gd
```diff
@@ -88,23 +96,78 @@ func _handle_touch(event: InputEventScreenTouch) -> void:
...
+			_sync_pinch()
 	accept_event()
 
 
 func _handle_drag(event: InputEventScreenDrag) -> void:
 	_touch_positions[event.index] = event.position
+	# Safety net: if a pinch finger's release was lost (app backgrounded / OS
+	# cancel) the count falls below two while _pinch_active lingers — end it here
+	# so a single finger can look around again instead of being swallowed.
+	if _pinch_active and _touch_positions.size() < 2:
```
regenesis-claw: [P2] This net can't fire for the case the comment describes. `_touch_positions` is only ever written in two places: set on press/drag, and erased in the release branch — and that release branch already calls `_end_pinch()` at the `size() < 2` check above. Since line 116 *inserts* `event.index` before this check, the count here can never be lower than it was on entry, so `_pinch_active and size() < 2` is unreachable in practice.

The real lost-release case also isn't covered by it: if a release event never arrives (app backgrounded, OS touch cancel), the stale index is never erased, so `_touch_positions.size()` stays at 2 and this guard sees nothing wrong. The next press then re-seats onto a stale coordinate via `_sync_pinch()` and the first `_update_pinch()` reads a large bogus spread.

The thing that would actually close it is clearing the touch state on focus loss, e.g.:

```gdscript
func _notification(what: int) -> void:
	if what == NOTIFICATION_APPLICATION_FOCUS_OUT:
		_touch_positions.clear()
		_two_fingers = false
		_drag_index = -1
		if _pinch_active:
			_end_pinch()
```

Not blocking — the stale-`_touch_positions` behavior predates this PR (`_two_fingers` had the same exposure) and it needs a backgrounding mid-gesture to hit. Worth either adding the handler or softening the comment so it doesn't promise a guarantee the code doesn't provide.

regenesis-claw (approved):
Delta re-review of `33de4fb → b06d588` ("fix: reset pinch zoom on real transitions, not parcel crossings").

### Previous findings

| # | Finding | Status |
|---|---------|--------|
| P1 | Reset hooked to `on_change_scene_id`, which fires on parcel-boundary crossings while walking | ✅ Fixed |
| P2-1 | Scene-forced `CameraModeArea` silently destroys the persisted zoom | ✅ Fixed |
| P2-2 | Pinch state stuck / un-restartable when a tracked finger lifts | ⚠️ Mostly fixed (see below) |
| P2-3 | Mode-crossing sound + `spring_length` pop when exiting FP mid-gesture | ➖ Not addressed (was advisory — fine to leave for the device pass) |
| P2-4 | `PINCH_ZOOM_SENSITIVITY` is DPI-dependent | ➖ Not addressed (matches existing `HORIZONTAL_SENS` precedent — noted only for tuning) |
| P2-5 | Test plan not in REVIEW.md §4 QA-executable format | ➖ Not addressed |

**P1 — correctly fixed.** `Global.loading_finished` is emitted from exactly one place, `async_hide_loading_screen_effect()` in `loading_screen.gd`, so it fires only when a loading screen is dismissed — i.e. Discover jump, teleport, realm change, and initial world entry. That is precisely the "deliberate transition" semantics the reset wanted, and it can no longer fire from ordinary parcel-to-parcel walking. The replacement comment on `_on_loading_finished()` documenting *why* `on_change_scene_id` was the wrong signal is a good addition — it stops the next person from re-deriving it.

**P2-1 — correctly fixed.** `stored_zoom_level_before_block` is captured under the same `if !camera_mode_change_blocked` guard as the mode (so a re-entrant block can't clobber it), restored *before* `set_camera_mode()` reads `_zoom_level` on unblock, and initialized to `THIRD_PERSON_CAMERA.z` so an unblock with no prior block is still sane. The FIRST_PERSON restore path overwriting it with the FP band value is correct — the user was in FP, so the scalar belongs there.

**P2-2 — the 3-finger half is fixed, the lost-release half isn't.** Re-seating onto the surviving pair via `_sync_pinch()` is a nicer solution than what I suggested — the gesture keeps running instead of dying, and the `if not _pinch_active` guard keeps `begin_pinch_zoom()` (and therefore the analytics baseline) from firing again on a re-seat. See the inline note for the residual: the drag-time safety net can't actually reach the case its comment describes.

---

### CI

Both red jobs are runner/infra failures unrelated to this diff, which is GDScript-only:

- **🤖 Android** — `error sending request for url (…/sentry-godot-1.6.0+4e3e3e5.zip): http2 error: stream error received: refused stream` — transient GitHub download flake.
- **🍏 iOS export** — `No Xcode >= 26 found on the runner`, then `protoc failed: … avatar_emote_command.proto: Input is shadowed in the --proto_path by …/lib/target/release/build/dclgodot-*/out/proto_patched/…` — a stale build directory on a dirty self-hosted macOS runner, plus a missing toolchain.

Neither can be produced by a change that touches only two `.gd` files, and the previous head (`33de4fb`) built green on **both** platforms with the same Rust/proto tree. Static checks, Gate, and Clippy are green on this head; Linux was still running at review time. Worth a re-run to get a clean board before merge, but nothing here blocks on code grounds.

---

### Residual P2s (non-blocking)

1. The `_handle_drag` safety net is unreachable — inline below.
2. `_reset_zoom_to_default()` still force-calls `set_camera_mode(THIRD_PERSON)`, so a user who picked first person from the Settings dropdown is put back into third person after any teleport/Discover jump. That is far more defensible now that it only happens behind a loading screen (and camera mode is deliberately non-persisted runtime state anyway), so I'm not blocking on it — but if QA reports it as surprising, restricting the reset to the *distance* while leaving the mode alone is the smaller behavior.
3. Test plan is unchanged, and the plan's own bullet *"First person → jump somewhere via Discover → after loading you're back in default third person"* is now the single most important case to actually execute, since it's the one the P1 fix rewired. Every box is still unchecked and the analytics deliverable is still marked pending a build — device behavior and the Metabase event remain author-stated, not demonstrated.

**Security:** no issues found — no new input parsing, network, auth, or persistence surface.

Approving. The blocker is genuinely resolved and the two substantive P2s were addressed properly; the rest are device-tuning and QA-format notes that don't need another round.

godot/src/logic/player/mobile_camera_input.gd
sdilauro: Fixed in 72176b35. Added the `_notification(NOTIFICATION_APPLICATION_FOCUS_OUT)` handler that clears `_touch_positions` / `_two_fingers` / `_drag_index` and ends any active pinch on focus loss, and removed the unreachable `_handle_drag` safety-net (which, as you noted, could never fire since the index is inserted before the count check).

Already tested on device before pushing: backgrounding the app mid-pinch and returning no longer strands `_pinch_active` or produces a bogus zoom jump — single-finger look recovers cleanly. Working as expected.

Ludmilafantaniella (commented):
## 🐛 With Issues

Tested on Android and iOS. Happy path works well overall:

**✅ Working correctly**
- [x] Smooth continuous zoom in/out, no stepping
- [x] Settings dropdown syncs correctly with pinch-driven POV changes
- [x] Camera stops correctly at the far limit, no runaway
- [x] No flicker at the 1st/3rd person boundary
- [x] Teleporting while in first person correctly returns to third person
- [x] Zoom persists correctly within the same scene

**🐛 Issues found (reproduced on both Android and iOS)**

1. **Pinch gesture gets stuck, blocking further zoom in/out** - reproduced intermittently, sometimes after reaching max zoom-out, sometimes just while sitting in third person with no clear trigger. Once stuck, neither pinch-in nor pinch-out responds. The only way to unstick it is going to Settings → Gameplay and toggling the POV dropdown manually.

2. **Camera clips through the floor** - happens at max zoom-out, when moving the camera angle toward the ground.

Both feel related to the pinch state getting desynced (similar territory to the stale touch-state issues fixed earlier in review - focus-loss handling, lost release events). Given how core the "camera doesn't clip through walls at any zoom" and "smooth zoom" requirements are to this feature, and the gesture is unrecoverable without a workaround, flagging this as blocking rather than a minor note.

Happy to try to narrow down a more consistent repro if useful (e.g. specific finger movement pattern, gesture speed, or app state before it happens).

cc @sdilauro 

Image

https://github.com/user-attachments/assets/b4753547-f578-4766-9aa7-befc11330491

regenesis-claw (dismissed):
Review retracted at the author's request — feedback moved to a private channel.

sdilauro:
Thanks a lot for the thorough pass @Ludmilafantaniella 🙏 Both issues are fixed now — here's what was going on and what changed:

**1. Pinch gets stuck / stops responding (only Settings → POV toggle recovers it)**
Two root causes, both addressed:
- A **zoom ↔ camera-mode desync**: the continuous `_zoom_level` scalar could end up parked in the first-person band while the camera was actually in third person (or vice-versa), a state where `_apply_zoom_level()` stops writing the spring length and the gesture reads as frozen. `begin_pinch_zoom()` now **reconciles the zoom level with the real camera mode at the start of every gesture**, so the state can't persist across gestures (no more Settings-toggle workaround). The toggle "worked" only because it reset `_zoom_level` as a side effect.
- **Stale touch state**: a dropped release (backgrounding, OS touch-cancel, or a touch-up during cinematic/pointer mode) left a phantom finger that a single real finger could then confirm into a spurious pinch. Touch state is now cleared on focus-out, and releases are processed unconditionally (before any early-return), so a phantom can't linger.

**2. Camera clips through the floor at max zoom-out**
Diagnosed with the built-in `DCL_CAM_DEBUG` probe — at the clip it logs:
```
[CamDebug] NO-COLLIDER on camera path (single-sided shell or cmask=0)
```
So it's **scene content**: some Genesis Plaza ground meshes have no usable collider (single-sided / `cmask=0`) — the **orange floor** clips, the **cobblestone path** (which has a proper collider) corrects fine. The camera's collision clamp is raycast-based, so if there's nothing to hit, it can't stop the camera — and the extended max zoom (6 m) is what makes that region reachable now.

Fix: a **floor guard** independent of scene geometry — the camera is kept above the **player's own ground contact** (the `CharacterBody3D` origin, which rests on whatever it's standing on), with a small clearance. Additive (only ever raises the camera), so wall handling is untouched. Verified on device against the orange floor.

The input path also changed quite a bit since your pass (pinch is now camera-area / free-touch only, the joystick zone shrank to the bottom-left, second-finger-in-joystick drives the camera) — details in the updated PR description. Given all that, a **fresh QA pass on the latest APK** would be ideal over a delta check. Thanks again!

leanmendoza:
Waiting for design definition

kuruk-mm:
Overseed by https://github.com/decentraland/godot-explorer/pull/2878

PR #2694, opened 2026-08-12
fix(xtask): retry flaky dependency downloads in CI

### LINKED ISSUE #2695: CI: dependency install fails on transient CDN errors (http2 refused stream)
CI jobs die on the very first network hiccup while installing dependencies. `cargo run -- install` downloads protoc, the Godot editor/templates, `@dcl/protocol` and the V8 bindings with **zero retries**, so a single refused HTTP/2 stream from GitHub's CDN fails the whole build job:

```
=== Installing Protocol Buffers Compiler ===
ℹ️  Downloading: https://github.com/protocolbuffers/protobuf/releases/download/v23.2/protoc-23.2-linux-x86_64.zip
Error: error sending request for url (...): http2 error: stream error received: refused stream before processing any application logic
❌ Failed: error sending request for url (...)
```

This is intermittent and unrelated to the change under test — it hit the 1.12.1 release-candidate PR several times, and most recently https://github.com/decentraland/godot-explorer/actions/runs/31644197256/job/94273983150. Every re-run costs a full build cycle.

Related sharp edges in the same code path:

- A connection dropped mid-body hits `response.chunk().await.unwrap()` and panics instead of returning an error.
- Response status is never checked, so a 404 writes the HTML error page into `tmp-file.zip` / the V8 binding file and fails later with a confusing zip/parse error.
- A failed download leaves a truncated file behind, which the next run treats as a finished download (`android_deps.zip`, the persistent download cache).

**Expected:** transient network failures are retried with backoff and CI only fails when the artifact is genuinely unavailable.

kuruk-mm:
CI builds keep dying on the very first network hiccup while installing dependencies — `error sending request ... http2 error: stream error received: refused stream before processing any application logic` while downloading protoc, which hit the 1.12.1 RC several times ([latest example](https://github.com/decentraland/godot-explorer/actions/runs/31644197256/job/94273983150)). `cargo run -- install` had zero retries: one refused HTTP/2 stream from GitHub's CDN and the whole job failed. This makes every download in xtask retry up to 5 times with exponential backoff (2s → 4s → 8s → 16s) and forces HTTP/1.1, which is what the CDNs actually flake on (multiplexing buys nothing for single large file downloads anyway). It also fixes three related sharp edges found along the way: a dropped connection mid-body used to `panic!` instead of erroring, non-2xx responses were never checked (a 404 happily wrote an HTML error page into `tmp-file.zip` / the V8 binding file), and a failed download left a truncated file behind that the next run would treat as a finished download (`android_deps.zip`, the persistent download cache).

Closes #2695

## Changes

- `src/download_file.rs` → retry helper with exponential backoff + HTTP/1.1-only clients; HTTP status check (retries 5xx/408/429, fails fast on other 4xx so a genuinely wrong URL still errors in <1s); no more `unwrap()` on the body stream; partial file removed on failure. Adds `http_get_bytes()` for blocking one-shot GETs.
- `src/install_dependency.rs` → the `@dcl/protocol` npm manifest + tarball requests go through the retrying client.
- `src/run.rs` → the V8 binding download (Android/iOS builds, also from GitHub releases) uses `download_file` instead of a bare `curl -L`, so it retries and no longer silently saves an error page as the binding.

## Test plan

No QA needed — no runtime behavior change, this is build tooling only.

Verified locally with a throwaway test module against the real endpoints (removed before commit):

- [ ] Real GitHub release asset (the exact protoc URL from the failing job) downloads over forced HTTP/1.1 — 2,916,347 bytes in 0.65s.
- [ ] 404 on a nonexistent release asset fails in 0.31s with `unexpected HTTP status 404 Not Found ...` and leaves no partial file (no pointless 30s of backoff).
- [ ] Unreachable host retries 5× with the 2/4/8/16s backoff, printing a `⚠️ ... Retrying in Ns...` line per attempt, then fails with `... failed after 5 attempt(s)`.
- [ ] `http_get_bytes` against the npm registry returns the `@dcl/protocol@next` manifest.
- [ ] CI on this PR: all `cargo run -- install` steps (linux/macos/windows/android/ios/clippy/coverage) pass.

### DIFF
--- a/src/download_file.rs
+++ b/src/download_file.rs
@@ -1,8 +1,18 @@
-use crate::ui::create_download_progress;
+use crate::ui::{create_download_progress, print_message, MessageType};
 use reqwest::Url;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
+use std::time::Duration;
+
+/// Attempts (first try + retries) for any single HTTP operation.
+pub const MAX_HTTP_ATTEMPTS: u32 = 5;
+/// Backoff before each retry: 2s, 4s, 8s, 16s.
+const RETRY_BASE_DELAY: Duration = Duration::from_secs(2);
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(60);
+/// Total timeout for the small blocking requests (npm manifest, protocol tarball).
+/// Large file downloads go through `download_file`, which has no total timeout.
+const BLOCKING_REQUEST_TIMEOUT: Duration = Duration::from_secs(300);
 
 enum DownloadEvent {
     TotalSize(u64),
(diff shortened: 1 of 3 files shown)


leanmendoza (approved):

EibrielInv (approved):
Looks good!

PR #2699, opened 2026-08-13
fix: avatar animation state machine traps + emote/glider prop toon shading

### LINKED ISSUE #2581: [Bug] Emote GLTF props render unshaded instead of using avatar toon material
## Description

The GLTF props attached to emotes (e.g. items held/spawned during an emote animation, such as the dog and the computer seen in the reference screenshots) are rendering **unshaded**, instead of using the same toon/avatar material as the rest of the avatar.

This causes a visible mismatch — the props pick up flat/unshaded colors (e.g. tinted by the sky/ambient light) instead of matching the avatar's shaded, toon-shaded appearance. Follow-up from thread discussion where "the dog should be white" and "also the computer" were called out as looking wrong in-scene.

## Expected Behavior

Emote GLTF props should use the same material/shader pipeline as the avatar (toon-shaded), so their appearance is visually consistent with the rest of the avatar instead of appearing flat/unshaded.

## Platform

Mobile (Android/iOS) — reported from in-scene screenshots.

## Related

- Related lighting/color context: #2516 (mobile day/night lighting over-orange), #998 (color imbalance / white balance)

---
Requested by Mateo via Slack

### LINKED ISSUE #2603: [Bug] Using an Action while Gliding stops gliding animation and triggers walking mid-air
### Platform
Mobile (device/OS not specified)

### App Version
v1.11.0

### Device Information
Iphone 16E

### Issue description
Using an Action while Gliding stops the gliding animation and triggers the walking animation mid-air, even though the player is still airborne and not touching the ground.

### Expected behavior
The gliding animation/state should persist while airborne after triggering an Action; the walking animation should not play mid-air.

### Screenshots/Media

https://github.com/user-attachments/assets/06bc3434-cfd1-4303-97f3-4eb744a48380

Image

### Steps to reproduce
1. Start gliding
2. While airborne and gliding, trigger an Action
3. Observe the gliding animation stops and the walking animation plays mid-air

### Occurrence
Confirmed by Rodri in Slack thread.

---
Requested by Rodri via Slack

manuelmaceira:
Closes #2603
Closes #2581

## #2581 — Emote GLTF props render unshaded

Emote props (`Armature_Prop`) and the glider prop kept the GLTF import pipeline's `PER_VERTEX` material — wearables get converted to the avatar toon shader in `avatar.gd`, these never did.

- New `AvatarMeshAssembler.apply_toon_material_recursive()`
- Applied to emote props on attach (`avatar_emote_controller.gd`) and to the glider prop on `_ready` (`avatar.gd`)

## #2603 — Action while gliding → walk/idle mid-air

Several stacked causes, all in the avatar animation stack:

- **Locomotion conditions unguarded**: `Idle -> Walk` beat `Idle -> Jump_Fall` while airborne (glide descends inside the rise/fall deadband, so `fall=false`). `walk/jog/run` conditions are now gated on `is_grounded` via `AvatarAnimHelpers.locomotion_conditions` (extracted for testability).
- **State machine traps**: `Emote`, `Emote_Mix`, `Jump_End`, `Run_Jump_End` had no outgoing transition for `fall` or `gliding` — landing there while airborne/gliding froze the avatar. Escape transitions added.
- **Glide reopen timing**: `-> Gliding_Start` transitions used the 0.5s `start_glide` pulse; if the machine reached `Idle` after the window, it stuck. All entries now use the sustained `gliding` condition; `Idle/Walk/Jog/Run -> Gliding_Start` added.
- **Remote avatars frozen at range**: `is_grounded` only arrived via full Movement messages; compressed updates left it stale `false`, freezing locomotion with the new gate. Now derived from vertical velocity in `set_target_position` / `snap_to_position` (full path still wire-authoritative).
- **Glide anims didn't loop**: `Gliding_Idle/Forward/TurnLeft/TurnRight` froze on their last frame after ~4s. Looped in `double_jump_and_glide.res`.
- **MID-LOD statue bug**: `AnimationPlayer autoplay="Idle"` kept writing Idle bone poses every frame; with the tree in MANUAL callback mode (MID throttle, advance every 3rd frame) the player stomped the tree's pose 2 of 3 frames. Autoplay removed — the state machine owns the pose.

## Tests (headless, `--script`)

- `test_avatar_locomotion_grounded` — grounded gating, Idle->Walk vs Jump_Fall ordering, glide reopen from Idle
- `test_avatar_state_machine_graph` — parses the real `avatar.tscn` graph, asserts every state has an escape path on `fall` and `gliding` (fails on the old graph with 23 trap states)
- `test_avatar_autoplay_stomp` — reproduces the autoplay stomp under MANUAL throttle, pins the fix
- `AvatarLODHelpers` renamed/moved to `AvatarAnimHelpers` (it now holds general avatar animation helpers, not just LOD)

## Test plan

1. **Action mid-glide**: open the app, enter Genesis Plaza, climb to a high point, jump twice and open the glider; while airborne and gliding, trigger an emote/action. PASS: avatar never plays walk/idle mid-air; gliding anim persists or glide closes into a fall anim.
2. **Ground graze**: glide low over rooftops so you briefly touch a roof edge, then keep airborne. PASS: gliding anim resumes after the graze; never stuck standing mid-air with the glider deployed.
3. **Remote glider at MID range**: with a second account/client, have the peer glide while you watch from 15-25m. PASS: peer plays the gliding animation (no standing-still "statue"); approaching to <15m and retreating keeps the anim correct both ways.
4. **Emote props**: play an emote with a GLTF prop (e.g. the dog / computer props from #2581 screenshots). PASS: prop is toon-shaded, matching the avatar, not flat/unshaded.
5. **Emote cut by glide** (expected change): start an emote, then jump twice and open the glider mid-emote. PASS: emote animation cuts and gliding starts (audio/prop may persist) — this is intended, do not file.

## Known behavior change

The new `Emote/Emote_Mix -> Gliding_Start` escapes have no `nemote` guard: starting a glide mid-emote cuts the emote animation (audio/prop keep playing). Intended per #2603 (action while gliding), flagged for review.


lib/src/godot_classes/dcl_avatar.rs
```diff
@@ -160,6 +160,7 @@ impl DclAvatar {
         self.rise = y_velocity > 1.0;
         self.fall = y_velocity < -1.0;
         self.land = !self.rise && !self.fall;
+        self.is_grounded = self.land;
```
regenesis-claw: [P1] `land` is `|y_velocity| <= 1.0`, and per the PR description a glide descends inside that deadband — so a gliding remote avatar is reported as grounded here, which re-enables `walk/jog/run` mid-air and (given `Idle -> Jog`/`Idle -> Walk` are ordered before the new `Idle -> Gliding_Start`) makes a moving remote glider jog through the air.

This also overwrites whatever `apply_wire_movement_state` just wrote, on every position packet, so the wire value is not authoritative in practice.

```suggestion
        // A glide descent sits inside the rise/fall deadband, so `land` alone
        // would report a gliding peer as grounded. glide_state persists from
        // the last full Movement message.
        self.is_grounded = self.land && self.glide_state == 0;
```

godot/src/decentraland_components/avatar/avatar_emote_controller.gd
```diff
@@ -634,6 +634,7 @@ func _load_emote_from_gltf_internal(
 					prop_anim_player.queue_free()
 
 				avatar.add_child(armature_prop)
+				avatar._mesh_assembler.apply_toon_material_recursive(armature_prop)
```
regenesis-claw: [P2] This reaches into the avatar's private `_mesh_assembler`. Children `_ready()` before parents in Godot, so this is only safe because the GLTF load awaits first — a fragile invariant for a one-liner. A public wrapper on `Avatar`, or a `if avatar._mesh_assembler == null: return` guard, would make the assumption explicit.

regenesis-claw (changes requested):
Solid piece of work — the root-cause write-up is unusually good, and the three headless regression tests (especially the one that parses the real `.tscn` transition graph and asserts every state has an escape edge) are exactly the right shape for this bug class. Reviewed against REVIEW.md.

Pre-flight: branch is up to date with `main` (0 behind), static checks green, Linux/Clippy/Android and the iOS export are still in flight at the time of review — the last posted build report (previous head) shows **Android APK: Build failed**, so please confirm the current run goes green before merging. This is GDScript + a 2-line Rust change with no platform branches, so an iOS artifact isn't strictly required, but one is building anyway via the distribute run.

One blocker, in the remote-avatar part of the fix.

---

### [P1] Deriving `is_grounded` from the rise/fall deadband re-creates #2603 for remote gliders

`set_target_position` now does `self.is_grounded = self.land`, and `land == !rise && !fall` means `|y_velocity| <= 1.0 m/s`. The PR description's own analysis says a glide descends *inside* that deadband — so for a remote avatar that is gliding, the derived `is_grounded` is `true`. With the new gate that means `walk/jog/run` are re-enabled while the peer is airborne and gliding.

That matters because of transition ordering, which this PR relies on elsewhere: `Idle -> Jog` and `Idle -> Walk` sit near the front of the `transitions` array and the new `Idle -> Gliding_Start` was appended at the end. A remote avatar gliding with horizontal movement therefore satisfies `jog`/`walk` first and takes those edges instead of `Gliding_Start` — a peer jogging through the air, which is the exact symptom of #2603 on the remote path.

Related: this also makes the "full path stays wire-authoritative" claim not hold in practice. `apply_wire_movement_state` writes the authoritative `is_grounded`, but the next position update (~10 Hz) overwrites it with the derived value, so the wire value only survives for at most one packet interval.

Suggested direction: exclude the glide states from the derivation and don't clobber a wire value with a weaker inference, e.g.

```rust
// glide descent sits inside the rise/fall deadband, so `land` alone would
// report a gliding peer as grounded.
self.is_grounded = self.land && self.glide_state == 0;
```

(`glide_state` persists from the last full Movement message, so it is available on the compressed path.) If a peer can glide without a preceding full message, the derivation needs an explicit "unknown" state rather than defaulting to grounded.

### [P1 — please verify] Dropping `autoplay = "Idle"` may leave never-drawn avatars in bind pose

Removing the autoplay is the right call for the stomp, but the `AnimationPlayer` was also what posed avatars whose `AnimationTree` is inactive. `resolve_anim_drive` returns `active: false` for off-screen **and** for `FAR`, and `apply_screen_freeze` deactivates the tree. An avatar that spawns already off-screen or already at FAR never gets a single tree evaluation, so nothing writes a pose — previously the autoplaying Idle covered that window.

Two consequences worth checking on a real device: a T-posed avatar for the frames between spawn and first on-screen evaluation, and — more importantly — an impostor baked from a bind-pose mesh for avatars that are far away from the moment they appear. If that reproduces, seeding one Idle evaluation at spawn (before the first freeze) would keep the stomp fix without the regression.

### [P2] `avatar._mesh_assembler` accessed from the emote controller

The emote controller reaches into the avatar's private `_mesh_assembler`. Children `_ready()` before parents in Godot, so the field is only guaranteed non-null because the GLTF load is async — a fragile invariant for a one-line call. A small public wrapper on `Avatar` (or a `if avatar._mesh_assembler == null: return` guard) makes the ordering assumption explicit.

### [P2] The new regression tests aren't wired into CI

None of the `--script` avatar tests (including the pre-existing throttle one) run in any workflow — Linux runs `cargo test` + `check-gdscript`, static checks run the formatters and the WS shim node tests. These three tests are the most valuable artifact in the PR and today nothing re-runs them. A follow-up issue to add a headless step is fine; the graph test in particular would catch a re-introduced trap state on any future `.tscn` edit.

### [P2] Graph test treats unconditional edges as always traversable

`_test_escapes` walks through edges with an empty condition, but an at-end transition out of a looping state only fires when the animation ends. A state whose only "escape" is an at-end edge on a looping animation would pass the test while still trapping at runtime. Not worth reworking now — worth a comment so the next reader knows the check is necessary but not sufficient.

### Minor

- Stray double blank line before the `AnimationNodeStateMachineTransition_em2jf` sub-resource in `avatar.tscn`.
- The known behavior change you flagged (`Emote`/`Emote_Mix -> Gliding_Start` with no `nemote` guard) reads fine to me — cutting the emote when the glider opens matches the issue's intent. Worth one line in the test plan so QA doesn't file it as a bug.
- Test plan: the four cases are the right ones, but they're one-liners. Per REVIEW.md §4, QA runs these by hand — each case should start from "open the app", name the entry point (e.g. which parcel/world has a glider), and state an observable pass/fail result.

---

Requesting changes on the `is_grounded` derivation; everything else is either quick or a verify-and-tell-me. Happy to re-review as soon as it's pushed.

regenesis-claw (approved):
Re-review of `3a8165f` (delta vs `e1d0711`).

|---|---|---|
| P1 | `is_grounded = self.land` reports a gliding remote as grounded | ✅ Now `self.land && self.glide_state == 0` on both paths (`set_target_position` + `snap_to_position`), with a new `_test_glide_beats_walk_when_airborne` case pinning that gliding wins over the earlier-ordered `Idle -> Walk` edge |
| P1 | Removing `autoplay = "Idle"` could leave never-drawn avatars in bind pose | ✅ `_ready` seeds one evaluation (`animation_tree.active = true` + `advance(0.0)`) before the first freeze can land, so an avatar spawning off-screen/FAR has a pose for the impostor bake |
| P2 | Emote controller reaching into `avatar._mesh_assembler` | ✅ Public `Avatar.apply_toon_material_to()` with a null guard |
| P2 | Graph test treats at-end edges as always traversable | ✅ Caveat documented at the top of the test |
| P2 | Test plan not QA-executable | ✅ Rewritten as 5 numbered cases with entry points and explicit PASS criteria, including the intended emote-cut case marked "do not file" |
| P2 | Headless `--script` avatar tests not wired into CI | ⚠️ Not addressed — fine as the follow-up we agreed on |
| nit | Stray blank line in `avatar.tscn` | ✅ Removed |

CI is green across the board on this head: Static checks, Clippy, Linux, Android, Docker, and the iOS export all pass — the Android failure from the previous head is gone.

One small thing for whenever you next touch the file: the new `apply_toon_material_to()` was inserted directly under the `## Setup trigger detection for this avatar …` doc comment, so that docstring now sits on the wrong function. Moving the new function above the comment block (or below `setup_trigger_detection`) restores it. Not worth a push on its own.

Residual, non-blocking: the derivation still reports a remote peer as grounded at the jump apex (vertical speed inside the ±1 m/s deadband, `glide_state == 0`). No `-> Walk` edge exists out of `Jump_Mid`/`Jump_Fall`, so nothing acts on it today — worth remembering if a locomotion edge is ever added to an air state.

Approving. Note that a bot approval does not satisfy this repo's codeowner requirement, so a human reviewer is still needed to unblock the merge.

Ludmilafantaniella (approved):
## ✅ Approve

Tested on Android and iOS.

- [x] Action mid-glide (jump twice, open glider, trigger emote/action while airborne) - avatar never plays walk/idle mid-air, gliding anim persists or glide closes into a fall anim correctly
- [x] Ground graze (glide low, briefly touch a roof edge, stay airborne) - gliding anim resumes correctly after the graze, never stuck standing mid-air
- [x] Remote glider at MID range (second account, watched peer glide from 15-25m) - peer plays gliding animation correctly, no "statue" freeze; correct both approaching and retreating across the 15m threshold
- [x] Emote props (GLTF props like dog/computer) - toon-shaded correctly, matching avatar material, no longer flat/unshaded
- [x] Emote cut by glide (start emote, then jump twice + open glider mid-emote) - emote cuts and gliding starts as expected (intended behavior, not a bug)

https://github.com/user-attachments/assets/b7cec719-2f6c-47d5-bded-2e528a013ceb

kuruk-mm (approved):
great work!

PR #2700, opened 2026-08-13
feat(avatar): pool-based avatar LOD, billboards only on over-population

### LINKED ISSUE #2713: feat(avatar): pool-based avatar LOD — billboards only on over-population
## Context

Billboard impostors are currently used as a distance penalty: past 30m an avatar drops to a static billboard even if it's the only one in the scene. Caps were also fixed (8 full + 32 throttled) regardless of graphics profile.

## Changes

- 3D mesh pool driven by graphics profile (new AvatarPoolPolicy): Very Low=8, Low=16, Medium=32, High=40. Read every tick, so profile switches apply live without a restart.
- Animation split unchanged on High: 8 closest avatars run full-rate, the next 32 keep the mesh with throttled animation (~20fps).
- Billboard only on over-population (rank >= pool) or past 80m (safety valve). 30m forced-FAR band removed; CROSSFADE/dither band removed.
- Overflow silhouette rule (pool + real impostor layers) unchanged.

## PR

Implemented in #2700 (draft, base main).

manuelmaceira:
## Summary

Billboards become an **over-population** fallback, not a distance punishment.

- **Flat 40-avatar mesh pool** (new `AvatarPoolPolicy`, pure RefCounted): the
  8 closest in-frustum avatars run full-rate animation, the next 32 keep the
  mesh with throttled animation (~20fps) — exactly the old
  `MAX_FULL_AVATARS=8 / MAX_THROTTLED_AVATARS=32` split and budget.
- **30m forced-FAR band removed**: distance bands are now <15m FULL, 15–80m
  MID (mesh, throttled), ≥80m FAR (safety valve only). Billboards otherwise
  appear only when in-frustum avatars exceed the pool.
- Coordinator partition is inlined (zero per-tick allocation on the main
  thread); the pure partition lives in `AvatarPoolPolicy.assign_caps` and is
  pinned by a headless test.
- Overflow silhouette rule (pool + real impostor layers) unchanged; impostor
  textures remain static catalyst snapshots (#2206).

## Product decisions (reviewer questions)

- **No per-profile pool sizes.** The old system used the same 8+32 budget on
  every graphics profile; profile only tunes render quality. This PR keeps
  that flat budget — pool is deliberately NOT tied to `graphic_profile`.
- **Hard pop at 80m is intended.** The old CROSSFADE/dither ramp never
  worked visually; at 80m+ the swap is imperceptible. No fade window or
  hysteresis added — dead CROSSFADE branch removed instead.
- **Nearby billboards possible under over-population** (>40 in frustum on any
  profile), same as the old rank overflow. Accepted.

## Tests

- New headless `godot/src/test/avatar/test_avatar_pool_policy.gd`: pool
  constant + helpers, FULL/MID throttle split, assign_caps boundaries +
  overflow.
- `test_avatar_anim_throttle.gd`, `test_avatar_impostor_hidden_parent.gd`
  still pass (run locally; CI does not run the headless suites — noted by
  review).

## Test plan (manual, per §4)

1. Enter a crowded scene on High; walk toward/away from a group — verify no
   popping of nearby meshes and full-rate animation on the 8 closest.
2. Same scene on Very Low; closest avatars still animate; verify billboards
   only beyond ~40 visible avatars or 80m.
3. Flip the graphics profile mid-session — mesh count unchanged (pool is
   flat); no mass pop.

---
Closes #2713


godot/src/decentraland_components/avatar/impostor/avatar_impostor_config.gd
```diff
@@ -1,26 +1,21 @@
 class_name AvatarImpostorConfig
 extends Object
 
-const DISTANCE_FAR: float = 30.0
-const DISTANCE_NEAR: float = 25.0
+# Distance safety valve: beyond DISTANCE_FAR an avatar becomes a billboard
+# even with a free pool slot (the pool only governs near/mid-range crowding).
+const DISTANCE_FAR: float = 80.0
```
regenesis-claw, line 6: [P1] With the CROSSFADE band gone, crossing this threshold flips `dither_alpha` 1.0 → 0.0 and hides the mesh in one tick — a visible pop, and a repeating one for an avatar hovering around 80 m (LOD is re-evaluated every 6 frames with no hysteresis). Please either keep a fade window below `DISTANCE_FAR` (e.g. ramp dither/fade over 72–80 m, reusing the ramp that was deleted from `_update_lod`) or add an exit threshold a few metres below the entry one.

godot/src/decentraland_components/avatar/impostor/avatar_pool_policy.gd
```diff
@@ -0,0 +1,54 @@
...
+# (~20fps, see AvatarImpostorConfig.MID_ANIM_ADVANCE_EVERY_N_FRAMES).
+const FULL_RATE_CAP := 8
+
+# Pool size per graphics profile (GraphicSettings.PROFILE_DEFINITIONS order):
+# 0 Very Low, 1 Low, 2 Medium, 3 High. Custom (4) and invalid indices clamp
+# to High. Tuning knobs live here only — the coordinator reads them each tick
+# so DynamicGraphics profile switches apply live.
+# High = 40 matches the old fixed budget (MAX_FULL_AVATARS 8 + MAX_THROTTLED_AVATARS
+# 32 meshes: FULL_RATE_CAP full-rate + 32 throttled).
+const POOL_SIZES: Array[int] = [8, 16, 32, 40]
```
regenesis-claw: [P1] On the lower tiers this makes close-range billboards reachable for the first time: with pool 8, a 20-avatar huddle leaves 12 avatars a couple of metres away rendered as static catalyst snapshots — no animation, no emotes, and `set_click_active` off. Distance used to guarantee billboards were ≥25 m away regardless of crowding.

Consider a distance floor so rank never demotes an avatar that's very close, e.g. skip the pool demotion below ~8–10 m and apply it only beyond that. The perf win comes from wide crowds, not from the handful of avatars right in front of the camera.

godot/src/decentraland_components/avatar/impostor/avatar_pool_policy.gd
```diff
@@ -0,0 +1,54 @@
...
+
+# Partition a distance-sorted in-frustum avatar list into pool decisions.
+# Returns one [lod_cap, overflow] entry per rank:
+#   rank < FULL_RATE_CAP                  -> FULL (mesh, full-rate anim)
+#   rank < pool_size                      -> MID (mesh, throttled anim)
+#   otherwise                             -> FAR (billboard impostor)
+#   rank >= pool_size + max_real_impostors -> overflow = true (borrowed
+#     silhouette texture, no capture — the real layers are a finite resource)
+static func assign_caps(entry_count: int, pool_size: int, max_real_impostors: int) -> Array:
+	var caps: Array = []
```
regenesis-claw, line 48: [P2] This allocates one `Array` per avatar on every coordinator tick, where the previous code allocated nothing — ~100 Variant arrays 10×/s in a busy scene, on the main thread. The function stays just as testable if it returns packed data instead, e.g. a `PackedInt32Array` of caps plus the overflow start rank, or expose `full_rate_for(pool_size)` and let the coordinator branch inline.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md` (pre-flight applied: branch is a single commit on top of `main`, no submodule drift, GDScript-only with no platform branches so an iOS build is *not* required per §0.2 — the `build` label was added anyway and iOS/Android are still pending at the time of writing; Static checks / Linux / Clippy / Docker are green).

The architecture is right: making billboards an over-population fallback instead of a distance punishment is the correct model, the policy is cleanly extracted as pure statics, and the new headless test pins the partition boundaries properly (I traced every case by hand — all of them match the implementation). The profile-read-every-tick claim also checks out: `DynamicGraphics` really does write `config.graphic_profile` through `GraphicSettings.apply_graphic_profile()`, and its 2–5 min cooldowns mean the pool won't thrash frame to frame.

Two things block merge for me, both about what the user actually sees.

### 🔴 P1 — the MID→FAR transition is now a hard pop

The CROSSFADE band was the only thing making the mesh→billboard swap invisible: `dither_alpha` ramped 1→0 and `fade_alpha` 0→1 across 25–30 m, so by the time FAR was reached the mesh had already faded. With that band deleted, crossing `DISTANCE_FAR` (80 m) flips `dither_alpha` 1.0 → 0.0 and `set_meshes_visible(false)` in a single LOD tick. An avatar walking near 80 m (re-evaluated every 6 frames, no hysteresis) will pop mesh↔billboard repeatedly.

The same applies at the pool boundary: two avatars at near-identical distance will swap ranks around `pool_size` on every coordinator tick and flicker.

Minimum fix: reintroduce a fade window (reuse the existing dither ramp over, say, 72–80 m) or add rank/distance hysteresis (enter FAR at 80 m, leave at 76 m; require a rank to be beyond `pool_size` for N consecutive ticks before demoting). Either is small; shipping the hard flip is the part I'd hold on.

### 🔴 P1 — nothing prevents a billboard at 2 m on lower profiles

Previously the mesh budget was 40 on *every* profile, and distance guaranteed billboards were always ≥25 m away. Now on Very Low (pool 8) / Low (16) / Medium (32), an over-populated scene turns nearby avatars into billboards — in a 20-avatar huddle on Very Low, 12 avatars standing a couple of metres away become static, camera-facing snapshots. Since the impostor texture is a static catalyst snapshot (#2206), those avatars won't move, emote or match their equipped state, and they also lose `set_click_active` / particles.

That's the core trade the PR is making and it is not mentioned in the description. Two asks: (a) confirm this was a deliberate product decision, and (b) consider a floor — never demote an avatar closer than ~8–10 m regardless of rank, applying the pool only beyond that. That keeps the perf win (the expensive case is a wide crowd, not the 5 people in your face) while removing the worst visual artefact.

### 🟡 P2 — per-tick allocation in the coordinator loop

`assign_caps()` builds one `Array` per avatar every coordinator tick, replacing a zero-allocation branch chain. In a 100-avatar scene that's ~100 Variant arrays 10× per second on the main thread, in a PR whose purpose is mobile perf. Suggestion: keep `assign_caps` as the testable pure function but have it return two `PackedInt32Array`s (or a single packed cap array plus an overflow start index), or expose `full_rate_for(pool_size)` and let the coordinator branch inline.

### 🟡 P2 — leftovers and description gaps

- `CROSSFADE` is now unreachable but still branched on in `_apply_lod_state` (`state == LODState.FAR or state == LODState.CROSSFADE`) and referenced in the skinning-throttle comment near the top of the avatar script. Keeping the enum value for compatibility is fine; the dead branch and stale comments should go.
- The description points at design notes in a root `NOTES.md`, but no such file is in the diff — either commit it or drop the reference. (A root-level `NOTES.md` is also a questionable home for this; `docs/` would fit the repo better.)
- **No `## Test plan`.** Per §4, QA runs these by hand on a real phone, and this is a purely visual/perf change with no automated coverage of the rendering side. Please add executable cases — at minimum: (1) enter a crowded scene on High, verify no popping while walking toward/away from a group; (2) same scene on Very Low, verify the closest avatars still animate; (3) flip the graphics profile in Settings mid-session and verify the mesh count changes live without a mass pop.
- FYI: no CI workflow runs the headless GDScript suites, so `test_avatar_pool_policy` is manual-only — the PASS claim for the two existing suites is unverified from CI.

No security-relevant surface in this diff (no network, auth, user input or serialization): **no security issues found**. No public/SDK contract change — `AvatarImpostorConfig.MAX_FULL_AVATARS`, `MAX_THROTTLED_AVATARS`, `DISTANCE_NEAR` and `MID_ANIMATION_PLAYBACK_SPEED` were removed, and I confirmed the only remaining references in the tree were the ones this PR deletes, so nothing else breaks.

Requesting changes on the two P1s. Happy to re-review as soon as the fade/hysteresis is back and the close-range question is answered.

manuelmaceira:
Thanks for the thorough review — both P1s addressed as deliberate product decisions, plus the P2s.

**P1 — MID→FAR hard pop:** agreed it's a hard flip at 80m, and that's intended. The old CROSSFADE/dither ramp never read as a fade in practice, and at 80m+ (or pool-overflow) the swap is imperceptible — we deleted the dead branch instead of reviving a ramp that never worked. No fade window / hysteresis added. PR body updated under "Product decisions".

**P1 — billboards near camera on lower profiles:** no longer applies the way it did. The pool is now **flat 40 on every profile** (old budget restored — 8 full-rate + 32 throttled), not 8/16/32/40. Profile only tunes render quality, exactly as before. Nearby billboards can still happen under >40 in-frustum avatars on any profile, same as the old rank overflow — accepted, documented in the PR body.

**P2 — per-tick allocation:** fixed. Coordinator now branches inline with `full_rate_for()` / `overflow_start()` — zero per-avatar allocation per tick. `assign_caps` stays as the tested pure partition for headless tests, per your suggestion.

**P2 — leftovers:** dead CROSSFADE branch in `_apply_lod_state` and stale comments removed; enum value kept for compatibility. NOTES.md reference dropped from the description (file removed from the branch).

**Test plan:** added executable manual cases to the PR body (§4) — crowded scene on High/Very Low, mid-session profile flip.

One FYI you noted: no CI workflow runs the headless GDScript suites — confirmed, the two existing suites are run locally only. Happy to wire the new test into the Linux build job if you want it covered; say the word.

regenesis-claw (approved):
Re-review of `8544447` (the branch was amended, not appended, so this is a fresh read of the whole diff rather than an incremental one).

### Delta on previous findings

| # | Finding | Status |
|---|---------|--------|
| P1 | Hard pop / no hysteresis at the MID→FAR boundary | ⚠️ **Answered, not changed** — accepted as an explicit product decision (see below) |
| P1 | Close-range billboards reachable on lower graphics profiles | ✅ **Fixed** — per-profile pools dropped for a flat `POOL_SIZE = 40`; the coordinator no longer reads `graphic_profile` at all, so the crowding threshold is identical to `main` on every tier |
| P2 | Per-tick Array-of-Arrays allocation in the coordinator | ✅ **Fixed** — partition inlined in the loop, allocation-free; `full_rate_for()` / `overflow_start()` keep the boundary arithmetic shared with the tested pure function |
| P2 | Dead `CROSSFADE` branch and stale comments | ✅ **Fixed** — `_apply_lod_state` now branches on `FAR` only; enum value kept for compatibility, which is fine |
| P2 | `NOTES.md` referenced but not committed | ✅ **Fixed** — reference removed |
| P2 | Missing `## Test plan` | ✅ **Added** (notes below) |

On the 80 m pop: your call, and the reasoning holds — at that distance the avatar is a few pixels tall and the old ramp was doing very little for the cost of an extra state. Not re-litigating it.

### New, non-blocking notes

1. **Stale comment in the avatar script.** The pool-cap comment above `_lod_rank_cap` still says the throttled tier is sized "(per graphics profile)", which the flat-40 revert made untrue. Same block, the skinning-throttle comment further up still reads "imperceptible at 15-30m distance" — that band is now 15–80 m.
2. **The headless test now covers a parallel implementation, not the shipped path.** `assign_caps` is only called from the test; production runs the inlined loop. The shared helpers mean the *boundaries* can't drift silently, but the branch structure can. Cheap insurance if you ever touch this again: have the test assert the inlined ordering too, or note it in the file header.
3. **The 15–80 m MID band is the one thing worth a number before this ships.** Skinning cost is distance-independent, so a scene with 40 avatars spread across 80 m now runs 8 full-rate + 32 throttled skinned meshes where `main` rendered mostly billboards past 30 m. Mesh LOD thresholds and the 20 fps throttle should absorb most of it, and the 40-mesh ceiling is unchanged, but a quick A/B fps capture on the A54 in that exact layout (sparse-and-wide, not dense-and-close) would confirm the trade is free. Follow-up is fine.
4. **Test plan format** (§4): the three cases read as reviewer notes rather than QA steps — no "open the app" starting point, no per-case expected result, and "enter a crowded scene" gives a tester no way to reach 40+ in-frustum avatars. Case 3 ("flip the profile, mesh count unchanged") is now asserting a no-op, so it's really a regression check — worth saying so. Not blocking, but QA will likely come back with questions on case 2.

One thing outside this diff, purely FYI: `avatar_anim_helpers.gd` still has a `LOD_MID, LOD_CROSSFADE:` branch. It's unreachable now, harmless, and not this PR's job to clean up.

### CI

Static checks green on this SHA; Linux (which runs GDScript validation), Clippy, Android and the iOS export are still pending at the time of writing. Nothing in the diff can plausibly break Rust, but worth eyeballing the Linux job before merging since it's the only thing that parse-checks the new GDScript.

No security-relevant surface, no SDK/public contract change — the removed `AvatarImpostorConfig` constants have no remaining references in the tree.

**Approving.**

Ludmilafantaniella (approved):
## Approved by QA ✅ 

Tested on both iOS and Android in Genesis Plaza with ~12 users present.
- Closest avatars animated at full-rate as expected, no visual issues.
- Switched graphics profile mid-session (Very Low → High) — no popping, mesh count stayed consistent since the group was small.
- Compared against prod: in prod, avatars visibly fade/turn into billboards after moving just a few meters away. In this PR, avatars stayed as full animated meshes even at distance — did not reach the 80m FAR threshold or the 40-avatar pool overflow with this crowd size, so didn't observe the mesh→billboard transition itself.

No issues found with this scenario. Would be good to also test with 40+ avatars in frustum to validate the pool overflow behavior described in the PR.

cc @manuelmaceira

https://github.com/user-attachments/assets/19a2a75d-96ca-4a42-b3d7-26cdcdde2db7

https://github.com/user-attachments/assets/f6a7af9a-17ac-4332-b167-aaf7fb04c9fb

kuruk-mm (approved):
great work!

Issue #2714, opened 2026-08-18
Reduce crash rate from >10% to <5%


kuruk-mm:
We are above **10% crash rate**. Target is **under 5%**. This issue tracks the reduction work itself — the measurement and live reporting system that gets us there and keeps us there is #2712. The two run in parallel; **this one must not wait on #2712**, because the app stores already give us enough to start today.

## Target

- **>10% → <5%**, sustained for 7 consecutive days, measured **per platform** — iOS and Android have different failure modes and very different fix latencies, so a blended number hides the problem.
- Intermediate checkpoint: **<8%** before we call the first pass done.

## Step 0 — pin the baseline (blocking, ~half a day)

"10%" means different things depending on where it came from. Before any fixing, post in this issue:

- **Source**: Google Play Console vitals (user-perceived crash rate), App Store Connect metrics, or Sentry.
- **Definition**: per session or per user? user-perceived or total? over what window?
- **Per-platform** current number + 30-day trend.
- Whether **preview/dev sessions are included** — today they're indistinguishable from real ones in both Segment and Sentry (see #2712), so part of the 10% may not be users at all.

Play Console and App Store Connect give this **today with zero instrumentation**, so they are the working source of truth until #2712 lands.

> **Store risk — check this first.** Google Play's "bad behaviour" thresholds are ~**1.09%** user-perceived crash rate and ~**0.47%** ANR rate; exceeding them can suppress store visibility and ranking. Verify our exact position in Play Console. If we're over, this stops being only a quality issue.

## Triage — classify before fixing

Rank crash groups by **volume × users affected**, then tag each with:

1. **Layer** — native signal (Godot/Rust SIGSEGV/SIGABRT) / Rust panic / GDScript / JS scene runtime / GPU driver. This sets the fix latency: **Android hotfixes ship in hours**, but anything in Rust needs a **full iOS resubmission (~3 days)** — OTA can only swap the PCK, not Rust or resources.
2. **Device concentration** — broad (min-spec / low-RAM) or one model / SoC / driver? Cross-reference the **devices the team actually owns**, so each group becomes either a reproducible ticket or a procurement ask instead of a dead end.
3. **Reproducible yes/no** — with steps if yes.

**Two failure modes that may be a large share of the 10% but won't appear as "crashes" in Sentry:**
- **ANRs** (Android) — main-thread stalls; counted by Play, invisible to most crash SDKs.
[...]

PR #2716, opened 2026-08-18
Release Candidate 1.12.2 (crash guards + Sentry quota)

kuruk-mm:
Hotfix RC on top of shipped 1.12.1 (`90050d7`) — 5 code commits plus the version bump, no engine bump and no new features. Two threads aimed at the same problem: we are over **10% crash rate** (#2714), and we could not read the crashes properly because Sentry was drowning in noise and burning its quota.

**Crashes.** GDScript's `if node:` is not a validity check — a freed instance isn't null, so the test passes and the *next* line faults. Cross that with `await`, which hands control back to the engine and gives anything a window to free the node the function is about to touch, and it is a use-after-free. The **release** export template we ship has none of the debug template's freed-object validation, so it lands as a SIGSEGV on the phone instead of an error in the log — which is why these never showed up in testing. Guarded in `ModalManager` (`_async_create_*` assigns `current_*` then awaits two frames, with 20 callers touching it on resume), in `DropdownList`, and in the emote loader, where `AvatarEmoteController` is `RefCounted` and so gets none of the engine's implicit safety nets for a freed owner.

**Sentry.** The `NOISE_PATTERNS` filter had been dead code: it read `event.message`, which is always empty for logger-captured events, so all 16 patterns never matched and ~78% of accepted production events were noise the filter was meant to drop. On top of that the logger's limiter was left at addon defaults (20 events/10s, same line re-reported every 1s), letting one device emit ~172k events/day — a few hundred concurrent users drained the monthly quota in hours. Both fixed, and reporting now defaults to **crash-only**, with a remote flag to turn error events back on for a debugging window.

- **Base:** `release` · **Head:** `release-1.12.2` (= `release` @ `90050d7` + 6 commits)
- **Version:** `1.12.2` in `lib/Cargo.toml`, `lib/Cargo.lock`, `export_presets.cfg`. Store build numbers come from the Cloudflare worker at build time and are untouched here.

## Changes

**Crash guards (#2714)**
- **`godot/src/utils/node_guard.gd`** *(new)* — `NodeGuard.is_alive(node, site)`: `is_instance_valid()` plus capped telemetry (first hit per site per session → Sentry + Segment `STALE_NODE_GUARD`), so we learn how often the race really fires instead of silently swallowing it. Documents the two rules: never truthiness-test a node reference, always re-validate after an `await`.
- **`modal_manager.gd`** — 57 truthiness tests become `is_instance_valid()`; the 6 `_async_create_*` refuse to return an instance freed across their awaited frames; the 20 call sites re-validate on resume; `_remove_*` and the `tree_exited` handlers clear their member unconditionally, so a freed-but-not-cleared reference can't survive a close.
- **`dropdown_list.gd`** — `_popup_panel` was read after an awaited frame, behind a truthiness test on `_shadow_rect`.
- **`avatar_emote_controller.gd`** — every remote-player emote spans a network fetch plus two loader awaits and then touched the avatar unchecked. A player leaving mid-download freed it, and `avatar.add_child()` on the freed avatar is a native crash — matching the SIGSEGVs Sentry groups under `Node::_update_children_cache` and `Node::set_name`.

**Sentry quota & signal**
- **`project_main_loop.gd`** — noise filter now reads `get_exception_value(0)` (falling back to `message`), so the 16 patterns actually match; logger limiter clamped to ~5 events/min per device (~7k/day) with a given source line re-reported at most once a minute; `_before_send` drops sub-fatal exception events unless the new flag is on. Explicit `capture_message` events (NodeGuard) and FATAL crashes always pass, and errors still attach to crashes as breadcrumbs.
- **`feature_flags.gd`** — new fail-closed `sentry-error-events` flag re-enables ERROR reporting remotely. With the firehose gated, `DEFAULT_SENTRY_SAMPLE_RATE` rises `0.1 → 1.0` so early-session crashes aren't randomly dropped; `sentry-sample-rate` stays the remote kill switch.

## Test plan

The guards are bail-outs on state that would otherwise have crashed, so the happy path should be identical — §1 checks that, §2 is the actual fix. The Sentry changes are telemetry-only and are **not** hand-testable on a device; they are verified in the Sentry dashboard after release. Run on Android and iOS, including one low-end Android (where the crashes concentrate).

**0. Version** — Launch the app → the lobby screen and **Settings → About** read **`v1.12.2.<build>`**.

**1. Nothing regressed** — each modal must show with the right title, body and buttons, and the buttons must do what they say:
- [x] `/goto 0,0` in chat → teleport modal with place name, creator and image; **JUMP IN** teleports.
- [x] `/world <a-real-world>.dcl.eth` in chat → world modal with name and image; **JUMP IN** enters it.
- [x] Tap a scene link that opens an external URL → **"Open external link?"**; **OPEN LINK** opens the browser, **CANCEL** closes.
- [x] Turn airplane mode on in-world → **"Connection lost"**; turn it off → recovers. (iOS shows **RETRY** only.)
- [x] **Settings → Gameplay → Camera Mode** dropdown → opens with its drop shadow drawn; picking an item closes it and applies.
- [x] As a guest: **Settings → Account → Add Email** → email modal, then code modal, then reward modal, in sequence.
- [x] Play three of your own emotes from the emote wheel → each plays fully, with its props and sound.

**2. The race itself** — each of these frees a node while an `await` is in flight, which is what used to crash:
- [x] Open a Settings dropdown and close it on the same frame (fast double-tap), 10× → no crash, and it still opens normally afterwards.
- [x] `/goto 0,0`, then tap the travel modal's **X** before its image finishes loading, 10× → no crash; the last one still teleports.
- [x] `/goto 0,0`, then **sign out** from the navbar while the modal is on screen → back to the lobby, no crash; sign in again → modals still work.
- [x] In a busy Genesis Plaza, with other players visibly emoting, teleport out and straight back 5× → no crash; other players' emotes still play after each return.

**3. Regression** — shared UI code, so:
- [x] Sign in with a wallet and as a guest → both reach the world.
- [x] Open and close **Backpack**, **Profile**, **Settings**, **Map**, **Emote wheel**, **Chat** → no crash, no stuck overlay.
- [x] Teleport between three places including Genesis Plaza and a World → each loads, the previous unloads.

---

**After release (not QA).** Three numbers decide whether this worked, and none of them are visible before it ships:
1. Sentry event volume should drop sharply — if it doesn't, the noise filter is still missing its targets.
2. Hits on `NodeGuard: … resumed on a freed node` mean the guards are catching real races. Each names its call site and is a follow-up bug, not a regression; a flood from one site means that path needs a real lifecycle fix rather than a bail-out.
3. Play Console / App Store Connect crash rate against 1.12.1 over 7 days — post it on #2714.

Crash-only reporting is the one thing to keep an eye on: if we lose visibility of a non-crash failure we care about, flip `sentry-error-events` on in the bff rather than shipping a build.

### DIFF
--- a/godot/src/feature_flags.gd
+++ b/godot/src/feature_flags.gd
@@ -11,9 +11,9 @@ extends Node
 ## gracefully (e.g. tearing archipelago down while keeping scene rooms alive).
 ##
 ## Fail-open: on timeout, network error or malformed payload every flag keeps
-## its default and the feature stays enabled. Exception: `pulse` is fail-closed —
-## the transport activates only when the payload explicitly enables it (see
-## _apply_flags).
+## its default and the feature stays enabled. Exceptions: `pulse` and
+## `sentry-error-events` are fail-closed — they activate only when the payload
+## explicitly enables them (see _apply_flags).
 
 signal flags_loaded
 
@@ -24,6 +24,9 @@ const FLAG_ARCHIPELAGO := "archipielago"
 const FLAG_PULSE := "pulse"
 # Sentry error-event sampling, served as a number in [0, 1].
 const FLAG_SENTRY_SAMPLE_RATE := "sentry-sample-rate"
+# Report ERROR-level Sentry events (the engine/Rust error firehose). Fail-closed:
+# absent flag = crash/fatal-only reporting — see ProjectMainLoop._before_send.
+const FLAG_SENTRY_ERROR_EVENTS := "sentry-error-events"
 # The bff also serves `sentry-traces-sample-rate`, but sentry-godot exposes no
 # performance-tracing API yet — there is nothing to apply it to until the SDK
 # grows one.
@@ -117,3 +120,4 @@ func _apply_flags() -> void:
 		main_loop.set_sentry_sample_rate(
 			get_number(FLAG_SENTRY_SAMPLE_RATE, ProjectMainLoop.DEFAULT_SENTRY_SAMPLE_RATE)
 		)
+		main_loop.set_sentry_error_events_enabled(is_enabled(FLAG_SENTRY_ERROR_EVENTS, false))
--- a/godot/src/ui/components/molecules/dropdown_list/dropdown_list.gd
+++ b/godot/src/ui/components/molecules/dropdown_list/dropdown_list.gd
@@ -215,7 +215,10 @@ func _async_open_popup() -> void:
 
 	# Sync shadow rect with popup panel (position, size, and offset)
 	await get_tree().process_frame
-	if _shadow_rect:
+	# Closing the dropdown during that frame frees the popup — see NodeGuard.
+	if not NodeGuard.is_alive(_popup_panel, "DropdownList._async_open_popup"):
+		return
+	if is_instance_valid(_shadow_rect):
 		# Get corner radius from popup style
 		var popup_style_shadow: StyleBoxFlat = (
 			_popup_panel.get_theme_stylebox("panel") as StyleBoxFlat
--- a/godot/src/ui/components/organisms/modal/modal_manager.gd
+++ b/godot/src/ui/components/organisms/modal/modal_manager.gd
@@ -159,9 +159,11 @@ func _ready() -> void:
 ## Shows an EXTERNAL_LINK type modal
 ## @param external_url: The external URL to open
 func async_show_external_link_modal(external_url: String) -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
 			print("NOT CREATED MODAL")
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_external_link_modal"):
 			return
 
 	current_modal.set_title(EXTERNAL_LINK_TITLE)
@@ -180,9 +182,11 @@ func async_show_external_link_modal(external_url: String) -> void:
 
 ## Shows a SCENE_TIMEOUT type modal
 func async_show_scene_timeout_modal() -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
 			print("NOT CREATED MODAL")
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_scene_timeout_modal"):
 			return
 
 	current_modal.set_title(SCENE_TIMEOUT_TITLE)
@@ -201,9 +205,11 @@ func async_show_scene_timeout_modal() -> void:
 ## Shows a CONNECTION_LOST type modal
 ## @param hide_buttons: If true, hides all buttons (used on iOS after retry fails)
 func async_show_connection_lost_modal(hide_buttons: bool = false) -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
 			print("NOT CREATED MODAL")
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_connection_lost_modal"):
 			return
 
 	current_modal.blocker = true
@@ -237,6 +243,8 @@ func async_show_teleport_modal(location: Vector2i, realm: String = "") -> void:
 	var destination_realm = realm if not realm.is_empty() else DclUrls.main_realm()
 
 	if not await _async_create_travel_modal():
+		return
+	if not NodeGuard.is_alive(current_travel_modal, "ModalManager.async_show_teleport_modal"):
 		return
 
 	current_travel_modal.closed.connect(close_travel_modal)
@@ -279,6 +287,8 @@ func async_show_world_modal(world_name: String) -> void:
 
 	if not await _async_create_travel_modal():
 		return
+	if not NodeGuard.is_alive(current_travel_modal, "ModalManager.async_show_world_modal"):
+		return
 
 	current_travel_modal.closed.connect(close_travel_modal)
 	current_travel_modal.jump_in_pressed.connect(_on_world_jump_in.bind(world_name))
@@ -303,6 +313,8 @@ func async_show_world_modal(world_name: String) -> void:
 func async_show_change_realm_modal(realm_name: String, _message: String = "") -> void:
 	if not await _async_create_travel_modal():
 		return
+	if not NodeGuard.is_alive(current_travel_modal, "ModalManager.async_show_change_realm_modal"):
+		return
 
 	current_travel_modal.closed.connect(close_travel_modal)
 	current_travel_modal.jump_in_pressed.connect(_on_change_realm_primary.bind(realm_name))
@@ -318,8 +330,10 @@ func async_show_change_realm_modal(realm_name: String, _message: String = "") ->
 ## Shows a SCENE_CRASH type modal
 ## @param entity_id: The entity ID of the crashed scene
 func async_show_scene_crash_modal(entity_id: String) -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_scene_crash_modal"):
 			return
 
 	current_modal.blocker = true
@@ -347,8 +361,12 @@ func async_show_scene_crash_modal(entity_id: String) -> void:
 func async_show_low_memory_warning_modal(
 	entity_id: String, footprint_mb: int = -1, available_mb: int = -1
 ) -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(
+			current_modal, "ModalManager.async_show_low_memory_warning_modal"
+		):
 			return
 
 	current_modal.blocker = true
@@ -386,8 +404,10 @@ func async_show_low_memory_warning_modal(
 func async_show_ban_pre_check_modal() -> void:
 	_force_hide_loading_screen()
 
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_ban_pre_check_modal"):
 			return
 
 	current_modal.blocker = true
@@ -409,8 +429,10 @@ func async_show_ban_pre_check_modal() -> void:
 func async_show_private_world_modal(world_name: String) -> void:
 	_force_hide_loading_screen()
 
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_private_world_modal"):
 			return
 
 	current_modal.blocker = true
@@ -432,8 +454,10 @@ func async_show_ban_kicked_modal() -> void:
 	if _suppress_ban_kicked:
 		_suppress_ban_kicked = false
 		return
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_ban_kicked_modal"):
 			return
 
 	current_modal.blocker = true
@@ -484,8 +508,10 @@ func async_show_disconnected_modal() -> void:
 func _async_show_disconnect_modal(
 	title: String, body: String, primary_label: String, primary_handler: Callable
 ) -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager._async_show_disconnect_modal"):
 			return
 
 	current_modal.blocker = true
@@ -519,8 +545,10 @@ func _on_session_ended_secondary() -> void:
 
 ## Shows a low-spec iPhone warning modal (lobby popup)
 func async_show_low_spec_iphone_modal() -> void:
-	if not current_modal:
+	if not is_instance_valid(current_modal):
 		if not await _async_create_modal():
+			return
+		if not NodeGuard.is_alive(current_modal, "ModalManager.async_show_low_spec_iphone_modal"):
 			return
 
 	current_modal.set_title(LOW_SPEC_IPHONE_TITLE)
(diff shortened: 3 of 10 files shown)


godot/src/ui/components/organisms/modal/modal_manager.gd
```diff
@@ -1146,41 +1206,45 @@ func _async_create_input_modal() -> InputModal:
...
+	if is_instance_valid(_input_canvas_layer):
 		_input_canvas_layer.queue_free()
-		_input_canvas_layer = null
+	_input_canvas_layer = null
 
 
 func async_show_code_modal(email: String = "") -> CodeModal:
 	var modal = await _async_create_code_modal()
-	if not modal:
+	if not is_instance_valid(modal):
```
manuelmaceira, line 1234: why sometimes is_instance_valid and not NodeGuard path?

EibrielInv (approved):
Looks good! 👍🏽

godot/src/utils/node_guard.gd
```diff
@@ -0,0 +1,69 @@
...
+##    suspending says nothing about the state on resume.
+##
+## `is_alive()` is rule 2 plus telemetry. Swallowing the stale reference silently would fix
+## the crash and hide how often the race actually fires, so every guarded site reports the
+## first hit of the session to Sentry and Segment — enough to size the problem per build and
+## per device, capped so a device that reopens a modal in a loop cannot flood either.
+
+## Hits reported per site per session. The counter in `_hits` keeps rising past this;
+## only the outbound event is capped.
+const REPORTS_PER_SITE := 1
```
regenesis-claw, line 29: [P2] Capped at one report per site per session for both Sentry and Segment, so one hit and ten thousand hits produce identical data. The test plan's §4 signal ("a flood from one site ... deserves a real lifecycle fix") can't be read from that. Either report on powers of two, or include the running count in the single event you do emit:

```suggestion
const REPORTS_PER_SITE := 1
```
(keeping the cap is fine if the count travels in the payload — see the Segment call below)

godot/src/utils/node_guard.gd
```diff
@@ -0,0 +1,69 @@
...
+	push_warning(message)
+
+	# Telemetry is off in asset-server / CI builds, where SentrySDK was never initialized.
+	if not DclGlobal.is_telemetry_disabled():
+		SentrySDK.capture_message(message, SentrySDK.LEVEL_WARNING)
+
+	# Segment carries the same hit so the rate can be tracked per release alongside the
+	# store crash rate, which is the number this guard is meant to move.
+	if Global.metrics != null:
+		Global.metrics.track_screen_viewed(
```
regenesis-claw, line 67: [P2] `track_screen_viewed` writes into the Segment `Screen Viewed` stream, which is what the Metabase loading/onboarding funnels are built on (`LOADING_END`, `LOADING_TIMEOUT`, `UPGRADE_OTP_*`). Injecting `STALE_NODE_GUARD` there inflates screen-view counts for anyone querying that event without a name filter. Worth a dedicated event, or at least a heads-up to whoever owns the funnel queries.

Also: passing `hits` here would restore the flood signal that `REPORTS_PER_SITE := 1` removes.

godot/src/ui/components/molecules/dropdown_list/dropdown_list.gd
```diff
@@ -215,7 +215,10 @@ func _async_open_popup() -> void:
 
 	# Sync shadow rect with popup panel (position, size, and offset)
 	await get_tree().process_frame
-	if _shadow_rect:
+	# Closing the dropdown during that frame frees the popup — see NodeGuard.
+	if not NodeGuard.is_alive(_popup_panel, "DropdownList._async_open_popup"):
```
regenesis-claw, line 219: [P2] This catches the lifetime race but not the state race, and the state race is the one test plan §2 describes. `_close_popup()` frees nothing — it only sets `_is_open = false` and `_popup_layer.visible = false`. So a close during the awaited frame leaves `_popup_panel` perfectly valid, this guard passes, and execution falls through to `_popup_layer.visible = true` at the end of the function, re-showing the popup the user just dismissed with `_is_open` now out of sync.

```suggestion
	if not NodeGuard.is_alive(_popup_panel, "DropdownList._async_open_popup"):
		return
	# Closing during that frame doesn't free anything, it just flips the state —
	# bail out too, or we re-show a popup the user already dismissed.
	if not _is_open:
		return
```

godot/src/feature_flags.gd
```diff
@@ -24,6 +24,9 @@ const FLAG_ARCHIPELAGO := "archipielago"
 const FLAG_PULSE := "pulse"
 # Sentry error-event sampling, served as a number in [0, 1].
 const FLAG_SENTRY_SAMPLE_RATE := "sentry-sample-rate"
+# Report ERROR-level Sentry events (the engine/Rust error firehose). Fail-closed:
+# absent flag = crash/fatal-only reporting — see ProjectMainLoop._before_send.
+const FLAG_SENTRY_ERROR_EVENTS := "sentry-error-events"
```
regenesis-claw, line 29: [P1] I couldn't find this flag in `decentraland/mobile-bff`. The seeded set is `pulse`, `dual-channel`, `sentry-sample-rate`, `sentry-traces-sample-rate` (`migrations/1786320000000_feature_flag_types.ts`, `openapi.yaml`, `test/mocks/feature-flags-db-mock.ts`) — no `sentry-error-events` row, no migration adding one.

Because the flag is fail-closed, that means 1.12.2 ships with ERROR reporting off and the remote switch described in `ProjectMainLoop`'s comment doesn't exist — re-enabling would need a new store build. Please land the bff seed migration alongside this RC, or default this one to fail-open until the bff serves it.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`.

**Pre-flight**
- Base `release`, head `release-1.12.2`, `mergeStateStatus: CLEAN`, not behind base. ✅
- iOS build present and green (`iOS export -> R2` passed, mobile distribution to TestFlight triggered on `84738fe`). ✅ Android APK green.
- All checks green (Static, Clippy+version check, Linux, Docker, Android, iOS).
- No submodule drift, no `.claude/` files, no debug prints added, no non-English comments.

The node-guard work itself is good and the reasoning in `node_guard.gd` is the kind of "why" comment this repo wants. Two blockers below are about **scope and the remote kill switch**, not the guards.

---

### [P1] `sentry-error-events` does not exist in mobile-bff — crash-only mode is not remotely revertible

The flag is fail-closed and the comment in `project_main_loop.gd` promises it "turns ERROR reporting back on remotely (e.g. to debug a release for a day) without shipping a build". I searched `decentraland/mobile-bff`: the seeded flags are `pulse`, `dual-channel`, `sentry-sample-rate`, `sentry-traces-sample-rate` (`migrations/1786320000000_feature_flag_types.ts`, `openapi.yaml`, `test/mocks/feature-flags-db-mock.ts`). There is **no** `sentry-error-events` row and no migration adding one.

So as shipped, 1.12.2 turns off every sub-FATAL exception event — every GDScript `push_error`/`printerr` and every Rust `tracing::error!` — with **no way to turn it back on except a new store build**. Please land the bff seed migration before or with this RC, or make the flag fail-open for this release and flip it to fail-closed once the bff serves it.

### [P1] Undisclosed scope: the description says "exactly one code change", the diff has two

The Changes table lists 3 files. The diff also contains:

- `project_main_loop.gd` + `feature_flags.gd` — a full Sentry reporting overhaul: crash-only `_before_send` gate, new fail-closed `sentry-error-events` flag, `DEFAULT_SENTRY_SAMPLE_RATE` 0.1 -> 1.0, `logger_limits` clamps, and the `_event_text()` fix (real bug: `event.message` was always empty for logger-captured events, so `NOISE_PATTERNS` never matched anything — worth calling out on its own).
- `avatar_emote_controller.gd` — two more NodeGuard sites.

QA runs the test plan by hand off this description; none of the above is in it. On a release-bound hotfix RC that is a blocker for the description, not the code. Please either update the summary + test plan to cover the telemetry change, or split it out of the RC.

Related: after this lands, 1.12.2's Sentry **error** volume is not comparable to 1.12.1's for reasons unrelated to the fix. Worth a line in #2714 so the post-release analysis isn't confounded (store crash rate from Play Console / ASC is unaffected).

### [P2] `DropdownList._async_open_popup` — the guard doesn't cover the close-during-await case

`_close_popup()` doesn't free anything: it sets `_is_open = false` and `_popup_layer.visible = false`. So if the user closes during the awaited frame, `is_alive(_popup_panel)` is **true**, the function runs to completion and sets `_popup_layer.visible = true` + the pressed stylebox again — the popup the user just closed reappears, with `_is_open == false`. That is exactly the test-plan §2 case ("close it on the same frame you opened it ... no stuck overlay"), and the validity guard doesn't catch it because it's a state race, not a lifetime race. Needs an `_is_open` re-check alongside the validity check.

### [P2] `REPORTS_PER_SITE := 1` makes the "flood" signal in test plan §4 unobservable

Both the Sentry message and the Segment event are capped at one per site per session, so a site that fires once and a site that fires 10,000 times are indistinguishable in the data. §4 says "a flood from one site means that path is being hit constantly and deserves a real lifecycle fix" — with this cap you can't tell. Consider reporting on powers of two, or carrying the running `hits` count in the payload of the one event you do send.

### [P2] `hit_counts()` has no callers

The doc comment says "read by tests and the debug hub", but no test is added in this PR and nothing calls it. Per REVIEW.md §15 either wire it up or drop it.

---

### Verified, not issues

- `_async_load_emote` / `_load_emote_from_gltf_internal`: `animation_player` / `animation_tree` are passed into `_init` from `avatar.gd` and live and die with the avatar scene, so guarding `avatar` alone is sufficient — the comment slightly oversells it but the code is right.
- `_async_load_travel_modal_data` / `_async_load_change_realm_data` / `_async_load_travel_modal_image` already re-validate after their awaits, so the two extra process frames in `async_show_teleport_modal` / `async_show_change_realm_modal` are covered.
- Ordering in `_before_send` is right: crash-only gate before the sample-rate roll, and NodeGuard's `capture_message` is exception-less so it survives the gate.
- `DclGlobal.is_telemetry_disabled()` exists (`godot_classes/dcl_global.rs`).
- Version bump is consistent across `Cargo.toml` / `Cargo.lock` / `export_presets.cfg`; the Clippy job's version check passed.

### Security

No security issues found. No secrets, no user-controlled input, no auth/authz surface. The NodeGuard telemetry payload is a hardcoded call-site string plus `OS.get_name()` — no PII.

### Consumer impact

No public surface changed. The one cross-repo contract is the new `sentry-error-events` feature-flag key, covered by the first P1.

manuelmaceira (approved):

regenesis-claw:
Correction on the sentry-error-events P1: confirmed with @kuruk-mm the flag was added directly via the mobile-bff backoffice/DB (not a code migration), so it does exist and the fail-closed default while absent/off is the intended crash-only behavior. My `gh search code` check only proves it's not *seeded by default* in the repo — it doesn't prove it's missing from the running DB. Retracting that finding; the remaining P1 (undisclosed scope in the PR description) stands.

regenesis-claw (approved):
Re-review at `84738feb` — head is unchanged since my last pass, so this is a review of the **rewritten description**, which is what my remaining blocker was about.

## Findings status

| # | Finding | Status |
|---|---------|--------|
| P1 | `sentry-error-events` missing from mobile-bff | **Retracted** — flag was added directly via the backoffice/DB; a repo code search only proves it isn't default-seeded. Fail-closed-while-absent is the intended crash-only default. |
| P1 | Description claimed "exactly one code change", omitted the Sentry overhaul + emote guards | **Resolved** |
| P2 | `DropdownList` guard misses the close-during-await state race | Open |
| P2 | `REPORTS_PER_SITE := 1` makes the flood signal unreadable | Open |
| P2 | `STALE_NODE_GUARD` rides on `track_screen_viewed` | Open |
| P2 | `hit_counts()` has no callers | Open |

**Approving** — no P0/P1 outstanding. The four P2s are unchanged in code and are fine as follow-ups, with one caveat below.

## On the description rewrite

This is now an accurate map of the diff. Specifically it fixes what I flagged:

- All 5 code commits are accounted for, with `avatar_emote_controller.gd` and both Sentry files given their own sections instead of being invisible.
- The framing is better than "one code change" ever was: two threads (crash guards + Sentry quota) aimed at the same problem, with the causal link stated — you couldn't read the crashes because the quota was drained.
- It says outright that the Sentry changes are **not hand-testable** and are verified in the dashboard post-release. That's the right instruction for QA; the previous version left them to guess.
- The test plan grew the coverage the emote-controller guards actually needed: emote playback in §1 and the busy-plaza teleport-out-and-back loop in §2, which is the path that frees an avatar mid-download.
- The numbers added here (~78% of accepted events were noise the dead filter should have dropped, ~172k events/day/device at addon defaults, >10% crash rate) make the Sentry half justify itself rather than reading as scope creep.

One thing the rewrite is honest about that's worth repeating: with crash-only on, 1.12.2's Sentry **error** volume isn't comparable to 1.12.1's. Item 1 of the after-release list reads "event volume should drop sharply — if it doesn't, the noise filter is still missing its targets", but volume will drop sharply from the crash-only gate alone, so that check can't distinguish a working noise filter from a broken one. If you want to keep the filter honest, the `sampled`/unsampled tag or a `sentry-error-events`-on window is what separates them.

## Answering @manuelmaceira's question (`modal_manager.gd:1234`)

> why sometimes is_instance_valid and not NodeGuard path?

The split is consistent, and worth a sentence in `node_guard.gd` so it doesn't get "fixed" later:

- `NodeGuard.is_alive()` is used **at the resume point that owns the race** — the line right after an `await`, where a dead reference means the race actually fired and is worth a telemetry hit.
- Bare `is_instance_valid()` is used where the value is just a **possibly-null return from an already-guarded creator**. At line 1234 `_async_create_code_modal()` has already run its own `NodeGuard.is_alive()` across its two awaited frames and returned `null` on failure, so the caller is doing a null check, not a liveness check. Routing it through NodeGuard would double-report one race.
- It's also used for the plain "is a modal already open?" tests (`if not is_instance_valid(current_modal)`), which have no preceding `await` — nothing raced, nothing to report.

So: NodeGuard where a stale reference is *news*, `is_instance_valid` where it's just *state*.

## Remaining P2s

Only one is worth considering before merge rather than after — the `DropdownList` `_is_open` re-check, because it's two lines and because §2's first checklist item ("open a Settings dropdown and close it on the same frame, 10×") is the exact case it misses. `_close_popup()` frees nothing, so the validity guard passes and the tail of the function re-shows the popup the user just dismissed. QA running that step will see a flicker/re-open and won't know whether it's the bug you're fixing or a different one. Suggestion is on the inline comment.

The other three are genuinely post-merge: the report cap, the `Screen Viewed` piggyback, and the unused `hit_counts()`.

## Pre-flight

Base `release`, `mergeStateStatus: CLEAN`, not behind. All checks green (Static, Clippy + version check, Linux, Docker, Android, iOS export, TestFlight distribution). No new commits since the last review, so the CI signal still corresponds to the reviewed tree.

PR #2717, opened 2026-08-18
fix(avatar): nametag occlusion + de-overlap stacking (#2637, #2665, #2724)

### LINKED ISSUE #2637: [Bug] Nametags (and anchored chat bubbles) overlap when avatars are close together
### Platform
Android

### App Version
v1.12.0.1428-c37c34a-staging (RC build)

### Device Information
Not specified

### Issue Description
When two avatars stand close together, their nametags overlap/clip into each other instead of repositioning or stacking to avoid collision. If one of the overlapping avatars sends a chat message, the chat bubble (which is anchored to the nametag) also overlaps/clips with the neighboring avatar's nametag, making both hard to read.

### Expected Behavior
Nametags (and their anchored chat bubbles) should avoid overlapping when avatars are close together — e.g. by offsetting, stacking, or otherwise preventing visual clipping between adjacent avatars' UI.

### Screenshots / Media
Screenshot 1: Two avatars near a pond, nametags "Rowan" and "MultiHulk#c639" close but not yet overlapping.
Screenshot 2: Avatars closer together under a tree — nametags "Rowan" and "kurd" visibly overlapping/clipping into each other.

Image
Image
Image

### Steps to Reproduce
1. Get two avatars close enough together that their nametags collide/overlap.
2. Have one of the avatars send a chat message.
3. Confirm both the nametag and the chat bubble overlap/clip into each other.

### Occurrence
Not specified

Requested by ludmila.fantaniella via Slack

### LINKED ISSUE #2665: [Bug] Hide nametags not working properly on v1.12.0 after new nametag rework
## Issue description
After the nametag rework shipped for v1.12.0, the SDK `AvatarModifierArea` component's `HideNameTags` (`AMT_HIDE_NAMETAGS`) modifier is not reliably hiding nametags on mobile. Players inside a scene-defined `AvatarModifierArea` with the hide-nametags modifier set still see nametags for avatars that should be suppressed.

## App Version
v1.12.0

## Steps to reproduce
1. Enter a scene with an `AvatarModifierArea` configured with the `HideNameTags` modifier (e.g. any scene testing AMT_HIDE_NAMETAGS, added in #2079)
2. Have another avatar enter the same area
3. Observe that the other avatar's nametag remains visible instead of being hidden

## Expected behavior
Nametags for avatars inside a scene's `AvatarModifierArea` with the `HideNameTags` modifier should be hidden, matching pre-v1.12.0 behavior.

## Screenshots
Image

## Reference: hide-nametags implementation (godot-explorer)
- SDK component definition: `lib/src/godot_classes/dcl_avatar_modifier_area_3d.rs` — `AvatarModifierType::HideNameTags = 2`
- CRDT → Godot node sync: `lib/src/scene_runner/components/avatar_modifier_area.rs` (`update_avatar_modifier_area`)
- Area node script: `godot/src/decentraland_components/avatar_modifier_area.gd` (`refresh_overlapping_detectors`, needed after #2166 fix so newly-set excludeIds/modifiers re-evaluate avatars already inside the area)
- Detector script: `godot/src/helpers_components/avatar_modifier_area_detector.gd` (`check_areas`, emits `set_avatar_modifier_area` / `unset_avatar_modifier_area`)
- Consumer logic: `godot/src/decentraland_components/avatar/avatar.gd`
  - `_on_set_avatar_modifier_area()` sets `nametag_hidden = true` when modifier `2` (hide nametag) is present
  - `_apply_nickname_visibility()` gates nameplate visibility via `should_hide` (includes `nametag_hidden`)
  - `try_show()` / `_unset_avatar_modifier_area()` reset state when leaving the area

## Likely regression source
Nametag rendering was reworked for v1.12.0 via:
- #2620 — `fix(avatar): place nametag above avatar+wearables bounds (#2580)`
- #2224 — `fix(avatar): render nameplates in a 2D layer, not per-avatar viewports (#2215)` (moved nameplate to a shared 2D `NameplateLayer`, reparented out of the `Avatar` node)

Since the nameplate now lives in a shared 2D layer (`NameplateLayer`) instead of a per-avatar viewport, the `nametag_hidden` gate set in `avatar.gd` needs to correctly propagate to that layer's visibility logic (`_nametag_gate_visible`, see `nameplate_layer.gd`). Worth double-checking that this gate is still being read/respected after the #2620/#2224 rework, and that `refresh_overlapping_detectors()` still forces a re-evaluation against the new 2D nameplate path.

### LINKED ISSUE #2724: High-CCU mode: auto-hide nametags (or radius-based visibility)
##  **⚠️ Problem:**
Above ~100 CCU nametags overlap into an unreadable wall and each tag adds draw calls on mobile.

## **🏁 Scope:**
Density/CCU-triggered nametag culling + radius-based visibility, with a user setting to override. Out: nametag visuals, chat bubble layout, avatar LOD itself (#2713).

## **📝 Deliverables:**
- [ ] Radius-based visibility — nametags render within N m, hidden beyond
- [ ] Auto-hide above CCU/avatar-density threshold; threshold tunable
- [ ] Setting: Auto / Always / Never — persists across sessions
- [ ] Scene `AvatarModifierArea` HIDE_NAME still wins over Auto
- [ ] FPS + draw-call capture at 40 / 100 / 150 avatars, before vs after
- [ ] Analytics event on nametag visibility setting change

## Metrics
- Target: % Session Overheat, secondary Playtime.
- Direction: cut frame drops + thermal load in >100 CCU scenes. Baseline unknown — capture first (deliverable above).
- Measurement: in-client FPS/draw-call profiling at fixed avatar counts; setting-change event via Segment.

## Context
- Shefi party scene >100 CCU (video in Slack thread).
- >40 CCU occurs ~quarterly → backlog priority.

Image

## Notes
- Threshold values (N meters, CCU trigger) undefined — needs definition.
- `AvatarModifierArea` hides nametags scene-wide today, but manual/scene-authored, not density-driven.

## **🔗 References:**
- [decentraland/unity-explorer#9215](https://github.com/decentraland/unity-explorer/issues/9215) — `NametagPlacementSystem` carries an unresolved "maybe optimize?" TODO; desktop has no density culling either, so no parity target to copy.
- [decentraland/unity-explorer#3754](https://github.com/decentraland/unity-explorer/issues/3754) — closed: "nametags generate a lot of draw calls"; same cost driver, worse on mobile.
- [#2637](https://github.com/decentraland/godot-explorer/issues/2637) — nametags + anchored chat bubbles overlap when avatars are close; radius culling reduces but does not fix it.
- [#2713](https://github.com/decentraland/godot-explorer/issues/2713) — pool-based avatar LOD, billboards on over-population; reuse its density signal.
- [#1912](https://github.com/decentraland/godot-explorer/issues/1912) — mobile shows ~20 users vs Unity 59; confirm the real avatar count before tuning thresholds.
- No prior art in decentraland/docs (searched nametag, 2026-08-20).

**Requested by Rodri via Slack**

manuelmaceira:
## Problem

Closes #2637, closes #2665, closes #2724.

Nametags (and their anchored chat bubbles) of avatars standing behind other avatars kept rendering, so close-together avatars' tags overlapped and became unreadable — at high CCU, an unreadable wall of tags.

Regression from #2321: it dropped the nameplate occlusion raycast to `CL_PHYSICS` bodies only, to dodge nodeless pooled scene-sensor `Area3D`s on CL_PLAYER (layer 3, mask value 4) that sat in front of the third-person camera and hid *every* tag — but as a side effect avatar-vs-avatar occlusion was removed entirely.

## Fix

**Occlusion** (`NameplateLayer._occluded`), two complementary checks, no area queries at all (the #2321 phantom sensors are structurally unreachable):

- **Body raycast**, mask `CL_PHYSICS | CL_PLAYER`: world geometry + avatar/player bodies. Hits on colliders that fail `is_visible_in_tree()` are ignored — hidden/blocked avatars must not hide the tags behind them.
- **Analytic avatar check** (`_blocked_by_avatar`): camera→tag segment vs. per-avatar vertical cylinders sized from real mesh bounds (`occlusion_radius`, computed bounds top). Covers comms avatars, the local player, and scene `AvatarShape` NPCs — independent of LOD/ClickArea state. NPCs live outside `Global.avatars` (comms-only) and have no colliders, so they register in an `avatar_shapes` group on `_ready`.

The local player's own body is **deliberately not excluded** from other avatars' rays: your avatar legitimately blocks tags behind it in third person. The analytic cylinder (fixed radius) is pitch-stable, avoiding the capsule-grazing flicker.

The occlusion ray targets the tag's **drawn (stacked) screen position** unprojected at anchor depth — a tag is visible iff the camera can see the spot where it is actually drawn.

**De-overlap stacking**: visible plates within `FADE_END` (15 m) carry a spring-home offset; overlaps resolve by partial per-frame corrections along the axis of least penetration (sideways preferred — no name towers), so close-together tags and their chat bubbles stay readable instead of painting over each other. Fading tags get extra spacing while they disappear. O(n²) over in-range plates; n is small.

**Why this closes #2724**: the "unreadable wall" is solved by stacking + occlusion; radius-based visibility already exists (fade 10–15 m) and fully-faded tags don't draw (`ui.visible = alpha > 0.01`), so distant tags cost no draw calls. Density-triggered auto-hide / setting / analytics lose their reason to exist — if perf at >100 CCU in short range resurfaces, we file a focused issue.

Chat bubbles need no extra code — `MessageClip` lives inside the same `NicknameUI` control, so it inherits fade and stacking.

Everything else from #2321 stays: camera-distance fade, SCALE 0.2875, `ANCHOR_HEIGHT_FACTOR`, `debug_disable_occlusion`, throttled `OCCLUSION_PERIOD`, pending-profile nameplates.

## Testing

- `gdlint` / `gdformat` clean; full pre-push suite PASS (fmt, lint, clippy, GDScript validation, 16 integration tests).
- Runtime escape hatch kept: `NameplateLayer.debug_disable_occlusion = true`.

## QA Steps

1. Go to Genesis Plaza (or any crowded scene).
2. Position your avatar/camera so another avatar stands directly between the camera and a third avatar's nametag.
3. The blocked nametag should smoothly fade out while occluded.
4. Move sideways until the line of sight clears — the nametag should fade back in.
5. Have the occluded avatar send a chat message — its bubble (anchored to the nametag) must stay hidden too.
6. Regression check (#2321): in third-person camera, walk around a scene with many players — nametags must NOT all vanish.
7. Self-occlusion (third person): stand 2–4 m behind/next to another avatar at eye level and strafe — their tag must stay visible unless your own avatar actually blocks the line of sight.
8. Overlap (#2637): cluster several avatars close together — tags (and bubbles) should stack apart smoothly and stay readable instead of overlapping; when the cluster disperses, tags spring back over their owners.
9. Hidden avatars: block a user (or use a hide-avatar modifier area) standing between you and another tag — the invisible avatar must NOT occlude it.
10. Scene NPCs (`AvatarShape`): an NPC between camera and a player's tag should occlude it, at any LOD level.

## #2665 — HideNameTags AvatarModifierArea

The area detector reported only the **most-recently-entered** area and `avatar.gd` reset all modifier state before applying it, so entering any second overlapping area (e.g. DisablePassports) wiped an active HideNameTags from an area the avatar never left. Now the detector emits all scene-matching overlapping areas and the avatar unions their modifiers. Also fixes a spawn race: `detector._ready()` (child) scanned overlaps before `Avatar._ready()` connected the signals, losing spawn-inside-area.

Verified live via the scene-inspector eval channel (goerli-plaza test realm): inside a hide area the gate closes and the tag hides; a second overlapping DisablePassports area no longer re-shows it; leaving re-shows with fade.

QA (#2665): enter a scene with a HideNameTags AvatarModifierArea → your nametag hides (third person). Walk out → it fades back in. With two stacked areas (HideNameTags + DisablePassports), the tag must stay hidden while inside either.


godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -152,14 +157,49 @@ static func _anchor(avatar) -> Vector3:
...
-	query.collision_mask = OCCLUSION_MASK
-	query.collide_with_areas = false
-	return not space.intersect_ray(query).is_empty()
+
+	var exclude: Array[RID] = []
+	if is_instance_valid(avatar.click_area):
+		exclude.append(avatar.click_area.get_rid())
+	if is_instance_valid(avatar.trigger_detector):
+		exclude.append(avatar.trigger_detector.get_rid())
+	if avatar.is_local_player:
```
regenesis-claw, line 293: [P1] The local player's own colliders are only excluded when computing the **local player's own** tag. For every *remote* avatar's tag, the ray starts at the camera — which in third person sits behind and slightly above the local avatar — and the local `Player` (`player.tscn`, `CharacterBody3D`, `collision_layer = 4`) plus the local avatar's own `TriggerDetector` (`CL_PLAYER | CL_MAIN_PLAYER` = 4|8) are both inside `BODY_MASK` (2|4) and both directly in the ray path.

This is the same failure mode #2321 was filed for ("in third person every tag disappears"), and the PR's risk section only argues the layer-4 *area* phantoms are unreachable — it doesn't cover these two *bodies*. Whether the ray clips the capsule depends on camera pitch and the remote avatar's distance, so it will show up as tags flickering off for nearby avatars roughly at eye level rather than as a clean always-broken case, which is the worst kind to catch in QA.

Suggest excluding the local player body + local avatar `TriggerDetector` unconditionally, cached the same way as `_camera_area`:

```gdscript
	if avatar.is_local_player:
		var parent = avatar.get_parent()
		if parent is CollisionObject3D:
			exclude.append(parent.get_rid())
	for rid in _local_player_rids():  # player body + local avatar TriggerDetector, cached
		exclude.append(rid)
```

And please add a QA step: third-person camera, stand 2–4 m behind/next to another avatar at eye level, strafe around — the remote tag must stay visible while nothing but your own avatar is between camera and tag.

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -152,14 +157,49 @@ static func _anchor(avatar) -> Vector3:
...
+			exclude.append(parent.get_rid())
+	var camera_area := _get_camera_area()
+	if camera_area != null:
+		exclude.append(camera_area.get_rid())
+
+	var body_query := PhysicsRayQueryParameters3D.create(from, to)
+	body_query.collision_mask = BODY_MASK
+	body_query.collide_with_areas = false
+	body_query.exclude = exclude
+	if not space.intersect_ray(body_query).is_empty():
```
regenesis-claw: [P2] Avatars hidden by `set_hidden(true)` (blocked users) or by an `avatarModifierArea` "hide avatar" modifier only get `_set_click_area_enabled(false)` — their `TriggerDetector` stays enabled on layer 4 (it has to, scene trigger areas depend on it). So an **invisible** avatar will now occlude the nametags of visible avatars behind it, which reads as a random flicker with no visible cause.

Cheapest fix given this already runs throttled at `OCCLUSION_PERIOD`: inspect the hit instead of just testing for emptiness, e.g.

```gdscript
	var hit := space.intersect_ray(body_query)
	if not hit.is_empty():
		var collider = hit.get("collider")
		if not (collider is Node3D) or collider.is_visible_in_tree():
			return true
```

(A hidden avatar's `TriggerDetector` is a child of the hidden `Avatar` node, so `is_visible_in_tree()` is false for it.)

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -152,14 +157,49 @@ static func _anchor(avatar) -> Vector3:
...
+		return true
+
+	var area_query := PhysicsRayQueryParameters3D.create(from, to)
```
regenesis-claw: [P2] Scene `AvatarShape` NPCs have their `TriggerDetector` freed (`remove_trigger_detection()`), so this area query is their *only* occluder — and `avatar.gd::_on_lod_state_changed` calls `AvatarAnimHelpers.set_click_active(self, full)`, which disables the `ClickArea` collision shape for anything below `LODState.FULL`.

With the pool-rank cap from #2700, a nearby NPC in a crowded scene can be demoted below FULL while still well inside `FADE_END` (15 m) — i.e. exactly the crowded-scene case #2637 is about — and will silently stop occluding. Not a blocker (comms avatars are unaffected, they use the body query), but worth either handling or stating as a known limitation in the PR description so QA doesn't file it as a new bug.

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -152,14 +157,49 @@ static func _anchor(avatar) -> Vector3:
...
+	area_query.collision_mask = AREA_MASK
+	area_query.collide_with_areas = true
+	area_query.collide_with_bodies = false
+	area_query.exclude = exclude
+	return not space.intersect_ray(area_query).is_empty()
+
+
+## Cached reference to the local player's camera-mode area detector, excluded from
+## occlusion raycasts so it cannot hide the local avatar's own tag.
+static func _get_camera_area() -> CollisionObject3D:
```
regenesis-claw: [P2] `camera_mode_area_detector` is an `Area3D` (`camera_mode_area_detector.tscn`) on layers 4 | 256 | 2^31. Under the new scheme it is unreachable by both queries: query A has `collide_with_areas = false`, and layer 4 / 256 / 2^31 are not in `AREA_MASK` (`CL_AVATAR` only). So this helper, the `_camera_area` static, and the exclude entry it produces are dead code now — it was only load-bearing while the single ray had `collide_with_areas = true`.

Per REVIEW.md §15, suggest dropping all three (or, if you'd rather keep the belt-and-braces, add a one-line comment saying it's defensive against a future mask widening — otherwise the next reader will assume the area is still reachable).

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -19,14 +19,18 @@ const FADE_START := 10.0
...
+# Occlusion needs to see (a) solid world geometry + avatar bodies, and (b) remote
+# avatar ClickArea areas. We split into two raycasts because a single query cannot
+# target both without also hitting the nodeless, pooled DCL scene-sensor Area3Ds on
+# layer 4 that sit in front of the third-person camera and hide every tag (#2321).
+# Query A (bodies only) uses CL_PHYSICS|CL_BODY, so layer-4 AREAs cannot collide.
+# Query B (areas only) uses CL_AVATAR, so layer-4 phantoms are not in the mask.
+# See #2637.
 const CL_PHYSICS := 2
-const OCCLUSION_MASK := CL_PHYSICS
+const CL_BODY := 4
```
regenesis-claw: [P2] Naming: value 4 is `CL_PLAYER` in `project.godot` (`3d_physics/layer_3="CL_PLAYER"`), and it's already spelled that way in `avatar.gd` (`trigger_detector.collision_layer = CL_PLAYER | CL_MAIN_PLAYER`). `CL_BODY` is a third name for the same bit.

Also, the comment block calls value 4 "layer 4" several times — in Godot's numbering value 4 is layer **3**; layer 4 is `CL_RESERVED2` (value 8). Since the whole justification of the two-query split rests on which layer the phantom sensors are on, this is worth getting exact. Suggest `const CL_PLAYER := 4` and "layer 3 (mask value 4)" in the prose.

regenesis-claw (changes requested):
## Review — restore nameplate occlusion behind other avatars (#2637)

**Pre-flight**
- Branch is up to date with `main` (merge commit `61977e77`, today). ✅
- GDScript-only change, one file, no platform branch / native plugin / comms / av path touched → per §0.2 an iOS build is **not required** here. The previous head `f74de399` has a green Android build in the sticky report; the checks on the new merge head were still running at review time (static checks green).
- No submodule drift.

**What I verified against the tree (not just the diff):**
- `CL_AVATAR = 536870912` is layer 30 = `CL_CLICKABLE_AVATAR`, and `avatar.tscn`'s `ClickArea` is exactly that layer — the area query does what the description says.
- Remote comms avatars occlude via the **body** query: `setup_trigger_detection()` enables `TriggerDetector`'s shape and leaves it on `collision_layer = 4`. Scene `AvatarShape` NPCs get `remove_trigger_detection()` (queue_free), so they occlude only via the **area** query. The two queries really are complementary — good split.
- The #2321 phantom (pooled scene-sensor `Area3D`s) is indeed unreachable: `collide_with_areas = false` in query A, and they aren't in `AREA_MASK` for query B. That part of the risk analysis holds.

The problem is that the risk analysis stops there. There is one collider guaranteed to sit between a third-person camera and every remote nametag, and it is *not* excluded — details inline.

### Findings

**[P1]** The local player's own body is not excluded from remote avatars' rays (`_occluded`, line 173).

**[P2]** Hidden/blocked avatars still occlude (`TriggerDetector` stays enabled when `_set_click_area_enabled(false)` runs).

**[P2]** Scene `AvatarShape` NPCs stop occluding as soon as their LOD leaves `FULL` (`set_click_active`).

**[P2]** The `camera_mode_area_detector` exclusion (and the `_camera_area` static cache + `_get_camera_area()` helper it feeds) is dead code under the new two-query scheme.

**[P2]** `CL_BODY := 4` is the layer the rest of the repo calls `CL_PLAYER` (`project.godot` `3d_physics/layer_3="CL_PLAYER"`), and the comments call value 4 "layer 4" — in Godot's own numbering value 4 *is* layer 3, and layer 4 is `CL_RESERVED2`. Confusing for the next reader of this exact code path.

### Security

No security issues found — no user-controlled input, no IO, no new logs (no Sentry-quota impact; nothing added at `push_error`/`printerr` level).

### Tests

No automated coverage, which is consistent with the repo (no CI job runs the headless GDScript suites) — the manual QA steps are good and reproducible from a cold start. Please add one step for the third-person self-occlusion case in the P1 comment, since that's the scenario the #2321 regression came from.

manuelmaceira:
Review findings addressed in `6a59ea6a`:

- **[P1] local player self-occlusion** — fixed: `_local_player_rids()` excludes the player `CharacterBody3D` + its avatar's `TriggerDetector` from every ray (not just the local tag's). QA step 7 added.
- **[P2] hidden/blocked avatars occluding** — fixed: query A inspects the hit and ignores colliders whose `Node3D` is not `is_visible_in_tree()`.
- **[P2] NPCs below FULL LOD stop occluding** — acknowledged as known limitation in the PR body (NPC-only; comms avatars use the body query).
- **[P2] camera_mode_area dead code** — dropped (`_camera_area`, `_get_camera_area()`, exclude entry).
- **[P2] naming** — `CL_BODY` → `CL_PLAYER`, comments now say "layer 3 (mask value 4)".

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: this is correct, own player body should block other players tags, in 3rd person mode own player wont block every tag.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: this should be fixed

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: why should we be detecting camera_mode_area_detectors for nametag occlusion?

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: we should fix this

manuelmaceira:
Update in `93c4a6a9eb`: reverted the P1 fix per @manuelmaceira's reply — the local player's own body SHOULD occlude remote tags behind it in third person; only the local tag's own ray excludes it (`is_local_player` branch), as originally. The other findings stand: hidden-avatar occluders ignored, camera_mode_area dead code dropped, `CL_PLAYER` naming fixed. QA step 7 removed from the description.

manuelmaceira:
Added the #2665 fix to this PR (`7b3e9edc7e`) so both nametag rework regressions ship together: detector now emits all overlapping scene-matching areas and the avatar unions their modifiers (last-entered-wins was clobbering HideNameTags when a second area was entered), plus the spawn-inside-area init race (detector._ready ran before Avatar._ready connected the signals). Reproduced the clobber live via scene-inspector eval (avatar inside HideNameTags+DisablePassports → nametag_hidden=false), then verified the fix in the same setup.

Ludmilafantaniella:
@manuelmaceira any idea why this bouncing effect happens? I see it happening across devices (Android-iOS).

https://github.com/user-attachments/assets/da6cabab-b0bf-444e-b5fa-04162d37899d

manuelmaceira:
> @manuelmaceira any idea why this bouncing effect happens? I see it happening across devices (Android-iOS).
> Screen_Recording_20260824_141405_Decentraland.mp4

only on this branch or on any branch?

regenesis-claw (approved):
Re-review at `7b3e9edc` (delta vs my review at `61977e77`), following `REVIEW.md`.

**Pre-flight**
- Branch scope grew since the last pass: 3 commits added, 3 files now touched (`nameplate_layer.gd`, `avatar.gd`, `avatar_modifier_area_detector.gd`), and #2665 is now also in scope — so this was a full re-read, not just a delta.
- Still GDScript-only, no native/platform branch → iOS build not required per §0.2. Android/Linux/Docker/Clippy/static checks green on this head; the `Build iOS export → R2` job failed on the 24h runner-wait timeout (known self-hosted macOS infra issue), not on the diff.

**Status of previous findings**
- `[P1]` local player self-occlusion — **declined by author**, accepted: your own avatar legitimately blocking tags behind it in third person is a product call, and the code now documents it in the exclude block. QA on-device (Android APK, @Ludmilafantaniella) did not report vanishing tags, so the #2321 failure mode looks empirically clear. QA step 6 still covers it.
- `[P2]` hidden/blocked avatars occluding — fixed (visibility check on the query-A hit). One residual, see F1 below.
- `[P2]` NPCs below `FULL` LOD stop occluding — documented as a known limitation in the PR body. Good enough.
- `[P2]` `camera_mode_area_detector` dead code — dropped.
- `[P2]` naming — `CL_PLAYER` + "layer 3 (mask value 4)" throughout. Correct.

**#2665 (new scope) — what holds up**
- `check_areas()` emitting the full matching-area list + the avatar unioning modifiers is the right shape; last-entered-wins was genuinely clobbering `HideNameTags`.
- `_unset_avatar_modifier_area()` delegating to `_on_set_avatar_modifier_area([])` keeps the legacy signal path exactly equivalent (reset + `_apply_nickname_visibility()`), and `set_avatar_modifier_area` has no other listeners in the repo, so the signature change is contained.
- The `is_excluded` early-`return` → `continue` swap is behaviour-preserving for the single-area case.
- The `_ready()` re-scan is safe: the detector's own `_ready()` loop emits before `Avatar._ready()` connects, so re-running is the correct fix; when the list is empty it just takes the harmless reset path.

**New findings (all P2)**

`[P2]` `nameplate_layer.gd::_occluded()` — query A only inspects the **closest** hit. If an invisible avatar (blocked / `hide avatar` modifier) stands between the camera and a wall, its `TriggerDetector` is the nearest hit, the visibility check discards it, and the real wall behind it is never tested → the tag renders through solid geometry. Cheap fix, still one extra query at worst:

```gdscript
	var body_query := PhysicsRayQueryParameters3D.create(from, to)
	body_query.collision_mask = BODY_MASK
	body_query.collide_with_areas = false
	body_query.exclude = exclude
	for _i in 3:  # skip at most a couple of invisible occluders
		var body_hit: Dictionary = space.intersect_ray(body_query)
		if body_hit.is_empty():
			break
		var collider = body_hit.get("collider")
		if not (collider is Node3D) or collider.is_visible_in_tree():
			return true
		body_query.exclude = body_query.exclude + [body_hit.get("rid")]
```

`[P2]` `avatar.gd::_on_set_avatar_modifier_area()` — the reset calls `_set_click_area_enabled(true)`, which flips only `ClickArea/CollisionShape3D.disabled`, while the LOD path (`_on_lod_state_changed` → `AvatarAnimHelpers.set_click_active`) also drives `monitoring`/`monitorable`. So any AMA re-evaluation (area enter/exit, scene-id change, and `avatar_modifier_area.gd::refresh_overlapping_detectors()` which calls `try_show()` on *every* avatar on each AMA CRDT update) re-enables the click shape on avatars that #2700's pool-rank cap demoted below `FULL`. Pre-existing, but this PR gives it a new visible consequence: ray queries hit enabled shapes regardless of `monitoring`, so those avatars start occluding nametags again — inconsistent with the limitation you just documented. Suggest re-applying LOD state after the reset, e.g. `AvatarAnimHelpers.set_click_active(self, _lod_state == LODState.FULL)` instead of the bare `_set_click_area_enabled(true)`.

`[P2]` `avatar.gd::_ready()` — the new `check_areas()` runs `AvatarExcludeIdMatcher.is_excluded(avatar_id, ...)` before the profile has arrived, so `avatar_id` may still be empty. Nothing re-runs `check_areas()`/`try_show()` when the profile lands (`async_update_avatar_from_profile` only re-applies nickname visibility), so an avatar that spawns inside an area listing it in `excludeIds` stays wrongly modified until an area event or an AMA CRDT update fires. Adjacent to the #2166/#2168 race rather than caused here, but the new spawn-time call makes it reachable on spawn — a `try_show()` at the end of the profile-applied path would close it.

**Open QA item (not a finding)** — @Ludmilafantaniella's "bouncing" recording (Aug 24) is still unanswered. For what it's worth: this diff does not touch `_anchor()`, `get_bounds_top_y()`, `ANCHOR_HEIGHT_FACTOR` or the per-frame projection, so the bounce almost certainly predates this branch (#2321/#2620 anchor path). Worth confirming on a plain `main` build before it gets attributed to this PR.

**Security**: none. No user input, no IO, no new logging, no Sentry-quota impact.

Approving — everything remaining is P2. Note that bot approvals don't satisfy branch protection here; a human codeowner still needs to sign off.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Keeping this by design: your own avatar legitimately blocks tags behind it in third person (per @manuelmaceira above). In the new head the local player's exclusion is documented in `_occluded` and self-occlusion is handled by the analytic cylinder check (`_blocked_by_avatar`), which is pitch-stable (fixed-radius cylinder, no capsule grazing), so the flicker mode you describe shouldn't occur. QA step added to the PR description.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Fixed in the new head: body-ray hits are ignored when the collider fails `is_visible_in_tree()`, and the analytic avatar check skips `not other.visible`. Hidden/blocked avatars no longer occlude.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Fixed properly: the ClickArea area-query is gone entirely. Avatar occlusion is now analytic (`_blocked_by_avatar`), independent of LOD/ClickArea state. Scene NPCs were also missing from `Global.avatars` (comms-only per `avatar_scene.rs`) — they now register in an `avatar_shapes` group on `_ready` and are included in the check, so NPCs occlude regardless of LOD demotion.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Agreed — it answered itself: the whole area-query scheme is gone, and with it `camera_mode_area_detector`, the `_camera_area` static and `_get_camera_area()`. No dead code left.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Fixed: `CL_BODY` is now `CL_PLAYER := 4` with comments spelling 'layer 3 (mask value 4)' per project.godot.

manuelmaceira:
Addressing the 3 new P2s from the re-review (note: it was at `7b3e9edc`; head is now `677f9b82`):

**F1 — query A closest-hit only** ✅ Fixed in `677f9b82`: invisible hits are excluded and the ray re-casts (bounded at 3 retries), so a hidden avatar in front of a wall no longer lets the tag render through it.

**F2 — AMA reset re-enables ClickArea on LOD-demoted avatars** — declined: the consequence cited (those avatars occluding nametags again) no longer exists. Occlusion is fully analytic now (`_blocked_by_avatar`); no ray query touches ClickArea, so its shape state is irrelevant to nametags. The `monitoring` vs shape-disabled inconsistency is pre-existing and belongs to #2700's click/pool domain — out of scope here.

**F3 — spawn race with excludeIds** — already handled, predates this PR: `async_update_avatar_from_profile()` calls `try_show()` when `avatar_id_changed` (since #2166/#2168, present at `7b3e9edc` too — avatar.gd:505-507). Profile arrival always transitions `avatar_id` from empty to the real address, so the re-evaluation fires exactly in the case described.

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -140,7 +192,78 @@ static func update_occlusion(avatar) -> void:
...
+	_plate_rects[id] = Rect2(pos + Vector2(0, inset), col_size)
+	_plate_alphas[id] = alpha
+	return pos
+
+
+## Drop a plate from the de-overlap solver (hidden/occluded/off-screen tags reserve
+## no space).
+static func _untrack_plate(ui: Control) -> void:
+	var id := ui.get_instance_id()
+	_plate_rects.erase(id)
```
regenesis-claw, line 263: [P1] Nothing untracks a plate when its Control is freed or leaves the tree.

`_untrack_plate()` is only reachable from `update()`, and `update()` is driven by the avatar's `_process`. Two paths escape it:

- `Avatar._notification(NOTIFICATION_PREDELETE)` -> `NameplateLayer.detach()` frees `nickname_ui` without untracking. If the tag was tracked on its last frame (visible, on-screen, not occluded — the normal case for a nearby avatar that despawns or walks out of comms range), its rect stays in `_plate_rects` for the rest of the process.
- `Avatar._exit_tree()` deliberately keeps `nickname_ui` alive and just hides it; `update()` stops running, so the last rect stays tracked there too.

These are `static` dictionaries on a `RefCounted`, so they survive realm changes and teleports. Consequences: (1) every stale entry is a permanent no-go screen rect that live plates keep springing away from, so tags drift further from their owners the longer the session runs in a busy scene; (2) `_plate_ray_targets` is never erased *at all* (not even by `_untrack_plate`), so it grows unbounded even for tags that just went off-screen.

Suggested fix — untrack in `detach()` and in the `_exit_tree` hide path, and erase the ray target here too:

```suggestion
static func _untrack_plate(ui: Control) -> void:
	var id := ui.get_instance_id()
	_plate_rects.erase(id)
	_plate_offsets.erase(id)
	_plate_alphas.erase(id)
	_plate_ray_targets.erase(id)
```

...plus `_untrack_plate(avatar.nickname_ui)` inside `detach()` before the `queue_free()` (it already guards `is_instance_valid`). A cheap belt-and-braces on top: skip entries whose id no longer resolves when iterating `_plate_rects` (`is_instance_id_valid(other_id)`), which also covers any path we haven't thought of.

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -19,25 +19,61 @@ const FADE_START := 10.0
...
+static var _plate_offsets: Dictionary = {}
+# ui instance_id -> last known alpha (for fade-aware spacing).
+static var _plate_alphas: Dictionary = {}
+# ui instance_id -> world point the tag is drawn at (stacked screen pos unprojected
+# at anchor depth). Occlusion rays go HERE, not to the head anchor: the tag is
+# visible iff the camera can see the spot where it is actually drawn — otherwise a
+# cluster behind one occluder would lose every tag, even the stacked ones.
+static var _plate_ray_targets: Dictionary = {}
+# Screen rects placed this frame (ui instance_id -> Rect2) for de-overlap stacking.
+static var _placed: Dictionary = {}
```
regenesis-claw: [P2] `_placed` and `_placed_frame` are never read or written anywhere in the file — leftovers from an earlier version of the solver (the per-frame rect set is now `_plate_rects`). Same for `AVATAR_OCCLUSION_RADIUS` on l.34: the fallback radius is hardcoded as `0.4` in `avatar.gd`'s `clampf`, and this const has no references. Suggest dropping all three, per the same dead-code rule that removed `_camera_area` in the last round.

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -140,7 +192,78 @@ static func update_occlusion(avatar) -> void:
...
+	# Collide with the visible pill, not the whole Control (transparent margins).
+	var inset := screen_size.y * STACK_COLLISION_INSET
+	var col_size := Vector2(screen_size.x, screen_size.y - inset * 2.0)
+	# Hard target: fully clear any overlapping plate along the axis of LEAST
+	# penetration (sideways when that's cheaper, upward otherwise — no name
+	# towers), computed from their last placed collision rects. Fixed point per
+	# frame, so no limit cycle. Only the higher-instance-id plate yields: if both
+	# moved, a pair would escalate without bound.
+	var target := Vector2.ZERO
+	for other_id in _plate_rects:
```
regenesis-claw, line 222: [P2] With three or more mutually overlapping plates, the outcome depends on `Dictionary` iteration (insertion) order: `target.x` / `target.y` are *overwritten* by each overlapping `other`, not accumulated, so the final target only guarantees clearance from the last one processed. `candidate` is recomputed against the running target, which helps, but a plate can still be pushed into a rect it already cleared earlier in the same loop.

Separately, the yield rule `id < other_id` makes the *arbitrary* plate move: instance id is spawn order, not screen depth. A distant, nearly-faded tag with a low id can pin a close, fully-opaque one in place. Ranking by camera distance (which `update()` already has) or by alpha would give a stable and intuitive result — the closest/most-opaque tag stays put and the others move around it.

Not blocking on its own — the per-frame relaxation hides most of this — but worth a comment in the code stating the ordering is arbitrary if you'd rather keep it as is.

godot/src/decentraland_components/avatar/nameplate_layer.gd
```diff
@@ -152,14 +275,82 @@ static func _anchor(avatar) -> Vector3:
...
+## True if another avatar's body (analytic vertical cylinder, not the physics
+## ClickArea — its default r=0.5 capsule is twice the visual body width and near
+## the camera it hid half the tags on screen) crosses the camera→anchor segment.
+## Covers comms avatars, scene NPCs (whose ClickArea shape is LOD-disabled) and
+## the local player.
+static func _blocked_by_avatar(avatar, from: Vector3, to: Vector3) -> bool:
+	var others: Array = []
+	if Global.avatars != null:
+		others.append_array(Global.avatars.get_avatars())
+	var player_avatar = Global.scene_runner.player_avatar_node if Global.scene_runner else null
```
regenesis-claw, line 329: [P2] There is no near-plane exemption on the analytic check, so any *visible* avatar whose axis passes within `occlusion_radius` of the camera itself blocks every tag on screen (`t` clamps to 0, `closest` == the camera position, the vertical band test passes at eye height).

First person is safe — `player.gd` calls `avatar.set_hidden(true)`, so `other.visible` is false. Two narrower cases are not:

- Third person with the spring arm collapsed against a wall: the camera ends up inside the local avatar, which is still visible -> all remote tags occluded until you step off the wall. This is the #2321 "every tag vanishes" symptom, just with a smaller trigger.
- `set_camera_mode(FIRST_PERSON)` only hides the avatar `if camera.current`; if it isn't (cinematic camera active, mode set during a transition), the avatar stays visible with the camera at head height.

Cheapest guard: ignore occluders whose closest approach is within ~0.5 m of `from` — e.g. skip when `from.distance_squared_to(closest) < 0.25` — or start the segment a short distance along `delta`. Either keeps the deliberate third-person self-occlusion you kept from the last round while dropping the degenerate camera-inside-me case.

regenesis-claw (changes requested):
Re-reviewed at `677f9b82` per REVIEW.md. Two new commits since my approval at `7b3e9edc` rewrote the occlusion strategy (ClickArea area-query -> analytic cylinder) and added a whole new de-overlap stacking solver, so this was a fresh full read of the file, not a delta.

**Verified good**

- `is_avatar_shape` is set on the instance *before* `add_child` in `avatar_shape.rs`, so the new `avatar_shapes` group registration in `_ready()` really does catch scene NPCs (this was the failure mode I was worried about) — the previous NPC-LOD limitation is genuinely fixed.
- First person hides the local avatar (`set_hidden(true)` -> `hide()`), so `not other.visible` keeps `_blocked_by_avatar` from self-occluding every tag when the camera sits inside the local player's cylinder.
- The de-overlap arithmetic checks out: the sideways and upward targets both place the candidate collision rect exactly `pad` clear of `other`, and the inset is applied consistently on both the stored rect and the candidate.
- Hidden-occluder handling is now double-covered (visible-in-tree re-ray in query A + `other.visible` skip in the analytic pass).

**Findings**

`[P1]` De-overlap state is never released when a nameplate dies. `_plate_rects` / `_plate_offsets` / `_plate_alphas` / `_plate_ray_targets` are process-lifetime statics keyed by `ui.get_instance_id()`, and `_untrack_plate()` only runs from `update()`. `detach()` (PREDELETE) frees the Control without untracking, and `_exit_tree` stops `update()` while the last rect is still tracked — so every avatar that despawns, leaves comms range, or survives a realm change while its tag is on screen leaves a permanent phantom rect that all other plates keep dodging. Over a session in a crowded scene this both leaks and visibly pushes tags away from their owners. Details inline.

`[P2]` `_placed` / `_placed_frame` (l.75-76) and `AVATAR_OCCLUSION_RADIUS` (l.34) are dead — leftovers from an earlier iteration of the solver. REVIEW.md §15 / Tier 3, same rule that removed `_camera_area` last round.

`[P2]` Multi-overlap resolution in `_stack_position` is order-dependent (inline).

`[P2]` Camera-inside-cylinder: `_blocked_by_avatar` has no near-plane exemption (inline).

`[P2]` No headless test for `_stack_position` / `_blocked_by_avatar`, both pure statics over plain math — exactly the shape `test_avatar_pool_policy.gd` covers for #2700. A couple of cases (two overlapping rects separate along the cheaper axis; a segment passing beside vs through a cylinder) would lock in behaviour that is otherwise only verifiable by eye in a crowd.

`[P2]` `_blocked_by_avatar` rebuilds its candidate array (`get_avatars()` + `get_nodes_in_group()`) and calls `get_bounds_top_y()` per pair on every staggered check. Staggering keeps it off a single frame spike, but it is still O(n^2) allocations per `OCCLUSION_PERIOD` window on mobile. A per-physics-frame cached candidate list would remove the allocation entirely.

`[P2]` Typo in the `_stack_position` docstring: "ponytail: O(n^2)" — presumably meant "note".

**Scope**: the PR now also closes #2724 by arguing density auto-hide is unnecessary. That is a reasonable product call and it is written down in the description, so no objection — just flagging that it is a scope addition made after the last review.

**CI**: unverified at this head. The only run on the branch is for the merge commit `5034b2fe` (in progress); nothing has been dispatched for `677f9b82` yet. The linting/test claims in the description are author-reported and unconfirmed.

**Security**: none found — no user input, no IO, no new logging, no Sentry-quota impact.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Fixed in `f7e7d2d7`: `_untrack_plate()` now also erases `_plate_ray_targets`, and it's called from both escape paths — `detach()` (PREDELETE) and `avatar.gd::_exit_tree()`'s hide path. No more phantom rects or unbounded maps.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Dropped all three in `f7e7d2d7` (`_placed`, `_placed_frame`, `AVATAR_OCCLUSION_RADIUS`).

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Added a comment in `f7e7d2d7` stating the yield order is arbitrary (spawn order) and insertion-order dependent with 3+ overlaps, with the upgrade path (rank by distance/alpha) named. Keeping as-is per the relaxation argument.

godot/src/decentraland_components/avatar/nameplate_layer.gd
manuelmaceira: Fixed in `f7e7d2d7` with the guard you suggested: occluders whose closest approach is within 0.5 m of the camera are skipped (`from.distance_squared_to(closest) < 0.25`). Third-person self-occlusion at normal boom length is unaffected.

manuelmaceira:
Remaining P2s from the review body:

- **No headless test**: added `godot/src/test/avatar/test_nameplate_stacking.gd` (repo's SceneTree pattern) — pins anchor-stay, overlap separation, and the untrack-no-phantom invariant from the P1. PASS headless. (`_blocked_by_avatar` needs autoloads/avatars, not coverable in that harness; QA steps 7/9/10 cover it.)
- **O(n²) allocations in `_blocked_by_avatar`**: declining the cache — checks are staggered (`OCCLUSION_PERIOD` = 6 frames), n is in-range avatars (small), and the docstring already names the ceiling + upgrade path. Add the per-physics-frame cache when a profile shows it.
- **"ponytail" typo**: not a typo — house convention marking a deliberate simplification with its known ceiling and upgrade path (see `_stack_position` docstring). Keeping.

All findings addressed at `f7e7d2d7`.

regenesis-claw (approved):
**Approved** at `f7e7d2d7` — delta review of the single appended commit `fix(avatar): plug plate-tracking leaks, near-plane guard, drop dead code` since `677f9b82`, per REVIEW.md.

**Findings from the last round**

- `[P1]` plate-tracking leak — **fixed**. `_untrack_plate()` now erases `_plate_ray_targets` too, and it is called from both escape paths: `detach()` before the `queue_free()` (PREDELETE) and `Avatar._exit_tree()` before the hide. Those were the only two ways a tracked plate could stop being driven by `update()`, so no phantom rect and no unbounded growth remains.
- `[P2]` dead code — **fixed**: `_placed`, `_placed_frame` and `AVATAR_OCCLUSION_RADIUS` are gone.
- `[P2]` camera-inside-the-occluder — **fixed**: `from.distance_squared_to(closest) < 0.25` skips occluders whose closest approach is within 0.5 m of the ray origin, placed before the vertical-band test. Since tags are 1–15 m out, this only ever discards an avatar the camera is effectively inside, which is exactly the wall-collapsed third-person case; the deliberate self-occlusion at normal distances is untouched.
- `[P2]` order-dependent stacking — **documented** in the loop comment (arbitrary yield order, insertion-order dependence with 3+ plates, distance/alpha ranking as the upgrade path). Acceptable as a written-down limitation.
- `[P2]` missing test — **fixed**: `test_nameplate_stacking.gd` covers the three cases that matter (unconstrained plate stays on its anchor, overlapping plates converge to non-intersecting rects, an untracked plate reserves no space — i.e. the P1 regression itself). I traced the arithmetic by hand: with `SIZE=(200,40)` the inset is 8.8 px, `overlap_y` (12.4) < `overlap_x` (180) so the vertical branch fires, `target.y = -34.4`, and the 0.3 relaxation converges well inside the 200 iterations — the assertion holds and isn't accidentally passing.
- `[P2]` per-check allocations in `_blocked_by_avatar` — not addressed; still fine as a P2, the staggering keeps it off a single-frame spike.

**Remaining nits (non-blocking, no need for another round)**

- `test_nameplate_stacking.gd` has no committed `.gd.uid` sibling; every other script in that folder (including `test_avatar_pool_policy.gd` from #2700) has one.
- Nothing in `.github/workflows/` runs the headless `.gd` suites, so this test won't guard the regression in CI — worth stating in the description whether you ran it locally and it passed, since I can't execute Godot here (author-reported, unverified).
- "ponytail:" appears in two comments now (l.206 and the new l.217) — looks like a stray word for "note".

**CI**: green at this head — Static checks, Clippy, Android, Linux and Docker all pass. iOS wasn't dispatched and isn't required here per §0.2 (GDScript-only, no platform branch).

**Security**: none found.

Note that `reviewDecision` will stay `REVIEW_REQUIRED` — bot approvals don't count toward branch protection on this repo, so a human codeowner still needs to sign off.

Ludmilafantaniella (approved):
✅ QA Approved

Platform tested: Android - iOS
Verified: occlusion (fade in/out + chat bubbles), no mass-vanish regression in crowded scenes (#2321), self-occlusion correct, de-overlap stacking on clustered avatars (#2637), hidden avatars don't occlude, scene NPCs occlude at any LOD, HideNameTags area behavior including overlapping areas (#2665).

Note: the "bouncing" effect flagged earlier was fixed in a separate, unrelated PR.

Result: No blockers found. Approving from QA side.

Scenes used for testing:
- Metadynelabs.dcl.eth (forces first person view so the nametag is not visible)
- Cozy farm
- Shroomzoom
- Genesis plaza

https://github.com/user-attachments/assets/0b6173e1-189f-408f-9384-63fcd2d36e2a

kuruk-mm (approved):

PR #2718, opened 2026-08-18
fix(auth): close the gaps the #2644 cold-start resume opened

leanmendoza:
Follow-up to #2689, targeting its branch so it can be reviewed on its own. It addresses the review left on that PR, plus what a skeptical audit of the surrounding flow turned up.

Every change defends one property: **the sign-in resume screen must reach a real outcome and must never become a dead end**, because on a cold start it can be the first thing a user sees.

### Cancel did not cancel

`abort_try_connect_account` could only stop the desktop polling flow — its handle was the only one stored. On mobile, `complete_mobile_connect_account` spawned untracked, so Cancel aborted nothing and the sign-in landed moments later regardless. Now the handle is tracked.

Cancelling a resumed cold start also had no way back: `_ready` returned early past the session recovery in favour of the parked token, so the user was dropped on ACCOUNT_HOME signed out with a session still on disk, unrestored until the next launch. Cancel now runs the recovery it stood in for.

And the cancel did not survive the deep link that followed it. `abort_try_connect_account` clears `pending_mobile_auth`, which after #2689 makes the returning link indistinguishable from a cold start — so it completed the sign-in the user had just refused. A `mobile_auth_cancelled` marker fixes that, checked **ahead of** the pending branch: `start_mobile_connect_account`'s spawn is not abortable and re-sets `pending_mobile_auth` after the cancel clears it, so a cancel during the browser-opening window left the pending flag true.

That marker is gated on `mobile_auth_started`. `abort_try_connect_account` is shared with the Android native WalletConnect path (`login.gd:_on_button_wallet_connect_pressed`), which shows the same AUTH_BROWSER_OPEN screen — Cancel included — without ever starting a browser hop. Ungated, cancelling MetaMask discarded a legitimate late `?signin=` token: the exact loss #2644 exists to prevent.

### A parked token could reach the wrong session

`deep_link_signin_parked` has a single consumer, the lobby. A token arriving while the user is signed in and past the lobby parked with nobody to redeem it — and `Global.sign_out()` swaps in a **fresh** lobby whose `_ready` redeems it, well inside the 5-minute cap. The user asks to sign out and is signed straight back in, as whoever the link belongs to. The router now drops the token when a wallet is already connected in this process.

The late-arrival handler also declines while another identity flow is in flight, and consumes the token rather than leaving it parked for a later lobby. `_recovering_session` covers the window where `_ready` is committed to recovery but suspended on the IAP env await — neither `waiting_for_new_wallet` nor `loading_first_profile` is set yet there, so a token landing in it started a second concurrent identity flow.

### The resume could hang

`complete_mobile_auth` could panic: `EphemeralAuthChain::new` unwraps `LocalWallet::from_bytes` on the ephemeral private key the server returns, and the expiration arithmetic overflows `SystemTime` on a pre-epoch timestamp. That task's `JoinHandle` is dropped, so the panic died unobserved — **neither `wallet_connected` nor `auth_error` was emitted** and the spinner stayed up forever. Both are now validated into `Err`, which reaches `auth_error`.

`_show_auth_error` is the one screen setter that never called `show_panel`, which is what dismisses the global `SplashOverlay`. The cold-start fallback in `_on_auth_error` raises that splash *before* it knows whether `try_recover_account` will succeed, and on failure falls through to this screen — drawing it under an opaque, `MOUSE_FILTER_STOP`, layer-100 overlay with TRY AGAIN unreachable. Reachable whenever the stored session is expired, which is correlated: the expired session is *why* the user was signing in.

`logout()` now clears the pending/cancelled/started markers so none outlive the account that set them.

## Verified separately

`GET /identities/:id` is strictly single-use server-side (`decentraland/auth-server`, `identities.ts`): the happy path deletes the identity and marks it `consumed`, a second read returns 404, and storage expires after **1 hour**, not 5 minutes. Mobile identities skip IP enforcement (logged, not blocked). This bounds the blast radius of a replayed token but does not remove the wrong-account paths above, which turn on *unconsumed* tokens.

## Not addressed here

Pre-existing, unrelated to #2689, worth their own issues:

- The thirdweb guest-login task installs the wallet via `call_deferred` without checking the attempt token. If the 20s watchdog abandons the attempt and the user then signs in for real, the late guest task overwrites the real wallet and persists the wrong `session_account`.
- `try_recover_account` panics rather than returning `false` on a corrupt stored session, via the same `EphemeralAuthChain::new` unwrap.
- HTTP 204 ("identity not ready") is terminal — `complete_mobile_connect_account` does a single GET with no retry, unlike the desktop polling path. One 204 loses the sign-in permanently.
- `pending_mobile_auth` is written with a non-deferred `bind_mut()` from a Tokio worker, unlike every sibling callback. `mobile_auth_started` works around that race; deferring the write would remove it.
- Lifting the pending gate makes login-CSRF reachable: a `?signin=` link from a third party, tapped on a signed-out install, signs the victim into the sender's account. Inherent to the #2689 design rather than an implementation slip, but it deserves an explicit decision.

## Testing

`cargo clippy -- -D warnings` clean, `cargo fmt --all`, `gdformat`/`gdlint` clean over `godot/`.

**Not validated on device.** The Cancel and `JoinHandle` changes affect the warm sign-in path too, not only the cold start, so this wants hardware time before merging.


lib/src/auth/dcl_player_identity.rs
```diff
@@ -852,6 +901,8 @@ impl DclPlayerIdentity {
                 }
             }
         });
+
+        self.try_connect_account_handle = Some(complete_handle);
```
regenesis-claw: [P1] Tracking the handle narrows the Cancel race but does not close it, and the lobby side does not cover the remainder.

`abort()` only takes effect at a yield point. The only `.await` in this task is `complete_mobile_auth(identity_id).await`; everything after it — including `call_deferred("try_set_remote_wallet", ...)` — is synchronous and runs to completion. So the window this PR describes ("the sign-in landed moments later regardless") is still open whenever the fetch resolves just before the tap.

On the *warm* path that then goes all the way: `_on_button_cancel_pressed` clears `waiting_for_new_wallet` only inside the `_resumed_signin_from_deep_link` branch, and `login.gd:67` set it true for an ordinary browser sign-in. Cancel → `show_auth_home_screen()` → the deferred wallet lands → `_on_wallet_connected` → `_async_on_profile_changed` takes the `waiting_for_new_wallet` branch → Welcome Back for the sign-in the user just refused. Same symptom as before the PR, just rarer.

The existing `_guest_login_attempt` token in `lobby.gd` is the pattern that fits: bump a monotonic attempt counter in `abort_try_connect_account`, capture it into this task, and have the completion drop its `call_deferred` when the counter moved. Clearing `waiting_for_new_wallet` unconditionally in `_on_button_cancel_pressed` is worth doing either way, but on its own it only downgrades the outcome — the user still ends up silently signed in.

[P2] Separately, this assignment drops any handle already in the slot, and dropping a tokio `JoinHandle` detaches rather than cancels, so the earlier task survives un-abortable. Reachable today only via `try_connect_account`, but it quietly undoes the property this line exists to add:

```suggestion
        if let Some(previous) = self.try_connect_account_handle.replace(complete_handle) {
            previous.abort();
        }
```

godot/src/ui/pages/auth/lobby.gd
```diff
@@ -907,6 +914,17 @@ func _on_button_cancel_pressed():
 	Global.metrics.track_click_button("cancel", current_screen_name, "")
 
 	Global.player_identity.abort_try_connect_account()
+
+	# Cancelling a resumed cold start has to undo what _ready skipped for it (#2644): the
+	# parked token took precedence over session recovery and _ready returned before running
+	# it, so without this the user lands on ACCOUNT_HOME signed out with a session still on
+	# disk, unrestored until the next launch.
+	if _resumed_signin_from_deep_link:
```
regenesis-claw, line 928: [P1] `waiting_for_new_wallet = false` is inside the resumed-cold-start branch, but the state it undoes is set on **every** browser sign-in (`login.gd:67`), not just the resumed one.

Cancel on an ordinary warm sign-in therefore leaves the lobby armed for a wallet it no longer expects, which is what turns the abort race above into a full Welcome Back. Hoisting it costs nothing on the resume path, where it is already being cleared:

```suggestion
	Global.player_identity.abort_try_connect_account()
	# The user is done with this sign-in either way — don't leave the lobby armed for a
	# wallet_connected that a lost abort race can still deliver.
	waiting_for_new_wallet = false

	# Cancelling a resumed cold start has to undo what _ready skipped for it (#2644): the
	# parked token took precedence over session recovery and _ready returned before running
	# it, so without this the user lands on ACCOUNT_HOME signed out with a session still on
	# disk, unrestored until the next launch.
	if _resumed_signin_from_deep_link:
		_resumed_signin_from_deep_link = false
```

godot/src/ui/pages/auth/lobby.gd
```diff
@@ -942,12 +967,37 @@ func _async_resume_signin_from_deep_link() -> bool:
...
 
 func _show_auth_error(error_message: String):
 	track_lobby_screen("AUTH_ERROR")
+	# This is the only screen setter that does not route through show_panel, which is what
+	# dismisses the global SplashOverlay — and that overlay is opaque, on layer 100 and
+	# MOUSE_FILTER_STOP, so anything shown under it is unreachable. Callers are expected to
+	# have the sign-in panel up already (that is why this only toggles containers), so assert
+	# the one invariant that matters instead of re-showing panels behind their backs.
+	# Idempotent: fade_out no-ops once dismissed.
+	SplashOverlay.fade_out()
```
regenesis-claw, line 1005: [P2] This guarantees the overlay is gone but not that the screen underneath it is the one you just wrote to, and the comment asserts an invariant the function does not enforce.

`_show_auth_error` never calls `show_panel(control_signin)` or `container_sign_in_step2.show()`. On the cold-start path the caller does have step 2 up, so the case this PR targets is fine. The other `_on_auth_error` callers are not: a session recovery whose profile fetch errors reaches here with `_resumed_signin_from_deep_link` false and `DCL_SPLASH` showing, so `fade_out()` now reveals whichever panel was last shown while "Authentication failed" and TRY AGAIN are written into hidden containers, and `loading_first_profile` is left true.

Still an improvement — pre-PR that path sat under a stuck opaque overlay — but since the stated invariant is "must never become a dead end", make the function own its own screen rather than assume the caller did:

```suggestion
	SplashOverlay.fade_out()
	show_panel(control_signin)
	container_sign_in_step1.hide()
	container_sign_in_step2.show()
```

(`show_panel` already fades the overlay, so the explicit call becomes belt-and-braces.)

godot/src/deep_link_router.gd
```diff
@@ -229,12 +229,31 @@ func _route_teleport() -> void:
 ## The token is self-contained — complete_mobile_connect_account fetches the identity by id
 ## — so the only question is who drives the UI while that fetch runs.
 func _handle_signin_deep_link(identity_id: String) -> void:
+	# The user cancelled a sign-in in this process, so this token is the flow they refused.
+	# Checked ahead of the pending branch on purpose: abort_try_connect_account() clears
+	# pending_mobile_auth, but start_mobile_connect_account's spawn is not abortable and sets
+	# it again when it resolves, so a cancel during the browser-opening window leaves the
+	# pending flag *true* and the warm branch below would complete the sign-in regardless.
+	if Global.player_identity.was_mobile_auth_cancelled():
```
regenesis-claw, line 237: [P2] The marker is process-wide, and `start_mobile_connect_account` clears it — so the cancel stops sticking one tap later.

1. Sign in → Google. Browser opens (`mobile_auth_started` true).
2. Cancel. `mobile_auth_cancelled` true — a returning token is correctly refused.
3. Sign in → Google again. `start_mobile_connect_account` sets `mobile_auth_cancelled = false`, and its spawn has not resolved yet so `pending_mobile_auth` is still false.
4. Finish the **first**, still-open browser tab. Token #1 comes back: not cancelled, not pending, no address → parked → the lobby redeems the attempt the user cancelled.

Same account both times, so this is a correctness wart rather than a security one — but the property the marker exists to defend is gone in step 3. Keying it to the attempt (store the cancelled attempt's counter and compare, rather than a process-wide bool) would make it hold.

[P2] Unrelated but on the same invariant: `complete_mobile_auth` is a single `fetch_identity_by_id` with no timeout and no retry, so Cancel is still the only exit from "Finishing sign in…" if that request hangs — and per the description a single HTTP 204 loses the sign-in permanently. Both are #2644-class losses in a PR whose whole point is to stop losing logins. `_on_guest_login_watchdog_timeout` in `lobby.gd` is the existing watchdog pattern if you want it here; otherwise these two are worth their own issue alongside the others you listed.

regenesis-claw (changes requested):
Re-review of the follow-up to my #2689 review. Checked every claim in the description against the source at `7a5a510`, not against the PR text.

## Pre-flight (REVIEW.md)

1. **The branch is 10 commits behind its own base.** It was cut from `837dc97` (the pre-merge #2689 head), not from `398ea0b`. That is why **Linux is red**: the merge ref carries the `xtask test-avatar` step that landed on `main` with #2699, but this branch's xtask has no such subcommand — `error: The subcommand 'test-avatar' wasn't recognized`, exit 2. Not a real failure of this diff, but it also skips Docker, so Linux/Docker give no signal. Merge `fix/coldstart-deeplink-signin` in and let it re-run.
2. **iOS.** This is deeplinks + auth, so an iOS build is required. `Build iOS export → R2` is still **pending** and there is no `build`/`build-ios` label. I am not treating a pending job as a green artifact. Static checks, Clippy and Android are green.
3. **QA sign-off invalidation.** @Ludmilafantaniella approved #2689 on `398ea0b` on Aug 18, and her tests 1/2 are the *warm* sign-in path. The Cancel and `JoinHandle` changes here alter that same warm path, so merging this into #2689's branch retires that sign-off. Consistent with the description's own "Not validated on device" — flagging it so it does not get lost when #2689 is merged.

## Claims I verified against the source

- `EphemeralAuthChain::new` really does `LocalWallet::from_bytes(&ephemeral_keys).unwrap()` (`ephemeral_auth_chain.rs:33`), and the new guard validates **the same `Vec`** that is later moved into it. The panic was real and the fix is at the right place. `u64::try_from` + `checked_add` likewise close the pre-epoch/overflow path.
- The `mobile_auth_started` gate is necessary and correctly motivated: `login.gd:348` shows `show_auth_browser_open_screen("Opening " + wallet_name, "metamask_native"/"wallet_connect_native")` — same screen, same Cancel — without ever calling `start_mobile_connect_account`. Ungated, that Cancel would have poisoned a legitimate late token.
- `abort_try_connect_account` has exactly one caller (the lobby's `_on_button_cancel_pressed`), reachable both from the button and from Android hardware-back via `_handle_back_action`'s `AUTH_BROWSER_OPEN` case.
- The `_recovering_session` window is real: `deep_link_signin_parked` is connected at `lobby.gd:457`, well before the `await Iap.async_await_env_resolved()` at 514, so a token can land inside it. The new flag is the only thing covering it.
- `_show_auth_error` really is the only screen setter that skips `show_panel`, which is where `SplashOverlay.fade_out()` lives (`lobby.gd:122`).
- The `get_address_str()` drop does **not** strand a guest→authenticated upgrade, which was my first worry with it: that flow is OTP (`sign_in_with_email.gd` → `async_login_email_verify`), not the `?signin=` browser hop, and `async_login` only exists on the signed-out login page. Recording it so it does not get re-litigated.

## Findings

`[P1]` x2, `[P2]` x3 — on the diff lines below, plus one that has no line to hang on:

**[P1] The login-CSRF item needs a recorded decision, not a bullet.** "Not addressed here" lists it as inherent to #2689's design, and that is fair — but it is the one item on that list that is an *auth* consequence rather than a latent bug: with the pending gate lifted, a `?signin=` link from a third party, tapped on a signed-out install, signs the victim into the sender's account. The new `get_address_str()` guard closes the signed-in variant (the sign-out path I flagged last time); the signed-out variant is exactly the state #2644 optimises for. Please open a tracking issue and link it here before this merges, so the trade-off is on the record rather than in a PR body.

The `Cancel did not cancel` work, the panic guards, and the wrong-account fix all land where I asked. The two P1s below are the same properties this PR claims, narrowed rather than closed.

leanmendoza:
Closing in favour of applying these directly on `fix/coldstart-deeplink-signin`, one commit per fix, so each defect can be reviewed on its own. Nothing is dropped — the same fixes plus the feedback from this review land on #2689.

PR #2720, opened 2026-08-19
feat: preview hot-reload over protobuf WsSceneMessage, with per-model hot-swap

### LINKED ISSUE #2600: Preview hot-reload: migrate from legacy JSON SCENE_UPDATE to protobuf WsSceneMessage
## Context

The sdk-commands preview server broadcasts file-change events over WebSocket using two protocols:

- **Legacy JSON** (`__LEGACY__updateScene`): text frames `"update"` + `{"type": "SCENE_UPDATE", "payload": {"sceneId", "sceneType"}}`
- **Protobuf** `WsSceneMessage` (`decentraland/sdk/development/local_development.proto`): binary frames with `updateScene` / `updateModel` variants

Godot only understands the legacy protocol: `godot/src/logic/preview_websocket.gd` parses each packet as UTF-8 text → JSON and matches `"SCENE_UPDATE"` only; binary protobuf frames fail the JSON parse and are silently dropped.

decentraland/js-sdk-toolchain#1502 initially removed the legacy broadcast (which silently broke hot-reload here) — it now keeps it, but only until Godot and Bevy migrate (decentraland/bevy-explorer#1020 is the matching issue).

## Task

1. Handle binary frames on the preview WebSocket: decode `WsSceneMessage` (proto already available via @dcl/protocol; the Rust side could decode in `lib/` where prost bindings live, or GDScript can receive it from a Rust helper).
2. Map `updateScene` → the existing `scene_update` signal (`sceneId`). Keep the JSON path as fallback for older CLIs.
3. Optional follow-up: use the `updateModel` variant (per-file hash/src/change-type) for model-level hot-swap instead of full scene reloads.

EibrielInv:
## What

Godot only spoke the legacy JSON preview protocol: every packet was parsed as
UTF-8 JSON and matched against `SCENE_UPDATE`, so the binary `WsSceneMessage`
frames the preview server now broadcasts were silently dropped. js-sdk-toolchain
keeps the legacy broadcast alive only until Godot and Bevy migrate
(decentraland/bevy-explorer#1020 is the twin issue).

This decodes the protobuf protocol and uses its richer `updateModel` variant to
hot-swap a single model, instead of tearing down and refetching the whole scene.

## How

- `build.rs` compiles `decentraland/sdk/development/local_development.proto`
  (vendored with @dcl/protocol, but never in the prost compile set); exposed as
  `proto_components::sdk::development`.
- `DclPreviewMessage.decode(PackedByteArray) -> Dictionary` — returns an empty
  dict on garbage rather than panicking across the GDExtension boundary.
- `preview_websocket.gd` splits on `was_string_packet()`: text keeps the legacy
  JSON path, binary goes through the decoder. `updateScene` maps to the existing
  `scene_update` signal, so `scene_fetcher` is untouched on that path.
- `updateModel` purges the model's cache entries and calls a new
  `force_reload_gltf()` on the live containers using that `src`. Anything that
  can't be swapped precisely — unknown scene, non-swappable file, no live
  container — falls back to the existing full reload.

## The awkward part: preview hashes are not content hashes

`b64HashingFunction` base64-encodes the file *path* plus a machine id. It never
changes when the file does, and the client's caching assumes the production
contract where a hash identifies content. Three consequences, each needing an
explicit fix:

- Every layer keyed by the hash serves stale bytes, so `purge_file` must drop the
  `scene_<hash>` promise key, the `existing_files` entry (what the cache check
  actually consults — it reads that map, not the disk) and the derived `.scn`.
  `ResourceProvider::delete_file` only unlinks paths it tracks, and a
  saver-written `.scn` isn't one, so the file needs an explicit unlink too.
- The `.scn` path is reused, so the shared load needs `CACHE_MODE_REPLACE`.
- `change_gltf` early-outs when the src string is unchanged, which is exactly the
  swap case — hence `force_reload_gltf()`.

The message's `hash` is deliberately **ignored**: the watcher hashes the
project-relative path while the scene definition hashes the absolute one, so the
two never match and adopting it points the container at an unrelated cache entry.
Since a path-derived hash can't change when contents do, the mapping is already
correct for an edit; only a deletion touches it.

## Legacy suppression, and a known gap

The server sends both protocols for every change, so acting on both would follow
each precise swap with a full reload. Once a connection has produced a decodable
binary frame, its legacy frames are dropped (re-probed per connection).

The server debounces file events through a single shared 800ms timer
(last-args-win), so saving two models inside that window reports only the last.
The legacy frame used to paper over this by reloading everything; now the first
file stays stale until saved again. Best fixed upstream with a per-path debounce.

## Deletes, and why there is a grace window

Deleting a model emits *both* an immediate `UMT_REMOVE` and — from the debounced
generic handler — an `UMT_CHANGE` for the same, now missing, path. Dropping the
mapping entry on remove is right (the container falls to `NOT_FOUND`, which is
what a missing model should look like), but the entry cannot be rebuilt locally
if the file comes back: the hash derives from the absolute path plus a machine id
only the server knows. So a change event with no mapping entry falls back to a
full reload, which refetches the definition and restores the entry.

That rule alone can't distinguish "the file was re-created" from the delete's own
trailing event — both are "change event, no mapping entry" — so every delete was
followed by a pointless full reload. `_recently_removed_models` records removals
and swallows change events for the same path within `REMOVED_MODEL_GRACE_MS`
(1500ms, comfortably above the server's 800ms debounce). Re-creating a file
inside that window is treated as the trailing event and takes effect on the next
save; a human is very unlikely to hit it.

## Why only .glb is swapped

A `.glb` embeds its buffers and textures. A `.gltf` is JSON pointing at sibling
`.bin`/texture files, which the importer resolves through the scene's content
mapping. Those siblings are not models, so they get no `updateModel` of their
own, and the server's shared debounce usually drops their events entirely —
leaving the mapping without them.

Swapping a `.gltf` in place therefore either fails outright (`There are some
missing dependencies in the gltf`) or, when the replacement reuses the old
filenames, rebuilds the model from stale cached buffers — which renders as old
and new geometry at once. Only the full reload refetches the scene definition,
which is what repopulates the mapping, so `.gltf` takes that path.

This was caught by QA on this PR (thanks @Ludmilafantaniella) and reproduced
locally: a `.gltf` referenced by two entities, replaced by a model whose
dependencies the mapping had never seen, logged one `GLTF load error` per
container and left both invisible until the next save. With `.gltf` routed to the
full reload the same swap imports cleanly.

## QA — reproducing the GLTF hot-swap from the Creator Hub

1. **Open a scene in the Decentraland Creator Hub** that uses at least one
   `.glb` model, and hit **Preview**. The Hub starts its preview server
   and launches the explorer against it.

2. **Make sure the model is actually on screen.** The swap only applies to models
   with a *live* `GltfContainer` — if nothing is currently instantiated (a model
   that only spawns on interaction, for example), the client correctly falls back
   to a full scene reload and you won't be testing the swap path. Walk up to it /
   trigger it first.

3. **Replace the file on disk, keeping the same path and filename.** Either
   re-export from Blender over the existing file, or from a terminal:
   ```bash
   cp some_other_model.glb "<scene>/assets/models/<the_model>.glb"
   ```
   Use a model that looks *obviously* different — same-looking geometry makes a
   stale cache indistinguishable from a correct reload.

4. **Expected:** the model is replaced in place, within a second or so. The rest
   of the scene keeps running — no loading screen, no scene restart, other
   entities keep their state, and the JS runtime is not torn down. The console
   prints `preview-ws > server speaks protobuf; ignoring its legacy JSON frames`
   once per connection, and no scene-shutdown lines follow the swap.

5. **Delete the model file.** It should vanish from the scene with no reload —
   the container drops to `NOT_FOUND`. Restoring the file afterwards brings it
   back via a full reload (see the grace-window section above).

6. **Contrast with a non-model change:** move an entity in the Hub, or edit a
   `.ts`. That writes `main.crdt`/source rather than a model, so the server sends
   `updateScene` and the client does a full (loading-screen-free) reload, as
   before. Only `.glb` edits are hot-swapped — `.gltf` takes the full reload for
   the reason above.

To watch it in detail, launch with
`--rust-log="dclgodot::content=debug,warn"`: a successful swap shows
`Scene GLTF cache MISS` → `[HTTP] GET .../contents/<hash>` → `GLTF processed …
saved to …scn`, with a node/byte count matching the new model.

### Verified during development

Against a real `sdk-commands` preview server: binary decode, `src` matching and
legacy suppression confirmed on the wire; the no-live-container case falling back
to a full reload; and a genuine model swap re-fetching and re-importing
(`64 nodes / 243104 bytes` → `5 nodes / 64444 bytes`) with the geometry visibly
replaced and no scene reload — plus the same in reverse.

After QA's report, also verified with a `.gltf` converted from the same model and
referenced by two entities at once: replacing it now triggers the full reload and
imports cleanly (`5 nodes / 64486 bytes`, zero dependency errors, versus two
`missing dependencies` errors before the fix), while an equivalent `.glb` swap
still hot-swaps with no reload.

`UMT_REMOVE` verified too: deleting a live model removes it from the scene with
no reload and no stray fetch, and restoring the file brings it back through the
recovery reload.

Also 4 `#[itest]`s covering both message variants, the remove case, and
garbage/empty-oneof input.

Closes #2600


Ludmilafantaniella (commented):
Tested the GLTF hot-swap flow per the QA repro steps (Creator Hub, scene with multiple `Palm.glb` instances on screen, all pointing to the same `palm.gltf` — see entity tree).

**Repro:**
1. Scene has several palm tree entities (`Palm.glb`, `Palm.glb_2`, `Palm.glb_3`...) all referencing `palm.gltf` (same path).
2. Replaced `palm.gltf` on disk with a different model (`gamer_chair.gltf`, same filename/path).
3. Observed the live Preview.

**Expected:** The palm tree model is replaced in place by the gamer chair model.

**Actual:** Both models render simultaneously in the same spot — the old palm mesh and the new gamer chair mesh are both visible, overlapping. Not a one-frame flicker; it persists across the session and after subsequent reloads.

cc @EibrielInv

https://github.com/user-attachments/assets/b7ca7b4a-e1a2-49ca-9d15-69b404ab888f

Image

Ludmilafantaniella (approved):
**Re-tested after the latest commit (`f30ba0d` — reverse hot swap for gltf).**

Repeated the original repro: scene with multiple `Palm.glb` instances all referencing `palm.gltf`, replaced on disk with `Armchair.gltf` keeping the same path/filename.

**Result:** the model is correctly replaced across all instances — no more overlapping palm tree / chair meshes. The swap looks clean, no scene reload, no persistent flicker.

Also re-ran the basic cases from the original QA (`.glb` swap, delete + grace window, `.gltf` full reload) and they still work as described in the PR.

The issue I previously reported (duplicated/overlapping models) is resolved.

Approving ✅

https://github.com/user-attachments/assets/16972990-d564-4ae9-8d21-f6f7fd4ebce4

kuruk-mm (approved):
looks great! feel free to merge!

PR #2723, opened 2026-08-20
feat: HUD revamp — second iteration (orb buttons)

### LINKED ISSUE #2635: HUD: Iteration #2
## Problem:
After controls customization, mobile HUD needs to be re-thought and iterated to facilitate immersion and avoid conflicts with scene made HUD buttons.

## Goal:
implement the redesign of the new HUD (HUD v2) that modifies color schemas, button formats and overall HUD UI to facilitate scene-driven buttons and increase immersion.

### HUD

[Figma HUD](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=8-1456&t=tXyz4GK3GV6pUccG-1)
[Atlas](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=609-4486&t=tXyz4GK3GV6pUccG-1)

Image

sdilauro:
Closes #2635 

# HUD revamp — second iteration (orb buttons) [Figma file](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=994-11220&m=dev)

Second pass on the mobile HUD: one reusable **orb-style button** now backs every standalone HUD action, the joypad combo got trimmed, and the virtual joystick was redrawn to match the orb look (and now sits safe-area-aware).

## New: `HudButton` component
`src/ui/components/atoms/buttons/hud_button/` — a drop-in `Button` skinned with 3 shared background textures swapped by state (**normal / pressed / selected**-when-toggled) + a per-button icon on top. Separate from the joypad's `OrbSkin`.
- **Size presets** (`Size` enum, picked in the Inspector): `SMALL 66/46`, `MEDIUM 80/48`, `LARGE 96/67` (disc / icon px). The px are the visible **disc**; the background texture renders larger to hold the glow, overflowing the footprint (`custom_minimum_size` = disc, tap area = disc).
- **Per-state opacities** baked into the shared backgrounds: default **72%**, pressed **100%**, selected **81%** (radial gradient `#251A3B → #161518` at 90% colors × the state alpha), 1px `#DFD0FF` outline, lilac inner-shadow sheen.
- Assets: `hud_button/textures/` (shared bg) + `hud_button/icons/` (per-button SVG). Raw art drops in `_raw/` (`.gdignore`d), normalized to VRAM high-quality (BC7/ASTC) on import.

## Buttons converted to `HudButton`
- **Chat**, **flip-orientation**, **discover** (chatbar) — these scale `SMALL ↔ LARGE` with orientation.
- **Emote wheel** button (`MEDIUM`).
- **Show-UI** restore button (`SMALL`, close/X icon).

## Joypad
- Arc positions synced to the authored editor layout, editor-only alignment guide hidden at runtime, `Button_Jump` → `Button_Interact` reference fix.
- Refreshed `glide-hold` / new `plus-hold` icons, `base_scale=3` on the joypad SVGs for crispness.

## Virtual joystick (shader rewrite)
`godot/shaders/joystick.gdshader` — orb aesthetic:
- **Resting**: 170px base orb + a lilac (`#DFD0FF` @ 50%) centre thumb, 1px outline.
- **Dragging**: 80px thumb orb (50% fill, 1px outline, no inner dot) + a comet trail — `32, 32, 18, 12, 8` px, **40px tangent gaps**, fading **45% → 10%** toward the origin, no big base ring around the tap point.
- **Resting position is now safe-area-aware & device-adaptive**: `_joystick_default_position` is relative to the safe area (assign `margin_profile = hud_margins.tres`), offset by the live device inset. A Resource `@export` is used, not a node reference, because node-path exports get dropped from the `.tscn` when the editor re-saves an instanced scene.
- The joystick script is now `@tool`, so the resting position previews in the editor with the DCL Mobile Preview.

## Test plan
Mostly cosmetic (orb restyle across the HUD), so this is about catching regressions on a real phone. Input paths moved (chat/emote/discover/flip/show-UI are new orb buttons; the joypad combo default and the joystick resting position changed), so a fresh pass beats a delta review.

### Case 1 — Emote wheel opens with a second finger while moving
**Setup:** signed-in profile with at least one equipped emote.
**Steps:**
1. Open the app and wait for the explorer to load.
2. Enter Genesis Plaza.
3. Press and hold the movement joystick with your left thumb — keep it held.
4. With a second finger, tap the emote button (bottom orb).

- [ ] **Expected:** the emote wheel opens; while the finger is down the emote orb shows its brighter *pressed* look, and after lifting it stays in the lit *selected* look while the wheel is open.
- [ ] **Regression — single touch:** release the joystick, tap the emote orb once with one finger → the wheel opens; tap it again (or tap outside) → the wheel closes and the orb returns to its default look.

### Case 2 — Emote orb recovers if the HUD hides mid-press
**Steps:**
2. Press and hold the emote orb with one finger — keep it held.
3. With a second finger, tap the hide-UI (X) button at the top-left to hide the HUD, then bring the HUD back.

- [ ] **Expected:** when the HUD returns, the emote orb is back to its default look (not stuck bright) and opens the wheel normally on the next tap.

### Case 3 — Chat stays closed when tapped with a second finger while moving
**Steps:**
2. Press and hold the movement joystick with your left thumb — keep it held.
3. With a second finger, tap the chat orb.

- [ ] **Expected:** the chat panel does **not** open (chat only opens from a normal single tap, never from a second finger while moving).

### Case 4 — HUD orbs resize with orientation
**Steps:**
2. In landscape, note the size of the chat / flip / discover orbs.
3. Rotate the device to portrait.

- [ ] **Expected:** the chat / flip / discover orbs grow in portrait and shrink back in landscape; the icon stays centered and crisp and the tappable area tracks the visible disc.

### Case 5 — Joystick rests inside the safe area
**Steps:**
2. Without touching it, look at the resting joystick at the bottom-left.

- [ ] **Expected:** the joystick base sits a small margin in from the bottom-left safe area (roughly aligned with the joypad/chat frame), not jammed against the screen edge — on both notched and non-notched phones.
- [ ] **Regression — drag:** drag the joystick → the thumb follows your finger with a fading comet trail behind it, and no large ring is left at the original rest spot.

### Case 6 — Joypad combo shows 4 actions by default
**Steps:**
2. Tap the "+" button on the joypad (bottom-right cluster).

- [ ] **Expected:** the overflow opens showing 4 action buttons; the "+" is present only because there are more buttons than fit directly on the ring.

### Case 7 — A scene can hide joypad buttons
**Setup:** a scene/world configured to hide one or more on-screen action buttons.
**Steps:**
1. Open the app.
2. Enter that scene.
3. Look at the joypad button cluster.

- [ ] **Expected:** the hidden button(s) are gone and the remaining buttons reflow up with no empty gap; if few enough remain, the "+" disappears and the rest show directly on the ring.


godot/src/ui/components/atoms/buttons/hud_button/hud_button.gd
```diff
@@ -0,0 +1,162 @@
...
+func _on_toggled(_toggled_on: bool) -> void:
+	_refresh()
+
+
+func _apply_style_overrides() -> void:
+	for slot: StringName in _STYLE_SLOTS:
+		add_theme_stylebox_override(slot, StyleBoxEmpty.new())
+
+
+func _clear_style_overrides() -> void:
```
regenesis-claw, line 128: [P2] The pre-save strip only covers `_STYLE_SLOTS`, but `_ready()` / `_apply_size()` / `_refresh()` also write `icon`, `custom_minimum_size`, `expand_icon`, `icon_alignment` and the `icon_max_width` theme constant onto the node — and because the script is `@tool`, those run in the editor and get serialized into whatever scene instances the button.

That's already visible in this diff: both `Button_Chat`/`Button_Flip` in `chatbar.tscn` and `Button_ShowUI` in `explorer.tscn` carry `icon = ExtResource(...)` right next to the `icon_normal` that produced it. Harmless at runtime (`_refresh()` overwrites `icon` on ready), but it's the same instance-bloat class you went out of your way to prevent for the styleboxes, and a later change to `icon_normal` leaves a stale baked `icon` sitting in the host scene diff.

Consider clearing `icon` (and optionally the size overrides) in `_clear_style_overrides()` too, so a host scene save stays clean.

godot/src/ui/components/atoms/buttons/button_touch_action/orb_skin.gd
```diff
@@ -0,0 +1,195 @@
...
+):
+	set(value):
+		default_texture = value
+		if is_node_ready():
+			_refresh()
+@export var pressed_texture: Texture2D = preload(
+	"res://src/ui/components/atoms/buttons/button_touch_action/orb_pressed.png"
+)
+@export var hold_texture: Texture2D = preload(
+	"res://src/ui/components/atoms/buttons/button_touch_action/orb-hold.png"
```
regenesis-claw, line 73: [P2] `orb-hold.png` is kebab-case while its three siblings are `orb_default.png` / `orb_pressed.png` / `orb_disabled.png` — it's the only one in the set that breaks the convention, and the rename churned the uid (verified nothing still points at the old `53y0kswva0qr`). Worth renaming back to `orb_hold.png` while it's still only referenced from two places.

godot/src/ui/components/atoms/buttons/button_touch_action/button_touch_action.tscn
```diff
@@ -11,25 +15,39 @@ offset_bottom = 55.0
...
-offset_bottom = -6.0
 grow_horizontal = 2
 grow_vertical = 2
-visible = false
 mouse_filter = 2
+texture = ExtResource("4_orbdf")
 expand_mode = 1
 stretch_mode = 5
+script = ExtResource("3_orbsk")
+hold_texture = ExtResource("5_nxgap")
```
regenesis-claw: [P2] This pins `hold_texture` to the exact file `orb_skin.gd` already `preload`s as the `@export` default, so the value is duplicated in two places for no behaviour change — and if the shared hold orb is ever swapped in the script, this instance keeps the old one silently. Dropping the line (and the matching `ext_resource`) leaves the same result, unless the editor keeps re-adding it on save.

godot/src/ui/components/atoms/buttons/hud_button/_raw/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. Copy each file to `../textures/` (backgrounds) or `../icons/` (icons).
+2. Set the `.import` to **VRAM Compressed, high quality** — matches the orb textures
+   (`compress/mode=2`, `compress/high_quality=true`, `mipmaps/generate=true`), which is
+   what avoids gradient banding and what the CI check (`tests/check_asset_imports.py`)
+   requires. SVGs go to `DPITexture` instead.
+3. `cargo run -- import-assets` to reimport.
+4. Wire the normalized textures into `../hud_button.tscn` (or into each button's instance).
+
+Nothing here is meant to ship — the raw files can stay uncommitted; only the normalized
```
regenesis-claw, line 35: [P2] The README says the raw files "can stay uncommitted", but `hud-default.png`, `hud-pressed.png` and `hud-selected.png` (~500 KB total) are committed here. The `.gdignore` keeps them out of the import pipeline so there's no runtime cost, but they're permanent repo weight for a staging folder.

The doc is also written as instructions to a specific assistant ("say so and I'll import them as DPITexture", "I run this once the files are here") — as a repo-level doc it'd read better as a plain normalization procedure any contributor can follow.

godot/src/ui/components/molecules/profile_icon_button/profile_icon_button.tscn
```diff
@@ -2,77 +2,54 @@
...
+layout_mode = 2
+mouse_filter = 2
 texture = ExtResource("1_uo4jm")
 expand_mode = 1
-stretch_mode = 4
+stretch_mode = 5
 
-[node name="Panel_Border" type="Panel" parent="." unique_id=773213466]
-clip_children = 2
+[node name="TextureRect_Border" type="TextureRect" parent="." unique_id=773213467]
```
regenesis-claw, line 41: [P2] `TextureRect_Border` has no `visible = false` default, so the glow ring is on from instantiation and only turns off because `navbar.gd::_ready()` calls `set_glow(button.button_pressed)`. It works today (the navbar is the only host and `_ready()` lands before the first frame), but the component's resting state depends entirely on an external caller — same shape as the `Panel_LoadScenes` default that regressed earlier in this branch. Defaulting it hidden in the scene makes the off state the component's own property.

godot/src/mobile/joystick/virtual_joystick.gd
```diff
@@ -248,26 +257,41 @@ func _update_input_actions():
...
+	if margin_profile:
+		var portrait: bool = (not Engine.is_editor_hint()) and Global.is_orientation_portrait()
+		safe_left = float(margin_profile.portrait_left if portrait else margin_profile.left)
+		safe_bottom = float(margin_profile.portrait_bottom if portrait else margin_profile.bottom)
+	if not Engine.is_editor_hint() and (Global.is_mobile() or Global.is_emulating_safe_area()):
+		var sa: Rect2i = Global.get_safe_area()
+		var win: Vector2i = DisplayServer.window_get_size()
+		var vp: Vector2 = get_viewport().get_visible_rect().size
+		safe_left = maxf(safe_left, float(sa.position.x) * (vp.x / float(win.x)))
+		safe_bottom = maxf(safe_bottom, float(abs(sa.end.y - win.y)) * (vp.y / float(win.y)))
```
regenesis-claw, line 290: [P2] Both scale factors divide by `DisplayServer.window_get_size()` with no zero guard. GDScript float division by zero yields `INF` rather than erroring, so a transient `0` window size (Android backgrounding / surface recreation is the realistic case) propagates `INF` into `safe_left`/`safe_bottom` and parks the joystick base off-screen until the next `_reset()`.

Cheap guard, e.g. skip the whole block when `win.x <= 0 or win.y <= 0` and keep the `margin_profile` values as the floor.

regenesis-claw (approved):
## Full review — head `ff26b70`

Fresh full-diff pass (100 files, +2072/-586).

### Pre-flight (REVIEW.md §0)

- **Branch up to date with `main`** — `behind_by: 0` (merge `e2227bd`). ✅
- **iOS build present and green** — `🍏 iOS → R2 (unsigned export)` passed and mobile distribution ran (TestFlight, build 1654). Required here: the diff is touch-input- and HUD-wide. ✅
- **No submodule pointer drift**, no `.claude/` files, no editor-state leak in `project.godot`, no `print`/`push_error`/`printerr` added, no non-English comments or identifiers anywhere in the diff (including `.tscn`/`.gdshader`). ✅
- **CI**: Static checks, Clippy, Linux, Android, Docker and iOS all green. ✅

### Verdict

Approving — no P0/P1. Six P2s below, none blocking.

### Things I verified rather than took on faith

- **Joystick resting position isn't a travel-area regression.** `margin_control` was referenced only by `explorer.tscn`'s `editor_description` and was never actually assigned, so the old `offset_left/top/right/bottom` expansion branch was dead code. Removing it changes nothing about how far the stick can travel.
- **The resting spot does move, deliberately.** `explorer.tscn` drops the instance override `_joystick_default_position = Vector2(300, 150)`, so the value now comes from `virtual_joystick.tscn` (`154, 146`) plus the `hud_margins.tres` insets (108/45 landscape): effective rest goes from `(300, h-150)` to `(262, h-191)`. That matches Test plan Case 5.
- **No orphaned resource references.** The four deleted `assets/button-chat-*` / `assets/flip-button-*` SVGs are referenced only by `chatbar.tscn`, which this PR rewrites. The `orb_hold.png` → `orb-hold.png` rename and the `glide-hold.svg` redraw both churned uids (`53y0kswva0qr` → `c8jpdutgdubrt`, `dio6yg4glplys` → `c8jb31i47s7jy`); both `.import` files and `joypad.gd`'s `preload` were updated and no reference to either old uid survives anywhere in the branch.
- **`glide-hold.svg` is now consistent with its siblings** — 100×100, stroke 3/5, same paths as `glide-normal`/`glide-pressed`, all three on `base_scale=3.0`. The old 56×56 asset was the odd one out and would have rendered at a different relative weight.
- **Shader uniform prune is complete** (Tier 3 §15): `tip_radius` / `tip_center_radius` / `bubble_radius` / `max_bubble_distance` are gone from both the shader and `joystick_material.tres`, with `bubble_edge_gap` added — no orphan `shader_parameter/` lines left behind. `state == 2` (release) still renders nothing, same as the old shader.
- **The `set_pressed_no_signal` → `button_pressed` swaps don't recurse.** `emote_wheel.close()` early-returns on `not control_wheel.visible`, so the `toggled` → `_on_button_toggled` → `close()` bounce terminates without emitting `emote_wheel_closed` twice; `chatbar.tscn` connects only `pressed` (never `toggled`), so the equivalent swap there has no loop at all.
- **Re-rooting `ProfileIconButton` from `Panel` to `TextureButton` is safe for its only host.** `navbar.tscn`'s `Panel_Profile` instance carries no `Panel`-specific overrides (no `theme_override_styles/panel`), so nothing is orphaned, and `Control/Button` is still declared after it, keeping hit-test order unchanged.

### Security

No security-relevant surface in this diff — no network calls, auth, secrets, deserialization or user-supplied string handling, and no new dependencies. **No security issues found.**

### Test plan

Matches §4 properly: six cases, setup lines only where the state is non-obvious, steps starting from opening the app, observable pass/fail expectations, and regression lines on the shared paths. This is the format QA can actually execute — thanks for writing it that way.

### Findings (all P2)

**1. `button_touch_action.gd` didn't get the escape hatch `emote_wheel.gd` just got — and the joypad case is the worse one.** (Kept out of line comments: the affected lines aren't part of this diff.)

`_on_gui_input` claims the touch with `_touch_index = event.index` and latches `Input.action_press(trigger_action)`; only the matching release clears both. If the button is hidden or re-laid-out mid-press (hide-UI, orientation flip, a scene swapping `PBTouchScreenControls` slots) that release never arrives, so the orb stays pressed *and* jump/glide/primary stays held at the input level. `emote_wheel.gd` now solves exactly this with `visibility_changed` + `is_visible_in_tree()`; the same guard here would additionally need `Input.action_release(trigger_action)` and `_is_action_active = false`. Pre-existing rather than introduced here, so not blocking — but the two atoms are now inconsistent and this is the one with real input consequences.

The rest, inline:

2. `HudButton`'s pre-save strip only covers styleboxes; `icon` and the sizing overrides still get baked into host scenes.
3. `_raw/` ships ~500 KB of art the folder's own README says shouldn't be committed.
4. `orb-hold.png` breaks the snake_case naming of its three siblings.
5. `hold_texture` is pinned in `button_touch_action.tscn` to the same file the script already preloads as its default.
6. The profile glow ring has no `visible = false` default in its own scene.
7. Unguarded division by `DisplayServer.window_get_size()` in the joystick's safe-area math.

Ludmilafantaniella:
Tested this on a real device (build `v1.13.0.1654-ff26b70-staging`) and I don't think I was actually seeing the redesigned HUD described in this PR — the buttons, joystick, and joypad all look like the classic style (grey circular buttons, standard directional joystick), not the orb aesthetic described (glow, lilac outline, comet trail on drag, etc).

https://github.com/user-attachments/assets/1ae88a80-6b38-436d-8524-6e5b677657c1

⚠️ Also, for future redesign PRs like this one, it'd really help QA if you could attach a reference screenshot/mockup of the expected look (or a before/after). Since this is mostly a visual/cosmetic change, it's hard to tell what's "correct" vs. a regression without something to compare against.

@sdilauro

sdilauro:
You're right, I didn't add any details about the design other than that somewhat generic description. Actually, I thought you'd be testing more for functional errors than for consistency with the design. I'll add the link to the [Figma file](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=994-11220&m=dev) in the PR description and take your advice to include before-and-after screenshots.

@Ludmilafantaniella Sorry for such sloppy work!!

Ludmilafantaniella:
Hey! Tested build 1673 on mobile against the Figma - orb redesign looks correct now, all good.

🐛 Found one issue though:
Bug: Joypad buttons (2-4) don't come back after visiting a scene with no action buttons

STR:
1. Be in Genesis Plaza with joypad "+" showing the expanded 1/2/3/4 buttons normally
2. Teleport to a scene that has no [E]/[F]/[click]/action button options (e.g. Space Runner)
3. Teleport back to Genesis Plaza
4. Tap "+" to expand the joypad

Expected: buttons 1-4 show up around the "+" like before.
Actual: 2-3 don't appear only the "+" button shows and 1 misaligne.

Workaround: cancel the teleport back to Genesis Plaza (go to Discover) and reopen it from there, or restart the app - buttons come back either way.

✅  Rest of the test plan (Cases 1-6) all passed correctly.


https://github.com/user-attachments/assets/6cc158a4-739c-4b8a-b87b-7633db69312e

Screenshot_20260827_112527_Decentraland

cc @sdilauro

Ludmilafantaniella (approved):
## ✅ QA Approved

Platforms tested: Android & iOS
Verified fixes:
- Orb-style HUD redesign now matches the Figma reference (glow, lilac outline, comet trail on joystick drag) - resolves the visual mismatch flagged earlier in this thread.
- Joypad combo buttons (1-4) correctly reappear around the "+" after visiting a scene with no action buttons and teleporting back to Genesis Plaza - resolves the misalignment/missing-buttons bug reported.

Result: Both previously reported issues are confirmed fixed. No new blockers found. Approving from QA side.

https://github.com/user-attachments/assets/d1dd9d6b-d227-440c-9e4e-1e1817b4272c

ElzathRusso (approved):

PR #2728, opened 2026-08-20
fix(sdk): read AvatarLocomotionSettings from the PLAYER entity

### LINKED ISSUE #2682: [Bug] AvatarLocomotionSettings changes not applied on mobile (CleanTheClub sticky patch)
### Platform
Android / iOS (mobile client)

### App Version
v1.11.0

### Device Information
Not specified

### Issue description
`AvatarLocomotionSettings.createOrReplace(engine.PlayerEntity, {...})` has no effect on movement speed on mobile. On desktop, changing `walkSpeed`/`jogSpeed`/`runSpeed` correctly slows the player down; on mobile the player keeps moving at normal speed.

Repro scene: [CleanTheClub](https://github.com/CleanTheClub/CleanTheClub), published at `afkj.dcl.eth`. The scene's `stickyHazardSystem.ts` calls:
```ts
AvatarLocomotionSettings.createOrReplace(engine.PlayerEntity, {
  jogSpeed:  SLOW_JOG_SPEED,   // 0.4
  walkSpeed: SLOW_WALK_SPEED,  // 0.2
  runSpeed:  SLOW_RUN_SPEED,   // 0.4
})
```
when the player steps on a "sticky patch" hazard (polled locally via `Transform.getOrNull(engine.PlayerEntity)`, not a trigger area). This works correctly on desktop but does nothing on mobile.

### Expected behaviour
When standing on a sticky patch, the player's movement should be temporarily slowed (per the applied `AvatarLocomotionSettings` values), matching desktop behaviour.

### Actual behaviour
On mobile, the player continues moving at normal (unslowed) speed — the locomotion settings change appears to have no effect.

### Screenshots/Media

Image

### Steps to reproduce
1. Open `afkj.dcl.eth` on mobile.
2. Walk onto one of the sticky patches placed in the scene.
3. Observe that movement speed does not decrease (works fine on desktop).

### Occurrence
Always (per reporter)

### Theories
1. The Rust→GDScript locomotion pipeline (`avatar_locomotion_settings.rs` → `DclLocomotionSettings` → `player.gd::_apply_locomotion_settings()`) appears platform-agnostic in the source, but mobile's movement-speed branch in `player.gd` (`elif Global.is_mobile(): ...`) computes `effective_speed` differently from desktop — it lerps between `walk_speed` and `jog_speed` based on joystick `input_magnitude` and only reaches `run_speed` when `Input.is_action_pressed("ia_sprint")`. If joystick input magnitude/flags (`walk_disabled`/`jog_disabled`) behave differently on mobile touch input, the applied slow values may not propagate the same way as on desktop.
2. Possible timing issue: `locomotion_settings_changed` signal is emitted deferred and only when `scene.scene_id == current_parcel_scene_id` — worth verifying this still holds true for the mobile scene-loading/focus path.
3. Could be related to broader mobile scene-interaction/trigger issues (see #1707), though this specific script polls `Transform.getOrNull(engine.PlayerEntity)` directly rather than using SDK trigger areas, so it may be an unrelated regression specific to `AvatarLocomotionSettings` runtime application on mobile.

---
Requested by Rodri via Slack

### LINKED ISSUE #2727: AvatarLocomotionSettings is read from the ROOT entity instead of PLAYER
### Problem

`AvatarLocomotionSettings` has no effect. Scenes set it as the protocol specifies:

```ts
AvatarLocomotionSettings.createOrReplace(engine.PlayerEntity, { walkSpeed, jogSpeed, runSpeed })
```

`engine.PlayerEntity` is entity **1**. But `update_avatar_locomotion_settings` only looks at entity **0**:

```rust
// lib/src/scene_runner/components/avatar_locomotion_settings.rs
// Only process ROOT_ENTITY (id = 0)
if locomotion_dirty.contains(&SceneEntityId::ROOT) {
```

The dirty check never matches, `locomotion_settings_changed` is never emitted, and `player.gd::_apply_locomotion_settings()` never runs. The player keeps the engine defaults. No error, no warning.

This was a deliberate choice in the original implementation (#1378: *"process the component on ROOT entity"*), so the component has never worked for spec-compliant scenes.

### Why PLAYER is correct

- Every other player-targeted component in the same directory already uses `SceneEntityId::PLAYER` — `input_modifier.rs`, `physics_combined.rs` — as does the player transform the engine writes back (`update_scene.rs`). This one is the outlier.
- The reference implementation ignores the component on any other entity:
  ```csharp
  // Only apply settings if the component was added to the local player entity
  if (crdtEntity.Id != SpecialEntitiesID.PLAYER_ENTITY) return;
  ```

### Steps to reproduce

1. Load a scene that calls `AvatarLocomotionSettings.createOrReplace(engine.PlayerEntity, { walkSpeed: 1, jogSpeed: 1, runSpeed: 1 })`.
2. Walk around.
3. Speed is unchanged.

Repro scenes: https://github.com/iillee/mobile_test, and `afkj.dcl.eth` from #2682.

### Note

This is not platform-specific — the same Rust path runs everywhere. #2682 reported it as mobile-only because the comparison was against the reference explorer, not against a desktop build of this client.

leanmendoza:
Fixes #2727
Fixes #2682

`AvatarLocomotionSettings` was read from `SceneEntityId::ROOT` (id 0), but scenes set it on `engine.PlayerEntity` (id 1) as the protocol specifies. The dirty check never matched, so `locomotion_settings_changed` was never emitted and `player.gd` kept the engine defaults — `walkSpeed`/`jogSpeed`/`runSpeed` had no effect at all.

Read `SceneEntityId::PLAYER` instead. Every sibling component that targets the player (`input_modifier`, `physics_combined`, and the player transform the engine writes back) already does. Removal behaviour is unchanged: a missing entry or a cleared value resets to the engine defaults.

One file, no new API.

### Test plan

Any scene that sets the component on `engine.PlayerEntity` — e.g. https://github.com/iillee/mobile_test, or `afkj.dcl.eth` from #2682:

1. Enter the scene and confirm the avatar slows to the values the scene sets.
2. Leave the scene bounds and confirm the defaults come back.
3. Re-enter and confirm the settings are reapplied.

Worth checking on both desktop and mobile: on mobile the top speed is `jog_speed` (`run_speed` needs `ia_sprint`, which mobile has no input for), so a scene that only lowers `runSpeed` still won't change anything there. That is existing behaviour and out of scope here.

### DIFF
--- a/lib/src/scene_runner/components/avatar_locomotion_settings.rs
+++ b/lib/src/scene_runner/components/avatar_locomotion_settings.rs
@@ -9,7 +9,14 @@ use crate::{
     scene_runner::scene::Scene,
 };
 
-/// Updates the locomotion settings for the scene.
+/// Updates the locomotion settings for the scene from the `PBAvatarLocomotionSettings`
+/// component on the PLAYER entity (id = 1), which is where scenes address it:
+/// `AvatarLocomotionSettings.createOrReplace(engine.PlayerEntity, { ... })`.
+///
+/// This mirrors the other player-targeted components (`InputModifier`,
+/// `PhysicsCombinedForce`) and the reference implementation. The component is
+/// ignored on any other entity.
+///
 /// Returns `true` if the settings were changed, `false` otherwise.
 pub fn update_avatar_locomotion_settings(
     scene: &mut Scene,
@@ -17,25 +24,25 @@ pub fn update_avatar_locomotion_settings(
 ) -> bool {
     let dirty_lww_components = &scene.current_dirty.lww_components;
 
-    if let Some(locomotion_dirty) =
-        dirty_lww_components.get(&SceneComponentId::AVATAR_LOCOMOTION_SETTINGS)
-    {
-        // Only process ROOT_ENTITY (id = 0)
-        if locomotion_dirty.contains(&SceneEntityId::ROOT) {
-            let locomotion_component =
-                SceneCrdtStateProtoComponents::get_avatar_locomotion_settings(crdt_state);
+    let is_dirty = dirty_lww_components
+        .get(&SceneComponentId::AVATAR_LOCOMOTION_SETTINGS)
+        .is_some_and(|dirty| dirty.contains(&SceneEntityId::PLAYER));
 
-            if let Some(value) = locomotion_component.get(&SceneEntityId::ROOT) {
-                if let Some(proto) = &value.value {
-                    scene.locomotion_settings.bind_mut().set_from_proto(proto);
-                } else {
-                    scene.locomotion_settings.bind_mut().reset_to_defaults();
-                }
-            } else {
-                scene.locomotion_settings.bind_mut().reset_to_defaults();
-            }
-            return true;
-        }
+    if !is_dirty {
+        return false;
     }
-    false
+
+    let locomotion_component =
+        SceneCrdtStateProtoComponents::get_avatar_locomotion_settings(crdt_state);
+    let player_settings = locomotion_component.get(&SceneEntityId::PLAYER);
+
+    // A missing entry or a cleared value means the scene removed the component:
+    // fall back to the engine defaults.
+    let mut settings = scene.locomotion_settings.bind_mut();
+    match player_settings.and_then(|entry| entry.value.as_ref()) {
+        Some(proto) => settings.set_from_proto(proto),
+        None => settings.reset_to_defaults(),
+    }
+
+    true
 }


regenesis-claw (approved):
Reviewed against `REVIEW.md`. Single-file Rust change (+27/−20), verified line by line.

**Verdict: approve.** No P0/P1. The fix is correct and the description's claims check out against the code.

### Claims verified
- `SceneEntityId::PLAYER` is `reserved(1)`, `ROOT` is `reserved(0)` (`lib/.../dcl/components/mod.rs:60-61`) — so the old `contains(&SceneEntityId::ROOT)` dirty check could never match a spec-compliant `AvatarLocomotionSettings.createOrReplace(engine.PlayerEntity, …)`. The "no effect at all" diagnosis holds.
- "Every sibling player-targeted component already does this": confirmed — `input_modifier.rs` and `physics_combined.rs` both key on `SceneEntityId::PLAYER`, and `update_scene.rs` writes the player transform to the same entity. This file was the outlier.
- "Removal behaviour is unchanged": confirmed — `player_settings.and_then(|e| e.value.as_ref())` collapses missing-entry and cleared-value into the same `reset_to_defaults()` branch the old nested match had. `set_from_proto` also `unwrap_or(DEFAULT_*)`s every unset field, so a partial component never inherits stale values.
- Test-plan step 2/3 (defaults on exit, reapplied on re-entry): supported by the existing plumbing — `player.gd:231-256` re-reads `get_current_scene_locomotion_settings()` on `on_change_scene_id`, and `scene.locomotion_settings` persists per-scene, so re-entry restores the scene's values without needing a fresh dirty tick.
- The early-return refactor is behaviour-preserving: previously a dirty entry for a non-matching entity fell through to `false`; now `is_some_and(...)` produces the same `false`.

### P2 — non-blocking
1. **Silent behaviour change for any scene that adopted the old ROOT-entity quirk.** Since this client only ever honoured entity 0, a scene author who tested against this client specifically may have set the component on `engine.RootEntity`; after this PR it is ignored with no log. That is the right call (it matches the reference implementation, which returns early for non-PLAYER entities), so no change requested — just noting the compatibility surface in case a scene reports a regression.
2. **No regression test.** The bug was a one-constant mismatch that no test would have caught, and the same class of mistake can silently return. A small `#[cfg(test)]` case that puts the component on `PLAYER`, asserts `true` + the applied speeds, then puts it on `ROOT` and asserts `false`, would pin the entity contract in `cargo test`. Optional given the repo has no precedent for component-level unit tests here.
3. **Pre-existing, out of scope:** the emit in `update_scene.rs:504` is gated on `scene.scene_id == *current_parcel_scene_id`, so a global scene / portable experience setting `AvatarLocomotionSettings` still never reaches `player.gd`. Unrelated to this diff — worth a follow-up issue if global scenes are meant to drive locomotion.

### Pre-flight
- Branch is **up to date with `main`** (ahead 1, behind 0).
- **iOS build not required**: pure Rust scene-runner logic, no `plugins/dcl-godot-ios/**`, no `OS.get_name() == "iOS"` branch, nothing under `comms/` or `av/`. The mobile symptom in #2682 comes from the same platform-agnostic path.
- No submodule pointer drift; no `.claude/` files; no `project.godot` / export-preset changes; comments are English and explain *why* (entity contract + reference-impl parity), matching the CLAUDE.md guidance.
- No new logs added, so no Sentry-quota impact.

### Security
No security issues found. No new inputs, network calls, logging of user data, or public API surface; the change narrows which entity a scene can drive locomotion from.

### CI
`Clippy & Build Checks`, `Static checks` and `Gate` green at the time of review; Linux / Android / iOS still running. Nothing in the diff can plausibly break them, but please confirm Linux is green before merge.

---
Requested by Lean via Slack

kuruk-mm (approved):
fix makes sense! if you tested, lets merge it

PR #2754, opened 2026-08-24
feat: localization (ES + PT-BR ready)

### LINKED ISSUE #2667: Localization v1: ES + PT-BR support across entry flows and in-game menus
##  **⚠️ Problem:**
The mobile client ships English-only, so players acquired in LATAM land in a UI they cannot read — from the sign-in and onboarding screens through to the settings, backpack, and map menus. 90% of our marketing acquisition targets LATAM (Spanish and Portuguese-speaking markets), so every campaign dollar currently delivers users into an English-only client, which suppresses Playtime and inflates effective Cost per Acquisition.

## **🏁 Scope:**
Introduce a localization layer in the Godot client covering all first-party UI strings across the entry flow (splash, sign-in/guest, onboarding, loading screens, error and permission dialogs) and the in-game menus (settings, backpack, map, profile/passport, chat UI chrome, emote wheel, notifications), shipping English, Spanish, and Portuguese (BR). Out of scope: user-generated and scene content, creator-authored `TextShape` strings, server-driven copy (quests, events, place descriptions), RTL/CJK script support, and any language beyond the three named.

## **📝 Deliverables:**
- [ ] All hardcoded UI strings in entry-flow scenes replaced with translation keys resolved via Godot's `TranslationServer` / `tr()`
- [x] All hardcoded UI strings in in-game menu scenes replaced with translation keys
- [x] Translation source files committed and wired into the build (see Implementation Specs), with a POT/extraction step that fails CI if a scene contains a literal UI string
- [x] Locale auto-detected from device locale on first launch, with fallback to English when the locale is unsupported or a key is missing — never a blank string or raw key
- [x] Manual language override in Settings, persisted across sessions, applied without requiring an app restart
- [x] ES and PT-BR translations delivered and reviewed for the full key set
- [ ] Layout audit: no truncation, clipping, or overflow in any localized screen — ES/PT strings run ~20–30% longer than EN
- [x] Font coverage verified for the full Latin-1 Supplement range (á é í ó ú ñ ü ã õ ç) in all UI fonts, including the number/HUD fonts
- [ ] `language` (resolved locale) added as a property on session-start analytics so Playtime can be segmented by locale
- [x] `SETTINGS_LANGUAGE_CHANGED` event implemented, carrying `from_locale` and `to_locale`

## Metrics
1. **Target metric:** Playtime (secondary: Cost per Acquisition, via retained value per acquired user).
2. **Expected direction:** Increase median session length and D1 return rate for users on ES/PT-BR device locales, closing the gap against the EN-locale cohort. Baseline gap to be measured at implementation time — we do not currently segment Playtime by locale, which is why the analytics deliverables above are in scope.
3. **How it's measured:** Segment session events segmented by the new `language` property, compared EN vs ES vs PT-BR cohorts in Metabase/Snowflake, over the 4 weeks pre/post release. Paired with LATAM campaign cohorts to read the CPA effect.

## Implementation Specs:
**Recommendation: use Godot's native localization stack. No third-party plugin is required.**

| Concern | Approach |
|---|---|
| Runtime lookup | `TranslationServer` + `tr()` / `atr()`. Control nodes auto-translate their `text` property, so most menu strings need only a key swap. |
| Source format | **CSV** (`.csv` with `keys,en,es,pt_BR` columns) — Godot imports it natively into `.translation` resources, and it diffs and reviews cleanly in PRs. Gettext `.po` is the alternative if we want translator-tool ergonomics over diff readability. |
| Key extraction | Built-in POT generation (`Project → Project Settings → Localization → POT Generation`) to enumerate translatable strings and catch stragglers; run in CI to detect untranslated literals. |
| Locale detection | `OS.get_locale()` / `OS.get_locale_language()` on boot; `TranslationServer.set_locale()` for the manual override. Godot's locale fallback resolves `pt_BR → pt → en` automatically. |
| Plurals / interpolation | `tr_n()` for plurals; `String.format()` with named placeholders — never string concatenation, which breaks word order in ES/PT. |
| Fonts | Godot 4 dynamic fonts with `fallbacks` configured. ES/PT need only Latin-1 Supplement, which our current UI font almost certainly already covers — verify rather than assume. No CJK/RTL work, so `TextServerAdvanced` is not required for this scope. |
| Translation management | If we want a TMS rather than hand-editing CSV, **Crowdin** or **Lokalise** both import/export Godot CSV and gettext and support a GitHub sync. Optional — recommend deferring until we exceed 3 languages. |

Godot plugins reviewed and **not** recommended for this scope: `godot-localization-editor` and similar asset-library editors add a spreadsheet UI over the same underlying `TranslationServer`, introducing a maintenance dependency for convenience we don't need at three languages.

## **🔗 References:**
- [decentraland/unity-explorer#150](https://github.com/decentraland/unity-explorer/pull/150) — the desktop client's original localization system; the reference implementation and the parity target for key naming.
- [decentraland/unity-explorer#2117](https://github.com/decentraland/unity-explorer/issues/2117) — desktop's own list of surfaces that were never localized (chat, backpack, passport, map, badges, quests, settings, profile view). It is the same surface list we must cover here, and a warning: on desktop, localization was added and then let rot as features shipped without keys. Hence the CI extraction deliverable above.
- [decentraland/unity-explorer#270](https://github.com/decentraland/unity-explorer/pull/270) — ES and IT were *removed* from desktop because locales were enabled without translated content behind them, crashing users whose OS was set to Spanish. Do not enable a locale before its table is complete.
- [decentraland/unity-explorer#2062](https://github.com/decentraland/unity-explorer/pull/2062) — the same class of failure: a Spanish OS locale left the desktop loading screen stuck. Locale fallback on the entry flow must be tested explicitly, not assumed.

## Notes
- Translation sourcing (vendor, in-house, or community) is not settled and is not blocking implementation — keys and English strings can land first.
- No prior localization issue exists in godot-explorer (searched `localization`, `i18n`, `translation`, `language`, 2026-08-08); this is the first.

EibrielInv:
## Summary

Adds a localization layer to the Godot client. **Ships dark:** the language picker is gated to
non-production builds and every build resolves to English, so a store build behaves exactly as it
does today. Lifting the restriction is two flags in `locale_settings.gd`
(`FOLLOW_DEVICE_LOCALE`, `is_language_picker_available()`), both marked `TEMPORARY`.

- **697 translation keys** across the entry flow, in-game menus and shared components, fully
  translated into `es` and `pt_BR`
- Per-locale CSV catalogues in `godot/locale/` (`en`, `es`, `pt_BR`), imported by Godot's
  built-in `csv_translation` importer
- `TranslationKey` (`godot/src/config/translation_key.gd`) makes "passed English where a key
  belongs" a **parse error** rather than something a scanner has to notice
- Locale-aware number and date formatting (separators, 12h/24h, month order)
- Plurals via `tr_n()` with per-locale rules; `plural()` selects the form and `format()` fills it,
  as two steps
- **Placeholders are named, never positional.** `"I like %s %s"` cannot become *"las flores
  rojas"* — the adjective has to move — so `%` specifiers are rejected in the catalogue and
  `{named}` fields are required

Closes #2667

## Tooling (`tools/i18n/`)

Stdlib Python only — no new dependencies, so every check runs on Windows, macOS, Linux and CI
with nothing installed. ~1,150 lines of tooling plus ~720 lines of tests.

| File | What it does |
|---|---|
| `catalogue.py` | Shared model for the CSV format — parsing, canonical writing, plural continuation rows, placeholder extraction. The other two scripts build on it. |
| `extract_strings.py` | Scans `godot/src/ui` (plus `config/`, `logic/`, `notifications/`, `global.gd`, …) for user-facing strings and **fails CI on any new hardcoded UI text**. Also verifies every key used in the UI exists in `en.csv`, and that no `auto_translate_mode = 2` node holds a translation key — a node that never looks text up draws the key verbatim on screen. |
| `format_csv.py` | Canonicalizes and validates the catalogues: plural naming, form counts, `\n` escapes, named-field parity across locales, staleness against `translation_sources.json`, and that `project.godot` references every catalogue that has translations. |
| `test_extract_strings.py`, `test_format_csv.py` | 72 unit tests over both, each building a throwaway UI tree in a temp dir. |

Supporting data, all human-editable:

- `unkeyed_baseline.txt` — debt ratchet for strings not keyed yet. **Currently 0**; it may only shrink.
- `not_translatable.txt` — permanent exclusions (design-time placeholders, SDK error codes, version strings), each with a justification comment.
- `translation_sources.json` — hash of the English each translation was made against, so a reworded source flags its translations as stale.
- `glossary.md` — register (informal *tú* / *você*) and the terms that stay in English.
- `README.md` — the workflow, including the manual pseudolocale sweep CI cannot run.

Wired into CI as three steps in `static_checks.yml` (unit tests, catalogue check, keyed-string
check) plus `cargo run -- test-i18n` in `linux_builds.yml`, which boots headless Godot to test
`TranslationKey` against the real catalogue.

## QA Steps

**Setup:** set the device language to a non-English one (e.g. Spanish or Portuguese (Brazil)),
then fully close the app.

1. Open the app and sign in.
2. Browse the app as usual.

**Expected result:**
- [ ] The UI is in **English** everywhere, regardless of the system language.
- [ ] The UI looks the same as before — no clipped, missing or misplaced text.

## Future plans

- Lift the restriction: `FOLLOW_DEVICE_LOCALE = true` and drop
  `is_language_picker_available()`, after a native-speaker review of the ES/PT copy.
- Add a **"System default"** entry to the picker — it currently only emits a concrete locale, so
  once a user chooses one there is no way back to following the device.
- Two push-notification titles ("Event Starts in an Hour", "New Item Received") are composed
  server-side and are not translatable from this repo.


regenesis-claw:
## Review notes (comment only — no formal review, as requested)

Reviewed at head `febf282`, following `REVIEW.md`. Draft, so no verdict — but for what it's worth I found **no P0/P1**. The catalogue tooling, the `TranslationKey` type and the `auto_translate_mode = 2` sweep are unusually well-defended for a first localization pass.

**Pre-flight:** 🍏 iOS export is green; all static checks (including the three new i18n steps) pass. Branch is 2 commits behind `main` (diverged) — worth a merge before it leaves draft.

**What I verified independently (not just author-reported):**
- 664 keys × 3 locales: no missing keys, no extra keys, **zero empty translations** in `es` / `pt_BR` — so `SUPPORTED_LOCALES` isn't shipping a half-translated locale (the unity-explorer#270 failure mode the code comments cite).
- Named `{placeholder}` sets and BBCode tag sets match the English source for every translated form in both locales (`format_csv.py`'s `format_specs` only covers `%`-style, so I checked these separately).
- `unkeyed_baseline.txt` is empty — the ratchet starts at zero rather than grandfathering debt.
- SDK scene UI is opted out at both `base_ui` construction sites in `scene_manager.rs`, and 105 nodes carry `auto_translate_mode = 2`.

### P2 — worth fixing before this leaves draft

1. **`godot/project.godot`: `boot_splash/show_image=true` was removed.** Unrelated to i18n and it's local editor drift. It's a no-op (the property defaults to `true`), but per `REVIEW.md` Tier 1 §5 please restore the line so the diff stays scoped.

2. **`settings.gd::_populate_cache_size_items()` changed the dropdown item *ids*.** It was `add_item("1 GB", 0/1/2)`; now the id is `gigabytes` (1/2/4), which no longer matches the `CACHE_SIZE_MB` index that `_on_dropdown_list_max_cache_size_item_selected()` writes into `max_cache_size`. Nothing calls `DropdownList.get_item_id()` on this dropdown today, so it's inert — but it's a latent trap. Suggest iterating with the index:
   ```gdscript
   for i in [1, 2, 4].size():
       dropdown_list_max_cache_size.add_item(
           TranslationKey.new("SETTINGS_CACHE_SIZE_GB").format_named({"size": [1, 2, 4][i]}), i
       )
   ```

3. **No way back to "follow device locale".** `config.locale` starts `""` and `LocaleSettings.set_locale("")` explicitly supports clearing the override, but the picker only ever emits a concrete locale — once a user touches the dropdown they're pinned for good, even if they later change their phone language. A "System default" first item would close that.

4. **Re-entrancy on language change (needs a device check, reads fine on paper).** `DropdownList._on_item_pressed()` does `select()` → `item_selected.emit()` → `_close_popup(true)`. The new handler runs `set_locale()` → `TranslationServer.set_locale()` → `NOTIFICATION_TRANSLATION_CHANGED` → `_populate_language_dropdown_items()`, which `clear()`s and rebuilds *that same dropdown* (and `clear()` itself calls `_close_popup()`) while still inside the emit. It should survive — the outer `_close_popup(true)` early-returns on `_is_open == false` — but this is the one path I'd want confirmed on a real phone, popup open, mid-tap.

5. **`_setup_language_dropdown()` hides only the `DropdownList`,** leaving `Container_Language` and its `SectionTittle_Language` header on screen. Unreachable today (3 selectable locales), but if it ever fires the user gets a "LANGUAGE" section title with nothing under it — hide the container instead.

6. **Two untranslated leftovers in `notification_text_helper.gd`:** `metadata.get("title", "Notification")` in the `"item_sold", "bid_accepted", "bid_received", "royalties_earned"` branch and in the `_` default. Every neighbouring fallback got keyed, so these read as oversights rather than intent.

7. **`LocaleFormat.short_date()` calls `TranslationServer.translate("DATE_SHORT_FORMAT")` directly** while the rest of the file goes through `TranslationKey`. Safe (the scanner still sees the key), just inconsistent with the type contract the PR argues for two files earlier.

8. **Positional `%s` inside BBCode strings is the fragile spot in the catalogue design.** `NOTIF_TITLE_WANTS_TO_BE_YOUR_FRIEND_2/_3` and the `..._ACCEPTED_YOUR_FRIEND_REQUEST_2/_3` pair carry `[color=#%s]%s[color=#…]` — `format_specs()` compares an *ordered list of conversion types*, so two same-type placeholders swapped by a translator (colour ↔ name) validates clean and renders a colour hex as the username. ES/PT are correct today (checked). `format_named()` for the BBCode-carrying keys would make that class of error impossible.

9. **Checker blind spot:** `extract_strings.py` scans `godot/src/ui` plus `config/`, `logic/`, `global.gd`, `connection_quality_monitor.gd`, `notifications_manager.gd`. `godot/src/decentraland_components/` isn't scanned even though this PR keys strings there (`nickname_ui.tscn`, `text_label3d_renderer.gd`), so new hardcoded strings in that tree won't be gated. Worth adding to `EXTRA_ROOTS` or noting as a follow-up phase.

10. **`TranslationKey._init()` uses `push_error`** for a malformed key. It's debug-gated so release/Sentry isn't affected, but per `REVIEW.md` §13 a `push_warning` would carry the same developer signal without routing through the Godot error stream on debug device builds.

**Security:** no findings — no secrets, no network/user-controlled input in the Python tooling (it only reads repo files in CI), and translation data is repo-controlled.

---

regenesis-claw (approved):
## Delta review — head `3718500` (previous pass was `febf282`)

Re-reviewed against `REVIEW.md`. **No P0/P1.** Approving.

**Pre-flight:** 🍏 iOS export, 🤖 Android, 🐧 Linux, 🐳 Docker, Clippy and 📊 Static checks (incl. the three i18n steps) are all green. Branch is **1 commit behind `main`** (diverged) — worth a merge before you hit the button, but nothing in that commit touches this surface.

### Previous findings — status

| # | Finding | Status |
|---|---|---|
| 1 | `boot_splash/show_image` dropped from `project.godot` | ✅ restored |
| 2 | Cache-size dropdown ids no longer matched `CACHE_SIZE_MB` | ✅ fixed — ids are the index now, and the GB labels are derived from `CACHE_SIZE_MB` so the two can't drift |
| 3 | No way back to "follow device locale" | ⚠️ still open, now documented under *Future plans*. Moot while the picker is hidden — fine as a follow-up |
| 4 | Re-entrancy when changing language from the open dropdown | ✅ confirmed safe on paper: `DropdownList.select()` doesn't emit `item_selected`, and `clear()`'s `_close_popup()` is guarded by `_is_open`. Still the one path worth a real-device tap in a non-prod build |
| 5 | Hiding the dropdown left an orphan "LANGUAGE" header | ✅ fixed — `container_language.hide()` on both early-outs |
| 6 | Untranslated `"Notification"` fallbacks in `notification_text_helper.gd` | ✅ fixed — `NOTIF_HEADER_NOTIFICATION` |
| 7 | `short_date()` bypassed `TranslationKey` | ✅ fixed |
| 8 | Positional `%s` inside BBCode keys could silently swap slots | ✅ fixed at the root — `format_csv.py` now rejects every `%` specifier, `format_named` collapsed into `format`, and the parity check compares the *set* of named fields instead of an ordered list of conversion types |
| 9 | `decentraland_components/` outside the scanner's roots | ⚠️ still outside; `godot/src/notifications` was added instead. Verified that tree holds no user-facing copy today (only design-time placeholders and debug glyphs), so this is low-risk follow-up, not a gap that matters now |
| 10 | `push_error` in `TranslationKey._init()` | ⚠️ kept. It's `OS.is_debug_build()`-gated so no Sentry quota impact — leaving it as your call |

The new `keys_on_untranslated_nodes()` check in `extract_strings.py` is the best addition in this round: it closes exactly the hole that shipped `EMOTES_EMOTES` to the emote wheel, and it caught eight more (`COMMON_UNKNOWN_PLACE`, `COLOR_CARROUSEL_HAIR_COLOR`, `CODE_MODAL_*`, `DIALOGS_LOADING_NFT`, `REWARD_MODAL_*`, `COMMON_SOMETHING_WENT_WRONG`, `LOADING_ENJOY_EXPERIENCE`, `PROFILE_USE_THIS_SPACE_*`).

### What I verified independently (not author-reported)

- **Catalogue integrity, parsed by hand rather than trusting CI:** 698 keys × 3 locales — no missing keys, no extra keys, **zero empty translations**, form counts match, **zero `%` specifiers anywhere**, named-`{field}` sets identical across `en`/`es`/`pt_BR`, and BBCode tag sets identical too. The plural rules are `n != 1` (es) and `n > 1` (pt_BR), and all 11 plural entries carry the named field their caller supplies in **both** forms.
- **Ran the tooling locally at this head:** 72 unit tests pass, `format_csv.py --check` passes (697 keys / 3 locales), `extract_strings.py` passes (673 keys in use, 0 awaiting conversion).
- **The `plural()` contract change is fully propagated.** `plural()` no longer substitutes, so every call site must chain `.format(...)` — all 9 do (`event_pills_bar`, `credits_history_item`, `place_item` ×3, `notification_item` ×3, `mutual_friends`). No `format_named(` survives outside a docstring in the test fixtures.
- **Every scene `text = "KEY"` you deleted has a code assignment behind it** — checked all eight individually, including that `Label_SceneTitle` in the mini-map is unconditionally written on `on_change_scene_id`. No label ends up blank in English.
- **Consumer check for the `"View profile"` → `TOOLTIP_VIEW_PROFILE` wire change:** exactly two references exist repo-wide (`scene_manager.rs` emitting, `explorer.gd` matching) and both moved together, with `TooltipLabel.VIEW_PROFILE_KEY` as the single source. `tooltip_label._resolve()` translates only that one constant, so creator-authored PointerEvents hover text still passes through untouched — the important half.
- **The "ships dark" claim holds.** `resolve_locale()` returns `FALLBACK_LOCALE` up front whenever `is_language_picker_available()` is false, *before* reading `config.locale` — so a stored `"es"` written by a QA build on the same device can't surface in a store build. `apply_locale()` also can't enable pseudolocalization there.

### P2 — follow-ups, none blocking

1. **The whole gate rests on `is_production()` = `GODOT_EXPLORER_VERSION.contains("-prod")`.** That's the same gate the Dev Tools tab uses, so it's the right one — but it means any store build cut without `--prod` ships the language picker *and* the `xx-pseudo` option. Worth one glance at the release workflow before this goes out, since the blast radius is now larger than a hidden Dev Tools tab.
2. **`is_pseudolocale_available()` widened** from `OS.is_debug_build() and not is_production()` to just `not is_production()`. Correct fix for the QA-visibility problem, but it also means a QA tester can now leave a build stuck in `«____bracketed____»` mode and report it as a bug. A "System default" entry (already on your list) would double as the escape hatch.
3. **Mode-2 defaults resolved in `_ready()` don't re-translate on a live language switch** — loading screen place name, color carrousel title, NFT dialog title, reward modal body. `tip_card`, `emote_wheel` and `settings` all handle `NOTIFICATION_TRANSLATION_CHANGED`; these four don't. Invisible today (English only), relevant the moment the picker ships.
4. **`TranslationServer.translate()` used directly** in `place_item._recurrent_frequency_text()`, `tooltip_label._resolve()` and `notification_utils`. Same inconsistency finding 7 fixed one file over — the keys are declared via `# i18n-keys:` markers so nothing is unchecked, it's just two idioms for one job.
5. **Stale docstring** in `test_translation_key.gd`: the header now reads `text()/format()/format()/upper()` after the `format_named` rename.
6. **`NO_SERVICE_MESSAGE = "FRIENDS_SOMETHING_WENT_WRONG_AND_WE_COULDN"`** — an auto-generated key truncated mid-word. Harmless (keys are opaque), but it will read as a typo to the next person.

**Security:** no findings. The Python tooling only reads repo files during CI, translation data is repo-controlled, and no user- or server-supplied string reaches a translation lookup except through the single `== VIEW_PROFILE_KEY` equality test.

---
Requested by Mateo via Slack

leanmendoza (approved):

PR #2762, opened 2026-08-24
fix: sign the metadata the same way it is sent

LautaroPetaccio:
Signs the payload format `@dcl/crypto-middleware` 6.x verifies.

## What changed

Both signing paths folded the whole payload before signing, while `x-identity-metadata` delivered the metadata unfolded:

```rust
// lib/src/auth/wallet.rs
let payload = format!("{}:{}:{}:{}", method, uri.path(), unix_time, meta).to_lowercase();
// lib/src/comms/signed_login.rs
let payload = format!("post:{}:{}:{}", uri.path(), unix_time, meta).to_lowercase();
```

The signature only ever covered the folded form, so a key or value could be re-cased between signing and delivery and still verify — and services read the delivered header, so they were authorizing on bytes the signature never covered.

Now only the method and path are lowercased and the metadata is interpolated verbatim.

## One helper, not two fixes

There were two payload builders differing only in whether the method was hardcoded. Two copies of a signing format is how they drift, so both now call `signed_fetch_payload`, and that one function is what every Decentraland verifier reconstructs.

## Testing

Not built locally, and I would rather say so than imply otherwise: the build script reads generated protobuf components from an uninitialized submodule, so `cargo check` cannot run in this checkout.

What I did verify: both edited files parse and are `rustfmt --check` clean, and the new import mirrors the existing `crate::auth::wallet::sign_pulse_connect` cross-module import exactly. The compile is on CI.

## ⚠️ Deploy ordering — this one is ordering-sensitive

**Do not merge before every service this client calls verifies the 6.x payload.**

A verifier still on the old format folds both sides. It therefore accepts this signature while the metadata is all-lowercase, and rejects it the moment any uppercase appears — and this client's metadata is camelCase throughout (`realmName`, `sceneId`, `isGuest`, `realm.serverName`). So the breakage is not gradual: every signed request carrying real scene metadata starts failing at once.


kuruk-mm (approved):
Looks great

kuruk-mm (approved):

Issue #2670, opened 2026-08-08
Ad-Based Dynamic FTUE Bypass to Scene (Android Only)


rodrigo685:
##  **⚠️ Problem:**
User watches a scene ad video, installs, then lands on the generic FTUE Discover carousel — the scene they were sold is 2+ taps away and may not even be in the carousel. Ad→content discontinuity inflates Cost per Acquisition; install→first-scene-load conversion is not currently measurable per campaign.

## **🏁 Scope:** [See research #2331](https://github.com/decentraland/godot-explorer/issues/2331)
**Android only.** First launch: resolve the Android install attribution payload to a target scene, boot straight into that scene's loading screen, skip FTUE. Unknown token, expired window, resolver failure, offline, organic install, or reinstall → today's FTUE, unchanged. No iOS client code, no MMP SDK.

## **📝 Deliverables:**
- [x] `assetlinks.json` verified for every deep-link target domain — `mobile.dclexplorer.com` serves a valid one for `org.decentraland.godotexplorer`. Note `decentraland.org` and `decentraland.zone` still return SPA HTML, so ad groups must target the former.
- [x] GA4F flag `google_analytics_deferred_deep_link_enabled` in Android plugin manifest
- [x] Kotlin: read `google.analytics.deferred.deeplink.prefs` (listener + direct read at startup), expose to engine
- [x] Rust plugin wrapper, mirroring existing install-referrer accessor
- [x] Boot router: valid token + first launch → scene loading screen, FTUE skipped
- [x] Priority rule: Google Ads deferred deep link > Play Install Referrer when both present
- [x] Route through pre-boot private-world gate — bypassing it caused #2569
- [x] Fallback path implemented + tested explicitly: unknown / expired / resolver fail / no network
- [x] Debug override: a `?c=<token>` deeplink runs the resolution, boot and fallback paths without a live campaign
- [x] `Install Attribution` Segment event extended: `campaign_token`, `attribution_source` (`ga4f_deferred_deeplink` | `play_install_referrer` | `none`), `deferred_deep_link`
- [x] Sub-issue: campaign→scene mapping panel in mobile/BFF backoffice — #2669
- [ ] Ad platform: per-ad-group App URL set, deep links allowlisted, Deep Link Validator passed

**Client side in #2788.** The only remaining deliverable is ad platform configuration, which is not a code change.

## Why both attribution mechanisms

[...]

Issue #2669, opened 2026-08-08
Backoffice: Ad Campaign → Target Scene Mapping (BFF Remote Config)


rodrigo685:
##  **⚠️ Problem:**
Campaign→scene mapping lives in the client or the ad platform, so every new campaign needs an app release or an engineer. PM cannot launch or repoint an ad campaign's destination on their own.

## **🏁 Scope:**
New panel in the mobile/BFF backoffice: CRUD of `campaign token → target scene` pairs, served as remote config and read by the client. Content model is a token resolved server-side, not a destination embedded in the link.

## **📝 Deliverables:**
- [x] Backoffice panel: create / edit / delete token→scene pairs, no deploy required
- [x] Fields per entry: token, target (coords or world name)
- [x] Token uniqueness + target validation on save — reject unresolvable scenes
- [x] BFF endpoint serving the map
- [x] Client reads it and resolves the token before the FTUE, without racing it
- [x] Fetch failure → default FTUE, never a blank screen

**Shipped in:** decentraland/mobile-bff#81 (endpoint + schema), decentraland/mobile-hub#12 (panel). Client side in #2788.

## Cut from the original scope

Recorded here so the absences read as decisions rather than omissions.

- **Active window (start/end) and the enabled flag.** A campaign is live as soon as it exists; to stop one, delete it. The table is four columns: token, target type, target position, target world.
- **Audit trail.** Dropped with the window and the enabled flag — there is much less state left to attribute.
- **Cached last-known-good map.** Deliberate, not an oversight: which campaigns are live is decided server-side and the payload carries no end date, so a cached map is one that can resurrect a campaign the server already retired. The FTUE only runs on a first launch, where a cache is empty by definition.
- **"Expired or disabled entry resolves to default FTUE."** Moot without a window or an enabled flag. The map only ever carries campaigns that exist, and an unknown token falls back to the default FTUE.
- **"Cacheable, CDN-friendly."** The public GET has no cache headers. Consistent with `/feature-flags` today rather than a regression, but worth revisiting if a launch-day spike shows up in pg metrics.

## Changed from the original scope

- **The client fetches lazily, not on boot.** The map is only requested once a token actually needs resolving. Fetching on every boot would put every install on the endpoint — and while `urls::campaigns()` is pinned to the dev deployment (see below), that would be every production client hitting dev.

## Still open

[...]

PR #2770, opened 2026-08-25
feat(ftue): resolve an ad campaign token into a personalized FTUE or a direct scene boot

leanmendoza:
Client half of #2670. An install attributed to a campaign either boots straight into the scene the ad sold, or renders the FTUE with that campaign's content, instead of landing everyone on the same generic Discover carousel.

Depends on decentraland/mobile-bff#80 (serves `GET /campaigns`) and decentraland/mobile-hub#11 (the panel that fills it). Implements #2669's client side.

## Why the token is opaque

The ad link carries `?c=<token>` and the destination is resolved server-side. That is mechanically required, not just a flexibility preference:

`_should_go_to_explorer_from_deeplink()` (`lobby.gd`) keys on `location` / `realm` / `preview`, and fires `_async_redirect_by_deep_link()` from `async_close_sign_in()` and `_async_on_profile_changed()`. A deeplink carrying its own destination therefore boots the explorer and **skips avatar creation and naming** — the flow the FTUE hangs off, and the very screen this feature personalizes.

An opaque `c=` param touches none of those branches: it lands in `params`, and the router's `/open` handler treats a link with no navigation target as config-only. It is inert to everything that exists today.

## Shape of the change

- **`campaign_resolution.gd`** — pure payload parsing and target resolution. Free of `Global` and of node state so it can be exercised headless.
- **`campaigns.gd`** — the node that owns the IO. Same fire-and-forget fetch shape as `feature_flags.gd` (raced against a timeout, fail-open), plus a last-known-good cache and the token lifecycle.
- **Token capture** — once per install, from both deeplink paths (the `--fake-deeplink` path at boot and the live router), with a freshness bound. The Play install referrer survives 90 days and only changes on reinstall, so without one a reinstall months later would replay a dead campaign. Capture takes an explicit timestamp so the referrer path can pass the install time rather than the capture time.
- **`lobby.gd`** — a `bypass` campaign is routed through `_async_redirect_by_deep_link()`, which applies the pre-boot private-world gate from #2569. Booting the explorer locally instead would fork two behaviours that have to stay identical.
- **`ftue_screen.gd`** — applies the campaign copy and CTA and the curated carousel, and falls back to the campaign target when the carousel resolves to nothing. Today that button is simply dead when the carousel is empty.

World targets only resolve when the name matches `Realm.is_dcl_ens`; anything else would be treated as a realm URL by the teleport path and never reach the intended world, so it does not resolve at all.

## Fallbacks

Every failure path renders today's FTUE, which stays the default path rather than a degraded one: no token, unknown token, expired token, already-consumed token, resolver unreachable, unusable target, empty carousel.

Each one is named in the metrics context merged into the existing `DISCOVER_FTUE` screen-viewed and jump-in events, so personalized-vs-default is comparable without a new pipeline.

## Deliberately deferred

#2670 also asks to extend the Segment `Install Attribution` event with campaign / source / resolved. That belongs with the attribution sources (GA4F deferred deep linking and the Play install referrer), which are not in this PR — today nothing would populate those fields, since that event is only emitted on Android with a referrer. The campaign context rides the FTUE events instead, which is what the issue's Metrics section actually asks for.

## Testing

- `cargo run -- check-gdscript`: 382 scripts, 0 errors. `gdlint` and `cargo clippy -- -D warnings` clean.
- New headless test for the resolution logic: `res://src/test/campaigns/test_campaigns_resolution.tscn`.

It runs as a scene rather than with `--script` on purpose: a SceneTree script does not register the project's global classes, which is the same reason `Realm.new()` in `global.gd` fails under `--script`. The first version of this test passed vacuously for exactly that reason — worth knowing before adding tests next to it.

**Not device-verified.** The FTUE only appears on a fresh install after the full onboarding, and `GET /campaigns` is not deployed yet.

https://claude.ai/code/session_01QgGhLAcv2XdsBCXiZ91NjG


regenesis-claw (commented):
Review of the client half. Draft PR, so no formal verdict. Followed `REVIEW.md` (pre-flight, tiers, conventions) and cross-read the two companion PRs for contract agreement.

**Pre-flight — iOS build:** no `build` / `build-ios` label on this PR, so there is no iOS artifact. This change is mobile-visible (FTUE copy, CTA label, carousel contents, and a boot path that skips the FTUE entirely), so per REVIEW.md §0 the iOS build is required before merge, not optional. Android, Linux, Docker, Static checks and Clippy are all green. The description already states this is not device-verified — agreed, and that plus the missing iOS build is what keeps this in draft for me.

## Findings

**[P1] The disk cache can resurrect an expired campaign, contradicting the stated design** — `campaigns.gd`.

Both this PR and the BFF's OpenAPI say the window is evaluated server-side so "a stale cached map cannot resurrect an expired campaign". The code doesn't hold that line:

1. `_ready` seeds `_campaigns` from `campaigns_cache` on disk.
2. `_async_load` sets `_loaded = true` on the failure branch too (after `push_warning`), leaving the cached map in place.
3. `async_resolve_pending`'s only cache guard is `not _loaded and _campaigns.is_empty()` — with `_loaded` true and a non-empty cache, `resolve(token)` answers straight out of the stale map.

So an install that captured a token, then completes onboarding offline (or while the BFF is down) up to `TOKEN_MAX_AGE_SECONDS` (7 days) later, can boot into a campaign the server has since disabled or whose window closed. For `bypass` that means booting into a scene that is no longer supposed to be the destination.

The cache entry has no timestamp and no TTL, and the campaign bodies carry no `endsAt` (the public payload deliberately omits it), so the client cannot re-check the window locally even if it wanted to. Cheapest fixes, in order of preference: store a `campaigns_cache_at` alongside the JSON and ignore the cache past a short TTL; or only consult the cache while a fetch is still in flight and treat a failed fetch as `resolver_unavailable`; or include `endsAt` in the public payload and filter locally.

**[P2] An unusable bypass target is reported as `unknown_token`** — `lobby.gd`, `_async_start_ftue`. The description says every fallback path is named in the metrics context, and lists "unusable target" as one of them. It isn't — the bypass fallback reuses `FALLBACK_UNKNOWN_TOKEN`, so "marketing saved a target the client can't route" is indistinguishable from "the token isn't in the map" in the funnel. Those two need very different responses. Add `FALLBACK_UNUSABLE_TARGET`.

**[P2] `_capture_campaign_token`'s `occurred_at` parameter has no caller** — `global.gd`. Both call sites (the boot fake-deeplink path and the router) use the default. The install-referrer wiring the parameter exists for isn't in this diff — which is fine and consistent with the "deliberately deferred" section, except that the `config_data.gd` comment states it in the present tense ("the install-referrer path passes the install timestamp instead"), which reads as shipped. Either wire it or soften the comment to say it's the reason the parameter exists.

**[P2] The campaign is burned before it is delivered** — `lobby.gd`, `_async_start_ftue` calls `mark_consumed()` immediately after resolving, before `show_discover_ftue_screen`. If the app is killed, crashes, or is backgrounded and evicted while the FTUE is on screen, `campaign_consumed` is already persisted and the next launch renders the default FTUE — the personalization is silently lost for an install the ad was paid for, with no metric distinguishing it (it reports `already_consumed`). For `bypass` the current placement is right (the redirect is the delivery). For `ftue` mode, marking consumed on the terminal action — jump-in or skip — matches the intent better.

**[P2] Curated carousel fetches are serial** — the featured data provider's `async_fetch_places_by_ids` awaits each id in turn, up to `PLACE_IDS_MAX` (10, per the BFF). On a mobile connection that's 10 sequential round trips before the campaign carousel appears, on the one screen where the user is deciding whether to stay. `PromiseUtils` already gives you the primitives to issue them together and reorder by the input list afterwards.

**[P2] Post-`await` resume guard doesn't follow the current convention** — `ftue_screen.gd`, `_async_load_campaign_places` uses `is_instance_valid(self)`. Since 1.12.2 / #2714 the convention at a post-`await` resume point is `NodeGuard.is_alive(self, ...)`, which adds the capped telemetry that tells you whether the race ever fires in the wild. `carousel` also isn't re-checked before `carousel.fetch()` / `carousel.set_items(places)` on the resumed path.

**[P2] The cache isn't scoped per environment** — `campaigns_cache` is a single config key, while `DclUrls.campaigns()` resolves per `ServiceGroup::MobileBff` (org vs zone). A build switched between environments with `--dclenv` reads the other environment's map until the next successful fetch. Low impact in production, confusing when testing this feature — which is exactly what people will be doing with it.

## Verified (claims I checked rather than took on faith)

- **`Vector2i.MAX` really is the "no location" sentinel.** `DclParseDeepLink::is_location_defined()` is `location.x < 1000000`, so the world branch's `location = Vector2i.MAX` correctly reads as undefined, and `_async_redirect_by_deep_link`'s private-world gate (which requires `not is_location_defined()`) does apply to a world bypass. The claim that routing through the shared path preserves #2569 holds.
- **The genesis branch routes to the right realm.** It sets `realm = ""` with a defined location, and `explorer.gd`'s param resolution fills `realm_string` with `DclUrls.main_realm()` in exactly that case. Matches what `target_position_and_realm` returns for a genesis target.
- **The token is genuinely inert to existing deeplink routing.** `_should_go_to_explorer_from_deeplink()` keys only on location / realm / preview; `c=` lands in `params` and trips none of them, so a `?c=` link cannot skip avatar creation. The premise of the whole design checks out.
- **Capture ordering is right.** `_capture_campaign_token` runs in the router before any branch that consumes the deeplink, and first-capture-wins is enforced against the persisted value, so an in-session deeplink can't repaint an install's campaign.
- **Payload contract matches the BFF.** `parse_response` and `target_position_and_realm` agree field-for-field with what mobile-bff#80 serves (`ok` / `data.campaigns` / `mode` / `target.{type,position,name}` / `title` / `cta` / `placeIds`), and the world-name rule is the same regex on both sides plus `Realm.is_dcl_ens` here.
- **The new test genuinely runs.** Scene-based rather than `--script` for the stated reason, and it covers the fail-open shapes, the unroutable world names and the metrics context. Note it isn't wired into any workflow, so like the other `.gd` suites in this repo it is manual-only — no CI signal on regressions here.

## Security

No security issues found. The campaign map is authored behind the BFF's signed-fetch + `ALLOWED_USERS` gate, the target is format-validated on both sides, and the client refuses any world name it cannot route. The `?c=` token is attacker-supplyable, but the worst it buys is selecting an already-public active campaign for your own install, once. Nothing is logged that isn't already public.

## Test plan

Per REVIEW.md §4, the description's Testing section is CI + a headless unit test — no QA-executable steps. Since the feature only exists on a fresh install past full onboarding, that section is the one thing QA will need most. Two cases worth writing when the BFF is deployed: a `bypass` campaign booting straight into a world (with the private-world variant, to prove the #2569 gate still fires), and an `ftue` campaign with a curated carousel where the resolver is unreachable (airplane mode after capture), expecting today's default FTUE and a `resolver_unavailable` fallback reason.

_Reviewed via Slack request from Lean._

regenesis-claw (commented):
Re-review of `8cf9a5f4..1cbe9faa` (one commit). Every finding is addressed, and the P1 was resolved by deleting the cache rather than bolting a TTL onto it — the stronger of the two options I offered. Still a draft, so no formal verdict.

| # | Finding | Status |
|---|---------|--------|
| 1 | **P1** stale disk cache can resurrect a retired campaign | **Fixed** — `campaigns_cache` is gone entirely (config field, load, save, `_parse_cache`, `_store_cache`), and a new `_fetch_ok` flag separates "the attempt finished" from "the map is real". A failed fetch now reports `resolver_unavailable` instead of answering from disk. |
| 2 | Unusable bypass target reported as `unknown_token` | **Fixed** — `FALLBACK_UNUSABLE_TARGET` added and used at the bypass fallback. |
| 3 | `occurred_at` has no caller while the comment reads as shipped | **Fixed** — both the `config_data.gd` comment and the new `global.gd` doc block now say the deeplink path is the only one wired and that the parameter is the seam for the referrer path. |
| 4 | Campaign consumed before the FTUE renders | **Fixed** — `mark_consumed()` moved after `show_discover_ftue_screen()` for the FTUE path, and into `_try_boot_into_campaign_target` right at the commit point for bypass. |
| 5 | Curated carousel fetched serially | **Fixed** — `async_fetch_places_by_ids` now issues all fetches up front and awaits `PromiseUtils.async_all`. |
| 6 | `is_instance_valid` instead of `NodeGuard.is_alive` | **Fixed** — `NodeGuard.is_alive(self, "FtueScreen._async_load_campaign_places")`. |
| 7 | Cache not scoped per `dclenv` | **Moot** — no cache exists any more. |

## What I checked on the fix itself

- **`_fetch_ok` is the right split.** The old guard (`not _loaded and _campaigns.is_empty()`) could never fire once `_loaded` was set on the failure branch, which is exactly how the stale map got through. The new one keys on the fetch actually returning, so offline resolves to `resolver_unavailable` and the funnel can tell that apart from a token the server didn't know.
- **The concurrent-fetch rewrite is sound.** `PromiseUtils.async_all` accepts raw promises (`_Internal.async_call_and_get_promise` returns `f` unchanged when it is a `Promise`), and `AllAwaiter` writes each result into `results[i]` by index — so "async_all keeps the input order" in your comment is accurate, and the curated carousel order is preserved. `assert(size < 64)` in `AllAwaiter` is well clear of the BFF's `PLACE_IDS_MAX` of 10. Resolving with `null` rather than rejecting keeps the half-broken-carousel behaviour intact, since `async_all` has no partial-failure path.
- **Consumption timing is now correct on both branches**, and marking bypass inside `_try_boot_into_campaign_target` after the `CAMPAIGN_BYPASS` metric — rather than in the caller — keeps it tied to the commit point instead of to a code path that might grow another early return later.
- No leftovers: nothing in the head diff still references `campaigns_cache`. An existing settings file with the old key just carries a now-unread entry, which is harmless.

## Two small residuals (neither worth blocking)

1. **An unusable target still burns the campaign.** In `_async_start_ftue`, the local `campaign` is still the original non-empty dict when the bypass fallback runs (only `resolution["campaign"]` is cleared), so `if not campaign.is_empty(): mark_consumed()` fires. If marketing then fixes a mistyped target, that install never gets the campaign — it reports `already_consumed` forever. Arguably right (the install did burn a launch on it), but if the intent is "a server-side mistake shouldn't cost the install", this is the one path where not consuming would be defensible.
2. **A 200 with a malformed body classifies as `unknown_token`.** `_fetch_ok` is set before `parse_response` can reject the shape, so a BFF that answers with a bad envelope looks like a token miss rather than a resolver problem. One line (`_fetch_ok = not _campaigns.is_empty()` is *not* it, since an empty map is legitimate — better to have `parse_response` distinguish "bad shape" from "no campaigns"). Only matters if you ever have to debug this from the funnel.

## Still open from the first pass

Not findings, just the two gates: no `build` / `build-ios` label, so there is still no iOS artifact for a mobile-visible change (REVIEW.md §0), and the feature remains device-unverified with `GET /campaigns` undeployed. Static checks and Clippy are green on this head; Linux and Android were still running when I looked.

leanmendoza:
Superseded by #2788 — same feature, rebuilt on `main` as two commits so the diff is the feature rather than five rounds of me trimming its scope.

What changed since this branch: a campaign is now always a destination (no `mode`, no FTUE content overrides), and the map is just token → target (no window, no enabled flag, no audit trail) — the companion backend PRs landed that way and are already merged. #2788 also adds the half this branch never had: where the token comes from on an ad-driven install, which is the actual point of #2670.

Closing this rather than pushing a sixth revision onto it.

Issue #2732, opened 2026-08-21
[Bug] Players constantly bouncing off floor after landing from jumps


regenesis-claw:
### Platform
Not specified

### App Version
Not specified

### Device Information
Not specified

### Issue Description
Players are constantly bouncing off the floor after landing from jumps.

### Expected Behavior
Player should land smoothly on the floor without bouncing after a jump.

### Screenshots / Media
Not provided — none shared in the thread.

### Steps to Reproduce
1. Move around in the mobile client
2. Jump
3. Observe the player bouncing off the floor upon landing

### Occurrence
Not specified

### Related Issues
- #905 — Colliders Shape Consistency Review v Unity (open, related collider consistency work)
- #1553 — Jump Animation Review (open)

Requested by Rodri via Slack

manuelmaceira:
Closed by #2778

PR #2778, opened 2026-08-26
fix(avatars): smooth remote-avatar interpolation + landing-bounce fixes (#2732, #2734)

### LINKED ISSUE #2734: [Bug] Nearby players animate as discrete frames instead of continuously (v1605)
### Platform
iOS

### App Version
v1605 (internal build)

### Device Information
iPhone 15 Pro, iOS (latest)

### Issue Description
Nearby players are rendered choppily — instead of continuously/smoothly animating, they appear to snap between discrete frames/poses, as if only sparse position/animation updates are being applied rather than interpolated motion.

### Expected Behavior
Nearby players should move and animate smoothly/continuously, matching normal avatar animation playback.

### Screenshots / Media
A screen recording was shared in the originating Slack thread showing the issue (not attached here — GitHub CLI can't attach media; please upload manually if needed).

https://github.com/user-attachments/assets/3c206cdb-1352-4b1e-919e-f09246dff17a

### Steps to Reproduce
1. Join a world/scene with other nearby players present.
2. Observe other players' avatars while they move.
3. Movement/animation appears frame-by-frame instead of continuous.

### Occurrence
Not specified

---
Requested by Rodri via Slack

manuelmaceira:
## What

Fixes #2734 (nearby players render as discrete frames) and #2732 (players bounce off the floor after landing).

Remote avatars were interpolated with a zero-buffer lerp that assumed a fixed 10 Hz packet cadence, while Pulse delivers at 20/10/5 Hz by distance tier with 30–190 ms arrival jitter (measured on-device). The landing "bounce" was the landing animation replaying on 1-tick state flicker.

## Changes

- **Buffered interpolation** (`dcl_avatar.rs`): 8-packet ring buffer + render clock ~2 packet intervals in the past (slew-limited, no extrapolation). Position, rotation, and anim state all render on this one clock — the AnimationTree follows the visible model, not the network truth.
- **Locomotion/air state from wire data**: walk/jog/run and rise/fall are driven by the sender's actual velocity + `is_grounded` (carried in Movement packets) instead of local `dy/interval` estimates that spiked on wire quantization and jitter. Local estimate remains only for the legacy Position-packet path.
- **Broadcast at ~18 Hz** (was 10 Hz): 10 Hz input quantized onto Pulse's 50 ms tick grid was the main jitter source; 18 Hz stays under the server's 20 Hz input-rate disconnect cap.
- **Landing bounce, local** (`player.gd`): `is_grounded` gets a 150 ms coyote window (suppressed during jump + jump cooldown, so tap-jumps unground instantly). Also fixed the dead hard-landing check (`time_falling` was read after being reset).
- **Scene CRDT** (`avatar_scene.rs`): remote avatars' *rendered* transform is pushed to scene CRDT every frame — scene scripts polling avatar positions (e.g. flagtag's flag) no longer see per-packet jumps.

## Tests

22/22: unit tests + Godot itests (jittered-stream smoothness using the real captured gap pattern, anim/motion coherence, wire air state, idle floor) + GDScript grounded-resolver regression test. Pre-commit suite passes.

## QA steps

1. Join the same world with two clients (mobile or `--emulate-android`), stand near each other.
2. Walk/jog/run around the other client → motion should be continuous, animation should start when the body starts moving, no bouncing on landing.
3. Stop suddenly → no walking-in-place (moonwalk).
4. Rotate in place → no walking anim while stationary.
5. Walk slowly (slight analog tilt) → walk anim still plays.
6. Move 30+ m away → remote avatar stays smooth at distance tiers.
7. On flagtag: grab the flag → it follows the holder smoothly.


godot/src/logic/player/player.gd
```diff
@@ -536,7 +544,14 @@ func _physics_process(dt: float) -> void:
...
 	avatar.jump_count = jump_count
 	avatar.glide_state = glide_state
-	avatar.is_grounded = on_floor
+	# Debounced ungrounding (GROUNDED_GRACE_WINDOW): a 1-tick is_on_floor()
+	# flicker must not flip the AnimationTree out of grounded state. Suppressed
+	# while jump is held so real takeoffs unground immediately. The jump-pad
+	# override below (combined_vy > 0.3) runs after this and still wins.
+	avatar.is_grounded = (
+		on_floor
+		or (time_falling < GROUNDED_GRACE_WINDOW and not Input.is_action_pressed("ia_jump"))
```
regenesis-claw: [P1] This debounce only suppresses on `not Input.is_action_pressed("ia_jump")`. The existing `in_grace_time` check above (used for `avatar.land`) additionally requires `_time_since_last_jump >= JUMP_COOLDOWN` — specifically so a fresh jump isn't treated as "just landed". A tap-jump (button released before the next physics tick reads it, which is common at the 18-30 Hz mobile rates this PR targets) will satisfy `not is_action_pressed("ia_jump")` immediately, so `is_grounded` stays true for up to `GROUNDED_GRACE_WINDOW` (150 ms) into a real jump even though `rise=true`/`land=false` this same tick. Since `is_grounded` drives `nfall` and the walk/jog/run locomotion conditions in `avatar.gd`, this can misdrive the jump-rise animation on ordinary tap-jumps. Suggest adding the same `_time_since_last_jump >= JUMP_COOLDOWN` condition here.

regenesis-claw (changes requested):
Solid root-cause work on both bugs and good test coverage (jittered-stream sim is a nice touch) — but two things need fixing before merge.

**[P1] `is_grounded` debounce can delay the jump-rise transition on tap-jumps** — `godot/src/logic/player/player.gd` (see inline comment). The new grace window only checks `not Input.is_action_pressed("ia_jump")`, but the pre-existing analogous `in_grace_time` check a few lines above also requires `_time_since_last_jump >= JUMP_COOLDOWN` specifically to avoid this exact false-positive during a fresh jump. Any tap-jump (press+release within one physics tick, common at 18-30 Hz mobile rates — the exact regime this PR targets) will read `is_action_pressed("ia_jump") == false` on the very next tick, so `is_grounded` stays true for up to 150 ms into a real jump even though `rise=true`/`land=false`. Since `is_grounded` drives `nfall` and the run/jog/walk locomotion conditions directly (avatar.gd `_process`), this looks like it can reintroduce a jump-animation glitch for the most common jump input pattern.

**[P1] `PLAN-2732-2734.md` is a committed local planning doc — Spanish text + a hardcoded personal path.** It opens with `Worktree: /Users/lordmanuel/Projects/decentraland/godot-explorer/fix-issues-2732-2734` and is written entirely in Spanish. Per REVIEW.md Tier 1 #5 (hardcoded local paths) and #6 (English-only, no exceptions for comments/docs), this blocks approval — please drop the file from the PR (or move it out of the repo into an issue/gist if you want to keep it for reference).

**[P2]** `godot/project.godot` drops `boot_splash/show_image=true` — this looks unrelated to the fix. It's a no-op (the engine default is already `true`), so it's harmless either way, but worth confirming it wasn't an accidental local-editor diff.

**[P2]** `push_interpolated_transforms_to_scenes` now writes into every active scene's CRDT dirty set every frame for every `LerpTwoPoints` avatar in view (mitigated by the 0.001 m/rad unchanged-skip, so likely fine) — just flagging since scene-runner per-frame writes are a hot path the team watches closely; worth a sanity check in a busy area (Genesis Plaza) if not already covered by the on-device validation.

**iOS build:** no `build`/`build-ios` label on this PR, so there's no CI iOS artifact — the PR body reports manual on-device iOS (iPhone 16e) validation, which I can't verify from here (author-reported, unverified). The changed paths (`lib/src/avatars/`, `dcl_avatar.rs`, `player.gd`) aren't on REVIEW.md's explicit iOS-required list (native plugins/comms/av), so I'm not holding on this, but since the PR is explicitly fixing an iOS-observed bug (#2734 cites iOS v1605), a maintainer adding the `build` label before merge would be good confirmation.

CI is green (Linux, Docker, static checks, Clippy, Android). Branch is up to date with main (0 behind). No security issues found — this is client-side rendering/physics code with no new external input surface.

Requesting changes for the two P1s above; everything else is a nit/FYI.

manuelmaceira:
Thanks for the thorough review — both P1s addressed in `d908f79f9d`:

**[P1] is_grounded debounce vs tap-jumps** — valid catch, and exactly the regime this PR targets. Added the same `_time_since_last_jump >= JUMP_COOLDOWN` guard that `in_grace_time` uses a few lines above. Timing check: jump cooldown (300 ms) now fully covers the grace window (150 ms), so a fresh jump can never read as grounded, while ledge walk-offs (cooldown long past) still get the debounce.

**[P1] PLAN-2732-2734.md** — dropped from the repo (kept locally). The PR diff no longer contains it.

**[P2] project.godot** — confirmed accidental local-editor diff; `boot_splash/show_image=true` restored.

**[P2] scene CRDT per-frame writes** — acknowledged it's a hot path. The unchanged-skip (0.001 m / 0.001 rad) means stationary avatars cost one `get_global_transform` + LWW read per frame and zero dirty-set writes; only avatars actually moving on screen write, ~1 HashMap insert per active scene per frame. Happy to run a Genesis Plaza busy-area sanity pass on-device if a maintainer wants the extra confirmation before merge.

**iOS build label** — agreed; manual validation was iPhone 16e + Android, two-client Pulse sessions. A maintainer-triggered `build` label run would be good independent confirmation.

regenesis-claw (approved):
Delta re-review at `d908f79f` (one appended commit since `f8967b22`). All three findings resolved — approving.

| # | Finding | Status |
|---|---------|--------|
| P1 | `is_grounded` grace window missing the `JUMP_COOLDOWN` guard | ✅ Fixed |
| P1 | `PLAN-2732-2734.md` committed (Spanish + personal path) | ✅ Removed |
| P2 | `boot_splash/show_image=true` dropped from `project.godot` | ✅ Restored |
| P2 | Per-frame CRDT transform push — hot-path sanity check | ℹ️ Advisory, no change expected |

**Verification notes**

- `player.gd:553-561` now carries the same three-clause guard as the pre-existing `in_grace_time` check (`time_falling < window` + `not is_action_pressed("ia_jump")` + `_time_since_last_jump >= JUMP_COOLDOWN`) — the two are now consistent, which was the point. `_time_since_last_jump` is reset to `0.0` on both the ground-jump (L476) and air-jump (L395) paths and initialised to `1000.0`, so a cold start or a walk-off-ledge isn't accidentally suppressed.
- Confirmed the guard doesn't defeat the original fix: a normal jump's airtime exceeds `JUMP_COOLDOWN` (0.3 s), so by touchdown the cooldown has elapsed and the 150 ms debounce is active for the landing flicker it was written for.
- Walk-off-ledge behaviour is unchanged in shape and lines up with the existing 0.2 s `in_grace_time` land window, so no new divergence between `avatar.land` and `avatar.is_grounded`.
- Diff no longer contains `PLAN-2732-2734.md` or `godot/project.godot` — file list is now the 3 code files + 2 repro scripts (+ `.uid`s).

**[P2, non-blocking]** For a very short hop (airtime < `JUMP_COOLDOWN`), the debounce is still suppressed at touchdown — `on_floor` is true in that frame so `is_grounded` is correct anyway, and it only matters if `is_on_floor()` flickers inside that sub-300 ms window. Worth knowing if #2732 ever resurfaces specifically on rapid repeated hops; not worth code today.

**CI:** `📊 Static checks` and `🔧 Clippy` green on this head; `🐧 Linux` and `🤖 Android` still running at review time. The delta is GDScript + a file deletion + a `project.godot` line — no Rust changed since `f8967b22`, which was fully green. An iOS export job is now running too, which covers the iOS-build gap I flagged last round. Worth a glance at the Linux/Android results before merge.

Branch is up to date with `main` (0 behind). Nothing further from me.

manuelmaceira:
Added regression coverage for both bugs in `b0609699fb`:

**#2732 (bounce)** — `godot/src/test/player/grounded_grace_test.gd` (first GDScript test in the runner): the grounded resolver was extracted to a pure `resolve_is_grounded` and the test covers 1-tick flicker (debounced), sustained fall (ungrounds), jump-held and tap-jump (unground immediately), and ledge walk-off (grace applies). Writing it caught a real bug: my first version shadowed the `time_falling` member, which is a warning-as-error and would have broken `player.gd` compilation.

**#2734 (ping-pong/stutter)** — `itest_remote_avatar_smooth_under_pulse_jitter`: a real `DclAvatar` in the scene tree fed the captured Pulse gap pattern (30–190 ms) at 30 fps, asserting per-frame render step stays near the continuous ideal. The test-driven iteration hardened the interpolator beyond the original PR: no extrapolation past the newest packet (dead-reckoning snap-back = ping-pong), and the render delay is now slew-limited toward 2× a slow interval EMA — both the fast EMA and an adaptive max-gap delay visibly lurched the render clock on the captured pattern. Final sim result: max per-frame step 0.068 m vs 0.067 ideal.

Full suite: 18/18 (`cargo run -- run --itest` + cargo unit tests).

manuelmaceira:
Final coherence pass (`d89d6f32d3`) after on-device verification rounds (2 clients, Pulse, flagtag):

**Playtest-driven changes:**
- **Broadcast 18 Hz** (was 10 Hz): the 10 Hz input quantized onto Pulse's 50 ms tick grid was the source of the captured 30–190 ms arrival jitter. 18 Hz stays under the server's 20 Hz `INPUT_RATE_EXCEEDED` disconnect cap with margin, and halves the render delay peers need for this client.
- **Anim state rides the packet ring** — walk/jog/run/rise/fall/land/is_grounded apply from the segment being *rendered*, not at packet arrival. Fixes 'rotates, walks in place, then translates': the AnimationTree now follows the visible model (bevy-explorer uses the same trick).
- **Locomotion from wire velocity** — genuine slow walks animate, short fast bursts classify correctly; the 0.5 m/s idle floor only filters sway (quantization/broadcast-threshold micro-movements no longer play walk).
- **First packet**: skips both classification (zeroed-default spike) and the interval EMA seed (spawn-time pinned the delay at 0.5 s for ~15 s).

**Independent deep review** of the full diff ran clean on the core logic (segment search, anim choice, wire ordering, snap resets, decay interplay); the two notes it raised (compressed-path grounded gate pollution, first-packet EMA seed) are fixed in this commit.

Suite: 22/22 (4 unit + 5 itest + GDScript grounded test + pre-existing). On-device verified: smooth motion near/far, no moonwalk, no bounce, coherent anim start, natural rotation.

lib/src/avatars/avatar_scene.rs
```diff
@@ -1767,7 +1812,24 @@ impl AvatarScene {
...
+            .get(&entity_id)
+            .map(|avatar| avatar.bind().get_is_grounded())
+            .unwrap_or(true);
+
         self._update_avatar_transform(&entity_id, dcl_transform, false);
+        // Compressed packets carry velocity but no grounded/jump state:
+        // drive air state AND locomotion from it (see apply_wire_movement_state).
+        if let Some(avatar) = self.avatar_godot_scene.get_mut(&entity_id) {
+            let mut avatar = avatar.bind_mut();
+            avatar.apply_wire_air_state(grounded_gate, velocity.y);
```
regenesis-claw: [P1] This fixes `rise`/`fall`/`land` for compressed packets, but not `is_grounded`.

`apply_wire_air_state` never assigns `is_grounded`, and the only thing that wrote it earlier in this call is `_update_avatar_transform` → `set_target_position`, which sets `is_grounded = land && glide_state == 0` from the **local `dy / interval` estimate** — the quantization-and-jitter-spiky signal this commit is specifically replacing.

So after this block the avatar can sit in an inconsistent state: `land == true` (wire-derived, correct) while `is_grounded == false` (dy-derived, spiked). `avatar.gd`'s `nfall` condition reads `is_grounded` directly, so that flicker still replays `Jump_Fall → Jump_End` — the remote landing bounce, on the one path that has no wire grounded flag to fall back on.

Since the compressed wire carries no grounded state, deriving it from the just-computed air state keeps the flags coherent:

```suggestion
            avatar.apply_wire_air_state(grounded_gate, velocity.y);
            avatar.set_grounded_from_air_state();
```

(with a small helper doing `self.is_grounded = self.land && self.glide_state == 0`). Please don't put that assignment inside `apply_wire_air_state` itself — the uncompressed path calls it *after* setting `is_grounded` from the wire, and would get its authoritative value clobbered.

`itest_wire_air_state_no_landing_replay` only drives `apply_wire_movement_state`, so this path has no coverage — worth an itest that goes through `update_avatar_transform_with_rfc4_position`.

regenesis-claw (changes requested):
## Re-review at `7056949` (round 3)

Scope grew substantially since the approval at `d908f79f` — 5 new commits that go well past the original two issues: wire-authoritative air state + locomotion, anim state riding the packet ring, and a **broadcast rate change from 10 Hz to 18 Hz**. Fresh full read rather than a delta.

The core direction is right and the new test coverage is a genuine improvement. One P1 blocks.

---

### [P1] `MovementCompressed` path: `is_grounded` is still driven by the noisy local dy estimate

Inline comment on the call site. Summary: the commit message says this commit fixes the remote landing replay for *both* wire paths, but on the compressed path only `rise`/`fall`/`land` become wire-driven. `apply_wire_air_state` never writes `is_grounded`, and nothing else does after `set_target_position` — so `is_grounded` keeps the value derived from `dy / interval`, which is exactly the quantization-and-jitter-spiky estimate this commit set out to remove.

That matters because `avatar.gd`'s `nfall` condition reads `is_grounded` directly. The resulting state can be internally inconsistent — `land == true` while `is_grounded == false` — which replays `Jump_Fall → Jump_End`: the same visible bounce, on the path that has no wire grounded flag to protect it.

`itest_wire_air_state_no_landing_replay` only exercises `apply_wire_movement_state` (the uncompressed path), so the compressed path is uncovered.

---

### [P2] The dy-derived speed still feeds the EMA on the uncompressed path

The commit message states: *"Local dy/interval estimate remains only for the legacy Position-packet path."* That isn't what the code does.

On every `Movement` packet the EMA is updated **twice**:

1. `set_target_position` computes `instant_speed = target_forward_distance / interval` and blends it in (`0.6·s + 0.4·instant`), then calls `classify_locomotion`.
2. `apply_wire_locomotion` then blends the wire speed into the *result* of step 1.

Net weighting is `0.36·s_prev + 0.24·dy_speed + 0.4·wire_speed` — so the estimate you diagnosed as the root cause still carries ~24% of the signal. Wire velocity dominates, so this is unlikely to be visible, but the stated invariant is not upheld. Suggest gating the dy classification/EMA behind a flag that only the Position-packet path sets, so the wire paths use wire velocity exclusively.

---

### [P2] Per-frame scene push re-acquires the scene runner once per avatar

`push_interpolated_transforms_to_scenes` runs every frame and calls `push_avatar_transform_to_scenes` per changed avatar; that helper does `DclGlobal::singleton().bind().scene_runner.clone()` followed by `bind_mut()` on each call. Against the 40-avatar mesh pool that's 40 singleton lookups and 40 scene-runner borrows per frame in a crowded scene. Hoisting the `scene_runner` bind out of the loop and passing it in would make this a single borrow per frame.

(The `0.001` unchanged-skip helps but only after the borrow.)

---

### [P2] The new tests don't run in CI

The four `#[godot::test::itest]` functions and `grounded_grace_test.gd` are only reachable through `test_runner.gd`, which no workflow invokes. The Linux job runs `cargo test`, `check-gdscript`, and `cargo run -- test-avatar` — and `test-avatar` runs `test_avatar_state_machine_graph.gd`, not the runner. So of everything added here, only the plain `#[test]` functions execute in CI; "22/22" is a local-only result.

Wiring an xtask step for `test_runner.gd` would be a valuable follow-up (same gap flagged on #2717) — otherwise these regression tests can rot silently.

---

### [P2] `LerpState.clock` is an unbounded `f32`

`clock` accumulates for the avatar's entire lifetime and `arrival` timestamps are stored against it. `f32` ulp reaches ~1 ms at ~2.3 h of uptime and ~8 ms at ~18 h — against a 55 ms segment that's up to ~15% quantization in the interpolation `t`, i.e. the micro-stutter this PR exists to remove, reappearing on long-lived sessions. Cheap fixes: make `clock`/`arrival` `f64`, or rebase the buffer by subtracting the oldest arrival when it exceeds some bound.

---

### [P2] Remote rotation now hard-zeroes pitch and roll every frame

`process` sets `set_global_rotation(Vector3::new(0.0, new_yaw, 0.0))`, replacing the previous full-euler-from-wire-basis assignment. Yaw-only matches the `rotation_y` wire convention so this is very likely correct, but it now actively overwrites any X/Z rotation each frame rather than merely not setting it. Worth confirming nothing else tilts the avatar node (glider pitch, emote props).

Note the rotation approach churned across three commits (buffered → exponential chase → snap → back on the segment clock); the final state is rotation on the segment clock. The commit titled "snap remote rotation instantly, drop rotation interpolation" no longer describes the code.

---

### Verified, not flagged

- **18 Hz broadcast is within the server limit.** Checked `decentraland/Pulse` directly: `MovementInputRateLimiter` defaults are `MaxHz = 20` with `BurstCapacity = 16`, and a violation is an *instant disconnect* (`INPUT_RATE_EXCEEDED`). The 55 ms timer is 18.18 Hz sustained, so it passes — but the margin is ~9%, and the failure mode is a hard disconnect rather than degradation. Given that this also raises upstream traffic ~82% on the LiveKit path (where the Pulse tick-grid rationale doesn't apply), it's worth an explicit sign-off from a comms owner before merge; it's a wire-rate change riding along with two rendering fixes.
- **The removed parcel-offset block really was dead code.** Confirmed against `main`: `avatar_scene_transform` was computed and the `insert` used `dcl_transform.clone()` regardless. Removal is safe and does not change scene-facing coordinates.
- **`POSITION_THRESHOLD` rescale is consistent** — `0.0275 / 0.055 s` preserves the previous `0.5 m/s` detection floor, and `IDLE_KEEPALIVE_TICKS = 18` preserves the ~1 s keepalive.
- **No extrapolation past the newest packet**, and the first-packet EMA/classification skips are correctly handled.
- **Security**: no findings. Client-side rendering and physics; the one new input is a velocity vector already present in the packet, consumed only for animation classification with no unchecked indexing or trust decision attached.

---

### CI / pre-flight

- **Linux is currently red.** The failure is the known headless teardown segfault — `[test_avatar_state_machine_graph] PASS` prints, then exit `signal 11` at `string_name.cpp:264` during shutdown. That's a pre-existing flake, not this diff, but the job needs a re-run to get an actual green signal before merge.
- Static checks and Clippy pass; `cargo test` reports 262 passed. Android and the iOS export are still pending.
- **iOS**: this PR touches `message_processor.rs` under the comms tree, which REVIEW.md lists as iOS-build-required. An iOS export job is running — please confirm it lands green before merging.
- Branch is `MERGEABLE` and up to date with `main`.

Followed `REVIEW.md`.

manuelmaceira:
Thanks for the careful re-read — verdict per point, all addressed in `87e717fe21`:

**[P1] compressed-path `is_grounded`** — You're right, and the suggested shape is the correct one. Implemented exactly as `set_grounded_from_air_state()` called after `apply_wire_air_state` (not inside it, so the uncompressed path's wire value isn't clobbered). I had originally deferred this as "legacy path", but our own client does send compressed on LiveKit when `jump_count < 2 && glide_state == 0`, so the path is live. Coverage added: `itest_compressed_path_grounded_coherent` replicates the production sequence (first packet → pre-packet hysteresis gate → spiky dy landing tick → wire air state → grounded from air state) and asserts `land`/`is_grounded` coherence.

**[P2] double EMA blend** — Correct catch; the stated invariant wasn't upheld. New `wire_classification` flag: once a peer streams wire velocity (Movement or compressed), `set_target_position` skips the dy classification entirely. Position-packet path keeps the local estimate as the only available signal.

**[P2] scene-runner borrow per avatar** — Done: `push_interpolated_transforms_to_scenes` binds once per frame and passes the guard into `push_avatar_transform_with_runner`; the packet path keeps its per-call bind via the wrapper.

**[P2] tests not in CI** — Fair, and the "22/22" claim was local-only. I'm not wiring `test_runner.gd` into CI in this PR: the headless teardown segfault (same one flaking the Linux job) would make a new CI step intermittently red from day one. That needs the flake addressed first, and deserves its own change.

**[P2] f32 clock** — Good catch, this one I missed entirely. `clock`/`arrival` are now `f64` (delay/interval EMAs stay f32 — they're small magnitudes).

**[P2] rotation zeroing pitch/roll** — Confirmed safe: nothing else writes the avatar node's rotation (checked `avatar.gd` + `avatar_scene.rs`; lean/IK are bone-level via separate head_yaw/head_pitch fields, and the Position path's quaternion is body yaw in practice). Also acknowledged on the commit-title churn — final state is rotation on the segment clock; the intermediate "snap instantly" title no longer describes the code.

**18 Hz broadcast** — Noted and agreed it deserves comms-owner visibility: it's a wire-rate change (also raises LiveKit movement traffic ~80%, where the tick-grid rationale doesn't apply). Verified within the server limiter (`MaxHz = 20`, `BurstCapacity = 16`) at 18.18 Hz sustained. Flagging explicitly here so a comms owner can veto before merge if the margin is uncomfortable.

Suite: 23/23 (added the compressed-path itest). Will re-run the flaky Linux job once the current run settles.

lib/src/avatars/avatar_scene.rs
```diff
@@ -1767,7 +1827,25 @@ impl AvatarScene {
...
+            avatar.set_grounded_from_air_state();
```
regenesis-claw: [P1] This closes a self-referential loop and pins remote peers to `is_grounded == true` forever.

`grounded_gate` (snapshotted a few lines up) is the previous `is_grounded`. `apply_wire_air_state` computes `rise`/`fall` as `!grounded_gate && ...`, then `land = !rise && !fall`, and this new call sets `is_grounded = land && glide_state == 0`. So `is_grounded_{n-1} == true` ⇒ `rise = fall = false` ⇒ `land = true` ⇒ `is_grounded_n == true`. The first packet takes the `first` branch in `set_target_position`, which sets `is_grounded = true`, and `glide_state` is never written from the wire on this path — so it starts true and latches.

Combined with the new `wire_classification` skip (which correctly removes the dy estimate), there is no longer *any* independent signal that can make a compressed-path peer airborne: `rise`/`fall` are permanently false, so `avatar.gd`'s `nfall` never fires and remote jumps/falls don't animate at all.

The comment above says compressed packets carry no grounded state, but `MovementCompressed::temporal` has a `grounded` bit (plus `jump`/`falling`), populated by our own encoder in `Temporal::from_parts`. `message_processor.rs` already decodes `movement.temporal` for the timestamp and rotation — plumb `movement.temporal.grounded()` through and treat this path like the uncompressed one:

```suggestion
            avatar.apply_wire_air_state(wire_grounded, velocity.y);
            avatar.set_is_grounded(wire_grounded);
```

(with `wire_grounded` added as a parameter, sourced from `movement.temporal.grounded()` at the `message_processor.rs` call site) — and drop `set_grounded_from_air_state` entirely, since both wire paths then share the same shape.

Worth adding the inverse of `itest_compressed_path_grounded_coherent` too: `grounded = false` + `vy > 1` must yield `rise == true` / `is_grounded == false`. The current test only covers grounded→grounded, so it asserts the latch rather than catching it.

regenesis-claw (changes requested):
## Re-review — round 4 (head `87e717f`, delta vs `7056949`)

One appended commit, two Rust files. Delta review; I re-read the full compressed/uncompressed packet paths because the P1 fix changed how `is_grounded` is sourced.

### Round-3 findings — status

| Finding | Status |
|---|---|
| **[P1]** compressed path left `is_grounded` on the spiky dy estimate while `land` was wire-driven | ⚠️ Addressed, but the replacement introduces a worse bug — see below |
| **[P2]** dy estimate still blended ~24% into the speed EMA on the `Movement` path | ✅ Fixed — `wire_classification` latch, correct |
| **[P2]** `push_interpolated_transforms_to_scenes` bound the scene runner per avatar per frame | ✅ Fixed — one bind per batch |
| **[P2]** `LerpState.clock` unbounded `f32` | ✅ Fixed — `clock`/`arrival` are `f64`, casts at the f32 boundaries are all correct |
| **[P2]** rotation hard-zeroes pitch/roll; stale "snap rotation" commit message | ❌ Not addressed (still non-blocking) |
| **[P2]** new itests + `grounded_grace_test.gd` never run in CI | ❌ Not addressed — `cargo run -- run --itest` is documented in CLAUDE.md but no workflow invokes it; `linux_builds.yml` runs `cargo test` (unit only), `check-gdscript`, `test-avatar`. "23/23" is still local-only. |

---

## [P1] The new grounded derivation makes remote peers on the compressed path permanently grounded

`set_grounded_from_air_state()` derives `is_grounded` from `land`, and `land` on that path is derived from `grounded_gate`, which is the *previous* `is_grounded`. That closes a self-referential loop with no independent input:

```
gate_n        = is_grounded_{n-1}
rise_n/fall_n = !gate_n && vy ≷ ±1
land_n        = !rise_n && !fall_n
is_grounded_n = land_n && glide_state == 0     <-- new
```

Substituting: if `is_grounded_{n-1} == true`, then `gate == true` ⇒ `rise = fall = false` ⇒ `land = true` ⇒ `is_grounded_n = true`. **Once true it can never go false.** And it starts true: the first packet for a peer takes the `first` branch in `set_target_position`, which sets `land = true` / `is_grounded = glide_state == 0`. `glide_state` is never written from the wire on the compressed path either, so it stays `0`.

Net effect: for every peer arriving over `MovementCompressed`, `rise` and `fall` are now permanently `false` — remote jumps and falls never animate at all. `avatar.gd`'s `nfall` reads `is_grounded` directly, and `AvatarAnimHelpers.locomotion_conditions` gates walk/jog/run on it, so the air state machine is simply never entered for those peers.

Before this commit the gate came from the dy/interval estimate — spiky, but an *independent* signal, so it could break the latch on a real jump. The round-3 change removed the only independent source (`wire_classification` now also skips the dy branch) without replacing it. This is a regression introduced by the fix, and it's in the same family as #2732 (wrong remote air animation), just inverted: no bounce, but also no jump.

`itest_compressed_path_grounded_coherent` doesn't catch it because it only exercises the grounded→grounded case and asserts `is_grounded == true` — it encodes the latch as the expected behaviour. There's no test where a compressed-path peer leaves the ground.

### The wire flag you need already exists

The comment on the call site says *"Compressed packets carry velocity but no grounded/jump state"* — that isn't accurate. `MovementCompressed::temporal` is a bitfield that carries exactly this (`lib/src/comms/adapter/movement_compressed.rs`):

```rust
pub struct Temporal {
    pub timestamp: B15,
    pub movement_kind: Kind,
    pub sliding: bool,
    pub stunned: bool,
    pub grounded: bool,      // <-- 
    pub jump: bool,
    pub long_jump: bool,
    pub falling: bool,
    pub long_falling: bool,
    ...
}
```

`Temporal::from_parts` sets `.with_grounded(grounded)` on send, so the bit is populated by our own encoder. `message_processor.rs` already decodes `movement.temporal` for `timestamp_f32()` and `rotation_f32()` — it just never reads `grounded()` and never forwards it.

Suggested shape: pass `movement.temporal.grounded()` (and optionally `jump()` / `falling()`) into `update_avatar_transform_with_movement_compressed`, then use it as *both* the gate and the value, exactly like `apply_wire_movement_state` does on the uncompressed path — and drop `set_grounded_from_air_state()` entirely. That removes the derived loop and makes both wire paths structurally identical.

The follow-up test to add is the one that's missing: a compressed-path peer with `grounded = false` and `vy > 1` must produce `rise == true` / `is_grounded == false`.

### Smaller notes on the same commit

- The doc comment on `set_grounded_from_air_state` says *"the uncompressed path sets is_grounded from the wire AFTER calling it"* — it's actually before (`apply_wire_movement_state` assigns `self.is_grounded = is_grounded;` on the line above `self.apply_wire_air_state(...)`). Worth correcting whichever way this lands.
- **[verified, not a finding]** I checked whether the `wire_classification` latch can strand a peer that falls back to legacy `Position` packets: it can't. `update_avatar_transform_with_rfc4_position` already early-returns for any alias present in `last_movement_timestamp`, so Position packets are unreachable once a Movement has arrived, and avatars are freshly instantiated per alias in `add_avatar` (no pooling), so the flag can't leak across peers. The latch is consistent with the existing priority rule.
- **[verified]** The `f64` migration is complete: `clock`, `arrival`, `render_t`, and the segment search are all `f64`; `seg_dt` and `t` narrow to `f32` only after the subtraction, which is where the precision mattered.

---

## CI / platform

- Static checks ✅, Clippy & build checks ✅. Linux and Android still running at review time — worth confirming Linux is green rather than hitting the known headless teardown segfault again.
- This PR touches `lib/src/comms/adapter/message_processor.rs`. Per REVIEW.md §0.2 that's an iOS-build-required path, and there is no `build` label on the PR: *"No iOS build on this PR — a maintainer can add the `build` label to trigger one. I will not approve platform-sensitive iOS changes without a green iOS build."*

## Security

No findings. The delta is client-side animation state derived from already-parsed peer packets; no new parsing, no new external input surface, no trust-boundary change.

Followed `REVIEW.md`.

---

manuelmaceira:
Round 4 (`5d5da323ce`): you're right on all counts — the round-3 fix was a closed loop (gate = previous `is_grounded` → air state → next `is_grounded`), latching compressed peers permanently grounded, and my itest encoded the latch as expected behavior. My premise was also wrong: the temporal bitfield does carry `grounded` (and `jump`/`falling`), populated by our own `Temporal::from_parts`.

Fixed exactly as suggested: `temporal.grounded()` is decoded in `message_processor` and plumbed through; the compressed path now calls `apply_wire_movement_state` identically to the uncompressed one (jump_count/glide are zero by construction — the sender forces uncompressed when either carries info). `set_grounded_from_air_state` is gone, and the itest is now a lifecycle test (grounded → jump → fall → landing) that fails on the latch. 23/23.

regenesis-claw (commented):
## Re-review — round 5 (head `5d5da32`, delta vs `87e717f`)

One commit, does exactly what I suggested. The P1 is fixed correctly.

### [P1] Compressed-path grounded latch — fixed

`message_processor.rs` now decodes `movement.temporal.grounded()` and passes it through to `update_avatar_transform_with_movement_compressed`, which calls `apply_wire_movement_state(0, 0, is_grounded, velocity)` — same shape as the uncompressed `Movement` path. `set_grounded_from_air_state` (the self-referential loop) is deleted.

Verified the hardcoded `jump_count=0, glide_state=0` isn't a regression: in `communication_manager.rs::broadcast_movement`, `needs_uncompressed = jump_count >= 2 || glide_state != 0`, so a real compressed packet's `jump_count` is always 0 or 1 and `glide_state` is always 0 by construction — reconstructing 0 on receive matches the sender's actual state. The receive-side double-jump SFX in `avatar.gd` only fires on a rising edge to `jump_count >= 2`, which a compressed packet can never carry anyway (comment `#b8` confirms single ground jumps are intentionally still eligible for compression since rise/fall covers them) — so forcing 0 doesn't suppress any real SFX.

`itest_compressed_path_grounded_lifecycle` replaces the old test and now actually exercises the case that broke: grounded → jump (`rise=true`, `is_grounded=false`) → fall → landing (`is_grounded=true`, no `fall` flicker). This is the right regression coverage — it would have caught the round-3 bug.

### Remaining open items (both pre-existing P2s, non-blocking)

- Rotation hard-zeroes pitch/roll each frame; the "snap rotation" commit message from round 2 still doesn't describe current behavior. Cosmetic.
- The itest suite (including the new lifecycle test) still isn't invoked by any CI workflow — `linux_builds.yml` runs `cargo test --release -- --skip auth` (plain `#[cfg(test)]`), `check-gdscript`, `test-avatar`; `--itest` is local-only. Not this PR's job to fix, but the new coverage isn't actually enforced yet.

### Holding formal approval — iOS build gate

This PR touches `lib/src/comms/adapter/message_processor.rs`, which REVIEW.md §0.2 explicitly lists under "anything under `lib/src/comms/`" as iOS-build-required. Checked the sticky build-report comment (updated 2026-08-27 18:02 UTC, for this same commit range) — no `build` label, no iOS artifact. Per REVIEW.md:

> No iOS build on this PR — a maintainer can add the `build` label to trigger one. I will not approve platform-sensitive iOS changes without a green iOS build.

Code-wise this round is ready — only P2s remain and the P1 that's blocked every prior round is now correctly fixed. Add the `build` label and I'll approve once the iOS export job is green (or flag if you'd rather I approve now and treat the iOS build as a separate merge gate — your call, just wanted the reasoning explicit rather than silently approving over a REVIEW.md rule).

*CI*: Static checks + Clippy green on this head; Linux/Android still running.
*Security*: no new findings — this delta only threads an already-decoded wire bit through existing call paths.

Followed `REVIEW.md`.

---

regenesis-claw (approved):
iOS build confirmed green (workflow 33102129308, head `5d5da32`) — all jobs pass (Prepare distribution, iOS resolve, iOS export → R2, Android, Linux, Docker, static checks, Clippy). Approving per round-5 review: the compressed-path grounded latch (P1) is correctly fixed, and the remaining P2s (rotation pitch/roll zeroing, itests not wired into CI) are accepted as-is per author decision.

Note: my approval doesn't unblock godot-explorer branch protection — a human codeowner is still needed to merge.

kuruk-mm (approved):
looks great!

PR #2779, opened 2026-08-26
feat: native bug reporting flow with Intercom integration

### LINKED ISSUE #2652: Native bug reporting flow with Intercom integration - Implementation
## Overview
Implement native in-app bug reporting that creates support cases in Intercom, aligning with the desktop client initiative.

## Context
We previously had a bug reporting feature via an embedded Google Form accessible from Settings > Help, but removed it due to issues (e.g., requiring Apple users to sign in with Google to attach images).

This feature should:
- Provide a native form experience (no external web views)
- Create support cases directly in Intercom (Align with Stoyan)
- Allow users to report issues without leaving the mobile app
- Support image/screenshot attachments
- Improve support quality and user experience

## Related
Desktop client is implementing a similar initiative: https://app.notion.so/decentraland/3055f41146a580968f88eb23fe90246e?v=3055f41146a5816f9685000cd2220d04&p=3605f41146a5809dbcd4d15fc966a46c&pm=s

## Questions to Address
- How will support team visualize and handle back-and-forth communication? (In-app vs external?)
- What metadata should be auto-collected? (device info, app version, logs, etc.)
- Should we integrate with existing Intercom SDK or use API directly?

EibrielInv:
## What

  Replaces the "Report a Bug" Google Form deep link in Settings with a **native in-app bug report flow** that files Intercom tickets through the Decentraland `intercom-proxy` (issue #2652).

  The old flow bounced the player out to an external browser and required a Google sign-in to attach an image — so screenshots were almost never included. The new form is in-app, pre-fills a screenshot of the
  game, and lands in the same Intercom buckets as the Unity Explorer client.

  Closes #2652
  
  ### Bug report flow                                                                                                                                                                                 
  - **`BugReportService`** — signed-fetch `POST /intercom/tickets` against `intercom-proxy.decentraland.<env>`, with the `Origin` header the proxy allowlists. Ticket attributes mirror Unity's 
  `IntercomTicketClient.cs` exactly (an extra or misspelled key gets the whole ticket rejected). Attaches device info (OS + brand/model, GPU, RAM), client version and last parcel coordinates.
  - **`BugReportCategories`** — the 8 issue types the mobile design keeps, as Intercom option UUIDs ported from Unity's `BugReportIssueTypes.cs`.
  - **`BugReportCapture`** — grabs the rendered frame **when Settings opens**, not when the form opens, so the attached image shows the actual bug instead of the Settings panel. Wired from `global.gd` because
  Settings is reachable both in-world and from the lobby.
  - **`BugReportModal`** — issue-type dropdown, 300-char description with counter, and **one screenshot** (pre-filled from the capture; it can be replaced or deleted). The approved design has three slots, but the proxy's `evidence` field carries a single image, so the extra slots offered an upload that was silently dropped at submit — `MAX_SCREENSHOTS` is capped at 1 until the proxy accepts multiple.
  - Evidence is JPEG-encoded at q85; anything over the proxy's **3 MB** cap is dropped so an oversized image can't sink the whole ticket. Submit is disabled until a type and a description are present.

  ### Native photo picker (new capability)
  - **`ImagePickerService`** — GDScript facade over new iOS (`PHPickerViewController`) and Android (`ACTION_OPEN_DOCUMENT`) pickers. The native side does the decode, EXIF rotation, downscale to 1920px and JPEG
  re-encode, so GDScript never holds a full-res 12MP photo. Exactly one `image_picked` signal per call (success / cancel / error), with a watchdog and a concurrent-call guard.
  - Android picks up `androidx.exifinterface:1.3.7` (both `build.gradle.kts` and the export plugin's dependency list).

  ### Supporting changes
  - **`ModalShell`** — reusable modal chrome (dim backdrop, screen blur, rounded panel, safe padding, content slot), replacing the subtree every modal was re-declaring inline. Only the new bug report modal uses
  it; migrating the five existing modals is deliberately left as a follow-up.
  - New atoms/molecules: `FieldLabel`, `ModalActions` (busy state), `ScreenshotSlot`.
  - **`Global.async_signed_fetch`** gains optional `metadata_override` and `extra_headers` — the proxy signs `{}` and leaves the body unsigned, unlike most DCL services.
  - New `IntercomProxy` service group in `lib/src/env` + `DclUrls.intercom_tickets()` / `intercom_origin()`, so QA can point reports at `zone` while the rest of the app stays on `org`.

  ## QA

  Test on both **iOS and Android** devices.                                                                                                                                                           

  **Happy path**
  1. Sign in, walk into any scene, then open **Settings** from the in-world menu.                                                                                                                     
  2. Tap **Report a Bug**. → The form opens with the **first screenshot slot already filled with the in-world frame** (not the Settings panel, not a black image).
  3. Pick an issue type, type a description. → Submit stays disabled until *both* are set; the counter caps at 300 chars.
  4. Delete the pre-filled screenshot, then tap the **+** slot → the **native gallery picker** opens. Choose a photo. → It replaces the slot, correctly oriented (test with a photo shot in **portrait** and one in **landscape** — EXIF rotation is the thing being checked).
  5. With a screenshot attached there is **no + slot** (one image is the cap). Delete it → the **+** comes back.
  6. Submit. → Buttons go busy, modal closes, **"Report sent"** toast appears.
  7. In Intercom, confirm the ticket exists with: correct issue type, title `Bug Report: <label>`, the description plus `Coordinates: x,y`, OS **with device brand/model**, GPU, RAM, client version, and **the 
  image attached**.

  **Cancel / dismiss**
  8. Open the form and tap **Cancel** → modal closes, no ticket filed. Tapping the **backdrop** must **not** dismiss it (`dismissable = false`, so an accidental tap can't discard a typed report).
  9. Open the native picker and **cancel it**. → Returns to the form with slots unchanged, no hang, no duplicate slot.                                                                                

  **Error handling**
  10. Enable airplane mode, submit. → Modal stays open, a generic failure toast appears, no crash. Re-enable network and resubmit → succeeds.
  11. Attach a very large photo (>3 MB after re-encode). → Ticket is still filed, **without** the attachment, and a warning is logged rather than the submit failing.

  **Environment routing**
  12. Launch with `--dclenv=zone` and file a report → hits `intercom-proxy.decentraland.zone`. Confirm the per-group override `intercomproxy` also works while the rest of the app is on `org`.

  **Regression**                                                                                                                                                                                      
  13. Confirm the old Google Form no longer opens anywhere, and that the other modals (input, code, travel, reward, upgrade, connection-lost) are visually and behaviourally unchanged — `ModalShell` was added 
  alongside them, not swapped in.

  **Entry from the lobby**
  14. Open Settings from the **lobby menu** (no explorer running) and file a report → form opens and submits fine; the screenshot slot may be the lobby frame, which is expected.
  

  ---

  ## Sentry diagnostics link (added after initial review)

  The ticket description ends with an `Internal diagnostics:` line. Unity fills it with a Sentry deep link; ours hardcoded `unavailable`. It now carries a real link.

  - **`SentryUserFeedback`** — before the Intercom POST, captures a Sentry event carrying a **128 KB tail of `godot.log`** plus the screenshot, then files a **Sentry User Feedback** entry associated with that event. Mirrors Unity's `SentryUserFeedbackService`.
  - Links to the **event**, not the feedback entry. The Godot SDK's `capture_feedback()` returns nothing (documented behaviour), so the feedback id is unobtainable — Unity only gets it by reaching past its own facade via `HubAdapter`, and there is no equivalent here. The feedback hangs off the event via `associated_event_id`, one hop from the link.
  - Event ids are **normalised to the 32-hex dashless form** (the SDK emits both dashed and simple UUIDs); the nil id is treated as failure so a dropped event never produces a dead link.
  - `_before_send` **exempts these events from the remote `sentry-sample-rate` throttle**. Without it the ticket links to an event that was never sent — and it fails *silently*, because a dropped event still returns a valid-looking id. **The `0.0` remote kill switch still wins.**
  - The screenshot is JPEG-encoded **once** and shared: `_build_evidence` drops images over the proxy's 3 MB cap, which is precisely when the Sentry copy matters. Attachment is named `bug_report_tail.log`, not `godot.log`, to avoid colliding with the existing 1%-sampled `options.attach_log` upload.
  - Sentry failure **never blocks the ticket** — every path returns a string and falls back to `unavailable`.
  - **`SDK version`** ticket attribute — the current scene's `runtimeVersion`, matching Unity's `SceneSdkVersion`. Deliberately omitted (not sent empty) when there is no live scene, e.g. lobby/Discover.

  ### Verified on device (Pixel 7, staging build)

  A staging build is required: `_before_send` returns `null` for dev builds, so none of this is observable on a debug build (`DECENTRALAND_STAGING_BUILD=1`).

  - Ticket carried `Internal diagnostics: https://dcl-regenesis-labs.sentry.io/issues/?query=60ea003c…` and the link resolves
  - Sentry event `environment=staging`, `level=info`, with **both** attachments (`screenshot.jpg`, `bug_report_tail.log`)
  - **User Feedback entry is linked to the event** — confirms `set_associated_event_id()` accepts the normalised id format
  - Reporter display name and real parcel coordinates present on the ticket

  ## Known issues / flags for reviewers

  1. **Log privacy — needs a decision.** Each report ships up to 128 KB of a real user's `godot.log` to Sentry, from a button whose UI says nothing about logs. Two mitigating facts: this is *narrower* than what already ships (`options.attach_log` uploads the **entire** log for a random **1% of all production sessions**), and Unity does the same thing. Note the corollary: **an audit of whether `godot.log` can contain credentials applies to what we ship today**, independently of this PR.
  2. **Sentry issue grouping is messy — pre-existing, not caused by this PR.** Exception-less `capture_message`-style events collapse into a single issue regardless of their message. Our event grouped with three unrelated `[Sentry Test]` events from Aug 2 (release `0.47.0-alpha`, different user, three different titles), and re-opened that resolved issue as *Regressed*. `node_guard.gd` warnings land in the same bucket. This SDK exposes no `set_fingerprint`, so the fix is a **server-side Fingerprint Rule** in `dcl-regenesis-labs/godot-explorer` — no client change: `message:"Bug report diagnostics" -> bug-report-diagnostics`
  3. **The link is queued, not flushed.** The ticket asserts a link to an event that is only queued for transport; if the app dies before it flushes, the link 404s. The SDK exposes no flush API.
  4. **`capture_feedback()` failure is undetectable** (returns `Nil`). If the feedback entry fails, the link still resolves to an event carrying the log and screenshot.
  5. **Upstream request worth opening:** `getsentry/sentry-godot` should return the feedback id from `capture_feedback()`, as the .NET SDK does — that would let us link directly to the feedback entry (one string to change here).

  ## Deferred from review (agreed, tracked)

  6. **`BugReportCapture._latest` is retained for the process lifetime** (~8 MB `Image`; `clear()` has no callers). It is *bounded* rather than leaking — one allocation, overwritten by each capture, never accumulating. Clearing it on modal close would break reopening the form with the screenshot still populated, and there is no `close_settings` signal to hang it on; the correct fix is storing encoded JPEG bytes (~200 KB), which touches the capture, `set_initial_screenshot` and slot rendering. Follow-up rather than more churn in this PR.
  7. **Capture cost on the navbar path** (`navbar.gd:191`). `Viewport.get_texture().get_image()` is a synchronous GPU→CPU readback plus a bilinear resize, now run on **every** navbar open whether or not a report is filed. Not yet measured on a low-end device; gating it (e.g. only once Settings is reachable) is the obvious mitigation if it hitches.

  ### Not yet verified

  - `sentry-sample-rate = 0.0` → expect **no** event and `unavailable` in the ticket (proves the kill switch still wins)
  - A **>3 MB** screenshot → expect it absent from Intercom but **present** in Sentry (the entire reason the JPEG encode is shared)


ElzathRusso:
**Feedback**:

- [ ]  **Modal title:** The wrong font variable is being used. The title should match all other modals: Inter Bold.
Screenshot 2026-09-01 at 12 51 18 PM

---
- [ ] **Description text field:** The X icon used to clear the text is hard to tap. Please increase the interaction area.

---
- [ ] **Description text field:** The spacing between the title and the field is smaller than in the design.
Screenshot 2026-09-01 at 12 49 44 PM


---
- [ ] **Success modal:** After tapping SUBMIT, the confirmation modal indicating that the report was sent successfully is missing.
Settings-Accountexito


---
- [ ] **In-game:** The modal is cut off. It is not switching from landscape to portrait.


https://github.com/user-attachments/assets/305c57f2-f571-45c1-a6e4-a910ab3a813c

ElzathRusso:
Modal success: portrait and landscape
The icon is pixelated, smaller, and off-center.

Screenshot 2026-09-02 at 6 44 18 PM

Screenshot 2026-09-02 at 6 44 12 PM

Ludmilafantaniella:
Hey @EibrielInv! Tested on Android and iOS. Found one thing: tapping the backdrop doesn't close the modal, only "Cancel" does (step 8 expects both). Everything else passed. Holding off on approval until clarified. intentional or bug?

Ludmilafantaniella (approved):
## ✅ QA Approved

Platforms tested: Android & iOS
Verified: happy path (screenshot capture, issue type + description, photo picker with EXIF rotation, 3-image limit, submit + Intercom ticket fields), Cancel button dismiss, airplane mode error handling, large photo attachment fallback, zone routing, old Google Form removed, other modals unchanged, lobby entry point.

https://github.com/user-attachments/assets/b0fce6c0-51f0-40fb-bed4-978580d9793f

godot/src/ui/pages/settings/settings.gd
```diff
@@ -716,59 +716,34 @@ func _on_button_test_notification_pressed() -> void:
...
-	if params.size() > 0:
-		url += "?" + "&".join(params)
 
-	Global.open_url(url)
+func _on_bug_report_failed(message: String) -> void:
+	# The message is a proxy/transport error, not something a player can act on,
+	# so it goes to the log while the toast stays generic.
+	push_warning("Bug report failed: %s" % message)
+	NotificationsManager.show_system_toast(
+		"Couldn't send report", "Something went wrong. Please try again.", "system", "alert"
```
regenesis-claw: [P1] Player-visible copy as raw English literals. The i18n scanner can't see strings passed as function arguments, so CI stays green — but on `es` / `pt_BR` this toast renders in English. Every other `show_system_toast` call site uses `tr("KEY")`, and the body copy already exists as `COMMON_SOMETHING_WENT_WRONG_RETRY`.

```suggestion
		tr("TOAST_BUG_REPORT_FAILED_TITLE"), tr("COMMON_SOMETHING_WENT_WRONG_RETRY"), "system", "alert"
```
(new title key to be added to `en.csv` / `es.csv` / `pt_BR.csv` + `translation_sources.json`)

godot/src/ui/components/molecules/modal_actions/modal_actions.gd
```diff
@@ -0,0 +1,88 @@
...
+## so the pair picks up theme changes instead of pinning its own colours.
+
+signal secondary_pressed
+signal primary_pressed
+
+const LOADING_SPINNER = preload(
+	"res://src/ui/components/atoms/controls/loading_spinner/loading_spinner.tscn"
+)
+
+@export var secondary_text: String = "CANCEL":
```
regenesis-claw: [P1] These defaults clobber the translation keys the scene sets. `modal_actions.tscn` gives the buttons `text = "COMMON_CANCEL"` / `"COMMON_SUBMIT"`, but `_ready()` assigns `secondary_text` / `primary_text` over them, and `bug_report_modal.tscn` overrides neither export — so both buttons render literal English in every locale, and the `COMMON_SUBMIT` key added in this PR is never used. This is the `@export`-default case `REVIEW.md` §Tier 2/8 asks reviewers to reject.

Defaulting to the keys fixes it without a translation-changed hook, since `Button.text` auto-translates and re-resolves on locale change:

```suggestion
@export var secondary_text: String = "COMMON_CANCEL":
```
(and `primary_text` → `"COMMON_SUBMIT"`, plus the `_primary_label` default on line 34.)

godot/src/ui/components/organisms/navbar/navbar.gd
```diff
@@ -185,6 +185,10 @@ func _on_button_toggled(toggled_on: bool) -> void:
 	Global.send_haptic_feedback()
 	panel_profile.set_glow(toggled_on)
 	if toggled_on:
+		# Grab the frame BEFORE the open animation starts: this is the last
+		# moment the world is unobstructed, and it is what the bug report form
+		# pre-fills its screenshot slot with (issue #2652).
+		BugReportCapture.capture(get_viewport())
```
regenesis-claw, line 191: [P2] `get_texture().get_image()` is a synchronous GPU→CPU readback plus a bilinear resize, and this now runs on **every** navbar open — immediately before `_animate_open()` — for a screenshot that is discarded unless the player files a bug. Worth a measurement on a low-end Android: if it costs a frame here, the animation starts with a visible hitch. If it does, gating the capture (or capturing at a lower resolution) would keep the feature without paying on every toggle.

godot/src/bug_report/bug_report_capture.gd
```diff
@@ -0,0 +1,81 @@
...
+	var longest: int = maxi(image.get_width(), image.get_height())
+	if longest > MAX_DIMENSION:
+		var scale := float(MAX_DIMENSION) / float(longest)
+		image.resize(
+			maxi(1, int(image.get_width() * scale)),
+			maxi(1, int(image.get_height() * scale)),
+			Image.INTERPOLATE_BILINEAR
+		)
+
+	_latest = image
```
regenesis-claw: [P2] `_latest` is a static that nothing ever releases — `clear()` has no callers — so a ~1920px `Image` (≈8 MB uncompressed) stays resident for the rest of the process after the first Settings/navbar open, on a memory-constrained target. Consider calling `clear()` when the report modal closes, or storing the JPEG-encoded bytes instead of the raw `Image` and decoding only when the form opens.

plugins/dcl-godot-android/plugin/src/main/java/org/godotengine/plugin/android/template/GodotAndroidPlugin.kt
```diff
@@ -1031,6 +1040,197 @@ class GodotAndroidPlugin(godot: Godot) : GodotPlugin(godot) {
...
+    // registerForActivityResult() is not usable here: the Godot activity is
+    // already STARTED by the time onGodotSetupCompleted() runs, and
+    // registering after STARTED throws. Hence startActivityForResult plus the
+    // onMainActivityResult hook below.
+
+    // Longest edge and JPEG quality for the in-flight pick, captured at launch
+    // so the result handler doesn't need them passed back through the Intent.
+    private var pickImageMaxDimension: Int = 1920
+    private var pickImageJpegQuality: Int = 85
+    private var pickImageInFlight: Boolean = false
```
regenesis-claw, line 1076: [P2] `pickImageInFlight` is only cleared in `onMainActivityResult`. If the activity is recreated while the picker sheet is up (low-memory kill, config change), that callback never arrives and every subsequent `pickImageFromGallery` answers `already_picking` for the rest of the process — while GDScript's own `_picking` clears itself after the 300 s watchdog, so the two sides disagree permanently. A reset in an activity lifecycle hook (e.g. `onMainResume`) would bound it.

plugins/dcl-godot-ios/plugins/dcl_godot_ios/image_picker_service.mm
```diff
@@ -0,0 +1,191 @@
...
+        // GDScript guards against concurrent picks too, but a dropped signal
+        // would hang an awaiter forever, so the native side refuses as well.
+        if (dcl_active_picker_delegate != nil) {
+            // Emitted directly, bypassing the one-shot guard: this reports the
+            // *new* call, while the guard belongs to the pick still on screen.
+            dcl_emit_image_picked(nil, @"already_picking");
+            return;
+        }
+
+        UIViewController *rootVC = [UIApplication sharedApplication].keyWindow.rootViewController;
```
regenesis-claw, line 167: [P2] `keyWindow` is deprecated since iOS 13 and returns nil under scene-based configurations, which surfaces here as a silent `no_root_view_controller` with no picker. Prefer resolving the foreground-active window from `UIApplication.sharedApplication.connectedScenes` and falling back to `keyWindow` only if that yields nothing.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`.

**Pre-flight**
- Branch is up to date with `main` (13 ahead, 0 behind) ✅
- iOS: platform-sensitive (new `PHPickerViewController` plugin + `.gdip` plist/framework changes) — `🍏 iOS → R2 (unsigned export)` is green and the distribution run shipped a TestFlight build ✅
- No submodule pointer drift; all CI checks green ✅
- QA approved on Android + iOS (@Ludmilafantaniella) ✅

This is a strong piece of work — the Intercom attribute contract is ported field-for-field from Unity, the picker's one-signal-per-call contract is defended on both native sides *and* in GDScript (the swipe-dismiss + watchdog handling is the kind of thing that usually ships as a hang), the Sentry attachment window is kept synchronous on purpose, and the "why" comments are exemplary. The self-reported known-issues list is honest and I agree with the framing on all five.

Two blockers, both i18n — the exact blind spots `REVIEW.md` §Tier 2/8 says the scanner cannot see, which is why static checks are green.

---

### 🔴 P1 — Failure toast ships raw English (settings.gd:744)

Strings passed as **function arguments** are invisible to `extract_strings.py`, so CI can't catch this. Every other `show_system_toast` call site in the repo passes `tr("KEY")` (`connection_quality_monitor.gd:258`, `profile.gd:259/264/371/396`). `COMMON_SOMETHING_WENT_WRONG_RETRY` already exists in the catalogue and is exactly this copy; the title needs a new key.

### 🔴 P1 — `ModalActions` @export defaults clobber the button translation keys (modal_actions.gd:20/26)

`modal_actions.tscn` correctly sets `text = "COMMON_CANCEL"` / `"COMMON_SUBMIT"` on the two buttons, but `_ready()` overwrites both with the @export defaults `"CANCEL"` / `"SUBMIT"`, and `bug_report_modal.tscn` doesn't override either export. `REVIEW.md` calls this out by name: *"an `@export` default silently clobbers a correct key set in the `.tscn`"*. Net effect: the form's buttons render English on `es` / `pt_BR`, and the `COMMON_SUBMIT` key this PR adds to all three CSVs + `translation_sources.json` is dead. Fix is the defaults themselves — `Button.text` auto-translates a key and re-resolves on a language change, so no `NOTIFICATION_TRANSLATION_CHANGED` hook is needed.

---

### 🟡 P2 — worth fixing, not blocking

1. **Test plan is stale in two places QA actually executes.** Step 8 asks for backdrop dismissal, but `bug_report_modal.tscn` sets `dismissable = false` — this already cost a QA round-trip. Steps 4–5 describe adding a 2nd and 3rd screenshot and the `+` tile disappearing at 3, which `MAX_SCREENSHOTS = 1` makes unreachable. Please align the description with the shipped behaviour (and drop/park the "up to 3 screenshots" line in *What*).
2. **`BugReportCapture._latest` is never released.** A ~1920px `Image` (≈8 MB uncompressed) is retained for the process lifetime; `clear()` has no callers. Consider clearing it when the report modal closes, or storing the already-encoded JPEG bytes instead of the raw `Image`.
3. **Capture cost on the navbar path** (navbar.gd:191): `Viewport.get_texture().get_image()` is a synchronous GPU→CPU readback plus a bilinear resize, now run on *every* navbar open — right as `_animate_open()` starts — whether or not a report is ever filed. Worth measuring on a low-end Android before this lands; if it hitches, gating the capture (e.g. only after Settings is reachable) would keep the feature without paying on every toggle.
4. **Submit stalls before the spinner paints.** `_async_submit()` calls `set_busy(true)` and then `BugReportService.async_submit()`, whose first work — JPEG encode + 128 KB log read + `capture_event` — happens *before* the first `await`, so the frame with the spinner never renders. An `await get_tree().process_frame` after `set_busy(true)` would let the busy state show.
5. **Android: `pickImageInFlight` can latch true forever** (GodotAndroidPlugin.kt:1062). It's only cleared in `onMainActivityResult`; if the activity is recreated while the picker is up, that callback never lands and every later pick answers `already_picking` for the rest of the process. GDScript's watchdog resets its own `_picking` after 300 s, so the two sides drift. Consider clearing it in an activity lifecycle hook (`onMainResume`) as a backstop.
6. **iOS: `keyWindow` is deprecated since iOS 13** (image_picker_service.mm:167) and returns nil in some scene-based configurations, which turns into a silent `no_root_view_controller`. Prefer walking `connectedScenes` for the foreground-active window.
7. **Log-privacy decision (your known-issue #1) — please get an explicit sign-off before merge.** I agree it's narrower than today's `attach_log` 1% sampling and matches Unity, but the button's copy still says nothing about shipping 128 KB of `godot.log`. A one-line note under the Screenshot field, or a product/legal ack recorded on the issue, would close it.

---

### Verified / checked
- No security issues found: no secrets in the diff, the Intercom token stays server-side behind the proxy, the `Origin` header is built from `DclUrls` (not user input), `ServiceGroup::IntercomProxy` is wired through `index()`/`as_str()`/`from_str()`/`ALL` and `COUNT` bumped to 12 consistently, and the new URLs are covered by the `urls` test list. The picker paths are permissionless on both platforms (system photo picker / SAF), and `NSPhotoLibraryUsageDescription` is present.
- `_before_send`: the FEEDBACK exemption sits after the dev-build drop and after the crash-only gate (exception-less events already pass there), and honours the `0.0` kill switch — behaves as described.
- `_normalize_event_id` returning `""` for a dropped event, and `capture_feedback()` returning nothing, are taken as reported — I can't verify either from the diff.
- The rebuilt `.aar` / `.xcframework` / `DclSwiftLib` binaries are opaque in the diff; I'm trusting they were produced from the sources in this PR.

EibrielInv:
Thanks for the review — the two P1s were both real and both in the blind spots you named. Everything below is desktop-verified through the debug hub unless stated; a device build is in flight but not yet confirmed.

## P1 — fixed

**1. Failure toast (`settings.gd:744`).** Both arguments now go through `tr()`. `COMMON_SOMETHING_WENT_WRONG_RETRY` was an exact match for the body as you said; the title needed a new key, `TOAST_BUG_REPORT_FAILED_TITLE`, added to all three catalogues.
Verified in `es`: *"No se pudo enviar el reporte / Algo ha salido mal. Inténtalo de nuevo."*

**2. `ModalActions` export defaults.** Changed to the keys themselves. One addition to what you flagged: `_primary_label` also defaulted to `"SUBMIT"`, and `set_busy(false)` restores `button_primary.text` from it — so fixing only the two `@export`s would have left the label reverting to English after the first submit. All three now hold keys.
Verified in `es`: **CANCELAR / ENVIAR**, still translated after a full busy round-trip.

## P2

**1. Test plan** — description updated: the backdrop step now states that dismissal is deliberately disabled (`dismissable = false`), the 2nd/3rd-screenshot steps are rewritten for the single-slot cap, and "up to 3 screenshots" is gone from *What*.

**2. Retained `Image` — fixed rather than deferred.** Took your second suggestion. `BugReportCapture` now stores encoded JPEG bytes: **144 KB instead of ~6.8 MB**, and the whole capture including the encode measures **8 ms**. `BugReportService.async_submit()` takes bytes directly and no longer encodes anything.

**4. Spinner** — `await get_tree().process_frame` after `set_busy(true)`. Note this is largely subsumed by #2: the JPEG encode, previously the largest pre-`await` item, now happens at capture time when nobody is waiting.

**5. Android `pickImageInFlight`** — cleared in the existing `onMainResume()` as you suggested. `onActivityResult` is delivered before `onResume`, so the happy path is unaffected. AAR rebuilt; I confirmed the new code is in the compiled class rather than trusting the build (Gradle exits 0 on failure here). The failure it guards against is awkward to reproduce, so this is verified as *compiled*, not as *working*.

**6. iOS `keyWindow` — reverted, not fixed.** I implemented a `connectedScenes` walk, then found two problems. There is **no `UIApplicationSceneManifest` in this project**, so the app uses the legacy `UIApplicationDelegate` lifecycle and `keyWindow` returns a real window today — consistent with QA passing the picker on iOS. And my replacement dropped the `keyWindow` fallback, so an empty `connectedScenes` would have returned nil and broken the picker outright. That trades a dormant future failure for a possible present one on a path QA already signed off, so it's reverted.
The deprecation is still real: if anyone adds a scene manifest, adopts `UIWindowSceneDelegate`, or does proper iPadOS multitasking, this silently becomes `no_root_view_controller` and the gallery button stops working. Tracking as a known issue rather than shipping an unbuilt, untested rewrite.

**3. Navbar capture cost** — recorded in the description's known-issues. Still unmeasured on a low-end device.

**7. Log privacy** — ack is being obtained separately; treating it as a merge gate.

## Found while working, outside the review

- **The screenshot delete badge was the genuinely hard-to-tap control**, not the text-field ✕. It was 32×32 logical — **18.3 dp** on a Pixel 7, about a third of the 48 dp guideline — with the badge stylebox on the button itself, so hit area and artwork were identical. Now a transparent 84×84 button (**48 dp**) with the badge and glyph as `mouse_filter = IGNORE` children; badge centre unmoved, confirmed on device by the reporter.
- **The picker was decoding native JPEG and re-encoding it at submit** — a lossy round trip. `ImagePickerService` gained `async_pick_image_bytes()`, so attachments are now byte-identical to what the native side produces.
- Worth noting for future review: the text-field ✕ your item targeted is 48 logical = **~27 dp**, still under the guideline. Left as-is deliberately; raising it makes the invisible target overlap the text area.

## Verification gaps

- Everything above is desktop-verified. The device build carrying #2, #4 and the delete badge is still building.
- **`async_pick_image_bytes()` has never executed** — it is iOS/Android only, so the first gallery pick on a device is its first real test.

regenesis-claw (approved):
Delta review of `f50fc86` ("review fixes", 1 commit on top of `6323037`, 0 behind `main`).

### Round-1 findings

| # | Finding | Status |
|---|---|---|
| P1-1 | Failure toast shipped raw English | ✅ both args through `tr()`; `TOAST_BUG_REPORT_FAILED_TITLE` added to en/es/pt_BR + `translation_sources.json` |
| P1-2 | `ModalActions` @export defaults clobbering the keys | ✅ and you found the third one — `_primary_label` also defaulted to `"SUBMIT"` and is what `set_busy(false)` restores, so fixing only the two exports would have regressed the label after the first submit. Good catch |
| P2-1 | Stale test plan | ✅ step 8 now states backdrop dismissal is deliberately off, steps 4–5 rewritten for the single-slot cap, "up to 3 screenshots" gone |
| P2-2 | Retained ~8 MB `Image` | ✅ better than what I asked for — `BugReportCapture` stores JPEG bytes (144 KB), `async_submit()` takes bytes, and the encode moved off the submit path entirely |
| P2-3 | Navbar capture cost | ⚠️ recorded as a known issue, still unmeasured on a low-end device. Note the navbar path now also carries the JPEG encode (8 ms desktop) |
| P2-4 | Spinner never painted | ✅ `await get_tree().process_frame` after `set_busy(true)` |
| P2-5 | Android `pickImageInFlight` latching | ✅ cleared in `onMainResume()`; the ordering argument (`onActivityResult` before `onResume`) holds |
| P2-6 | iOS `keyWindow` | ✅ revert accepted — I verified there is no `UIApplicationSceneManifest` anywhere in the repo, so `keyWindow` resolves today, and shipping an unbuilt rewrite that drops the fallback would have been the worse trade. Tracking it as a known issue is right |
| P2-7 | Log privacy | ⏳ treated as a merge gate — agreed |

The two unrequested fixes are both improvements: the delete badge at 84×84 (48 dp) with the badge/glyph as `mouse_filter = IGNORE` children is the correct shape for that, and `async_pick_image_bytes()` removing the decode→re-encode round trip means the attachment is now byte-identical to what the native side produced. Thanks for flagging that the text-field ✕ is still ~27 dp — noted, and I agree it isn't worth overlapping the text area for.

---

### New in this commit — all P2, none blocking

1. **The hidden modal still holds a decoded preview.** `_shots[i]["image"]` plus the slot's `ImageTexture` live for the life of the modal, and `close()` only hides — `_remove_bug_report_modal()` doesn't run until the *next* open. So after one report there's still a full-size decoded frame resident, which takes back part of the win from #2. Clearing `_shots` (and calling `BugReportCapture.clear()`) in `close()`, or freeing the modal there, would close it out.
2. **`async_pick_image()` now has zero callers.** The only consumer moved to `async_pick_image_bytes()`; nothing else in the repo references it. It's a reasonable convenience wrapper to keep, but as of this commit it's dead code (`REVIEW.md` Tier 3/16) — worth either a comment saying it's kept for future callers, or dropping it.
3. **`capture()` leaves a stale frame on encode failure.** The early `return` after `push_warning` skips the assignment, so `_latest_jpeg` keeps the *previous* capture — the form would then pre-fill with a frame from an earlier screen rather than nothing. Clearing `_latest_jpeg` before the encode (or on failure) makes the failure mode "no screenshot" instead of "wrong screenshot".
4. **The 84×84 delete button will overlap its neighbour when the 3-slot design returns.** With `offset_right = 33` it extends 33 px past the tile's right edge, while `GAP_FULL`/`GAP_SPARSE` are 20/32 — so with `MAX_SCREENSHOTS` back at 3 it lands on top of the next slot's `Button_Add`, and since the next slot is a later HBox sibling it wins the touch, silently killing the right third of the delete target. Harmless at the current cap of 1 (there is never a filled tile *and* a following tile) — just worth a note next to `MAX_SCREENSHOTS` so it's re-checked when the cap is lifted.
5. Stale comment: the new `await get_tree().process_frame` in `_async_submit()` says `async_submit()` "does its JPEG encode … before its first await" — the encode moved to capture time in this same commit.

### Verification status (yours, restated as merge gates)

- The device build carrying the bytes-based attachment path, the spinner yield and the new delete badge is still in flight, and `async_pick_image_bytes()` **has never executed** — every one of those is an iOS/Android-only path, and QA's pass was against `6323037`. A fresh QA run on this head (add the `build` label for the iOS artifact) is worth having before merge; the picker → attach → Intercom round trip is the specific thing to re-run.
- CI on this head: Static checks and Clippy green, Linux and Android still running at review time; no iOS artifact yet on `f50fc86` (the delta touches no iOS native code — the `keyWindow` change was reverted — but the shared GDScript attachment path did change).
- Log-privacy ack outstanding, per your own gate.

No security issues found in the delta. Approving — the P1s are resolved and everything above is either cosmetic or a verification step you've already named.

ElzathRusso (approved):
🙌

PR #2788, opened 2026-08-28
feat(ftue): boot an ad-attributed install into the scene the ad sold

### LINKED ISSUE #2670: Ad-Based Dynamic FTUE Bypass to Scene (Android Only)

leanmendoza:
Closes #2670.

An install attributed to an ad campaign boots straight into the scene the ad sold, skipping the FTUE. Everything else gets today's FTUE, unchanged.

The ad link carries an opaque `?c=<token>` and the destination is resolved server-side against the mobile-bff — endpoint in decentraland/mobile-bff#81, backoffice panel in decentraland/mobile-hub#12, both merged (#2669). Opaque is required, not just flexible: a deeplink carrying `position=` or `realm=` trips `_should_go_to_explorer_from_deeplink()` and boots the explorer past avatar creation, skipping the very screen this personalizes.

An ad-driven install has no deeplink — the app did not exist when the ad was clicked — so on Android the token arrives through install attribution instead. Both mechanisms are implemented because they cover disjoint traffic: Google Ads replaces the Play referrer with a bare `gclid` (87% of attributed installs, measured on Segment over 30 days), and only a GA4F deferred deep link reaches those; organic installs keep their query string.

Every failure path — no token, unknown, expired, already consumed, resolver unreachable, unusable target — falls back to today's FTUE, which stays the default, not a degraded one.

**Review the squashed diff.** The later commits fix defects found in the earlier ones, so the first commit messages no longer describe the current code.

## Behaviour changes outside the feature

- **`rotate-guest=true` ships.** `is_production()` is true only for `release*` branches, so builds from `main` — TestFlight included — honour it. It wipes the local guest identity, with no off switch other than reinstalling; recoverable, since the native device anchor is untouched (verified on device across install → rotate → reinstall).
- **Lobby deeplink handling is narrower.** `_on_deep_link_received` now requires a destination the router acts on, so a token-only link stays inert.
- **Attribution is polled before the EULA gate.** The deadlines run on wall-clock from boot, so behind the gate the sources went unread until the user tapped through and GA4F's window had expired. Nothing leaves the device earlier — polling only queues; `process_and_send_events` stays gated.
- `Install Attribution` gains `attribution_source`, `campaign_token`, `deferred_deep_link` and `referrer_settled`.

## QA (Android)

Each scenario needs a **fresh install** *and* the **deeplink**: reinstalling clears app data, the link rotates the device anchor so a new guest is minted and the first-time flow is reachable. Skip either and the result is wrong.

| Link | Expected |
|---|---|
| `decentraland://open?rotate-guest=true` | The usual first-time screen |
| `decentraland://open?rotate-guest=true&c=aesironline` | Skips it, lands in `aesironline.dcl.eth` |

## Not verified, and one call to make

**The attribution path has never run on a device.** Every test above delivers the token by deeplink; ad spend has been paused since late August, so GA4F and the referrer are unexercised until an ad group is turned back on. A process killed in the first ~10-25s also loses the campaign.


lib/src/urls/mod.rs
```diff
@@ -248,6 +248,13 @@ pub fn feature_flags() -> String {
         suffix(ServiceGroup::MobileBff)
     )
 }
+pub fn campaigns() -> String {
+    // TODO: restore the per-environment form once the campaigns endpoint ships to
+    // production. It only exists on the dev deployment today (`.org` returns 404), so
+    // resolving by environment would break the org build rather than fall back:
+    //     format!("https://mobile-bff.decentraland.{}/campaigns", suffix(ServiceGroup::MobileBff))
+    "https://mobile-bff.decentraland.zone/campaigns".to_string()
```
regenesis-claw: [P1] This ships to production. `campaigns()` has no environment branch, so an `.org` build that captured a token queries the dev deployment and boots the user into whatever that dev row points at — a target chosen in the zone backoffice, on a real user's first launch.

Gating it off in production costs nothing, since the endpoint 404s there today and the client's failure path is already the default FTUE:

```suggestion
pub fn campaigns() -> Option<String> {
    // The endpoint only exists on the dev deployment today (`.org` 404s), so a production
    // build must not resolve against it. Restore the per-environment form once it ships:
    //     format!("https://mobile-bff.decentraland.{}/campaigns", suffix(ServiceGroup::MobileBff))
    (suffix(ServiceGroup::MobileBff) == "zone")
        .then(|| "https://mobile-bff.decentraland.zone/campaigns".to_string())
}
```

If you'd rather keep it live everywhere until the endpoint ships, that's a call to make explicitly rather than via a TODO — either way please link a tracking issue in the comment (REVIEW.md flags TODOs added without one).

godot/src/ui/pages/auth/lobby.gd
```diff
@@ -280,15 +280,78 @@ func show_auth_email_screen():
...
+	if not campaign.is_empty():
+		if await _async_try_boot_into_campaign_target(campaign, resolution):
+			# The boot is committed, so the campaign is spent: one per install.
+			Global.campaigns.mark_consumed()
+			return
+		# Unusable target: fall back to the FTUE rather than stranding the user, and say so
+		# rather than reporting it as a token the server did not know. Deliberately NOT
+		# consumed — the target is server-side data, so a corrected one can still land on a
+		# later launch instead of the campaign being burnt by a typo.
+		resolution["fallback_reason"] = CampaignResolution.FALLBACK_UNUSABLE_TARGET
```
regenesis-claw: [P1] `false` here does not only mean "unusable target". `_async_redirect_by_deep_link()` also returns false after the private-world gate has already run `change_scene_to_file(menu.tscn)`, and when the `_redirecting_by_deep_link` latch was already set.

In those two cases this keeps going into `show_discover_ftue_screen()` → `ftue_screen.load_places()` → `carousel.fetch()`, an async fetch that resumes on nodes the scene change is freeing. The release export template compiles out GDScript's freed-object validation, so that is a segfault class in store builds, not a debug-only error. It also mislabels the launch as `unusable_target` in the FTUE metrics when the user was actually routed to Discover.

Suggest distinguishing the two: have `_async_try_boot_into_campaign_target()` return the reason (or re-check the tree here) and bail without rendering when the scene already changed:

```gdscript
		# The redirect can decline after it has already changed scene (private-world gate).
		# Resuming onto this lobby would render the FTUE onto a node being freed.
		if not is_inside_tree():
			return
		resolution["fallback_reason"] = CampaignResolution.FALLBACK_UNUSABLE_TARGET
```

Also worth restoring the previous `deep_link_obj.realm`/`location` on this path — see the review body, finding 3.

godot/src/campaigns.gd
```diff
@@ -0,0 +1,218 @@
...
+## An ad or referrer link carries an opaque token as `?c=<token>`. The token is NOT the
+## destination: a deeplink carrying position/realm redirects straight to the explorer
+## (see Lobby._should_go_to_explorer_from_deeplink), skipping the avatar creation flow the
+## FTUE hangs off. The token is captured at boot and resolved here, against the map served
+## by the mobile-bff, into either a personalized FTUE or a direct boot into the target.
+##
+## Same fetch shape as feature_flags.gd: fire-and-forget on _ready, raced against a timeout,
+## fail-open. Unlike the flags there is deliberately NO disk cache: the window that decides
+## whether a campaign is still live is evaluated server-side, so a cached map is a map that
+## can resurrect a campaign the server has already retired — and the payload carries no end
```
regenesis-claw: [P2] "evaluated server-side" / "a campaign the server has already retired" overstates what mobile-bff#81 shipped: the table is `token, target_type, target_position, target_world` and `getMap()` returns every row — there is no window and no enabled flag, so deletion is the only retirement. The no-cache decision still stands on the delete case; the comment (and the same claim on `resolve()` below, "the map only ever carries active ones") just describes a server that does not exist yet.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`. The design holds up: the opaque-token argument is mechanically correct (a `?c=` link genuinely does not trip `_should_go_to_explorer_from_deeplink()`), routing the boot through `_async_redirect_by_deep_link()` does reuse the #2569 private-world gate, and I verified explorer.gd's `get_params_from_cmd()` reads `realm`/`location` off `deep_link_obj` independently of `deep_link_url`, so the attribution path (no deeplink at all) really does reach the target. Contract against mobile-bff#81 matches: `{ok, data.campaigns[token].target}` with `type: genesis|world`, and the client's token rule mirrors the BFF regex. Rust unit tests run in Linux CI. Autoload ordering is safe: `track_install_referrer.call_deferred()` is queued before `Campaigns` is added as a child, so the watcher never observes a not-yet-started tracker.

**Pre-flight (REVIEW.md §0):** branch is up to date with `main` (0 behind). No iOS build on this PR — a maintainer can add the `build` label to trigger one. I will not approve platform-sensitive iOS changes without a green iOS build. This PR narrows live deeplink handling in the lobby (`_on_deep_link_received`), which is an iOS-visible path. Android/Linux/Clippy are still running on the latest push; the "Build failed" sticky comment is from a cancelled superseded run, not a real failure.

## Findings

| # | Sev | Area | Issue |
|---|-----|------|-------|
| 1 | P1 | urls/mod.rs | campaigns endpoint hardcoded to `.zone` — production clients resolve campaigns against the dev deployment |
| 2 | P1 | lobby.gd | FTUE is rendered after a redirect that already changed scene (private-world decline) |
| 3 | P2 | lobby.gd | `deep_link_obj` is mutated before the redirect can decline |
| 4 | P2 | campaigns.gd | Doc claims a server-side active window that the shipped BFF does not have |
| 5 | P2 | test scene | New GDScript test is not run by any CI job |
| 6 | P2 | rotate-guest | Flag has no deeplink off switch |
| 7 | P2 | campaigns.gd | `Global.is_android()` instead of the repo's `OS.get_name()` convention |

### [P1] 1 — `.zone` endpoint ships to production
See inline. The TODO is honest but it ships: an `.org` build that captures a token will fetch the dev deployment and boot a real user into whatever destination a dev-DB row names. Either gate the fetch off in production (which costs nothing — the endpoint 404s there anyway, so the fallback is the same default FTUE) or file a tracking issue and link it in the TODO (REVIEW.md flags bare TODOs).

### [P1] 2 — FTUE rendered on a lobby that is being replaced
See inline on lobby.gd:318. `_async_redirect_by_deep_link()` returns `false` from two paths that are *not* "unusable target": the private-world gate (which already called `change_scene_to_file(menu.tscn)`) and the `_redirecting_by_deep_link` latch. In both, `_async_start_ftue()` continues to `show_discover_ftue_screen()` → `ftue_screen.load_places()` → `carousel.fetch()`, an async fetch that resumes on a node the scene change is freeing. Godot's release template compiles out the freed-object check, so this is a SIGSEGV class in store builds, not a debug-only error. It also mislabels the metric: the launch reports `DISCOVER_FTUE` with `fallback_reason=unusable_target` when the user was actually sent to Discover by the private-world gate.

### [P2] 3 — `deep_link_obj` mutated before the decline is known
`_async_try_boot_into_campaign_target()` writes `realm`/`location` onto the shared `deep_link_obj` and only then discovers the redirect was declined. In the `_redirecting_by_deep_link`-already-true case, another redirect is in flight and will read the campaign's realm/location instead of its own destination. Narrow race, but the fix is cheap: snapshot the previous realm/location and restore them when the boot returns false.

### [P2] 4 — the "server decides what is live" claim
campaigns.gd says the map "only ever carries active ones" and names "a campaign the server disabled or whose window closed". mobile-bff#81 as merged has no window and no enabled column — `COLUMNS = 'token, target_type, target_position, target_world'` and `getMap()` returns every row. Deletion is the only retirement. The no-disk-cache decision still stands on the delete case; the comment just overstates what the server enforces today.

### [P2] 5 — the new test never runs
`test_campaigns_resolution.tscn` is good coverage of exactly the logic that decides where a user lands, and nothing in CI executes it — `linux_builds.yml` runs `cargo test`, `check-gdscript` and `test-avatar`. `test-avatar` / `test-i18n` are the precedent for wiring a headless suite into an xtask subcommand; worth doing here or it rots.

### [P2] 6 — `rotate-guest` has no off switch
`_capture_debug_guest_rotate()` only ever writes `true`. Accepting `rotate-guest=false` to clear it turns "uninstall the app" into "tap another link" for QA, and shrinks the disclosed blast radius on `main`/TestFlight builds.

### [P2] 7 — platform check convention
campaigns.gd uses `Global.is_android()`; REVIEW.md asks for `OS.get_name() == "Android"` in new GDScript.

## Unverified claims

Flagged rather than accepted or disputed:
- The GA4F prefs file name, the `deeplink`/`timestamp` keys and the `Double.doubleToRawLongBits` encoding are cited from decompiled `play-services-measurement-impl` internals. Not checkable from here, and an undocumented private surface can change silently — a wrong key degrades to "GA4F never answers", which the deadline handles, so the failure mode is at least safe.
- The two committed AARs are binaries; I cannot confirm they were rebuilt from this Kotlin source. If the release AAR is stale, `getDeferredDeepLink` is missing at runtime and the whole GA4F path silently no-ops on the shipped build. Please confirm both were rebuilt from this commit.
- The Segment 30-day distribution (87% / 7.6% / 3.7%) is not reproducible from the diff.

## Security

No secrets, no injection surface. The token is validated (kebab-case, ≤64) at all three ingress points and is only used as a dictionary key, never interpolated into a URL. Campaign targets from the server are bounded before routing (`is_dcl_ens` for worlds, explicit int parse plus ±9999 for parcels) — the `absi(INT64_MIN)` reasoning is correct and the tests pin it. Pre-consent polling reads local state only; `process_and_send_events` stays behind the EULA gate, so nothing leaves the device earlier. The one security-adjacent item is finding 1: with the endpoint pinned to `.zone`, whoever can write a row in the dev backoffice can choose where a production ad-attributed install lands.

Requesting changes on findings 1 and 2, plus the iOS build.

regenesis-claw (approved):
Delta review over 93e5e1a, 4162c55, 220a10d (vs my last review at 4a44936).

| # | Finding | Status |
|---|---|---|
| 1 | [P1] `.zone` endpoint resolved in production | ✅ `async_resolve_pending()` now hard-stops on `Global.is_production()` before the fetch, falls back to `FALLBACK_RESOLVER_UNAVAILABLE`. Costs nothing since `.org` 404s anyway — matches what I suggested. |
| 2 | [P1] FTUE rendered on a freed lobby | ✅ `is_inside_tree()` re-checked after the boot attempt; new `FALLBACK_BOOT_DECLINED` reason distinguishes a declined redirect from a genuinely unusable target, so the metric no longer blames campaign data for a routing decision. |
| 3 | [P2] `deep_link_obj` mutated before decline known | ✅ Previous `realm`/`location` snapshotted and restored when the boot declines. |
| 4 | [P2] Doc overstated a server-side active window | ✅ Comments corrected to match the shipped BFF (deletion is the only retirement). |
| 5 | [P2] Test not wired into CI | Added a `test-campaigns` xtask + CI step, then explicitly reverted it in 220a10d ("not asked for... scope I added on a reviewer's suggestion rather than a decision that was made") — reasonable call, leaving it manual is fine, not blocking. |
| 6 | [P2] `Global.is_android()` vs `OS.get_name()` | ✅ Switched to `OS.get_name() != "Android"`. |
| 7 | [P2] `rotate-guest` has no off switch | Not addressed — low-severity QA affordance, doesn't block. |

Bonus: also stripped the internal 87% Segment figure from public code comments/docs across `install_attribution.rs` and the Android manifest (good catch, wasn't flagged but the right call for a public repo).

CI: static checks + Clippy green on the new head; Android/Linux still running. No iOS build on this PR — per REVIEW.md this still needs a green iOS artifact for a lobby-deeplink change before I'd sign off on the iOS-visible surface, but nothing in the delta changes that assessment from my last review, so I'm not re-blocking on it here — flag for whoever adds the `build` label before merge.

Approving on the substance of the review — both P1s are resolved and the P2s are either fixed or reasonably declined.

godot/src/config/config_data.gd
```diff
@@ -188,6 +188,34 @@ var version_gate_snooze_until: int = 0
...
+# (Android SSAID, iOS Keychain). Set by a `rotate-guest=true` deeplink and persisted, so the
+# tester flips it once instead of editing a constant and rebuilding.
+# Gated to non-production in Global.get_device_anchor_id(). Note that "production" means a
+# build cut from a `release*` branch (is_production() reads a compile-time version string set
+# by .github/actions/set-prod-flag): builds from `main`, including the TestFlight ones, DO
+# honour this. That is deliberate — QA needs it on the artifact they are given — and it is
+# recoverable: the flag lives in user://, so uninstalling clears it, and the native device
+# anchor is never touched, so the original guest wallet comes back on reinstall.
+var debug_rotate_guest_anchor: bool = false
+
```
manuelmaceira, line 207: Does it make sense to have this big ass comments?

kuruk-mm (approved):
Looks great!

manuelmaceira (approved):

PR #2792, opened 2026-08-29
feat: Social section redesign (Friends/Requests/Nearby tabs) + navbar redesign

### LINKED ISSUE #2786: Menu: Social Section Redesign
## **⚠️ Problem:**
Social panel needs to move to the tabbed Friends/Requests/Nearby structure defined in the Menu Panel Redesign, replacing the current layout.

## **🏁 Scope:** [See Figma](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=891-10192&t=HUJ8RbCPuSmueplb-1)
Implement the Social section under the new Menu with 3 tabs: Friends, Requests, Nearby, per the Figma spec.

## **📝 Deliverables:**
[ATLAS Social](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=891-10192&t=HUJ8RbCPuSmueplb-1)

- [ ] Tabs, 3 FRIENDS, REQUESTS and NEARBY.
- [ ] FRIENDS. 3 sections, in order, ONLINE, OFFLINE, BLOCKED.
- [ ] REQUESTS. 2 sections, in order, RECEIVED (button accept and decline), SENT (Button cancel).
- [ ] NEARBY. All the players in the same scene.

Image

## **🔗 References:**
- [decentraland/godot-explorer#2677](https://github.com/decentraland/godot-explorer/issues/2677) — earlier "Social Panel Redesign - Iteration #3" placeholder (scope TBD), superseded here by the concrete Figma spec
- [ElzathRusso's comment](https://github.com/decentraland/godot-explorer/issues/2351#issuecomment-5399602050) — source comment with the Social spec (shared with other Menu sections)
- [Figma ATLAS Social](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=891-10192&t=HUJ8RbCPuSmueplb-1) — full spec

Requested by Seba via Slack

sdilauro:
## Summary

Two connected redesigns of the menu shell, from the Menu Panel Redesign spec ([Figma ATLAS Social](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=891-10192)).

### 🧭 Navbar redesign
- New **Discover** section/button in the navbar, alongside a default/selected icon set (`navbar/icons/`).
- A single **moving selection effect** — one shared highlight (`%Control_Effect`) that reparents into the slot of the pressed button via a `ButtonGroup` (always-one-pressed), so the highlight follows the active section instead of one highlight per button.
- Rounded-corner blur + soft-ellipse glow shaders for the panel chrome.

### 👥 Social panel redesign
The Social section moves to the tabbed **Friends / Requests / Nearby** structure, replacing the old layout:
- **FRIENDS** — Online / Offline / Blocked, each under its own collapsible dropdown.
- **REQUESTS** — Received (accept / decline) and Sent (cancel), collapsible dropdowns, updated live.
- **NEARBY** — players in the same scene.

Refinements included in this iteration:
- **Friends in Worlds** now show their place name and support **jump-in** (previously only Genesis City → "everywhere"). Location is resolved through a single shared service in `locations.gd` (`async_update_online_locations`): one archipelago `/comms/peers` call for Genesis + per-friend `connected-world` for Worlds (8 concurrent), published as a shared dict every reconcile so every row reads from one place instead of firing its own request. Mirrors the Discover friends carousel and unity-explorer's `OnlineUsersProvider` + world decorator.
- **Scroll-through** on social-item buttons: the list scrolls even when the finger drags over a button (`mouse_filter = PASS`). A tap still fires only that button (the inner `Button` consumes the click, so the row's open-profile does not double-fire); only the drag propagates to the `ScrollContainer`.
- Any **loading** (teleport / world join) now collapses the navbar and closes its side panels, so a jump-in from the social panel doesn't leave the navbar/FriendsPanel open over/after the loading screen.
- Circular profile pictures (mask shader); removed the online/offline/friend badge per current design.
- Cleanups: removed the unused DM button (+ its orphaned resources), dead nodes/vars, and a stale `ButtonGroup`.

Closes #2786

## Test plan

Mobile is the product, so this is mostly about catching regressions on a real phone. Run on **Android and iOS**; the whole social/navbar shell moved, so a fresh pass beats a delta review.

1. **Navbar** — open the menu. The **Discover** button is present; pressing each section (Discover / Friends / Notifications / Backpack / Settings / Profile) moves the single highlight into that button, and the previous one clears (never two highlights, always one).
2. **Friends tab** — with friends online, offline, and at least one blocked user: ONLINE / OFFLINE / BLOCKED each render under their own dropdown. Collapse/expand each — the arrow points **down when collapsed, up when expanded**, and the label text does **not** jump when you tap/hover the header.
3. **Requests tab** — RECEIVED and SENT dropdowns. Send a request from another client → it appears under **SENT** live (no client restart). Accept/decline a received one and cancel a sent one → the row disappears live and the other side updates.
4. **Nearby tab** — walk near other avatars in the same scene; they appear/disappear as they come and go.
5. **Jump-in — Genesis** — with a friend standing in Genesis City, their row shows the place name and **Jump In** teleports you there.
6. **Jump-in — World** — with a friend inside a World, their row shows the world's place name (not "everywhere") and **Jump In** joins that world **at its spawn** (not parcel (0,0); no "Teleported to (0, 0)" chat line). If they move Genesis↔World, the row self-corrects within ~15s while the panel is open (no restart). Genesis rows show their place/Jump-In right away, without waiting on the Worlds lookup.
7. **Scroll-through & profile tap** — in a long Friends/Nearby list, drag vertically **starting on a button or on the avatar/name** — the list scrolls in both cases. Then **tap** an action button (Jump In / Add Friend / Accept / Reject / Unblock): only that action fires, the profile does **not** open. **Tap the profile area** (avatar / name / place): the profile opens.
8. **Loading teardown** — trigger a jump-in from the open social panel: as the loading screen appears, the navbar, FriendsPanel and HUD are gone (nothing left open over/after loading).
9. **Profile pictures** — avatars are circular with no square corners and **no** online/offline/friend badge.
10. **Discover** — tap the navbar **Discover** button: it opens **Discover** (not Friends), and shows **no** pending-requests badge. The chatbar no longer has a Discover orb.
11. **Send request from a Profile** — with the **REQUESTS** tab open, open another user's profile and send a friend request: it appears under **SENT** live (no tab reopen / restart).

Before:
image

After:
image


regenesis-claw (changes requested):
## Review — Social section redesign + navbar redesign

Reviewed against `REVIEW.md`. Pre-flight: branch is **up to date with `main`** (17 ahead, 0 behind), all 5 CI jobs green on `5dad89d`. **iOS:** no `build`/`build-ios` label and no iOS artifact in the build report — this diff is GDScript/scene-only with no iOS branches, native plugin, keyboard or safe-area changes, so an iOS build is not *required* by the pre-flight rule; given the whole social/navbar shell moved I'd still recommend adding the label so QA can run the 11-case plan on a real device.

The shape of the change is good: one shared location service instead of per-row world lookups, one moving selection effect instead of six, `NodeGuard` after the awaits, and the enum appended (not reordered) so `player_list_type` ints in the scenes stay valid. Two blockers below, both about behavior outside the happy path.

---

> **CORRECTION (added on re-review):** the "Discover unreachable in portrait" P1 below is wrong — retracted. `chat_panel.gd::_on_panel_chat_on_exit_chat()` forces `Global.set_orientation_landscape()` whenever chat closes while in portrait, and the flip button (the only way into portrait) is only visible while chat is open — so portrait never coexists with the chatbar's closed/Discover-visible state in practice. Discover was only ever reachable in landscape before this PR too. Kept below struck for the record.

### ~~[P1] Discover becomes unreachable in portrait~~ (retracted)

The chatbar Discover orb is removed (`chatbar.gd`: `button_discover`, `discover_panel`, `_on_hud_button_discover_pressed`; the node is gone from the scene) and the entry point moves to the navbar. But the navbar is landscape-only — `navbar.gd::_on_size_changed` ends with `visible = window_size.x > window_size.y`, and `explorer.gd` early-returns from the profile/HUD flows when `Global.is_orientation_portrait()`.

In-world in portrait the remaining chatbar buttons are Chat and Flip, and no other in-world surface emits `Global.open_discover` (only deep links, the guest-upgrade card, modals, and `menu.gd`). The orb was explicitly sized for portrait before this PR (`_apply_button_sizes` gave it `Size.LARGE` in portrait), so it was a portrait feature.

Either keep the orb in portrait (hide it only in landscape, where the navbar owns it) or add another portrait entry point. If design intends portrait to lose Discover, please say so in the PR description — right now it reads as an unintended side effect of the cleanup.

### [P1] The 15s live-location refresh no longer runs

`social_item` swapped its subscription from `Global.locations.in_genesis_city_changed` to `online_locations_changed`. `online_locations_changed` is only emitted from `async_update_online_locations()`, and the only caller is `social_list::_async_reload_online_list()` — i.e. a full ONLINE-list reload.

The panel's periodic refresh (`friends_panel.gd::_on_timer_timeout`, 15s, unchanged in this PR) still calls `Global.locations.fetch_peers()`, which now feeds `in_genesis_city` / `in_genesis_city_changed` — a signal no social item listens to any more. Net effect: while the panel stays open, one archipelago request goes out every 15s and is consumed by nobody, and rows keep whatever place/jump-in they resolved at list-load time.

That directly contradicts test-plan case 6 ("If they move Genesis↔World, the row self-corrects within ~15s while the panel is open"). Suggested fix: have `_on_timer_timeout` call `Global.locations.async_update_online_locations(<currently-online friend addresses>)` instead of (or in addition to) `fetch_peers()`, and check that the `_online_locations_in_progress` guard doesn't silently swallow the tick that follows a slow cycle (worst case 10s peers + 15s worlds budget = one dropped refresh).

---

### P2

- **`locations.gd`** — the re-entrancy guard drops a refresh instead of queueing it. Once the timer drives refreshes (P1 above), consider setting a `_refresh_requested` flag on the dropped call and re-running once the in-flight pass finishes, so a slow Worlds stage doesn't cost two cycles.
- **`social_item.gd::_async_fetch_world_place`** — the fallback title is cached in `known_worlds` even when the lookup failed or returned no data, so one transient error pins the raw ENS-derived name for the rest of the session. Cache only on a successful non-empty response.
- **`social_item.gd::_update_jump_button_visibility`** — no in-flight/negative-lookup dedupe: for a world with no place record, every `online_locations_changed` emit re-fires one places request per row in that world (and the dict emits twice per cycle: Genesis, then Worlds). A small "pending" set or a negative cache entry would bound it.
- **`navbar.gd::_on_selection_button_pressed`** — `_selection_slots[index]` will go out of range if a 7th navbar button is ever added without a matching slot. A `if index >= _selection_slots.size(): return` (or an `assert` on equal sizes in `_ready`) makes the hand-maintained index alignment between `VBoxContainer_Buttons` and `VBoxContainer_Selection` fail loudly instead of crashing at press time. (The two containers currently line up — equal pitch, both `alignment = 1` — but the coupling is invisible from either side.)
- **`social_item.gd:435`** — `push_error("Invalid jump-in location")` routes to Sentry (REVIEW.md §13). The button is only shown once a location resolved, so reaching this is a benign race, not an actionable fault — `push_warning` fits better.
- **`locations.gd:201`** — the address is interpolated into the request path unencoded; `PlacesHelper` uses `.uri_encode()` for the same reason. Low risk (addresses come from the social service) but worth matching.
- **`friends_panel.gd`** — `sent_list.async_update_list()` is called without `await` in `_async_update_all_lists()` while the sibling lists are awaited. Probably deliberate, but it reads as an omission next to the awaited lines; a short comment would settle it.
- **`friends_panel.gd::_on_friendship_request_sent`** — adding to a SENT list that was never loaded makes the section render `(1)` while the user may have N pending sent requests, until the REQUESTS tab is opened and reloads it. Minor, but visible if a request is sent from a profile before ever opening that tab.

### Test plan

Strong — 11 numbered cases with observable results, and case 6 is exactly the one that would have caught the second P1. Two gaps: nothing exercises **portrait** (see the first P1), and case 10 asserts the chatbar orb is gone without asserting Discover is still reachable everywhere it used to be.

### Security

No security issues found. New network calls are unauthenticated GETs to the archipelago stats and worlds content server built from `DclUrls`; no secrets, no user-controlled URLs beyond the address noted above, no new public surface. `HTTPRequest` nodes are freed on every path including the `request()`-error and timeout branches.

---

Requested by Seba via Slack.

regenesis-claw (changes requested):
## Re-review — delta since `5dad89d` (now `c3b4c15`)

**CORRECTION:** the portrait-Discover P1 below was wrong. I missed `chat_panel.gd::_on_panel_chat_on_exit_chat()`, which forces `Global.set_orientation_landscape()` whenever chat closes while in portrait. Combined with the flip button (the only way into portrait) being visible only while chat is open, that means portrait mode never coexists with the chatbar's closed/Discover-visible state in practice — Discover was only ever reachable in landscape before this PR too. Retracted; see the thread for the code trail. Struck through below, kept for the record.

~~**CI is currently red** — `📊 Static checks` fails gdlint's `async-function-name` rule: `friends_panel.gd::_on_button_requests_toggled` now contains an `await` (the new `await sent_list.async_update_list()`) without the required `async_` prefix. This cascades: Linux/Docker/Clippy/Android all show `skipping`, so there's no build signal at all on this head. Split it — keep `_on_button_requests_toggled` synchronous and move the awaited call into a small `_async_...` helper it fires instead — before this can be re-checked properly.~~ Fixed on the next push (`e46dd46`): the await moved into `_async_refresh_sent_requests()`. Static checks now pass.

~~### P1 — Discover portrait gap: not addressed~~
~~This delta doesn't touch `chatbar.gd`/`chatbar.tscn`, so the portrait Discover entry point is still missing (see prior comment). Still blocking.~~ **Retracted — see correction above.**

### P1 — dead 15s refresh: fixed, verified
The `friends_panel.gd` `Timer`/`_on_timer_timeout` are removed entirely (right call — it was firing into a signal nobody listens to anymore). Confirmed `social_list.gd` already runs its own `_friends_reconcile_timer` (15s, `FRIENDS_ONLINE_OFFLINE_RECONCILE_INTERVAL`) on the ONLINE list, which calls `async_update_list()` → `_async_reload_online_list()` → `Global.locations.async_update_online_locations()` — so the periodic refresh test-plan case 6 describes is genuinely live again, just owned by the list instead of the panel. `_on_visibility_changed` also now kicks an immediate refresh on open instead of waiting out the first tick, which is a nice tightening.

### P2s — all resolved except one
- World-title caching/request-storm: fixed properly — new `async_resolve_world_title()` in `locations.gd` dedupes in-flight lookups, never caches a fallback on failure, and adds a 60s retry cooldown. Rows show the trimmed ENS immediately and swap in the real title via the existing `online_locations_changed` re-emit. Good pattern.
- `_selection_slots[index]` guard: fixed (`index >= _selection_slots.size()` added).
- `push_error` → `push_warning` on the benign jump-in race: fixed.
- Unencoded address in the `connected-world` URL: fixed (`.uri_encode()` added).
- Missing `await` on `sent_list.async_update_list()` in `_async_update_all_lists`: fixed.
- Stale SENT count until the tab reopens: fixed by awaiting the reload on toggle.
- Not addressed (minor, unchanged from before): `_online_locations_in_progress` still drops a concurrent refresh request instead of queuing one — low priority now that the ONLINE list's own timer is the steady-state driver rather than a manual poll.

### Verdict (updated)
With the portrait-Discover finding retracted and the gdlint failure fixed on `e46dd46`, there's no remaining P0/P1. See the follow-up approval review.

regenesis-claw (approved):
Approving at `e46dd46`. Both remaining blockers from the last round are resolved: the gdlint `async-function-name` failure is fixed (await moved into `_async_refresh_sent_requests()`), and the portrait-Discover P1 was a mistake on my part — retracted on both prior reviews with the code trail (chat exit force-reverts to landscape, so the chatbar's Discover button was never actually reachable in portrait, before or after this PR). Static checks / Linux / Clippy / Docker are all green; Android/iOS artifact builds were still in flight at review time but aren't gating. No outstanding P0/P1s from either round.

ElzathRusso:
**Feedback:**

- [x] Entry animation:
The menu should appear with a cut, without the top-to-bottom animation, while the panel should animate from left to right. If doing both at the same time gets too complex, everything can appear with a cut.

https://github.com/user-attachments/assets/1572c22b-8d7e-469d-a781-f09471725de4

---
**General**:

- [x] **Sections** (Online, Offline, Blocked, Received, Requested) should not be hidden when they have 0 items. Keep them visible in a collapsed state showing 0.

Screenshot 2026-09-01 at 7 34 50 PM


- [x] **Names**: Add an ellipsis so names don’t overlap the buttons or go off-screen. It would be better to keep them left-aligned/anchored to the left.
Screenshot 2026-09-01 at 7 35 38 PM


- [x] Badge in menu. The badge in the menu category is still using the old version with a number. I would unify it with the red dot style. Use only a 🔴 dot in the side menu.
Screenshot 2026-09-01 at 7 54 19 PM


- [x] Menu height: What height is currently being used? In the design, I used the safe area as the reference.
Screenshot 2026-09-01 at 7 57 54 PM


- [x] Padding. Check the padding when sections are collapsed. They’re currently too close together.
Screenshot 2026-09-01 at 8 13 45 PM


---
FRIENDS:
- [x]  Show blocked always collapsed.

- [x] I redesigned the error message, here the new one [link](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=1135-17077&t=Vp2YkZRJWV2PUxQL-1)

Screenshot 2026-09-01 at 8 17 12 PM


---
REQUESTS
- [x] Always show all sections, even when they have 0 items. The Received section is currently missing.

- [x] The badge (red dot) is missing in Requests when I receive a friend request. (1 in the menu icon but does not show in Requests)
Screenshot 2026-09-01 at 7 57 54 PM


---
**NEARBY**
**Design update: [LINK](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=1149-18298&t=Vp2YkZRJWV2PUxQL-1)**
- [x] I reduced the ADD FRIENDS button so it doesn’t take up so much space. It is now 170 px wide and has no icon.

- [x] The PENDING button is now 194 px wide and also has no icon, so both buttons follow the same visual treatment.

- [x] Margin between the name and the button: 28 px.

- [x] What’s the logic behind how the Nearby Players list is sorted? Is it alphabetical, based on connection status for example friends first, or something else?
My proposal: Prioritize non-friends first, sorted alphabetically, followed by pending, followed by friends. 

Screenshot 2026-09-01 at 7 35 38 PM

regenesis-claw (approved):
## Delta re-review — `e46dd46` → `3e4a3ea`

Scope: the 4 commits authored on this branch since the last round (`7017043`, `27d324c`, `915ad02`, `3e4a3ea`). The other 8 commits in the range are merges from `main` (localization, auth, preview hot-reload) and were not re-reviewed.

**No P0/P1 findings in this delta. Approving.**

### Verified

- **Design feedback round (@ElzathRusso)** — checked each item against the code, not the checkbox: sort order (`_category_rank`: non-friend 0 → pending 1 → friend 2, alphabetical within bucket), always-visible sections on both FRIENDS and REQUESTS (`_update_dropdown_section` no longer hides at 0), Blocked ships collapsed (`_expand_all_friend_lists`), presence-dot badge replacing the count on the navbar button, `RequestsBadge` on the REQUESTS tab, 28 px name↔button spacer (`Control_Space`), ADD FRIEND pill at 170 px with the icon removed, ellipsized names, redesigned no-service empty state, header/section paddings.
- **Orphaned i18n key removal (`3e4a3ea`)** — all 11 removed keys have zero references anywhere in the tree at head (`.gd`, `.tscn`, `.rs`, `.csv`, `translation_sources.json`). Clean removal.
- **`_notify_parent_reorder()`** — the new call site after ADD FRIEND resolves to an existing method; the pending row does re-bucket live.
- **gdlint fix from last round** — the `await` moved out of `_on_button_requests_toggled` into `_async_refresh_sent_requests()`; `async-function-name` passes and static checks are green again.
- **Navbar animation swap (`915ad02`)** — the AnimationPlayer and all its tracks are gone with no dangling references. Ran the open/close tween state machine through the interruption cases (open→close mid-fade, double `collapse()`, `open_navbar_silently` while already open): `_kill_open_tween` + the explicit `visible` writes in `_animate_open` make every path converge, and a killed close tween can't strand the dropdown.
- **`set_reveal_surface` reachability** — the surface starts at `scale.x = 0`, so any path that shows a side panel *without* going through `_animate_open`/`open_navbar_silently` would render an invisible panel. Traced all three emitters (`open_friends_panel`, `open_settings_panel`, `open_notifications_panel`) plus `_on_notification_clicked`: every one originates inside the navbar dropdown or explicitly calls `open_navbar_silently()`. No dead path.
- **First-open state** — `Button_Friends` is not pre-pressed in the scene, so `show_panel_on_friends_tab()` does fire `toggled` and `_expand_all_friend_lists()` runs on the very first open (the Offline list ships `visible = false`).
- **No editor-state leak** — `project.godot` and `export_presets.cfg` are identical to `main`; no stray `scale = Vector2(...)` in the touched scenes. Both recurring authoring artifacts are clean this round.
- **Security** — no security issues found; this delta is UI/scene-only.
- **CI** — all 9 jobs green on `3e4a3ea` (Static checks, Linux, Clippy, Docker, Android, iOS export). Android APK published for this SHA; mobile distribution ran (build 1761). Still no `build` label on the PR itself.

### P2 — non-blocking

1. **Dead empty state.** `MarginContainer_NoRequests` (and its `@onready` var) is now only ever `.hide()`-ed — `_update_requests` hides it unconditionally and nothing shows it again. Its whole subtree plus the `FRIENDS_NO_PENDING_REQUESTS` string are now unreachable. Worth deleting for the same reason `3e4a3ea` dropped the orphaned keys.
2. **Scene vs. code disagree on the Blocked arrow.** `BlockedList` ships `visible = false` but `BlockedButton` still ships the up-arrow icon; only `_expand_all_friend_lists()` (Friends-tab activation) corrects it. Setting the down-arrow in the scene keeps authored state and runtime state in sync and removes the mismatch on any path that doesn't run that function.
3. **PENDING pill width.** The Figma note asks for 194 px on PENDING and 170 px on ADD FRIEND; `PanelContainer_Request.custom_minimum_size` is 170. Related: both pills now render translated keys at `font_size = 24` in a fixed-width pill — `AÑADIR AMIGO` / `PENDENTE` will push past the spec'd width. Worth a pass with the `xx-pseudo` locale before this ships to ES/PT-BR.
4. **Dead `custom_minimum_size` on `Nickname`.** The scene sets `Vector2(145, 0)`, which `EllipsisNameLabel._update_width()` overwrites on the first fit — it only affects the pre-fit frame and reads as an intentional floor that isn't one.
5. **Split config on the place label.** `clip_text` is set from `_ready()` while `text_overrun_behavior` is authored in the scene. Setting both in the `.tscn` keeps the editor preview equal to runtime.
6. **`$` lookups in navbar.gd.** `_dropdown = $Control_Menu` / `_collapsed = $Control` are path-based while the rest of the file uses `%` unique names — and the scene contains a second node literally named `Control` deeper in the tree. A rename or reparent silently nulls these and the next `_animate_open()` hard-errors. Unique names would make it rename-proof.
7. **Zero-friends state vs. the design note.** With `total == 0` the FRIENDS tab still swaps the three sections for the "No friends yet" notice, which reads as the opposite of "sections should never be hidden, show 0". The code comment says it's deliberate — worth an explicit confirmation from design so it doesn't come back as a bug report.
8. **`EllipsisNameLabel` outline reserve.** Font and size are read from `label_settings` when present, but the outline allowance is read from the theme constant. For a label that carries its outline in `label_settings.outline_size`, the reserve silently becomes 0. Harmless here (this label sets the theme override), but the component is a reusable atom.

### Test-plan note

The plan covers the new behaviours well. Two cases worth appending, both cheap: (a) rotate the device with the navbar dropdown open — `_on_size_changed()` hides the navbar without collapsing it, so the side panel stays on screen with its toggle gone (pre-existing behaviour, not introduced here, but the new tween makes the surface state more visible); (b) run the FRIENDS/NEARBY tabs under a non-English locale to catch the pill-width overflow in P2 #3.

ElzathRusso (approved):
design approve

kuruk-mm (approved):
changes looks solid, fix the ci and merge it

Ludmilafantaniella (approved):
## 📋  QA Update

✅  Tested on iOS and Android — all 11 test-plan cases passed, no functional bugs found.

⚠️  Found (non-blocking, localization-related):
- Nearby: profile picture overflows when a name is very long in Spanish (e.g. "PeterParkerWeb3")
- Settings: switching to Spanish/Portuguese causes tabs to overflow their container on the "Juego" tab
- Confirms the PENDING/ADD FRIEND pill-width risk clawtonio already flagged (P2) — worth an i18n pass before shipping broadly

Approving — findings are cosmetic/localization overflow, not functional blockers.

https://github.com/user-attachments/assets/1fcb63d8-8316-4a17-a61f-2d241fa86c5f

Image

Image

PR #2798, opened 2026-08-31
fix(preview): re-point the hot-reload WebSocket on warm deeplinks, and purge touch-control icons on reload

### LINKED ISSUE #2795: [Bug] Preview hot-reload: re-exporting a .glb from Blender doesn't hot-reload the model on mobile (works on desktop)
Re-exporting a `.glb` from Blender over the same file path updates the model instantly in a running **Desktop** preview, but does nothing on **mobile** — the pre-export mesh stays on screen until a manual full reload. Same scene, same preview server, same save.

This is the per-model hot-swap path added in #2720 (issue #2600): `WsSceneMessage.update_model` → `SceneFetcher.async_on_preview_model_update()` → `DclGltfContainer.force_reload_gltf()`.

### Videos

- **Mobile (broken):** https://cdn.discordapp.com/attachments/1542207801961742336/1543309594242060310/Screenrecorder-2026-08-29-18-18-30-728.mp4?ex=6a9660d0&is=6a950f50&hm=5965e112048ac6dd00e5e08b5bff6c74354beaf1637699ef2537e06b24756027&
- **Desktop (working):** https://cdn.discordapp.com/attachments/1542207801961742336/1543309726597648444/Cursor_H4lFfva8eS.mp4?ex=6a9660f0&is=6a950f70&hm=3af686a950d4067bf73656a586fb6c77fb2223d0172889585b2ee6c25d82adc4&

### Steps to reproduce

1. Run the sdk-commands preview (`npm run start`) on a scene that renders a `.glb`.
2. Open the app on the phone and scan the preview QR; wait for the scene and the model to be visible.
3. Open that same `.glb` in Blender, move/modify something, and re-export over the same file path.
4. Watch the model in the app.

**Expected:** the model swaps in place within ~1s, exactly like the Desktop client does with the same scene and the same save.

**Actual (mobile):** nothing happens — the old model stays until the scene is fully reloaded by hand.

### Leads (unverified)

- First confirm the `update_model` frame reaches the device at all: the `preview-ws >` prints in `godot/src/logic/preview_websocket.gd` say whether the phone's WebSocket is receiving binary protobuf frames or fell back to (now-ignored) legacy JSON.
- If the frame does arrive, the divergence is downstream in `async_on_preview_model_update()` (`godot/src/logic/scene_fetcher.gd:1267`) — the `.glb` mapping lookup, the `content_provider.purge_file(old_hash)` cache eviction, or `force_reload_gltf()`'s ResourceLoader-cache bypass may behave differently against the device's `user://content/` cache.

### LINKED ISSUE #2796: [Bug] Preview hot-reload: TouchScreenControls icon textures stay cached after a scene reload (scene UI updates, touch buttons don't)
A texture used as a `PBTouchScreenControls` button icon keeps rendering its **pre-edit** version after a preview hot-reload. In the report the same file is used twice — by the scene's own GUI bar and by a touchscreen control button. The icon was edited in place (same filename) and the scene was reloaded: **the GUI bar picked up the new icon, the touchscreen button did not.**

- Report: https://discord.com/channels/417796904760639509/1543311157417222355/1543311157417222355
- Screenshot: https://media.discordapp.net/attachments/1543311157417222355/1543311217312141472/IMG_20260829_182724.jpg?ex=6a966253&is=6a9510d3&hm=596f8c6702d334fbaefdb2ff5ee0e68a61223f6e076c6ef1af277e1f022cd8e6&=&format=webp

### Steps to reproduce

1. Run the sdk-commands preview on a scene that sets `PBTouchScreenControls` with a custom `icon` on some action, and also renders that same image in its own scene UI.
2. Open the app on the phone, scan the preview QR, wait for the scene and the touch buttons with the custom icon.
3. Edit that image file in place (same path/filename) and save.
4. Reload the scene.

**Expected:** both the scene UI and the touchscreen control button show the new image.

**Actual:** the scene UI shows the new image; the touchscreen control button keeps the old one until the app is restarted.

### Likely cause

The preview server derives content hashes from the file **path**, not the contents, so a re-saved image keeps its hash and every cache keyed by it has to be dropped explicitly. `SceneFetcher.reload_scene()` does exactly that — it walks the content mapping and calls `content_provider.purge_file(hash)` for each entry.

But the two consumers don't share a cache key:

- The scene's own UI goes through `fetch_texture` / `fetch_texture_by_hash`, whose promise is cached under the **plain hash** (`content_provider.rs:1522`) — which is what `purge_file` removes (`content_provider.rs:2588-2591`, it drops `<hash>` and `scene_<hash>` only). So it re-downloads → updated icon. ✅
- The touch-control icons go through `fetch_texture_by_url` → `fetch_texture_by_url_with_quality`, which caches under **`"<hash>_q<quality>"`** (`content_provider.rs:1701`). `purge_file` never touches that key, so the resolved `TextureEntry` from before the edit survives the reload and is handed straight back. ❌

That also matches the detail that `joypad.gd:507` first tries `get_texture_from_hash(icon_hash)` (plain key → correctly purged → miss) and then falls through to `fetch_texture_by_url`, hitting the stale `_q` entry.

### Suggested fix

Make `purge_file` drop every key derived from the hash, i.e. also `"<hash>_q<quality>"` for all texture qualities (or purge by `<hash>` prefix), and check whether the on-disk file written for the `_q` variant is removed too.

### Also affected

The same fetch backs the pointer tooltip's custom glyph (`tooltip_label.gd:136`), so it should go stale in the same way. Any other `fetch_texture_by_url` caller (icons fetched by URL rather than through a scene mapping) has the same exposure.

EibrielInv:
Two preview hot-reload bugs, both reproduced and verified fixed on a Pixel 7.

  ## #2795 — the preview WebSocket is never re-pointed on a warm deeplink

  Scanning the preview QR **without closing the app** switches the realm and loads the
  scene, but leaves the WebSocket pointed at the previous server. Hot-reload is then
  silently dead: no error, nothing logged, updates only after a manual reload.

  `set_preview_url` had two call sites: the desktop-only settings field
  (`settings.gd:371`) and `explorer.gd:334`, which runs **inside `explorer._ready()`**.
  A deeplink arriving while the explorer is already running is routed by
  `_route_teleport()`, which switches the realm but never touches the socket. That also
  explains the desktop/mobile asymmetry with no platform difference in the reload
  machinery: desktop enters the URL through the settings field, mobile scans a QR into a
  running app.

  **Fix:** call `set_preview_url` from `_route_teleport()`, where the realm is already
  derived from `deep_link_obj.preview`. No change to `PreviewWebSocket` — re-pointing a                                                                                   
  live socket is already supported (`set_url()` stores a pending URL; `_process` closes,                                                                                                              
  reconnects and resets `_speaks_protobuf`), which is how the settings button works today.

  | | before | after |
  |---|---|---|
  | `preview-ws > connected` after the deeplink | **never** | 0.6s later |                                                                                                                            
  | edit on the new server reaches the device | **nothing** | scene reload |
  
  ## #2796 — touch-control icons stay stale after a preview reload                                                                                                                                    

  The preview server hashes file *paths*, so a re-saved image keeps its hash and every
  cache keyed by it must be dropped explicitly. `reload_scene()` does that via
  `purge_file()`, which drops `<hash>` — but `fetch_texture_by_url` caches under
  `<hash>_q<quality>`, which `purge_file` never touches. The scene's own UI picked up an                                                                                                              
  edited icon while the `PBTouchScreenControls` button and the pointer-tooltip glyph kept
  resolve the scene's content mapping and fetch **by hash**, sharing the plain-hash cache
  slot that `purge_file` actually drops. The by-URL path stays as a fallback for an
  already-unloaded scene. Side benefit: one shared cache entry instead of a duplicate, and
  the `get_texture_from_hash` fast path starts hitting.

  ## Not fixed here

  The preview server collapses every file event into one message naming only the
  last-written file (one shared 800ms trailing debounce in `@dcl/sdk-commands`). Two model
  files written inside that window strand one of them. No client-side mitigation exists —
  preview hashes are path-derived, so a single `updateModel` is indistinguishable from one
  standing in for several changed files. Documented in `preview_websocket.gd`; belongs in a
  `js-sdk-toolchain` PR.

  ## QA — how to reproduce each bug (on `main`, then confirm gone here)

  **#2795**
  1. Start two previews on different ports with visibly different scenes:
     `npx sdk-commands start --no-browser --port 8000` and the same on `--port 8001`.
  2. On the phone, scan the `:8000` QR (or
     `decentraland://open?preview=http://<LAN-IP>:8000&position=0,0`). Wait for the scene.
  3. **Leave the app open** and scan the `:8001` QR.
  4. Edit any file in the `:8001` project and save.
  5. **Bug:** nothing happens — no reload, and `adb logcat | grep "preview-ws >"` shows no
     `connected` line after the second deeplink. **Fixed:** `connected` appears and the
     edit reloads the scene.

  **#2796**
  1. Preview a scene that uses the same image *twice*: in its own UI, and as the `icon`                                                                                                               
     of a `PBTouchScreenControls` action.
  2. Open it on a phone and wait for the scene UI and the touch buttons.
  3. Edit that image in place (same path/filename) and save. No manual reload needed —
     image edits are not hot-swapped (only `.glb` takes the in-place path), so the server
     sends `updateScene` and the client reloads the scene by itself within a second or two.
  4. **Bug:** the scene UI shows the new image, the touch button keeps the old one until
     the app is restarted. **Fixed:** both update. Same for the pointer-tooltip glyph.

  Closes #2795
  Closes #2796


kuruk-mm (approved):
if you tested, welcome to merge!

Issue #2800, opened 2026-09-01
[Bug] AudioStream and VideoPlayer ignore their spatial audio fields


leanmendoza:
### Platform
Mobile (Android/iOS)

### Issue Description

Both media components declare spatial audio fields in the protocol, and godot-explorer ignores all of them:

- `PBAudioStream` (1021): `spatial` (4), `spatial_min_distance` (5), `spatial_max_distance` (6, default 60)
- `PBVideoPlayer` (1043): `spatial` (7), `spatial_min_distance` (8), `spatial_max_distance` (9, default 60)

`grep spatial` over the whole media path — `lib/src/scene_runner/components/audio_stream.rs`, `lib/src/scene_runner/components/video_player.rs`, `lib/src/godot_classes/dcl_video_player.rs` and `godot/src/decentraland_components/video_player.gd` — returns zero hits. A scene that sets `spatial: true` on either component gets flat 2D audio at full volume regardless of where the listener stands.

### Why it is not a small fix

There are two independent reasons the audio cannot be positioned today, and both have to be solved:

1. **The nodes are not 3D.** `DclAudioStream` and `DclVideoPlayer` are both declared `#[class(init, base=AudioStreamPlayer)]`. Godot's 2D player has no distance attenuation or panning at all. The only 3D node in the audio path is `DclAudioSource` (`base=AudioStreamPlayer3D`), used by `PBAudioSource`.

2. **The native backends bypass Godot entirely.** On Android and iOS, video and (once #2799 is fixed) audio streams are decoded by ExoPlayer / AVPlayer, which render straight to the OS mixer. Their output never enters a Godot audio bus, so Godot's 3D panner and distance attenuation cannot act on it, no matter what node type wraps them.

Changing the base class alone therefore fixes nothing on the platforms we ship. Real spatial support requires decoding the media to PCM and pushing it into a Godot `AudioStreamPlayer3D` — which is what the removed ffmpeg path used to do for audio streams before #920, and what does not exist for video at all.

### Unity reference

Unity applies these fields in `MultiMediaPlayer.UpdateSpatialAudio`, and only on the AVPro branch — for LiveKit sources it is a no-op:

```csharp
audioSource.spatialBlend = args.isSpatial ? 1f : 0f;   // binary, not a blend
audioSource.minDistance  = args.minDistance;           // default 0
audioSource.maxDistance  = args.maxDistance;           // default 60
[...]

Issue #2496, opened 2026-07-09
Implement AssetLoad (pre-load resources) SDK component on mobile


regenesis-claw:
## Description

Implement the SDK/protocol **AssetLoad** (pre-load resources) API on mobile, so scenes can instruct the client to pre-load assets ahead of time and avoid flickering/pop-in during gameplay.

Currently this is only implemented in `bevy-explorer` (desktop). Mobile (`godot-explorer`) has no support for it yet.

## Protocol / SDK Reference

- Protocol PR: https://github.com/decentraland/protocol/pull/339 — "feat: main asset pre-load" (merged 2026-01-30). Adds `PBAssetLoad` and `PBAssetLoadLoadingState` components.
- Proto definitions:
  - `asset_load.proto`
    ```proto
    message PBAssetLoad {
      repeated string assets = 1; // paths of assets to preload
    }
    ```
  - `asset_load_loading_state.proto`
    ```proto
    message PBAssetLoadLoadingState {
      LoadingState current_state = 1; // current loading state
      string asset = 2;               // asset scene path
      uint32 timestamp = 3;           // monotonic counter
    }
    ```
- SDK (TypeScript) wiring: `decentraland/js-sdk-toolchain` — `@dcl/ecs` `AssetLoad` component + system, lets a scene set `assets` on an entity and subscribe to `AssetLoadLoadingState` events per asset.

## Reference Implementation (bevy-explorer)

- https://github.com/decentraland/bevy-explorer/pull/672 — "Preload protocol" (closes https://github.com/decentraland/bevy-explorer/issues/667)
- Follow-up tracking: https://github.com/decentraland/bevy-explorer/issues/862 — "Apply preload assets API to prevent flickerings"

## Goal

Implement the equivalent behavior on mobile: entities with the `AssetLoad` component should trigger client-side asset pre-loading, and the client should emit `AssetLoadLoadingState` updates back to the scene as assets load/finish/fail, matching the bevy-explorer implementation.

## Acceptance Criteria

- [ ] Scenes can set `PBAssetLoad` on an entity with a list of asset paths and mobile begins pre-loading them
- [ ] Mobile emits `PBAssetLoadLoadingState` updates (loading/finished/failed) per asset back to the scene
- [ ] Verified against a sample scene using the SDK `AssetLoad`/`AssetLoadLoadingState` components

---
Requested by Rodri via Slack

leanmendoza:
https://github.com/decentraland/godot-explorer/pull/2567

PR #2801, opened 2026-09-01
fix: SDK audio parity with unity-explorer

### LINKED ISSUE #861: SDK Audio Components: Unity Parity Audit and Volume Curve Alignment
##  **⚠️ Problem:**
A creator tunes a scene's audio on desktop and the same scene sounds wrong on mobile — identical `AudioSource.volume` and identical slider positions produce noticeably different loudness, and some SDK audio behaviour on desktop doesn't exist on Godot at all. This forces creators to author for one client, suppressing `Playtime` in audio-driven scenes (venues, events); the drop-off size is not currently measured.

## **🏁 Scope:**
Audit every SDK audio surface in Godot against Unity — `AudioSource`, `AudioStream`, `AudioEvent`, avatar/emote audio, settings buses — and fix the volume and attenuation math so a given slider position and SDK `volume` produce perceptually equivalent loudness on both clients. Implementing the missing `AudioEvent` component is **out of scope**; it is split into #2673.

## **📝 Deliverables:**
- [ ] Parity matrix published covering `PBAudioSource`, `PBAudioStream`, `PBAudioEvent` and avatar/emote audio: field × implemented in Godot? × matches Unity?
- [ ] Slider→gain curve matched to Unity on all six buses. `audio_settings.gd` currently maps the setting **linearly onto dB** (`db = -80 + 80*pct/100`), so 50% ≈ −40 dB instead of ≈ −6 dB
- [ ] SDK `volume` → dB fixed. `audio_source.gd` uses `20 * log(dcl_volume)`, but GDScript `log()` is **natural** log — `0.5` gives ≈ −13.9 dB, not −6.0 dB. Use `linear_to_db()` consistently across `audio_source.gd`, `audio_stream.gd`, `video_player.gd`; handle `volume = 0.0` (`log(0)` → `-inf`)
- [ ] Spatial attenuation matched: document Unity's rolloff/min/max distance and set Godot's `unit_size` / `max_distance` / attenuation model to land within ±3 dB at equal world distance
- [ ] Bus routing made explicit — `Scene`, `VoiceChat`, `UI` have `send = &""` while `Music` and `AvatarAndEmotes` send to `Master`
- [ ] Reference test scene checked in (fixed-amplitude tone at fixed distances) and A/B verified on iOS + Android vs desktop at slider 25 / 50 / 75 / 100

## Notes
- Does Unity apply SDK `volume` as raw linear amplitude, or through a mixer group with its own curve? The audit must answer this first — the Godot curve is only correct relative to a measured Unity baseline.
- Confirm with design whether matching Unity is right at the quiet end, or whether mobile speakers need a compressed floor above −80 dB.

## **🔗 References:**
- [decentraland/protocol#195](https://github.com/decentraland/protocol/pull/195) — adds `AudioEvent` and the shared `MediaState` enum (merged Apr 2024). The change the original ticket asked us to check; **not implemented in godot-explorer** → #2673.
- [AudioEventsSystem.cs](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/AudioEventsSystem.cs) — Unity's `PBAudioEvent` implementation; reference for #2673.
- [unity-explorer#4790](https://github.com/decentraland/unity-explorer/pull/4790) — Unity's crossfade fix for neighbouring-scene audio bleed; parity target for the `dcl_enable` → `max_db = -100` hack Godot uses today.
- [unity-explorer#2589](https://github.com/decentraland/unity-explorer/pull/2589) — how master and world volume compose onto a stream's volume on desktop; reproduce this order in Godot.
- [unity-explorer#7428](https://github.com/decentraland/unity-explorer/pull/7428) — audio outside the mixer group ignored the slider entirely; same class of bug for any Godot player not routed to a named bus.

### LINKED ISSUE #1184: Implement Audio Analysis Component #511
https://github.com/decentraland/protocol/pull/338

PR list:

- Explorer - https://github.com/decentraland/unity-explorer/pull/6361
- Js-Toolchain - https://github.com/decentraland/js-sdk-toolchain/pull/1256
- Protocol - https://github.com/decentraland/protocol/pull/328
- Example scene - https://github.com/decentraland/sdk7-goerli-plaza/pull/273

### LINKED ISSUE #2673: SDK Audio: Implement AudioEvent (PBAudioEvent) Component
## Dependency
Split out of #861 (SDK audio parity audit).

##  **⚠️ Problem:**
A creator who reacts to audio state — waiting for a track to load before starting a cutscene, showing a buffering spinner on a stream, or advancing a game beat when a clip finishes — can do it on desktop but not on mobile, because Godot never writes `PBAudioEvent` back to the scene. Those scenes either break or get authored desktop-only, which costs `Playtime` on mobile in exactly the audio-driven scenes (venues, events, guided experiences) that keep sessions long; the affected scene count is not currently measured.

## **🏁 Scope:**
Implement the outbound `PBAudioEvent` component (ECS id `1105`) in godot-explorer for both `AudioSource` and `AudioStream` entities, mirroring Unity's `AudioEventsSystem`: emit a `MediaState` + monotonic `timestamp` on the grow-only set whenever the entity's playback state changes. This follows the existing `PBVideoEvent` write-back in `video_player.rs` — no new transport, no changes to inbound audio components, and no changes to volume behaviour (that stays in #861).

## **📝 Deliverables:**
- [ ] `PBAudioEvent` emitted for `AudioSource` entities, appended to the grow-only set with the same dedupe pattern as `poll_video_events()` in `lib/src/scene_runner/components/video_player.rs` — write only on state change, increment `timestamp` per emission
- [ ] `PBAudioEvent` emitted for `AudioStream` entities, reusing the stream's existing state where available
- [ ] State mapping implemented and matching Unity's `GetAudioSourceState`: clip still loading → `MS_LOADING`; URL set but no clip → `MS_LOADING`; no URL → `MS_NONE`; playing → `MS_PLAYING`; otherwise → `MS_READY`. Fetch failure maps to `MS_ERROR`
- [ ] `MediaState` enum values match the proto exactly (`MS_NONE=0`, `MS_ERROR=1`, `MS_LOADING=2`, `MS_READY=3`, `MS_PLAYING=4`, `MS_BUFFERING=5`, `MS_SEEKING=6`, `MS_PAUSED=7`)
- [ ] No CRDT spam: verified that a steadily-playing source emits nothing after its initial `MS_PLAYING`, matching Unity's `LastPropagatedAudioState` guard
- [ ] Test scene checked in that subscribes to `AudioEvent` on both an `AudioSource` and an `AudioStream` and logs the state sequence, run side by side against Unity desktop to confirm identical sequences

## Metrics
**Target:** `Playtime` in audio-driven scenes. There is no baseline for how many deployed scenes read `AudioEvent`, so acceptance is a **parity gate**: for the same scene and the same interaction, the `MediaState` sequence and ordering emitted by Godot mobile matches Unity desktop, verified with the test scene above.

## Notes
- Unity stamps `timestamp` with `sceneStateProvider.TickNumber`, while Godot's `PBVideoEvent` uses a per-entity incrementing counter. Pick one and state it — the proto only requires a monotonic counter, but creators comparing clients will notice if the semantics differ.
- `MS_BUFFERING` / `MS_SEEKING` / `MS_PAUSED` are reachable for streams but not for a plain `AudioSource`. Confirm during implementation whether Godot's stream backend can report them, or whether they stay unused for now.

## **🔗 References:**
- [decentraland/protocol#195](https://github.com/decentraland/protocol/pull/195) — the PR that adds `AudioEvent` (merged Apr 2024) and introduces the shared `MediaState` enum, replacing the duplicated `VideoState`. Review discussion explains why `AudioEvent` is deliberately smaller than `VideoEvent`.
- [proto/decentraland/sdk/components/audio_event.proto](https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/audio_event.proto) — the contract to implement: `PBAudioEvent { MediaState state = 1; uint32 timestamp = 2; }`, ECS component id `1105`.
- [proto/decentraland/sdk/components/common/media_state.proto](https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/common/media_state.proto) — the eight `MediaState` values the mapping must produce.
- [Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/AudioEventsSystem.cs](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/SDKComponents/AudioSources/Systems/AudioEventsSystem.cs) — Unity's implementation and the behavioural target: `PropagateAudioSourceEvents` / `PropagateAudioStreamEvents`, the `LastPropagatedAudioState` dedupe, and the `GetAudioSourceState` mapping this issue reproduces.
- [lib/src/scene_runner/components/video_player.rs](https://github.com/decentraland/godot-explorer/blob/main/lib/src/scene_runner/components/video_player.rs) — `poll_video_events()` is the in-repo pattern for an outbound event component: read node state, compare against last known, `append()` to the grow-only set, bump `timestamp`. Follow it rather than inventing a new shape.

### LINKED ISSUE #2741: [Bug] AssetLoad preloading audio breaks AudioSource playback on mobile
### Platform
Mobile (Android/iOS)

### App Version
Not specified

### Device Information
Not specified

### Issue Description
Using `AssetLoad` (SDK7 component, `@dcl/sdk/ecs`, protocol id 1213) to preload an audio file breaks `AudioSource.playSound` for that clip on mobile. Once the audio file has been preloaded via `AssetLoad`, subsequent attempts to play it via `AudioSource` fail — this only happens on mobile; desktop is unaffected.

A minimal reproduction repo has been created by a community member (stom) to validate the bug:
<https://github.com/stom66/dcl-bugtest-assetload-breaks-audio> — includes a video comparison of mobile vs. desktop behavior.

`AssetLoad` was implemented for mobile as part of #2496 (closed) — this appears to be a regression/gap introduced by that implementation specifically affecting audio assets.

### Expected Behavior
Preloading an audio file with `AssetLoad` should not prevent `AudioSource.playSound` from working for that same clip — behavior should match desktop.

### Steps to Reproduce
1. Clone/deploy the repro scene: https://github.com/stom66/dcl-bugtest-assetload-breaks-audio
2. Preview/run the scene on a mobile device
3. Trigger playback of an audio clip that was preloaded via `AssetLoad`
4. Compare against desktop, where the same clip plays correctly

### Occurrence
Not specified (repro repo indicates consistent occurrence on mobile)

### Related Issues
- #2496 — Implement AssetLoad (pre-load resources) SDK component on mobile (closed)

Requested by Rodri via Slack

### LINKED ISSUE #2799: [Bug] AudioStream (radio modules) doesn't play audio on mobile
### Platform
Mobile (Android/iOS) — also reproduced running the Godot mobile client build from a desktop computer

### App Version
Not specified

### Device Information
Not specified

### Issue Description
`AudioStream` components (used by the radio modules in Genesis Plaza) don't play audio on the mobile client. All affected radio modules point to streaming mp3 URLs hosted on the same CDN, e.g. `https://dcl-radio.b-cdn.net/K-pop_1.mp3`. The pointer-event hover indicator on the radio shows correctly (see screenshot in the originating thread), but no sound is produced.

This is the live-stream `AudioStream` SDK component (`lib/src/scene_runner/components/audio_stream.rs`, backed by `av_sinks`/`stream_processor`), distinct from the local-file `AudioSource` component — #2741 tracks a separate, unrelated `AudioSource` regression caused by `AssetLoad` preloading.

### Expected Behavior
`AudioStream` should play the configured mp3 stream on mobile, matching desktop client behavior.

### Steps to Reproduce
1. Open Genesis Plaza (or any scene with a radio module using `AudioStream`) on the Godot mobile client — reproduced both on a physical mobile device and running the mobile client build from a desktop.
2. Approach a radio module streaming from `dcl-radio.b-cdn.net` (format: `https://dcl-radio.b-cdn.net/<track>.mp3`).
3. Observe no audio plays, despite the pointer/hover indicator rendering correctly.

### Occurrence
Reported as consistent (100%) across both test environments (mobile device + desktop-run mobile client).

### Related Issues
- #2741 — `[Bug] AssetLoad preloading audio breaks AudioSource playback on mobile` — different component (`AudioSource`, local files) and different trigger (`AssetLoad` preloading), but same "audio silently fails on mobile" symptom class; worth cross-checking during investigation.
- #2251 — `[Bug] Stream stops playing when audio device changes (AirPods switching devices)` — another open `AudioStream`/streaming reliability bug in the current sprint.
- [decentraland/unity-explorer#4062](https://github.com/decentraland/unity-explorer/issues/4062) — `[QA] Audio Stream | Audio stream is not currently working as it should` (closed) — prior, non-mobile-specific report of `AudioStream` streaming failures (MP3/WAV/OGG, valid URLs) affecting the desktop client; confirms `AudioStream` streaming has been fragile across clients, not just Godot mobile.

Requested by Lean via Slack (originally reported by Omar Bobrow)

leanmendoza:
Closes #861, closes #2673, closes #2799, closes #1184, closes #2741

# What?

- **Volume and attenuation match Unity.** The settings slider maps through `linear_to_db` instead of a linear dB ramp, SDK `AudioSource` drops to `unit_size = 1` (Godot's default of 10 is +20 dB past a metre), and avatar steps / effects / double-jump / glider / emote were audited against the Unity prefabs.
- **Godot's attenuation high-shelf is switched off** on every 3D player. It is wired to the attenuation multiplier, so lowering `unit_size` also cut up to 22 dB above 5 kHz. Unity's rolloff is gain-only.
- **`AudioStream` plays again**, through the same native backends as video. Silent since #920 removed the software decoder.
- **`PBAudioEvent`** for `AudioSource` and `AudioStream`, mirroring Unity's state machine.
- **`PBAudioAnalysis`** for `AudioSource`: 8 bands + amplitude, ported from unity-explorer's `native/audio-analysis`.
- **An `AssetLoad` preload no longer breaks `AudioSource`.** The raw-file promise cache key is namespaced, so a typed fetcher can no longer be handed a promise that resolves to nothing.

# Why?

Everything spatial was roughly +20 dB louder than Unity from a metre out, because nothing set `unit_size`. `AudioStream` had been silent on every platform for over a year. `PBAudioEvent` and `PBAudioAnalysis` had no implementation at all, so scenes using them got nothing back. And a preloaded audio clip failed silently while reporting itself ready.

Two known limits, both pre-existing and shared with `VideoPlayer`: `AudioStream` is `Noop` on desktop, and neither component is spatial — its audio never enters Godot. Tracked in #2800.

# Testing

Build **1734** has this changes, compare against build **1724** which doesn't.

**Where.** Everything except Case 1 and Case 3 runs in the world **`aesironline.dcl.eth`** (MannakiaWorld). Paste this into a note or a Slack DM on the phone and tap it:

```
decentraland://jump?realm=aesironline.dcl.eth
```

If the deeplink does not fire, open the in-app chat and send `/changerealm aesironline.dcl.eth`.

The world opens on the **SDK Audio Parity (#861 / #2673 / #2799 / #1184)** scene by default; the blue **Select a scene here** button at the top switches to the others (Case 8 uses it). It draws a panel with three numbered sections — `1 TONE`, `2 ANALYSIS`, `3 STREAM` — a row of ground markers at 1, 2, 5, 10 and 20 m from the speaker, each labelled with the dB it should lose, and live readouts of the distance, the media-state events and the analysis bands. Every case below is driven from that panel: no console, no logs, no code.

**A/B baseline.** These are mostly louder / quieter / muffled judgements, and they read far better side by side. Run each case again on **build 1724** (or any build from before this PR) — a second phone, or the same one before updating. Where the two builds should differ, the case says so.

### Case 1 — The volume slider has a usable range
**Steps:**
2. Open Settings → Audio.
3. Drag the **General** slider to 50%.
4. Drag it to 25%.

- [ ] **Expected:** at 50% the audio is clearly audible, at roughly half the loudness of 100%. At 25% it is quieter but still clearly audible.
- [ ] **Against 1724:** the same 50% is nearly silent — everything audible is crammed into the top of the slider.

### Case 2 — An SDK sound fades with distance and stays bright
**Steps:**
1. Open the app and enter the world with the deeplink above.
2. On the panel, under `1 TONE`, press **play**.
3. Walk to the ground marker labelled **1 m**, then to 2 m, 5 m, 10 m and 20 m, standing still a couple of seconds at each.

- [ ] **Expected:** the panel prints `tone is X.X m away -> expect -Y.Y dB`, and every doubling of the distance roughly halves the loudness. At 20 m the tone sounds the same, only quieter — not muffled or underwater.
- [ ] **Against 1724:** clearly louder from 5 m out, and it loses its highs as you walk away.
- [ ] **Regression — global:** press **global on**. The tone plays at full volume and does not change as you walk. **global off** brings the distance fade back.

### Case 3 — Avatar footsteps
**Steps:**
2. Switch to third person.
3. Walk, then run, then jump and land.
4. Walk away from another avatar until you are about 50 m from them.

- [ ] **Expected:** your own steps sound natural and open, not boxed in or muffled. The other avatar's steps fade out and are gone by ~50 m.
- [ ] **Against 1724:** your own steps sound muffled, and other avatars' steps stay audible far across the parcel.

### Case 4 — The radio stream plays (Android / iOS only)
**Setup:** phone only. On desktop this stays silent by design.
**Steps:**
1. Open the app and enter the world with the deeplink above.
2. On the panel, under `3 STREAM`, press **play**.
3. Wait a few seconds.
4. Press **pause**, then **play** again.

- [ ] **Expected:** music starts within a few seconds, and the `events ... stream:` line goes `LOADING` → `PLAYING`. **pause** stops the audio and the line shows `PAUSED`; **play** resumes it.
- [ ] **Against 1724:** nothing ever plays — the stream is silent on every platform.

### Case 5 — A broken stream URL does not take anything down
**Setup:** phone only, following Case 4.
**Steps:**
1. Under `3 STREAM`, press **url BROKEN**.
2. Wait about 15 seconds.
3. Under `1 TONE`, press **play**.
4. Press **url ok** and then **play** under `3 STREAM`.

- [ ] **Expected:** no audio from the stream, and the app keeps running normally — the tone in step 3 still plays, and the panel keeps updating. Step 4 recovers the radio.
- [ ] **Known limitation, not a bug:** the `stream:` line stays on `LOADING` rather than reporting `ERROR`. The native players report no error for a URL that resolves but 404s, so there is nothing to surface yet.

### Case 6 — Audio events fire once, not continuously
**Steps:**
1. Open the app and enter the world with the deeplink above.
2. Under `1 TONE`, press **loop off**, then **clear log**.
3. Press **play** and let the clip run to its end without touching anything.
4. Keep watching the `events ... tone:` line for another 15 seconds.

- [ ] **Expected:** one `PLAYING` appears when the clip starts and one `READY` when it ends. The line stops growing after that — no repeated states while the clip plays, and nothing new after it finishes.
- [ ] **Against 1724:** the line stays empty. The component reported nothing at all.

### Case 7 — The analysis bands follow the music
**Steps:**
1. Open the app and enter the world with the deeplink above.
2. Under `1 TONE`, press **loop on**, then **play**.
3. Watch the `analysis  amp ...  bands ...` line for about 30 seconds, through several loop repeats.
4. Under `2 ANALYSIS`, press **mode LOG**, then **mode RAW**.

- [ ] **Expected:** the amplitude and band numbers move in time with the sound and settle at zero when you press **stop**. They keep moving across every loop point — they must not freeze after the first repeat. In **mode LOG** every band stays between 0 and 1.
- [ ] **Expected — no cost:** the frame rate does not visibly drop while the analysis runs.
- [ ] **Note:** the bands only track the tone, not the radio in `3 STREAM`. Streams are decoded by the platform's native player, so their audio never reaches the analysis — that is expected, and is one of the limits tracked in #2800.

### Case 8 — A preloaded clip still plays
**Steps:**
1. Open the app and enter the world with the deeplink above.
2. Tap the blue **Select a scene here** button at the top of the screen and pick **AssetLoad Preload Demo (#2496)**.
3. In the panel, under `AUDIO CLIP (#2741)`, press **PLAY CLIP**. This is the control — the clip has not been preloaded yet.
4. Press **PRELOAD CLIP** and wait until the `clip:` line reads `preloaded, ready`.
5. Press **PLAY CLIP** again.

- [ ] **Expected:** the clip is audible in step 3 **and** in step 5, and the `events:` line appends `PLAYING` then `READY` on each press.
- [ ] **Against 1724:** step 3 plays but step 5 is silent — that is the bug. The `events:` line also stays empty there, since `PBAudioEvent` had no implementation at all.

**Numbers.** The dB figures the panel prints are arithmetic on the two rolloff formulas, not measurements, and nothing here has been A/B'd against the Unity client yet.


godot/src/decentraland_components/audio_stream.gd
```diff
@@ -1,27 +1,185 @@
...
+			_async_init_native_backend("res://src/decentraland_components/av_player.tscn")
+		_:
+			_init_noop_backend()
+
+
+func _async_init_native_backend(scene_path: String):
+	stream_state = STREAM_STATE_LOADING
+
+	native_player = load(scene_path).instantiate()
+	add_child(native_player)
```
regenesis-claw, line 64: [P1] `exo_player.gd` and `av_player.gd` both define `func _process(_delta): update_texture()`, so from this `add_child()` onward the wrapper polls the plugin every frame for a node that has no texture and no surface.

Two consequences:

1. Per-frame JNI (`exoPlayerHasVideoSizeChanged` + `exoPlayerHasNewHardwareBuffer`) or ObjC calls per audio stream, for nothing.
2. Worse: if the URL carries a video track — and nothing stops a scene pointing `AudioStream.url` at an HLS/MP4 — `update_texture()` sees `HasVideoSizeChanged` with non-zero dimensions and calls `_reinitialize_surface()`, which runs `exoPlayerInitSurface` and allocates an `ExternalTexture` nobody will ever render. On failure it also `push_error`s, so that path bills Sentry.

The description's premise ("`init_texture()` is the only video-specific call") holds for the calls *you* make, but not for the ones the wrapper makes on its own. Suggest disabling processing on the wrapper right here:

```suggestion
	native_player = load(scene_path).instantiate()
	# Audio-only: the wrapper's own `_process` drives `update_texture()`, which
	# polls the plugin every frame and will allocate a video surface if the
	# source turns out to carry a video track.
	native_player.set_process(false)
	add_child(native_player)
```

godot/src/decentraland_components/audio_stream.gd
```diff
@@ -1,27 +1,185 @@
...
+
+	await get_tree().process_frame
```
regenesis-claw, line 66: [P1] Nothing survives this `await` intact. `_init_backend_impl` can be re-entered before the next frame (a scene changing `AudioStream.url` twice, or entity removal calling `_backend_dispose()`), and both cases break here:

- **Disposed during the await:** `_backend_dispose()` → `_free_native_player()` sets `native_player = null`, then line 62 calls `native_player.set_source_url(...)` on Nil — a runtime error that aborts the coroutine, and it lands in Sentry.
- **Re-inited during the await:** the stale coroutine resumes and drives the *new* player — a second `set_source_url` / `set_looping` / `play()` on a player another coroutine is already setting up.

An epoch counter is the pattern that closes this (same shape as `lobby.gd`'s `_guest_login_attempt`): bump `_init_epoch` at the top of `_init_backend_impl`, capture it before the await, and bail after it if `_init_epoch != captured or not is_instance_valid(native_player)`. Note `if native_player:` is not a validity check on a freed instance — `is_instance_valid()` is.

regenesis-claw (commented):
Commenting rather than voting because this is still a draft — but as it stands I would hold approval on the two P1s below.

Followed `REVIEW.md`. Pre-flight: branch is **up to date with `main`** (`behind_by: 0`). This touches native iOS/Android media interop, so **an iOS build is required** — `🍏 iOS → R2 (unsigned export)` is currently *pending* on this head, and the only green Android APK in the sticky build report is for `48c6198`, an earlier commit. Not a blocker yet, just don't read the build report as covering `4a91491`.

I treated every number in the description as a claim to check, as you asked. Where I could reach the source, I did:

### Verified against source (not taken on trust)

| Claim | Verdict |
|---|---|
| `AudioEvent` must be a GrowOnlyValueSet | ✅ `js-sdk-toolchain/scripts/protocol-buffer-generation/generateIndex.ts` lists it in `GROWN_ONLY_COMPONENTS`. (Side note: that list also has `ExplorerUiEventsResult`, which `lib/build.rs` still registers as LWW — separate issue, not this PR.) |
| `max_db = volume_db` reproduces `volume * min(1, unit_size/d)` | ✅ `audio_stream_player_3d.cpp::_get_attenuation_db` is `att += volume_db; if (att > max_db) att = max_db;`, and the caller uses `db_to_linear(att)` as the whole multiplier. The trick is correct and the comment is accurate. |
| `PBAudioStream` was dead on every platform | ✅ `av::backend::av_sinks` delegates unconditionally to `noop::av_sinks`. |
| Writing `playing = false` back on `PBAudioSource` is Unity parity | ✅ `AudioEventsSystem.cs::WriteBackNaturalFinish` does exactly this, and your `IsNaturalFinish` conditions match Unity's line for line. |
| Godot's `max_distance` adds a linear ramp across the whole range | ✅ `audio_stream_player_3d.cpp` L445-447: `multiplier *= MAX(0, 1.0 - (dist / max_distance))`. Which is why finding 4 below matters. |
| The `file_hash` promise-key collision | ✅ Real, and both call sites (`fetch_file`, `scene_fetcher.gd`) route through the namespaced key, so the fix is complete. |

### Unverified

Every dB figure in the table, the `0.09% of a core` decode benchmark, and "Unity's state machine never produces `MS_LOADING`/`MS_READY` for a stream". The last one I could partly check: Unity's `GetAudioStreamState` is a bare `(MediaState)videoState` cast, so the claim rests entirely on AVPro's state machine, which I can't see from here — see finding 6.

### Findings

**[P1] The native wrappers are not as video-free as the description says.** `exo_player.gd` / `av_player.gd` both run `_process(_delta): update_texture()`, and `update_texture()` calls `_reinitialize_surface()` whenever the player reports a video size. `init_texture()` is not the only video-specific entry point — `_process` is a second one, and it runs unconditionally the moment you `add_child()` the wrapper. Details inline.

**[P1] Re-entrancy in `_async_init_native_backend`** across the `await get_tree().process_frame`. Details inline.

**[P2] `MS_ERROR` is probably unreachable for the failure that actually happens.** The description says Android's `setSourceUrl` "just sets the media item and prepares" — which means its return value tells you nothing about whether the URL resolves. `_update_stream_state` never writes `STREAM_STATE_ERROR`, so a 404 / bad codec sits in `LOADING → MS_NONE` forever, and the only path to `MS_ERROR` is a non-`http(s)` scheme. That makes test case 2 ("a deliberately broken URL to force `MS_ERROR`") pass only if "broken" means "not a URL". Either wire an error callback out of the wrappers or correct the test plan.

**[P2] Test plan shape.** `REVIEW.md` §4 asks for Setup / numbered Steps from "Open the app" / Expected, because QA runs these by hand off the TestFlight build. "In rough order of risk" is a good reviewer checklist but it is not executable by a tester, and it depends on a scene in `mannakia-dcl-scene` they have no way to load. Worth converting the six items before this leaves draft.

Rest is inline. Positives worth saying out loud: the archaeology on the AudioStream regression is right, the `max_db` trick is a genuinely clever way to get Unity's rolloff out of Godot's clamp, the three FFT tests actually pin behaviour (band placement + RMS in one assertion) rather than just executing the code, and the promise-key namespacing is a real bug fix that happens to be in the neighbourhood.

godot/src/decentraland_components/audio_stream.gd
```diff
@@ -1,27 +1,185 @@
...
+
+	# No init_texture() here: that call is the video surface, and nothing else
+	# in the wrapper depends on it.
+	if not (_source.begins_with("http://") or _source.begins_with("https://")):
+		push_error("AudioStream: only remote sources are supported, got ", _source)
```
regenesis-claw: [P2] Both of these (`57` and `63`) fire on scene-authored data: the URL comes straight from `PBAudioStream.url`. A creator typing a `file://` path or an unreachable host is not an actionable client fault, but `push_error` routes to Godot's error stream and opens a Sentry issue per occurrence (`REVIEW.md` §5). `push_warning` for both — the `stream_state = STREAM_STATE_ERROR` beside them is already the real signal, and it reaches the scene as `MS_ERROR`.

godot/src/decentraland_components/audio_stream.gd
```diff
@@ -1,27 +1,185 @@
...
+
+
+func _update_stream_state():
+	var duration: float = native_player.get_duration()
+	var is_native_playing: bool = native_player.is_playing()
+
+	if is_native_playing:
+		stream_state = STREAM_STATE_PLAYING
+	elif duration <= 0:
+		stream_state = STREAM_STATE_LOADING
```
regenesis-claw: [P2] `duration <= 0` as the loading test misfires on the main use case for this component. A live radio stream (Icecast/SHOUTcast, endless HLS) reports duration `0` or `inf` for its whole life, so:

- pausing it takes the `duration <= 0` branch before ever reaching `STREAM_STATE_PAUSED` → the scene sees `MS_NONE`, never `MS_PAUSED`;
- and since `STREAM_STATE_ERROR` is never written from here at all, a stream that fails after `set_source_url` succeeded is indistinguishable from one still buffering.

Using `_is_playing` (what we asked for) rather than duration to separate paused-from-loading would fix the first half; the second half needs an error signal out of the wrappers.

godot/src/decentraland_components/avatar/avatar_sfxs.tscn
```diff
@@ -7,8 +7,16 @@ script = ExtResource("1_q3p8f")
 
 [node name="AudioPlayer_Steps" type="AudioStreamPlayer3D" parent="." unique_id=257952697]
 transform = Transform3D(1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 1.31567, 0)
+max_distance = 50.0
```
regenesis-claw, line 10: [P2] These nodes keep `max_distance = 50`, which means they get the exact multiplier your own comment in `audio_source.gd` cites as the reason to set `max_distance = 0.0` for SDK sources — `audio_stream_player_3d.cpp` L446: `multiplier *= MAX(0, 1.0 - (dist / max_distance))`.

So footsteps are `(1/d) * (1 - d/50)`, not the plain `1/d` the description says Unity's curve resolves to: ≈ -1.9 dB at 10 m and ≈ -6 dB at 25 m against the reference. Same for the three `glider_prop.tscn` players.

If the intent is Unity-exact, `max_distance = 0` here too (`1/d` is already ≈ -34 dB at 50 m, so nothing audible survives anyway). If the ramp is wanted as a cheap culling cutoff, worth a line saying so — as written it reads like the two files disagree. `AudioPlayer_Emote` is unaffected: `attenuation_model = 3` + `max_distance = 15` is exactly the linear ramp you wanted there.

lib/src/scene_runner/components/audio_stream.rs
```diff
@@ -62,113 +65,116 @@ pub fn update_audio_stream(
...
+/// Mirrors Unity's `GetAudioStreamState`, which is a direct cast of the media
+/// player state. Unity's state machine never produces `MS_LOADING` or
+/// `MS_READY` for a stream — a stream that is loading or finished reads
+/// `MS_NONE` — so neither is emitted here.
+fn media_state(stream_state: i32) -> MediaState {
+    match stream_state {
+        STREAM_STATE_PLAYING => MediaState::MsPlaying,
+        STREAM_STATE_PAUSED => MediaState::MsPaused,
+        STREAM_STATE_ERROR => MediaState::MsError,
+        _ => MediaState::MsNone,
```
regenesis-claw, line 152: [P2] This collapse is the one parity claim I couldn't confirm. Unity's `GetAudioStreamState` is `return (MediaState)videoState;` — a bare cast, no filtering — so "Unity's state machine never produces `MS_LOADING` or `MS_READY`" is a statement about AVPro's `VideoState`, not about that function, and it isn't checkable from this repo.

It also disagrees with the sibling component: `poll_video_events` in `video_player.rs` maps *every* `VIDEO_STATE_*` through, including `VsLoading` and `VsReady`, off the same wrappers and the same state constants. Given `_update_stream_state` sets `LOADING` for the entire buffering period of a live stream, the practical effect here is that a scene watching a radio start up sees `MS_NONE` until audio arrives.

Either map them through (matching `video_player.rs`), or keep the collapse and cite the AVPro state list that justifies it — the comment currently attributes the behaviour to a Unity function that doesn't do it.

lib/src/scene_runner/components/audio_stream.rs
```diff
@@ -62,113 +65,116 @@ pub fn update_audio_stream(
...
+                        if playing {
+                            audio_stream_node.bind_mut().backend_play();
+                        } else {
+                            audio_stream_node.bind_mut().backend_pause();
+                        }
+                    }
+                    _ => {
+                        audio_stream_node.bind_mut().backend_dispose();
+
+                        // Streams have no LiveKit variant: `from_source` only ever
```
regenesis-claw: nit: the comment doesn't match `av::backend::BackendType::from_source`, which returns `BackendType::LiveKit` for any source starting with `livekit-video://` before it reaches the platform branches. The outcome is harmless — `to_gd_int()` gives `0`, which falls into the GDScript `match` default and initialises the noop backend — but a reader trusting this comment would not expect that URL to be reachable at all.

lib/src/scene_runner/components/audio_source.rs
```diff
@@ -84,4 +93,89 @@ pub fn update_audio_source(
...
+            let Some(mut value) = audio_source_component
+                .get(&entity_id)
+                .and_then(|entry| entry.value.clone())
+            else {
+                continue;
+            };
+            if value.playing != Some(true) {
+                continue;
+            }
+            value.playing = Some(false);
```
regenesis-claw, line 177: [P2] The write-back itself matches Unity (`AudioEventsSystem.cs::WriteBackNaturalFinish`), so no argument there — but this is the first place in this repo where the renderer `put`s a *scene-authored* LWW component. Everything else that writes LWW state back (`tween_state`, `gltf_container_loading_state`, `raycast_result`, `asset_load_loading_state`) is renderer-owned.

That matters because `LastWriteWins::set` (`lib/src/dcl/crdt/last_write_wins.rs`) drops anything that isn't *strictly* greater and implements no payload tie-break:

```rust
fn is_timestamp_greater(...) -> bool { timestamp > entry.timestamp }
```

So: entry at `T`, this `put` bumps the renderer to `T+1`, and a scene that sets `playing = true` in the same tick (a replay-the-SFX pattern, `playSound()` right after the clip ends) sends `T+1` as well — equal, dropped, sound never replays, and the two sides stay out of sync until the scene writes again. The js SDK's CRDT resolves equal timestamps by comparing payloads; ours doesn't.

Same mechanism, wider window, for `audio_analysis.rs`: it `put`s every tick a source produces samples, so a scene later flipping `mode` or `bands_gain` has a decent chance of landing on an equal timestamp and being silently ignored.

Not asking you to fix the CRDT here — but it's worth deciding whether the natural-finish write-back is worth that exposure given `PBAudioEvent`'s `MS_PLAYING → MS_READY` transition already tells the scene the clip ended, and whether the tie-break belongs in a follow-up issue.

regenesis-claw (commented):
Delta review on `70c2ed9` (previous review was `4a91491`). Single commit, all six findings from the AudioStream path addressed:

| Finding | Status |
|---|---|
| [P1] `set_process(false)` on the native wrapper | ✅ Added at instantiation, with a comment explaining why. |
| [P1] Re-entrancy across `await get_tree().process_frame` | ✅ `_init_epoch` bumped in `_backend_dispose()`, checked with `is_instance_valid()` on resume. Re-init disposes first, so it also invalidates a stale coroutine — covers both directions I flagged. |
| [P2] `duration <= 0` misreads a live stream as permanently loading | ✅ Replaced with `_source_ready`/`_has_played`, driven off `_is_playing`/`is_playing()` instead of duration. Traced the four branches — `PLAYING` when actually playing, `LOADING` when asked-to-play-but-not-yet, `PAUSED` once it has played before, `READY` otherwise. Correctly separates a paused live stream from one that never started. |
| [P2] `MS_LOADING`/`MS_READY` collapsed to `MS_NONE`, wrong Unity citation | ✅ Now maps the whole set through, matching `poll_video_events`. Added an honest note that `MS_BUFFERING`/`MS_SEEKING` stay unreachable because the wrappers don't expose that signal — correct, and better than silently mapping them to something wrong. |
| [P2] `push_error` on scene-authored bad URLs | ✅ Both downgraded to `push_warning`. |
| Stale "streams have no LiveKit variant" comment | ✅ Corrected to describe what `from_source`/`to_gd_int()` actually do. |
| `avatar_sfxs.tscn`/`glider_prop.tscn` keep `max_distance = 50` vs `audio_source.gd`'s `0` | ✅ Resolved with a comment rather than a code change, which is fine — it explains the tradeoff (per-avatar cutoff vs. Unity-exact rolloff, ~-1.9 dB at 10 m / -6 dB at 25 m) and states it's deliberate. That was the ask. |

Deferred, both reasonably: the LWW equal-timestamp exposure on the `PBAudioSource`/`PBAudioAnalysis` write-backs is filed as #2807 instead of fixed inline — correct call, that one's a CRDT-level change and out of scope for an AudioStream-path commit. `MS_ERROR` being unreachable for a real network failure (only a non-http scheme reaches it) is still true and wasn't addressed — understandable since fixing it needs new signal out of the Kotlin/Swift plugins, but worth a line in the PR body or a tracking issue so it doesn't get lost, since the current `Testing` section (d) still just says "it should play" without a case for a broken URL.

Nothing else changed outside the four files in this commit. CI on `70c2ed9`: Static checks, Clippy, Linux, Android, Docker all ✅. No iOS artifact for this head (no `build`/`build-ios` label) — still holding on that per REVIEW.md pre-flight for a PR that touches the native media wrappers, same as last time. Still a draft, so comment-only again.

leanmendoza:
builb number **1734**

kuruk-mm (approved):
code looks great! if you tested and other people tested looks well to merge

PR #2812, opened 2026-09-02
feat: graphic profiles budgets - view distance, particles, lights (#2672)

### LINKED ISSUE #2672: Graphic Profiles: Wire View Distance, Particles and Dynamic Lights per Profile
##  **⚠️ Problem:**
Scene lights and particles now render on mobile, but every profile pays the same cost — a Very Low device runs the same light count, particle budget and draw distance as High. 

## **🏁 Scope:**
Add limitos for the new graphic capabilities per graphic profile, by adding 3 keys to `PROFILE_DEFINITIONS` in `godot/src/config/graphic_settings.gd` — render distance, particle budget, light budget — and apply them in `apply_graphic_profile()`.

## **📝 Deliverables:**
- [x] `view_distance` key per profile → camera far + fog end. Values set by dev — High matches Unity 300m, lower profiles err low
- [ ] Verify no scene pop-in regression at each distance in Genesis Plaza
- [x] `particle_quality` key per profile → max particles + emitter cap. Disabled on Very Low
- [x] `CPUParticles3D` vs `GPUParticles3D` selection follows profile, not device only
- [x] `max_active_lights` + activation-range multiplier per profile. 0 on Very Low, 2 Low, 4 Medium, 8 High
- [x] Profile values override Dev Tools light settings from #2007 unless profile = Custom
- [x] Custom profile exposes all 3 as individual controls
- [ ] A/B gpu ms + fps per profile, Genesis Plaza + `rituals.dcl.eth`

## Metrics
% Session Overheat. Target: Very Low/Low sessions hitting thermal FPS cap ↓. Measured via existing thermal state API (`get_thermal_state()`, PR #1236) + Segment session events. Baseline needs a capture run before close.

## **🔗 References:**
- [decentraland/godot-explorer#2633](https://github.com/decentraland/godot-explorer/pull/2633) — open PR, LightSource with mobile perf guardrails; light budget lands here
- [decentraland/godot-explorer#2007](https://github.com/decentraland/godot-explorer/pull/2007) — Max Active Lights, Auto Activation Range, Global Light Budget as Dev Tools toggles; this issue promotes them to profile-driven
- [decentraland/godot-explorer#2588](https://github.com/decentraland/godot-explorer/pull/2588) — merged ParticleSystem SDK component; per-scene budget defined, not per-profile
- [decentraland/godot-explorer#1018](https://github.com/decentraland/godot-explorer/issues/1018) — preset matrix these 3 keys extend
- [decentraland/unity-explorer#6294](https://github.com/decentraland/unity-explorer/pull/6294) — desktop cut load radius to 20 for all quality settings for perf; reason to stay low on mobile

manuelmaceira:
Closes #2672

## Done
- `view_distance` key per profile (60/100/200/300m — Low matches the previous fixed 100m `camera.far`, High matches Unity 300m) → camera far + fog begin/end (0.6×/0.9× far, hides far-plane pop-in)
- `particle_quality` key per profile (0-3) → Rust scene particle system: scene budget + per-emitter cap (0 / 2k+500 / 15k+2k / 50k+5k), 0 = disabled on Very Low
- `CPUParticles3D` vs `GPUParticles3D` follows profile: Low uses CPU emitters (params mapped directly on node, bursts skipped — no `emit_particle` on CPU); Medium/High use GPU
- `max_active_lights` per profile (0/2/4/8), profile overrides Dev Tools light settings unless Custom
- Custom profile exposes view distance, particle quality, max lights as individual controls
- Unit tests: `test_graphic_profiles.gd` (all profile keys land on their targets)

## Pending (validation, needs device runs)
- [ ] Verify no scene pop-in regression per distance in Genesis Plaza
- [ ] A/B gpu ms + fps per profile (Genesis Plaza + rituals.dcl.eth)
- [ ] Baseline thermal capture run

## Test plan
- [x] `test_graphic_profiles.tscn` headless: profile keys land on config, lights, avatar gate **and** Rust particle budgets (via `DclGlobal.get_particle_profile_budgets()`)
- [x] Pre-commit suite: fmt, clippy, GDScript validation (389 scripts), Rust unit tests, integration tests
- [x] i18n: `extract_strings.py --check` + `format_csv.py` canonical
- [ ] Device: switch profiles in Settings with Genesis Plaza loaded → particles stop/start immediately (Very Low off, Low = CPU emitters), no rebuild flicker on budget-only changes
- [ ] Device: relaunch on Very Low → particles stay off from boot
- [ ] Device: walk out/in of a scene on Very Low → particles stay off

## Notes for reviewers
- High fog band 180/270m: intentional — issue #2672 deliverable was updated in place ("High matches Unity 300m, lower profiles err low"), matching Unity's 300m on High was requested after the original wording.
- Idle scenes (~120s tick): profile changes land on their next tick. Acceptable — their emitters aren't emitting while the user is elsewhere (reconcile gate).


regenesis-claw (changes requested):
Reviewed against `REVIEW.md`. The shape of this is right — profile definitions stay data-driven, the Rust budgets are atomics rather than consts, the CPU/GPU split is honest about what it drops, and there's a headless test. But the profile→Rust wiring isn't actually connected, so most of the second bullet in the description doesn't happen outside the Custom dropdown.

### Pre-flight
- **Branch is stale and conflicting.** `main...HEAD` = ahead 3 / **behind 10**, `mergeStateStatus: DIRTY`. Please merge `main` — in particular `PROFILE_NAMES` is now a list of catalogue keys on `main` (`SETTINGS_GRAPHIC_PROFILE_CUSTOM`, …) while this branch still has plain `["Very Low", …]`, and `settings.gd` filters on the literal `"Custom"`.
- **No CI on the current head.** The sticky build report is from `0e7681e`; there are no workflow runs for `d76bfcb`. Nothing here is verified by a build.
- iOS build not required (no `plugins/dcl-godot-ios/**`, comms or av paths touched) — noting it explicitly per pre-flight step 2.

---

## Blocking

**[P1] `apply_particle_quality()` is never called from `apply_graphic_profile()`.** `graphic_settings.gd` sets `config.particle_quality` and calls `AvatarAnimHelpers.apply_particles_enabled(...)`, then stops. The only call site of `apply_particle_quality()` in the repo is `settings.gd::_on_custom_particles_changed()`. So for Very Low / Low / Medium / High, `DclGlobal.set_particle_profile_budgets()` never fires: the atomics keep their 50 000 / 5 000 / GPU defaults and **no CPU emitter is ever created**. The description's headline particle bullets are only live on the Custom profile.

```gdscript
	# Avatar move/jump/land dust particles follow the profile.
	AvatarAnimHelpers.apply_particles_enabled(profile.particle_quality > 0)
	apply_particle_quality(profile.particle_quality)
```

**[P1] `particle_quality = 0` does not disable scene particles.** In `apply_particle_system()` (and the CPU twin) the budget-0 guard runs *before* the playback-state match:

```rust
if SCENE_PARTICLE_BUDGET.load(Ordering::Relaxed) == 0 { node.set_emitting(false); }
…
_ => { node.set_speed_scale(1.0); node.set_emitting(active); }   // re-enables it
```

The default playback arm unconditionally re-enables emission a few lines later. And `compute_amount()` ends at `amount = remaining_budget.max(1)`, so with a 0 budget every emitter still gets 1 live particle instead of 0. Net effect on Very Low: emitters are still created, still emitting, still allocating a process material + draw mesh. Fold the budget into the emitting decision (`let emitting = active && budget > 0;`) rather than pre-setting it, and consider skipping process-material/mesh construction entirely when the budget is 0.

**[P1] Nothing restores the particle state at boot.** `view_distance` is re-applied on load (`explorer.gd` sets `camera.far`, `environment_selector._ready()` sets fog), but `AvatarAnimHelpers.particles_enabled` defaults to `true` and the Rust atomics default to 50k/5k/GPU — and both are only written inside `apply_graphic_profile()` / the Custom handler. A device that picked Very Low last session relaunches with avatar dust particles on and the full GPU scene budget until the user re-picks a profile. Apply the persisted `config.particle_quality` once at startup (same place the profile is otherwise honoured).

**[P1] Budget-clamp `tracing::warn!` becomes a per-emitter Sentry stream on low profiles.** `compute_amount()` warns every time an emitter is clamped, and it runs on every `PBParticleSystem` component update. With `particle_quality` 0 or 1 (budget 0 / 2 000) essentially *every* emitter in a scene trips it, on every CRDT update — a Godot warning per hit, captured by Sentry (REVIEW.md §5). Downgrade to `tracing::debug!`, or log once per entity when the clamped value changes.

---

## Non-blocking

**[P2] `fog_light_color` write in `set_view_distance()` is dead.** `sky_base.gd:196-197` reassigns `world_environment.environment.fog_light_color` from `fog_color_gradient` every `_process` frame. The hardcoded `Color(0.7, 0.75, 0.8)` is overwritten on the next frame — drop the line (good news: the day/night tint is not broken). Also note `fog_enabled = true` / `fog_sky_affect = 0.0` now unconditionally override whatever `game_environment.tres` shipped.

**[P2] The fog band is a visual change for every High-profile user, and #2672 says the opposite.** Today fog is fixed at begin 76 / end 100; on High this becomes 180 / 270 (effectively no fog at normal play distances). High is the default profile on mobile. Issue #2672 explicitly cautions *"err low, do not match Unity 300m"*, while the description says *"High matches Unity 300m"*. Worth reconciling in the issue before this lands, since the pending A/B items in the description are exactly what would settle it.

**[P2] New user-facing strings are literals.** `"View Distance"`, `"Particles"`, `"Max Active Lights"`, `"Off"/"Low"/"Medium"/"High"` are passed as function arguments / `add_item` labels, which `tools/i18n/extract_strings.py` cannot see. `main` already localises this exact surface (`SETTINGS_GRAPHIC_PROFILE_*` in `en/es/pt_BR.csv`), so after the merge these need catalogue keys.

**[P2] Are the Custom rows reachable?** `settings.gd` builds the profile dropdown with `if profile != "Custom"`, and I found no code path that sets `config.graphic_profile = ConfigData.PROFILE_CUSTOM` (`dcl_dynamic_graphics_manager.rs` only reads it). If Custom can't be selected, the three new rows, `_sync_custom_profile_controls()` and `apply_particle_quality()`'s only live call site are all unreachable UI. How do you get there?

**[P2] The CPU↔GPU swap is lazy, not eager.** `update_particle_system()` only runs for entities in the dirty set, so the class swap and the new caps land on an existing emitter *whenever the scene next updates that component* — which for a static emitter is never. The commit message claims "Profile switch CPU<->GPU recreates existing emitters"; the code doesn't do that. Either walk `scene.particle_systems` on profile change or soften the claim.

**[P2] Interaction with the dynamic graphics manager.** It auto-downgrades/upgrades the profile at runtime. With this PR that now snaps `camera.far` and the fog band mid-play, and — once the P1 above is fixed — tears down and recreates every emitter as a different node class. Was that considered? A visible fog/draw-distance snap during a frame-rate dip is the kind of thing QA will file.

**[P2] The test doesn't cover the thing that broke.** `test_graphic_profiles.gd` asserts the config fields and the avatar-particle gate, but not that the profile reaches `set_particle_profile_budgets` — an assertion on the Rust mapping would have caught P1-1. (Also, as with the other `godot/src/test/**` suites, nothing in CI runs it.)

**[P2] `## Test plan` per REVIEW.md §4.** The "Pending (validation)" list is a personal checklist, not steps QA can execute on a phone. Worth adding at least: pick each profile → observe the fog/draw-distance change in Genesis Plaza; pick Very Low → confirm scene particles and avatar dust are gone; relaunch → confirm the setting stuck.

**Security:** no security issues found — no new network, auth or user-controlled input surface; the new `#[func]` only stores three clamped integers into atomics.

Requesting changes on the four P1s; the rest are follow-up-or-argue.

---

regenesis-claw (changes requested):
Re-review at `a102dd1` (branch was rebased onto `main` — fresh full read, not a delta diff). Four of the five previous P1s are properly fixed, and the generation-counter approach to the lazy CPU↔GPU swap is a better answer than the one I suggested. Two things still block: a hole in the budget-0 gate, and the new Custom rows aren't hooked into the translation-changed path.

### Previous findings

| # | Finding | Status |
|---|---------|--------|
| P1 | `apply_particle_quality()` never called from `apply_graphic_profile()` | ✅ wired, and the test now asserts the Rust mapping via the new `get_particle_profile_budgets()` |
| P1 | budget 0 doesn't disable particles (pre-set `set_emitting` overwritten) | ⚠️ fixed in `apply_particle_system`/`_cpu` (`active && budget > 0`), but re-opened elsewhere — see P1-A |
| P1 | no boot restore of particle state | ✅ `global.gd::_ready` re-applies `config.particle_quality` to both the Rust atomics and the avatar gate |
| P1 | clamp `warn!` → Sentry stream on low profiles | ✅ downgraded to `tracing::debug!` |
| P2 | dead `fog_light_color` write | ✅ removed |
| P2 | unlocalized settings strings | ⚠️ keys added to en/es/pt_BR and existing keys reused — but see P1-B |
| P2 | Custom rows unreachable | ✅ `_populate_graphic_profile_items()` now includes Custom, and `_on_dropdown_list_graphic_profiles_item_selected` already sets `graphic_profile = PROFILE_CUSTOM` for it |
| P2 | lazy CPU↔GPU swap contradicts the commit message | ✅ `PARTICLE_PROFILE_GENERATION` + per-scene re-apply block |
| P2 | test doesn't assert the Rust mapping | ✅ |
| P2 | High fog band vs #2672's "err low, do not match Unity 300m" | ❌ unaddressed, body still says "High matches Unity 300m" |
| P2 | dynamic-graphics-manager interaction | ❌ unaddressed (and now heavier — see P2-C) |
| P2 | no QA `## Test plan` | ❌ body unchanged |

---

## Blocking

**[P1-A] `reconcile_user_presence()` re-enables emission regardless of the budget.** The `(false, true)` arm — user walks back into the scene — does `ps_set_emitting(&item.node, true)` gated only on `item.playing`. On Very Low (or Custom → Particles: Off) the budget-0 gate you added in `apply_particle_system` is bypassed the first time the player leaves and re-enters the scene, and stays bypassed until something dirties the component. `item.playing` is derived from `last_value.active`, not from the budget, so it's still true.

```rust
            (false, true) => {
                if item.playing && SCENE_PARTICLE_BUDGET.load(Ordering::Relaxed) > 0 {
```

A small `fn particles_enabled() -> bool` next to `get_particle_profile_budgets()` used by both call sites would keep the two in sync.

**[P1-B] The three Custom rows never re-translate.** `settings.gd` has an explicit convention for this — the doc comment above `_populate_camera_mode_items()` and the `_notification(NOTIFICATION_TRANSLATION_CHANGED)` handler that rebuilds every dropdown whose items are translated strings. The new code resolves at build time and is not in that handler:

- `_custom_particles_dropdown.add_item(tr("SETTINGS_PARTICLES_OFF"), 0)` and the three profile keys — finished strings, exactly the case the comment describes.
- `_make_custom_profile_row(..., tr("SETTINGS_VIEW_DISTANCE"))` etc. assign `label.text` from GDScript, so the label doesn't re-translate either.

`_setup_custom_profile_controls()` only runs from `_ready`, so after a language switch these three rows stay in the old locale for the lifetime of the page. Add a `_retranslate_custom_profile_rows()` (re-set the three `label.text`s + rebuild the particles dropdown, preserving `selected`) and call it from `_notification`.

---

## Non-blocking

**[P2-A] Every new scene rebuilds its emitters once, immediately after creating them.** `Scene::new` initialises `particle_profile_generation: 0`, while the global counter is already ≥ 1 after the boot restore in `global.gd`. So on the first `update_particle_system` tick of any scene, the dirty loop creates the emitters and then the generation block tears them down and recreates them in the same call — plus a duplicate `fetch_texture_by_hash`. Seed it from the current value instead:

```rust
            particle_profile_generation: PARTICLE_PROFILE_GENERATION.load(Ordering::Relaxed),
```

(needs a small `pub fn current_particle_profile_generation()` accessor, since the static is private to the module.) Applies to both constructors in `scene.rs`.

**[P2-B] Profile switches land late on idle scenes.** Per the doc comment on `reconcile_user_presence`, `update_particle_system` ticks idle scenes roughly every ~120s — so the re-apply is "eager" only for the active scene. Worth one line in the PR body so QA doesn't file a bug when a background scene keeps GPU emitters for a while after switching to Low.

**[P2-C] The dynamic graphics manager now triggers a full emitter rebuild.** It auto-downgrades/upgrades the profile at runtime (disabled only on Custom), and each change bumps the generation → every live emitter in every scene is freed and rebuilt, `started_at`/burst cycles reset, textures re-fetched, plus a `camera.far` and fog-band snap. During a frame-rate dip that's a visible hitch on top of an already-struggling frame. Was that considered? Skipping the rebuild when only the budget changed (class unchanged) would cover the common case.

**[P2-D] Still unaddressed from round 1:** the High fog band (76/100 → 180/270 for the default mobile profile) versus #2672's explicit "err low, do not match Unity 300m", and the missing QA-executable `## Test plan` (the "Pending validation" list is your checklist, not something a tester can run). One case per profile switch, one for Very Low → no scene/avatar particles, one for relaunch-persistence would cover it.

**CI:** `📊 Static checks` green; Linux, Clippy and Android still pending at review time — the Rust changes here are non-trivial (new `AtomicU32`, `VarArray` return, borrow churn in the re-apply block), so worth confirming Clippy before merge.

**Security:** no issues — no new network/auth/user-input surface.

---

regenesis-claw (approved):
Re-review at `b4a88d20` (3 commits ahead of `a102dd1` + a merge from main pulling in unrelated files — `avatar.gd`, `nameplate_layer.gd`, `avatar_modifier_area_detector.gd`, `test_nameplate_stacking.gd` are from other merged PRs, not this one).

### Round-2 findings

| # | Finding | Status |
|---|---------|--------|
| P1-A | `reconcile_user_presence` re-enables emission on scene re-entry regardless of budget | ✅ `ps_set_emitting(&item.node, SCENE_PARTICLE_BUDGET.load(...) > 0)` |
| P1-B | Custom rows never re-translate | ✅ `_retranslate_custom_profile_rows()` added and wired into `_notification`, `_populate_custom_particles_items()` extracted and reused in both places |
| P2-A | fresh scenes seed `particle_profile_generation: 0` and immediately rebuild | ✅ both `Scene::new`/constructor sites now seed from `current_particle_profile_generation()` |
| P2-C | dynamic graphics manager auto-switches cause a full emitter teardown/rebuild mid-play | ✅ better than asked: `set_particle_profile_budgets` now no-ops (skips the generation bump) when the values are unchanged, and the per-scene re-apply only recreates the node when the CPU/GPU *class* changed — a same-class budget update re-applies in place |
| P2-D | High fog band vs #2672 wording | ✅ re-read #2672: it says *"High matches Unity 300m, lower profiles err low"* — that's exactly what this PR does, no actual conflict (my earlier read of the issue was off) |
| P2-D | missing QA `## Test plan` | ✅ added, with device cases for profile switching, relaunch persistence and walk-out/walk-in — the three cases this PR most needed |

Everything from both rounds is resolved. No new issues found in the delta.

**CI:** `📊 Static checks` green; Linux / Clippy / Android still pending at review time — nothing in this diff should trip Clippy, but worth a final glance once they land given the borrow-heavy Rust changes.

Approving — nice turnaround across both rounds, and the no-op/generation-skip design for the dynamic-graphics-manager case is a cleaner fix than what I suggested.

---

godot/src/config/graphic_settings.gd
```diff
@@ -27,6 +27,8 @@ const PROFILE_DEFINITIONS: Array[Dictionary] = [
 		"fps": ConfigData.FpsLimitMode.FPS_18,
 		"scale": 0.5,
 		"mesh_lod_threshold": 8.0,
+		"view_distance": 60.0,
```
kuruk-mm:
```suggestion
		"view_distance": 100.0,
```

kuruk-mm (changes requested):

godot/src/config/graphic_settings.gd
manuelmaceira: done, changed to 100/150/200/300

kuruk-mm (approved):
Looks great!

PR #1471, opened 2026-02-17
fix: plane mesh UV and unlit alpha texture

manuelmaceira:
## Summary

- Fix plane mesh UV Y-coordinate conversion: `1.0 - y` instead of `-y`, correctly mapping from Decentraland (0,0=bottom-left) to Godot (0,0=top-left). Negative UVs caused plane mesh textures (like Genesis Plaza event panel arrows) to be invisible.
- Use albedo texture alpha channel (`.a`) instead of alpha texture red channel (`.r`) in the unlit alpha shader, matching Unity's URP Lit behavior.
- Add ETC2 compression error handling with dimension padding and graceful fallback.
- Update iOS deploy tooling to use `xcrun devicectl` instead of `ios-deploy`.

Closes #1275
Partially resolves #1291 — fixes invisible buttons/textures on the try-on panel, but full try-on functionality still needs testing.

## Test plan

- [x] Verify Genesis Plaza event panel arrows are visible on desktop
- [x] Verify transparency renders correctly (no white bars)
- [ ] Test on iOS device
- [ ] Test try-on panel button visibility

PR #2813, opened 2026-09-02
fix(sdk): match Unity UV offset/tiling V direction in materials

### LINKED ISSUE #2719: [Bug] Material texture offset/tiling V direction inverted vs Unity (uv-direction parity)
### 📄 Issue description

Material texture `offset`/`tiling` moves the texture in the **opposite V direction** vs Unity Explorer. With a tiled color-grid texture, Unity and Godot display **different grid cells** on the Y axis for the same scene.

Repro scene: https://github.com/stom66/dcl-bugtest-uv-direction — a plane with `tiling: 0.25`, `offset: (0.25, 0.25)` set via `Material.Texture`. Unity shifts the texture right+up; Godot shifts it right+down.

### 🔍 Root cause

Our primitive mesh UVs are stored in Godot convention, V-flipped relative to DCL/Unity (`v_godot = 1 − v_unity`, see `mesh_renderer.gd` / PR #1471). But the material offset/tiling is applied **raw** in Godot's V-down UV space, so the V component lands in the wrong place whenever `offset.y ≠ 0` or `tiling.y ≠ 1`:

- `lib/src/scene_runner/components/material.rs` — `set_uv1_offset(offset.x, offset.y)` / `set_uv1_scale(tiling.x, tiling.y)` (both the unlit and PBR branches)
- `godot/assets/shaders/dcl_unlit_alpha.gdshader` — `vec2 uv = UV * uv_scale + uv_offset;`

To match Unity, the Godot Y offset must be **`1 − tiling.y − offset.y`** (identity for the defaults tiling=1/offset=0, which is why untiled textures look fine).

Example with the repro values: Unity samples the V window `[0.25, 0.5]`, we sample `[0.5, 0.75]` — one grid row off.

### 🧩 Scope

- StandardMaterial3D path (unlit + PBR) in `material.rs`
- Unlit alpha shader path (`apply_unlit_shader_properties` + `dcl_unlit_alpha.gdshader`)
- `TextureMove` / `TextureMoveContinuous` tween (PR #946) drives the same offset — its Y direction is inverted too and must go through the same compensation

### ✅ Expected behavior

Same scene shows the identical texture cell / offset direction in Godot and Unity Explorer.

### Related

- #1112 / PR #1470 (box UV mapping) and PR #1471 (plane UV conversion) fixed the base UV orientation but not the offset/tiling compensation.

manuelmaceira:
## Fixes #2719

Material texture `offset`/`tiling` moved the texture in the opposite V direction vs Unity Explorer whenever `offset.y ≠ 0` or `tiling.y ≠ 1`.

### Root cause

Primitive/glTF UVs are stored V-flipped relative to Unity (`v_godot = 1 − v_unity`, see `mesh_renderer.gd` / PR #1471), but the material offset/tiling was applied raw in Godot's V-down UV space. Compensate at write time:

`offset_y_godot = 1.0 − tiling.y − offset.y`

Identity for the defaults (tiling=1, offset=0) — why untiled textures were never affected.

### Changes

- `material.rs`: new helper `dcl_uv_offset_y()`, applied in unlit + PBR `StandardMaterial3D` paths and in `apply_unlit_shader_properties` (unlit alpha `ShaderMaterial`)
- `tween.rs`: `apply_uv_to_mesh` compensates on write, so `TextureMove`/`TextureMoveContinuous` get the fix too; internal tween state stays in Unity space so continuous delta signs need no change
- `.gdshader` untouched — compensation happens CPU-side, zero GPU cost

### Tests

- New regression test `uv_offset_y_matches_unity_texture_window` (defaults identity, repro values, pure scroll)
- `cargo fmt`, `clippy`, GDScript lint/format, unit + integration tests all pass (pre-push suite)
- Manual: repro scene [stom66/dcl-bugtest-uv-direction](https://github.com/stom66/dcl-bugtest-uv-direction) — plane showed cells **E3/E4/F3/F4** before, now shows **C3/C4/D3/D4**, matching Unity Explorer


lib/src/scene_runner/components/material.rs
```diff
@@ -479,7 +487,11 @@ fn apply_unlit_shader_properties(shader_mat: &mut Gd<ShaderMaterial>, unlit: &Dc
     if let Some(tex) = &unlit.texture {
         shader_mat.set_shader_parameter(
             "uv_offset",
-            &Vector2::new(tex.offset.0.x, tex.offset.0.y).to_variant(),
+            &Vector2::new(
+                tex.offset.0.x,
+                dcl_uv_offset_y(tex.tiling.0.y, tex.offset.0.y),
```
regenesis-claw, line 492: [P2] Correct for the transform, but note `dcl_unlit_alpha.gdshader` declares both samplers `repeat_disable`, so any resulting uv outside `[0,1]` clamps instead of wrapping. `tiling.y > 1` was already broken on this path, and this change newly produces negative offsets when `offset_y > 1 - tiling_y` (e.g. `t=1, o=0.3` -> `-0.3`), which clamps at the low end. Worth a follow-up to switch those uniforms to `repeat_enable` so the unlit-alpha path can actually reach parity.

lib/src/scene_runner/components/tween.rs
```diff
@@ -93,7 +96,7 @@ fn apply_uv_to_mesh(mesh: &mut Gd<MeshInstance3D>, tex_anim: &TextureAnimation)
             if let Ok(mut base_material) = material.try_cast::<BaseMaterial3D>() {
                 base_material.set_uv1_offset(Vector3::new(
                     tex_anim.uv_offset.x,
-                    tex_anim.uv_offset.y,
+                    dcl_uv_offset_y(tex_anim.uv_scale.y, tex_anim.uv_offset.y),
```
regenesis-claw, line 99: [P2] Verified this is safe: `TextureAnimation::default()` sets `uv_scale = (1,1)`, so a TMT_OFFSET-only tween gets the identity compensation.

Separate pre-existing gap, flagging so it isn't assumed fixed by #2719: this function writes `uv1_scale` purely from `tex_anim.uv_scale`, ignoring the `Material` component's own `tiling`. An entity with `Material{tiling.y = 4}` plus a TMT_OFFSET tween has its tiling reset to 1 and its V compensation computed from `tiling_y = 1`. Self-consistent, so nothing renders garbage, but it still won't match Unity.

regenesis-claw (approved):
## Review — approved (no P0/P1)

Followed `REVIEW.md`.

**Pre-flight**
- Branch is up to date with `main` (merge commit `f5cafe6`, today).
- **iOS build not required** for this PR: it touches only `material.rs` / `tween.rs` under the scene runner — no native iOS plugin, no `OS.get_name() == "iOS"` branch, no deeplink / keyboard / safe-area / comms / av code. Android APK built green on the previous head. Static checks green; Linux / Clippy / Android still running at review time.
- No submodule drift.

### Math verified independently

I re-derived the compensation rather than trusting the description. With `v_g = 1 - v_u` and Godot texture row = `V_g` vs Unity texture row = `1 - V_u`, requiring `V_g = 1 - V_u`:

```
(1 - v_u)*t + o_g = 1 - (v_u*t + o_u)
t + o_g          = 1 - o_u
o_g              = 1 - t - o_u
```

The `v_u` terms cancel, so `dcl_uv_offset_y()` is **exact for every UV**, not an approximation, and identity at `t=1, o=0`. Tiling magnitude is unaffected by the flip, so leaving `set_uv1_scale` raw is correct.

**Coverage check:** I grepped every write of `uv1_offset` / `uv_offset` in `lib/` and `godot/` at this head — the four sites the PR touches (unlit `StandardMaterial3D`, PBR `StandardMaterial3D`, `apply_unlit_shader_properties`, tween `apply_uv_to_mesh`) are all of them. The two reset-to-zero branches are correctly left alone. `dcl_unlit_alpha.gdshader` computes `UV * uv_scale + uv_offset`, i.e. identical semantics to `uv1_scale`/`uv1_offset`, so the same formula applies there — confirmed, not assumed.

**Tween default check:** `TextureAnimation::default()` sets `uv_scale = (1,1)`, so `dcl_uv_offset_y(1.0, 0.0) = 0.0` — a TMT_TILING-free tween is unchanged at rest. Good; a `Vector2::ZERO` default would have silently produced a `1 - o` offset here.

**Consumer impact:** `dcl_uv_offset_y` is `pub(crate)` — no public surface change. This *is* an SDK-visible behavior change though: any live scene whose author hand-compensated for the old inverted V on mobile will now render shifted. Given Unity is the parity ground truth (REVIEW.md §1) that's the intended outcome, but it's worth mentioning in the release notes.

**Security:** no security issues found — pure arithmetic, no I/O, no user-controlled input path, no new deps.

---

### Findings (all P2, none blocking)

**[P2] Custom-UV boxes use `-v`, not `1 - v`, so the new formula is off by one tile there.**
`mesh_renderer.gd` writes custom plane UVs as `Vector2(u, 1.0 - v)` but custom **box** UVs as `Vector2(u, -v)`. Under `v_g = -v_u` the compensated result becomes `-v_u*t + (1 - t - o_u)`, which differs from the target by exactly `t`. With texture repeat that wraps away for integer `t` (including the default `t = 1`), so it is invisible in the common case — but a custom-UV box with fractional `tiling.y` (e.g. an atlas at `0.25`) will still not match Unity. This is a pre-existing inconsistency, not something the PR introduced, but the PR's correctness now depends on it. Aligning `set_box` to `1.0 - v` like `set_plane` would make the assumption uniform. Fine as a follow-up issue.

**[P2] The unlit-alpha shader still can't tile at all.**
`dcl_unlit_alpha.gdshader` declares both samplers `repeat_disable`. Any `uv` outside `[0,1]` clamps to the edge instead of repeating, so `tiling.y > 1` renders as edge stretch on that path regardless of this fix. The fix also newly produces *negative* offsets whenever `offset_y > 1 - tiling_y` (e.g. `t=1, o=0.3` → `-0.3`), which clamps at the other end. The geometry is now right, the sampling still isn't. Switching those uniforms to `repeat_enable` (or `fract()`-ing the uv) would let the unlit-alpha path actually reach parity — worth a follow-up.

**[P2] Tween + Material tiling still diverge.**
`apply_uv_to_mesh` writes `uv1_scale` from `tex_anim.uv_scale` only, ignoring the `Material` component's own `tiling`. An entity with `Material{tiling.y = 4}` plus a TMT_OFFSET tween gets its tiling reset to 1 *and* its V compensation computed from `tiling_y = 1`. Self-consistent, so nothing renders garbage, but it won't match Unity. Pre-existing and out of scope — flagging so it isn't mistaken for fixed by #2719.

**[P2] Missing `## Test plan`.**
REVIEW.md §4 asks for QA-executable cases (start from opening the app, one action per line, observable pass/fail). The `### Tests` section here gives a local repro and a before/after cell reference, which is good evidence for a reviewer but not something QA can tick off on a phone. A short case — open app → enter the repro world → look at the tiled plane → expect cells C3/C4/D3/D4 — plus a regression case for an untiled texture and one for a `TextureMoveContinuous` scroll would cover it.

---

Unit test is narrow (it exercises the pure helper only) but it does verify the Unity-window equivalence numerically rather than restating the formula, which is the right call given the rest needs a running renderer. Approving; please let Linux/Clippy/Android finish green before merging.

kuruk-mm (approved):
looks great!

PR #2826, opened 2026-09-03
feat(notifications): revamp Notifications section (Menu redesign)

### LINKED ISSUE #2785: Menu: Notifications Section Redesign
## **⚠️ Problem:**
Notifications need a revamp under the Menu Panel Redesign: current design doesn't cover all 4 notification types or read/unread states cleanly.

## **🏁 Scope:** [See Figma](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=930-23060&t=HUJ8RbCPuSmueplb-1)
Revamp the Notifications section for the 4 notification types and unread/read states per the Figma spec.

## **📝 Deliverables:**

[ATLAS Notifications](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=930-23060&t=HUJ8RbCPuSmueplb-1)

- [ ] Revamp notifications 4 types: 
   Player accept your request
   Player send you a request
   Event started: only 1 for event, when started.
   New Item Receive 

- [ ] 2 states, Unread / Read

Image

## **🔗 References:**
- [decentraland/godot-explorer#1549](https://github.com/decentraland/godot-explorer/issues/1549) — earlier Notifications Panel Design Polish issue, related surface
- [ElzathRusso's comment](https://github.com/decentraland/godot-explorer/issues/2351#issuecomment-5399602050) — source comment with the Notifications spec (shared with other Menu sections)
- [Figma ATLAS Notifications](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=930-23060&t=HUJ8RbCPuSmueplb-1) — full spec

Requested by Seba via Slack

sdilauro:
Revamps the **Notifications** section for the Menu Panel Redesign, covering the 4 notification types and the read/unread states per the Figma spec.

Closes #2785
📐 Figma: [ATLAS — Notifications](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=930-23060)

## The 4 notification types (per issue)
| Type | Title | Body | Tapping it |
|---|---|---|---|
| Player sends you a request | player nickname (title color #A0ABFF) | "Want to be your friend!" | Social panel → **Requests** tab |
| Player accepts your request | player nickname (title color #A0ABFF) | "Accepted your friend request" | Social panel → **Friends** tab |
| Event started | event name (`metadata.name`) | "Event started!" | **Travel modal** for the event's place (same as a chat place/world link) |
| New item received | "New Item Received" (title color #DF9CFF) | item (wearable / emote) name | **Backpack** — Emotes tab for emotes, Wearables otherwise |

## Summary
- **Title/body** — title is the **subject** (nickname / event name / item label), action moves to the body. Title is a plain `Label` with a **per-type title colour** applied via `add_theme_color_override` (no BBCode → nothing to escape) and ellipsis on overflow.
- **Per-type image** — friend → round `ProfilePicture`; event/fallback → `AsyncImage` with a rounded outline; item → `AsyncImage` over an `ItemRarityBackground` (the rarity gradient the backpack draws), so a transparent thumbnail sits on its rarity fill.
- **Tap routing** (`menu`, `explorer`, `friends_panel`) — per the table above; navbar collapse lives in the menu handler so panel and toast entry points match.
- **Timestamp** — full relative scale (`Now` / `{n}m` / `{n}h` / `Yesterday` / days / weeks / months / years), re-rendered on locale change.
- **Read/unread** — read notifications dimmed (`modulate`); list updates incrementally (reused by id) so thumbnails don't re-fetch on refresh.
- **`AsyncImage` (shared)** — image textures use their own alpha (skeleton panel only shows while loading, then hides), per-instance rounded-corner shader driven by `border_radius`, new `border_width` export. Fixed the rounded shaders (`texture_rect_rounded_px`, `texture_rect_rounded`) dropping the parent `modulate`.
- **`ProfilePicture`** — new `@export var show_border` (defaults `true`, non-breaking).
- **i18n** — en/es/pt_BR keys for the new titles/bodies/timestamp units; removed superseded keys.

## QA review follow-ups
Applied from the design/QA pass on this PR (device-tested on iOS):
- **Event copy fixed** — title is the event's own name (`metadata.name` per the DCL `CommonEventsMetadata` schema); body is a fixed short **"Event started!"** instead of the server's long description sentence. Events are only surfaced **once they start** (`events_starts_soon` dropped from the supported list).
- **Per-type title colours** — friend request/accepted `#A0ABFF`, event `#73FFAF`, new item `#DF9CFF` (replaces the previous avatar-tinted nickname).
- **Emoji support in titles** — event names with emojis now render (Inter 600 + NotoColorEmoji fallback, same setup as the Discover cards) instead of tofu boxes.
- **List bottom margin** — 24px of breathing room after the last card so it isn't flush with the panel edge when scrolled to the bottom.

## Validation
- `gdlint godot/` ✓ · `gdformat` ✓ · `cargo run -- check-gdscript` ✓
- i18n: `format_csv --check` ✓ · `extract_strings --check` (0 pending, no orphans) ✓ · `test-i18n` ✓
- All notification scenes + the shared `AsyncImage`/shader consumers (Discover cards, profile, navbar, joypad) instantiate headless ✓

### Case 1 — Friend request → opens Requests tab
**Setup:** a second Decentraland account that isn't a friend of this one yet.
**Steps:**
1. Open the app and sign in.
2. From the second account, send this account a friend request.
3. On this account, tap the **bell** in the navbar to open the Notifications panel.
4. Tap the friend-request notification.

- [ ] **Expected:** title is the sender's nickname in their avatar colour, body "Want to be your friend!", round avatar. Tapping opens the social panel on the **Requests** tab.

### Case 2 — Friend request accepted → opens Friends tab
**Setup:** a second account; from this account, send it a request first.
**Steps:**
1. Open the app and sign in.
2. Have the second account accept this account's pending friend request.
3. Tap the navbar **bell** to open the Notifications panel.
4. Tap the acceptance notification.

- [ ] **Expected:** title is the sender's nickname (coloured), body "Accepted your friend request". Tapping opens the social panel on the **Friends** tab.

### Case 3 — Event started → opens the travel modal
**Setup:** an event that is **live now** at a Genesis City parcel (e.g. mark yourself Interested on a starting event at events.decentraland.org so its "started" notification fires).
**Steps:**
1. Open the app and sign in.
2. Tap the navbar **bell** to open the Notifications panel.
3. Tap the "Event started" notification.

- [ ] **Expected:** title is the event name, thumbnail with a rounded outline. Tapping opens the **travel modal** (Jump In / Cancel) for the event's place — the same modal a `name.dcl.eth` link in chat opens — and Jump In teleports there.

### Case 4 — New item received → opens the backpack
**Setup:** a reward assigned to this account (claim a campaign/POAP reward) so a "New Item Received" notification arrives; use an **emote** reward to check the tab.
**Steps:**
1. Open the app and sign in.
2. Tap the navbar **bell** to open the Notifications panel.
3. Tap the "New Item Received" notification.

- [ ] **Expected:** title "New Item Received", body the item name, thumbnail on its rarity gradient (backpack look). Tapping opens the **Backpack** on the **Emotes** tab for an emote (Wearables tab otherwise).

### Case 5 — Read / unread + timestamps
**Steps:**
1. Open the app and sign in with an account that has both read and unread notifications.
2. Tap the navbar **bell** to open the Notifications panel.
3. Tap **Read All**.

- [ ] **Expected:** unread rows are fully opaque; read rows are dimmed **including their thumbnail**. Timestamps read Now / {n}m / {n}h / Yesterday / days / weeks / months / years for their age.

### Case 6 — Regression: shared AsyncImage + rounded shader
The `AsyncImage` rework and the rounded-corner shaders are shared by Discover cards, the profile avatar, the navbar avatar and the joypad.
**Steps:**
1. Open the app and sign in.
2. Open **Discover** (compass) and scroll the place carousels.
3. Open the **Backpack**, then the **profile** editor.
4. Look at the navbar profile avatar and, on mobile, the joypad arc.

- [ ] **Expected:** Discover cards, the profile avatar and the navbar avatar render with correct rounded corners and no darkened / over-saturated / washed-out images; a Discover card whose thumbnail fails to load shows the placeholder glyph over its tinted rounded backdrop.
- [ ] **Regression — i18n:** switch to the **Pseudolocale (QA)** language (Settings, non-prod), reopen the panel — no clipping/overflow in es/pt titles.

After:
image


regenesis-claw (changes requested):
## Review — `feat(notifications): revamp Notifications section` (head `a490dbf`)

The structural call here is the right one: replacing the runtime StyleBox surgery with three pre-built, mutually-exclusive image nodes is a real simplification, it kills the grey-background-behind-the-avatar bug at the source, and reusing `ProfilePicture` / `AsyncImage` instead of hand-rolling a fourth image widget is exactly the direction `COMPONENT_AUDIT.md` asks for. The i18n work is clean too — the removed keys (`NOTIF_HEADER_EVENT_STARTED/_ENDED`, `NOTIF_HEADER_FRIEND_REQUEST_*`, `NOTIF_HEADER_REWARD_ASSIGNED`, the six `NOTIF_TITLE_WANTS_TO_BE_YOUR_FRIEND*` / `..._ACCEPTED_YOUR_FRIEND_REQUEST*` variants) have zero remaining references anywhere in `godot/` or `lib/`, and `translation_sources.json` matches all three catalogues.

**Pre-flight**
- Stacked on `feat/revamp-menu-social`, `ahead 2 / behind 0` — up to date with its own base. Reviewed as a delta against that branch, not `main`.
- iOS: not platform-sensitive (GDScript UI + locale data only, no `OS.get_name()` branch, no native plugin, nothing under `lib/`), but a green `iOS → R2 (unsigned export)` artifact is present anyway.
- CI: `Linux / Build Linux` is red on the **Avatar regression tests** step — both scripts printed `PASS` and the job then died on `exited with code <signal 11>` during teardown. That is the known headless `test-avatar` teardown segfault, and a GDScript-UI/locale diff cannot cause it. Please re-run the job; not counted against this PR.
- No submodule / `Cargo.lock` / `rust-toolchain.toml` / `project.godot` drift. No `.claude/` files, no debug prints added, no non-English comments or identifiers. No new `push_error`/`printerr` — Sentry quota unaffected.

---

## 🔴 P1 — blocking

### 1. The event thumbnail and event title both hang on metadata keys nothing in the repo confirms

`notification_text_helper.gd:108` reads the event title from `metadata.title`, and `notification_content.gd:78` reads the image from `metadata.image → thumbnailUrl → thumbnail`. The only two event payload shapes visible in this repo do not support that:

- `notifications_manager.gd::_generate_fake_notification`, `events_started` fixture: `{title, description, link, CommunityId, communityName}` — has `title`, but **no image key of any kind**.
- `notifications_manager.gd::_create_local_notification`: `metadata = {title, description, link}` — again no image.

The only other consumer of event-notification metadata in the client (`pages/discover/discover.gd:293`) reads `link` and nothing else, so it gives no independent confirmation either. The pre-rework code never touched `metadata.title` and never tried to load an event image, so this reliance is new.

The failure mode is silent: `AsyncImage.load_from_url("")` takes the `_finish_with_error()` path, which is a legitimate defined fallback but renders the generic "no image" placeholder with the `#E8B9FF` outline around an empty box. So if the server payload has no `image`/`thumbnailUrl`, **every event notification ships as an empty framed box** rather than the Figma thumbnail — and nothing logs, so QA has to catch it by eye on a live event.

What I'd like before approving: paste one real `events_started` (or `events_starts_soon`) payload from the notifications service, or confirm the key names from its schema. If the image genuinely isn't in the payload, this needs either a type-specific default image restored for events or an explicit design decision to ship the placeholder — and if the title field is `name` rather than `title`, the title silently degrades to the generic `NOTIF_HEADER_EVENT` ("Event") for every event, which is the one thing the redesign was meant to remove.

---

## 🟡 P2 — non-blocking

### 2. English copy is wrong in both new strings (and disagrees with es/pt)

`godot/locale/en.csv:400` — `NOTIF_HEADER_NEW_ITEM_RECEIVED` = `"New Item Receive"`, missing the `d`. Spanish and Portuguese both say "recibido" / "recebido", so English is the odd one out.

`godot/locale/en.csv:382` — `NOTIF_BODY_ACCEPT_YOUR_FRIEND_REQUEST` = `"Accept your friend request"` reads as an imperative (as if the user is being asked to accept). The notification means the *sender* accepted. es = "Aceptó tu solicitud de amistad", pt_BR = "Aceitou seu pedido de amizade" — both past tense. Suggest `"Accepted your friend request"`.

If the Figma spells them this way, worth pushing back on the spec rather than shipping the typo — the key name (`..._ACCEPT_...`) would want renaming with it.

### 3. BBCode injection through the sender's display name

`notification_text_helper.gd:28-35` interpolates the server-supplied `sender.name` straight into a BBCode string that lands on a `bbcode_enabled = true` `RichTextLabel`. A display name containing `[img]https://…[/img]`, `[url=…]` or `[color=…]` renders as markup in the notification title. Unclaimed names are free text, so this is reachable.

This is a carry-over — the old `NOTIF_TITLE_WANTS_TO_BE_YOUR_FRIEND_2/_3` entries substituted `{name}` into BBCode the same way — but this PR is now the only remaining site, so it's cheap to close here: escape the name (`sender_name.replace("[", "[lb]")`) before wrapping it in the colour tag.

### 4. `reward_assigned` is not a real notification type — three unreachable branches

`notifications_manager.gd::SUPPORTED_NOTIFICATION_TYPES` is a six-entry whitelist (`ENABLE_NOTIFICATION_FILTER = true`): `events_starts_soon`, `events_started`, `reward_assignment`, `reward_in_progress`, `social_service_friendship_request`, `social_service_friendship_accepted`. `reward_assigned` appears nowhere else in the repo. So the new `"reward_assigned"` arms in `notification_text_helper.gd:115` and `notification_content.gd:55`, plus the pre-existing body arm at `notification_text_helper.gd:294`, can never run. Worth dropping all three rather than carrying a second spelling forward.

(For what it's worth I traced the other side of this and it checks out: `reward_assignment` has no body arm, falls to the `_: return metadata.get("description", "")` default, and the fixture carries `description` — so "body = server description" works as the PR describes. Same for `events_starts_soon`.)

### 5. Orphans left behind by the rework

- `notification_toast.tscn` still declares `[sub_resource type="StyleBoxFlat" id="StyleBoxFlat_image"]`. Its only consumer was the `ImageContainer` override this PR deleted, so it is now dead.
- Seven PNGs under `godot/assets/ui/notifications/` are now referenced by nothing: `CoauthorNotification`, `FriendNotification`, `LandRentedNotification`, `ProposalFinishedNotification`, `RewardNotification`, `WorldAccessRestoredNotification`, `WorldUnaccessibleNotification`. Only `DefaultNotification.png` survives (still used by `alert_toast.gd:12`). Delete them together with their `.import` siblings.

### 6. Local editor state leaked into two scenes

- `notifications_panel.tscn:91` — `metadata/mobile_preview_orientation = "landscape"` (mobile-preview plugin state), plus `offset_right = 1002.0` on the root.
- `notification_item.tscn:20` — `offset_right = 715.0` / `offset_bottom = 135.0` on a `PanelContainer` that only ever lives inside the list `VBoxContainer`, so the offsets are inert but misleading. In the same node `custom_minimum_size = Vector2(280, 83)` is now stale: with the 80px avatar and the 22px margins on all four sides the row can't be shorter than ~124px.

Same class as the `project.godot` editor-state leaks from the earlier HUD PRs — worth a pass before merge.

### 7. Nothing re-renders the panel on a language change

`LabelTitle`, `LabelDescription` and `LabelTimestamp` are all `auto_translate_mode = 2` and filled from GDScript (`TranslationServer.translate` / `TranslationKey`), and neither `notification_item.gd` nor `notification_content.gd` implements `_notification(NOTIFICATION_TRANSLATION_CHANGED)`. Switching language with the Notifications panel open leaves every row in the old locale until a notification refresh rebuilds the list. Pre-existing shape, but this PR moves seven more strings onto that path — a one-liner calling `_update_ui()` from the hook would close it, and it also makes the Pseudolocale QA sweep in your test plan meaningful without a restart.

### 8. `NotificationsPanel` root stylebox lost its `bg_color`

`StyleBoxFlat_panel_bg` dropped `bg_color = Color(0, 0, 0, 0.8)` and now only carries the two right-side corner radii, so it falls back to Godot's default opaque white. It is invisible today only because `clip_children = 2` uses the panel as its own mask and the gradient `TextureRect` covers the full rect. That works, but it means any future change to the gradient child's layout exposes white. Setting the bg colour explicitly (transparent or the gradient's dark base) makes the intent readable.

---

## Test plan

The six cases are in the right shape — they start from opening the app, name the on-screen entry point (navbar bell), and each has an observable expected result, which is more than most PRs here manage. Two gaps worth adding, both aimed at the P1 above:

- An **event** case that names a specific live event and states what the thumbnail should look like when it loads vs. when the server sends no image — right now case 3 says "thumbnail with `#E8B9FF` rounded outline", which a tester would tick as passing even when the image is the empty placeholder, since the outline is present either way.
- A **read/unread regression** case: mark all as read with the button, reopen the panel, confirm the rows stay dimmed and the button is now disabled rather than hidden (that visible-vs-disabled swap in `notifications_panel.gd` is a user-visible behaviour change and isn't covered).

---

## Verdict

**Requesting changes** on finding 1 only — the event-payload key names need confirming before this can ship, because the failure mode is a silently empty image box on one of the four types the PR exists to deliver. Findings 2-8 are all small and none of them block; 2 (shipping copy) and 5 (orphans) are the ones I'd fold into the same push.

No security issues beyond finding 3. No SDK contract, CRDT, autoload-ordering or resource-lifetime concerns — the panel frees and rebuilds its items on every refresh, and toasts `queue_free()` themselves, so neither the `ProfilePicture` nor the `AsyncImage` instances are reused across notifications and there is no stale-texture or late-await-clobber path to worry about.

regenesis-claw (approved):
## Re-review — delta against the previous round (`a490dbf` → `ffd9f6c`)

Followed the repo's `REVIEW.md`. The PR base moved from the social branch to `main`, so this was reviewed as a file-level delta on the notification scope rather than a compare-range diff.

**All 8 findings from the previous round are resolved.** No P0/P1 remains — approving. The P2 list below is new material found in this round.

---

### ✅ Verified fixed

| Previous finding | Status |
|---|---|
| Event title/thumbnail on unconfirmed metadata keys (was the blocker) | Now degrades gracefully — `_is_loadable_url()` rejects `""`/`"https://"` and falls back to `DEFAULT_IMAGE`, and the title falls back to `NOTIF_HEADER_EVENT`. The `events_started` fixture confirms `metadata.title`. The image key is still unconfirmed by any in-repo payload, but the failure mode is a placeholder rather than an empty framed box, so it's no longer a blocker. |
| English copy (`New Item Receive`, `Accept your friend request`) | Fixed in all three catalogues — `New Item Received` / `Accepted your friend request`, matching es (`Aceptó`) and pt (`Aceitou`). |
| BBCode injection via `sender.name` | Escaped with `.replace("[", "[lb]")`. |
| `reward_assigned` unreachable branches | Removed from all 3 sites; whitelist is now exactly the 4 designed types (`events_starts_soon` dropped, `reward_in_progress` disabled with a reason). |
| Orphan `StyleBoxFlat_image` + 7 unreferenced PNGs | Sub-resource and all 7 PNGs (+ `.import`) deleted; grepped for every basename — zero remaining references. |
| Editor-state leak (`metadata/mobile_preview_orientation`, stray root offsets) | Both gone; `project.godot` is clean at `viewport_width=1600`. |
| Missing `NOTIFICATION_TRANSLATION_CHANGED` hook | Added in both the content and the item (the latter re-renders the relative timestamp). |
| `NotificationsPanel` root stylebox lost its `bg_color` | Restored. |

**i18n re-verified:** 82 `NOTIF*` rows in each of en/es/pt_BR, the two removed keys have zero references anywhere in `godot/` or `tools/`, and `translation_sources.json` hashes match all three catalogues.

**Also good in this round:** the `int()` cast in the `get_notifications()` sort (String timestamps would have raised at runtime); the id-keyed incremental item reuse, which is the right fix for the cancelled in-flight thumbnail load; and the `AsyncImage.background_color` default flip — I checked every other instance and only one unrelated scene sets that property, and the loaded/error colours for default instances are byte-identical to the old behaviour, so the shared atom change is non-breaking.

---

### 🟡 P2 — non-blocking

**1. Event and item titles still reach the BBCode label unescaped.** `_get_sender_name_colored()` escapes `[`, but the event branch returns `metadata.get("title", "")` and the reward branch returns a translated constant straight into `LabelTitle`, which is `bbcode_enabled`. Event titles are creator-authored, so the same escape applies. Consider hoisting the escape into a small `_escape_bbcode()` helper used by every branch that emits server data.

**2. `_items_by_id` collapses notifications that carry no `id`.** The key is `str(notif.get("id", ""))`, so two id-less notifications map onto `""` and only one row renders (the second `refresh()`es the first). The manager's own dedupe guards this case explicitly (`if "id" in notif`), so it's considered reachable there. A fallback key — index plus timestamp — closes it.

**3. The rarity backdrop can show square corners.** `ItemRarityBackground` is a plain `TextureRect` on `anchors_preset = 15` behind an `ItemImage` with `border_radius = 8`, and the parent `ImageContainerWrapper` is a `Control` with no `clip_children`. Unless the rarity PNGs already have transparent corners, the gradient's square corners will sit proud of the rounded frame. Worth an eyeball on device; a rounded `PanelContainer` (or `clip_children` on a rounded parent) would fix it if so.

**4. A language switch re-downloads every visible thumbnail.** The new `_notification()` hook calls the full `_update_ui()`, which re-runs `_update_image()` and therefore `load_from_url()` — the skeleton flashes and every image refetches. The item already models the cheaper split (`refresh_timestamp()` vs a full rebuild); a text-only path here would match it.

**5. The event route now depends on the deep-link host allowlist, and fails silently.** `_async_open_event_travel_modal()` reads the id via `DclParseDeepLink.parse_decentraland_link()`, which only accepts `decentraland.org`, `decentraland.zone` and `mobile.dclexplorer.com` — any other host returns a default result with empty params. The fixture link (`decentraland.org/jump/events?id=`) parses fine, but the code this replaced scanned for `?id=` host-agnostically and logged on failure. Both early returns here are silent, so a tap that produces nothing leaves no trace. Adding the two `printerr` lines back would keep that visible.

**6. Reward taps from a toast no longer sync the navbar.** `Global.open_navbar_silently.emit()` was dropped from the reward branch and the compensating `Global.close_navbar.emit()` lives only in the panel's click handler. The panel path is covered; `notification_toast.gd` also emits `Global.notification_clicked`, so a toast tap opens the backpack with the navbar left in whatever state it was in. Intentional is plausible — just flagging that the two entry points now diverge.

**7. Two stale sizing values survive.** `NotificationItem.custom_minimum_size = Vector2(280, 83)` against a row that is now ~124px (22px margins + an 80px image), and `NotificationsPanel.custom_minimum_size = Vector2(370, 500)` alongside `offset_right = 1002.0`.

**8. `LabelTitle` has no overrun behaviour.** It's a `fit_content` `RichTextLabel`, so a long event name or item label grows the row vertically instead of clipping — the description below already does the right thing with `clip_text` + `text_overrun_behavior = 3`. The `EllipsisNameLabel` atom introduced in the social work covers this pattern.

---

### One copy question

English now reads **"Accepted your friend request"**, which is what es/pt already said. If the design spec's literal string is "Accept your friend request", the two need reconciling one way or the other — flagging so it doesn't drift back on the next pass.

### CI

`Static checks` green. `Linux`, `Clippy` and `Android` were still running at review time — the previous round's red `Linux` was the known headless `test-avatar` teardown segfault (both scripts printed `PASS`, then `signal 11`), not a code failure; worth confirming this run goes green rather than assuming.

regenesis-claw (changes requested):
## Re-review — delta `ffd9f6c` → `299506a` (3 commits, ahead 3 / behind 0)

**All 8 P2s from the last round are cleared**, several of them structurally rather than locally — the title becoming a plain `Label` with a separate colour override kills the whole BBCode-escaping class instead of patching one call site, and moving the navbar collapse into the menu handler gives the panel and toast entry points real parity. Nice.

The third commit, though, reaches outside the notification scope, and that's where the problem is.

---

### 🔴 P1 — `texture_rect_rounded_px.gdshader` now squares the texture

```glsl
vec4 tex_color = texture(TEXTURE, UV);
float aa = fwidth(dist);
tex_color.a *= 1.0 - smoothstep(-aa, aa, dist);
COLOR *= tex_color;
```

The diagnosis behind this change is right — `COLOR = tex_color` was throwing the modulate away, which is exactly why dimmed read items stayed solid. But the incoming `COLOR` in a `canvas_item` fragment is **`modulate * texture(TEXTURE, UV)`**, not the modulate alone. Multiplying by a fresh sample applies the texture a second time:

```
rgb = modulate.rgb * tex.rgb * tex.rgb      // squared
a   = modulate.a * tex.a * tex.a * mask     // squared
```

Every image drawn through this shader comes out visibly darker and over-saturated, and semi-transparent pixels get doubly transparent.

The repo has its own proof of what `COLOR` contains on entry — `addition_shader.gdshader`, used by `backpack_landscape`, `backpack_portrait` and `menu.tscn`:

```glsl
// Double the node's own colour so the additive contribution is stronger.
COLOR.rgb *= 2.0;
```

That line only does anything because `COLOR` already carries the node's texture. `soft_ellipse.gdshader`, two folders over, applies a mask the way this one should:

```glsl
COLOR.a *= mask;
```

This matters beyond notifications: the shader/material is used by `profile.tscn`, `button_profile`, `snap_carousel_card.tscn` (the Discover cards) and now every `AsyncImage` texture and the rarity backdrop — so the regression lands repo-wide, not just on this feature.

The minimal fix keeps the modulate *and* drops the double-sample:

```suggestion
	float aa = fwidth(dist);
	// COLOR already carries modulate * TEXTURE; only carve the rounded corners out of its alpha.
	COLOR.a *= 1.0 - smoothstep(-aa, aa, dist);
```

(The `vec4 tex_color = texture(TEXTURE, UV);` line above then becomes unused and can go.)

---

### 🟡 P2

**1. The toast's friend avatar is now oversized and off-centre.** In the content scene `ProfilePictureFriend` gained `custom_minimum_size = Vector2(80, 80)` with `offset_left/top = -30` and `offset_right/bottom = 50` on a centre anchor — an 80×80 rect whose centre sits 10px down-right of the slot centre. The item scene overrides it back to a symmetric `-40 … 40`, but `notification_toast.tscn` only overrides `picture_size`, so the toast inherits an 80×80 avatar shifted +10,+10 inside a 50×50 wrapper. Either make the base offsets symmetric or add the override to the toast.

**2. Corner radii disagree in the item slot.** `ItemRarityBackground`'s shader material bakes `corner_radius_px = 8.0`, while `ItemImage` now uses `border_radius = 5`. The backdrop will round more than the image sitting on it.

**3. Editor-drag values back on the content root.** `NotificationContent` picked up `offset_right = 69.0` / `offset_bottom = 56.0`. Inert under a container, but it's the same local-state class you cleaned out of the two other scenes last round.

**4. `ItemRarityBackground` lost `visible = false` and gained a baked `BaseThumbnail` texture,** so it starts visible in both scenes that instance the content. `_update_image()` hides it, but that's one instantiation frame later — a base rarity square can flash behind friend and event notifications.

**5. The `AsyncImage` no-image state changed for its other consumer.** `TextureRect_NoImage` moved out of the (now hidden) `Panel` and went from shrink-centred to full-rect with `stretch_mode = 3`, and the error backdrop is gone with `NO_IMAGE_BACKGROUND`. On `snap_carousel_card.tscn` the fallback used to be a small centred glyph on a tinted panel; it'll now be a large glyph on whatever is behind the card. Worth eyeballing a Discover card with a dead thumbnail before this merges.

**6. `_apply_corner_radius()` only runs from `_ready()`,** so setting `border_radius` from code after the node is ready leaves the shader on the old value. Both current consumers set it in the scene, so this is latent — a setter would close it.

**7. The sibling `texture_rect_rounded.gdshader` still has the original `COLOR = tex_color`,** i.e. the same dropped-modulate bug you just fixed here. Out of scope for this PR, but worth a follow-up so the two don't behave differently.

---

### CI

`Static checks` and `Clippy` green; `Linux` and `Android` still running at review time.

Everything except the shader is cosmetic — happy to flip to approve as soon as the double-sample is out.

regenesis-claw (approved):
## Re-review — delta `299506a` → `51f982c` (4 commits, ahead 4 / behind 0)

Followed the repo's `REVIEW.md`. **Every finding from the last round is resolved, including the P1.** No P0/P1 remains — approving.

### ✅ Verified fixed

**P1 — the shader double-sample.** `texture_rect_rounded_px.gdshader` is now:

```glsl
float aa = fwidth(dist);
// COLOR already holds modulate * texture(TEXTURE, UV); only mask the corners on its alpha …
COLOR.a *= 1.0 - smoothstep(-aa, aa, dist);
```

The `tex_color` re-sample is gone, so the texture is applied once and the parent modulate survives — a dimmed read notification fades, and no image is squared. The sibling `texture_rect_rounded.gdshader` got the matching treatment (`COLOR.a = 0.0` in place of `COLOR = tex_color`).

I checked the blast radius on both, since these are shared surfaces: the non-px shader reaches `navbar_profile_button.tscn`, `profile_icon_button.tscn` and `joypad.tscn`, and the px one reaches `profile.tscn`, `button_profile`, `snap_carousel_card.tscn`, `async_image.tscn` and `notification_content.tscn`. None of those scenes — or their scripts — set `modulate` or `self_modulate`, so the change is a no-op on every existing screen and correct for the new dimmed-item case.

| Other findings | Status |
|---|---|
| Toast friend avatar oversized/off-centre | Base offsets are symmetric `-40 … 40` and the toast overrides to `-19 … 19` for the 38px size. `ProfilePicture._update_size()` rewrites `custom_minimum_size` from `picture_size`, so the inherited 80×80 doesn't leak into the toast. |
| Rarity backdrop radius mismatch | Material `corner_radius_px` 8.0 → 5.0, matching `ItemImage.border_radius`. |
| Editor-drag offsets on the content root | Removed; wrapper is a consistent 80×80. |
| `ItemRarityBackground` visible by default | `visible = false` restored. |
| `AsyncImage` no-image state | New `_apply_no_image_style()` brings the rounded `NO_IMAGE_BACKGROUND` panel back, and the `Panel` node moved ahead of the two `TextureRect`s so the backdrop draws *behind* the glyph rather than over it. |
| `_apply_corner_radius()` only ran from `_ready()` | `border_radius` is a setter now, re-applying the shader parameter and the current state's style behind an `is_node_ready()` guard. |

**Also good here:** the `_is_uuid()` check before the event id is concatenated into the events API path — that's real hardening, not just a shape check. And renaming `async_display_notifications` → `display_notifications` is the right call now that no `await` is left in it; I grepped for stale references and there are none.

---

### 🟡 P2 — nits only

**1. `Panel` in `async_image.tscn` still carries `clip_children = 2` but no longer has any children.** The two `TextureRect`s moved out to siblings, so the clip mode just costs a canvas-group allocation for nothing. Safe to drop.

**2. `_is_uuid()` compiles a fresh `RegEx` on every call.** A lazily-initialised static would avoid recompiling the pattern per notification tap. Also worth a thought: if the events API ever hands back a non-UUID id, the tap becomes a silent no-op — the `push_warning` makes that traceable, so this is more a note than a concern.

**3. `border_radius` gained a setter; `border_color` and `border_width` didn't.** Changing either of those from code still waits for the next load transition to take effect. Not a problem for the current consumers (both are set in the scene), just an asymmetry now that the pattern exists.

---

### CI

`Static checks` and `Clippy` green; `Linux` and `Android` still running at review time.

regenesis-claw (approved):
## Re-review — delta `51f982c` → `a4aa0bb` (1 commit, ahead 1 / behind 0)

Followed the repo's `REVIEW.md`. All three nits from the last round are fixed, nothing new. Approving — no findings this round.

| Nit | Fix |
|---|---|
| `clip_children = 2` on the now-childless skeleton `Panel` | Dropped. |
| `_is_uuid()` recompiling its `RegEx` per call | Cached in a lazily-initialised `static var _uuid_regex`. Matches the existing pattern in the repo (`glyph_map.gd`'s `_styled_cache`), and `static var` is used widely enough elsewhere (`mesh_renderer.gd`, `light_source_component.gd`) that there's no engine-version risk. |
| `border_radius` had a setter, `border_color`/`border_width` didn't | Both now re-apply the current state's style behind the same `is_node_ready()` guard. |

One thing worth noting since it's a behaviour change rather than a pure nit fix: `notification_content.gd::_show_item()` assigns `item_image.border_color` before calling `load_from_url()`, so the new setter now fires on every item render. In the fresh-instance case it just re-runs the loading style a moment early (harmless), and on a reused item it repaints the border with the new rarity colour immediately instead of waiting for the next load transition — which is the improvement you were after. No path where it fights the load sequence.

**CI:** `Static checks` green; `Linux`, `Clippy` and `Android` still running at review time.

ElzathRusso:
Feedback:

- [x] bottom margin, add 24px on the bottom in the las notification when finish the scroll

Screenshot 2026-09-07 at 6 08 19 PM

---

- [x] Event notification, text in the correct order
**title**: "Event name" 
**copy:** Event started!

Screenshot 2026-09-07 at 6 14 58 PM

- [x] Some events have emojis in the name; can we support emojis in the notifications?

---

- [x] Title colors: each type of notification has an assigned title color

Screenshot 2026-09-07 at 6 13 36 PM

ElzathRusso:
- [ ]  I received a Wearable; it shows in the backpack, but the notification does not show

---

- [ ] Can we reset the scroll when change the section and back again in notification?

https://github.com/user-attachments/assets/9adabd2e-e15b-4afe-ab85-8de5a8ae24c0


---

- [ ] emojis in Events names, The emojis look slightly vertically misaligned. Is that caused by the font, or can it be adjusted?

Screenshot_20260908_163000_Decentraland

ElzathRusso (approved):

PR #2829, opened 2026-09-04
feat(iap): instrument the credits flow

leanmendoza:
Two things: a generic `Request Result` Segment event wired through the credits flow, and the Option D gate fix that instrumentation exposed.

**Instrumentation.** `Request Result` (opt-in by `context`) is used by `iap_quote`, `iap_verify`, `iap_register` and the StoreKit purchase round-trip. A business refusal (HTTP 200 + `ok:false`) reports as not-ok with the server code; a rejected promise — which is every non-2xx — carries the server reason. Endpoints are normalised so no wallet reaches analytics. `Click Button` gains `OPEN_CREDITS`, tracked where the page is instantiated so all three shop entry points are covered (two had none), and `BUY_PACK`. All of them carry `Iap.analytics_context()`: resolved dclenv, whether the Option D switch ran and which gate stopped it, StoreKit environment/source, credits-server in use, balance and `products_loaded`.

**Gate fix.** `_apply_storekit_env` gated on `OS.is_debug_build()`, which reflects the Godot export template — and the iOS export runs without `--release`, so it was true on every shipped build and the sandbox env switch never ran on TestFlight or App Review. Now gated on `Global.is_dev()`, so `-prod` and `-staging` switch while dev builds keep the developer chosen env.

**Why it matters.** Left on `.org`, `POST /credits/iap/quote` answers `unknown_product` for the `credits_tier_b*` packs the storefront sells — the org catalog does not have them, zone does. That is the failure App Review hit: the shop opens, the pack does nothing, and no telemetry recorded any of it.

## Test plan

1. **Gate fires (needs a `-prod`/`-staging` artifact — cannot be done locally, release requires distribution provisioning).** Install from TestFlight on a device signed into a sandbox Apple ID. Open the credits shop. Expect the packs to load and the balance to render, and in Segment an `OPEN_CREDITS` whose `extra_properties` shows `option_d_applied: true` with `credits_server` on `.zone`.
2. **Purchase completes.** Press a pack. Expect `BUY_PACK` -> `iap_quote ok` -> `iap_verify ok` -> `iap_purchase ok`, and the balance to increase by the pack amount.
3. **Failure is legible.** Press a second pack past the daily cap. Expect `iap_quote` `ok:false` `error: cap_exceeded` and a matching `iap_purchase` failure, not a silent no-op.
4. **Regression, production users.** On a build installed from the App Store, `option_d_skip_reason` must read `not_sandbox` and `credits_server` must stay on `.org`.

Segment debugger tip: its search is plain substring matching, so `option_d_skip_reason` filters to exactly these events.

## Verification so far

Device-verified on a `-dev` build: both shop entry clicks, `iap_register`, a denied quote propagating into a failed purchase, a full successful purchase (`iap_quote` -> `iap_verify` -> `iap_purchase`) once forced onto the Option D env by a baked deeplink, and the `cap_exceeded` path.

Not exercised: the gate itself letting a `-prod`/`-staging` build through, the `cancelled`/`pending` outcomes, and the `marketplace_cta` / `wearable_item` entry points.

### DIFF
--- a/godot/src/iap/iap_manager.gd
+++ b/godot/src/iap/iap_manager.gd
@@ -408,7 +408,13 @@ func _track_request(
 func _on_purchase_outcome(
 	outcome: String, product_id: String, reason: String, credits: int
 ) -> void:
-	var extra := {"product_id": product_id, "outcome": outcome}
+	# purchase_completed also fires for transactions StoreKit redelivers at launch, which
+	# never went through purchase(); flag them so they aren't counted as presses.
+	var extra := {
+		"product_id": product_id,
+		"outcome": outcome,
+		"user_initiated": _purchase_started_ms > 0,
+	}
 	if credits >= 0:
 		extra["credits_granted"] = credits
 	# 0 when the press never reached purchase(); report the outcome anyway.
@@ -949,7 +955,11 @@ func _async_signed_iap(path: String, method: int, body: String, context: String 
 		# after a purchase would emit that many identical error events. The retry loop
 		# already recovers from it.
 		print("[IAP] ", path, " transport/auth error: ", response.get_error())
-		_track_request(context, endpoint, method_name, false, -1, "transport", started)
+		# The promise is rejected for ANY non-2xx, so this is not only transport
+		# failures: the reason carries the server body, or the canonical status text
+		# when it was empty. Truncated like the unparseable-body log below.
+		var reason := str(response.get_error()).substr(0, _ERROR_BODY_EXCERPT_CHARS)
+		_track_request(context, endpoint, method_name, false, -1, reason, started)
 		return null
 	var status: int = response.status_code()
 	var json = response.get_string_response_as_json()
--- a/godot/src/iap/iap_manager.gd
+++ b/godot/src/iap/iap_manager.gd
@@ -76,8 +76,7 @@ const _CREDITS_BY_PRODUCT := {
 	"credits_tier_b3": 260,
 }
 
-# HTTP verb names for the Request Result event. Godot has no built-in mapping and
-# only these two are used by the IAP endpoints.
+# Godot has no built-in verb mapping; only these two are used here.
 const _METHOD_NAMES := {
 	HTTPClient.METHOD_GET: "GET",
 	HTTPClient.METHOD_POST: "POST",
@@ -185,16 +184,11 @@ var _transaction_history: Array = []
 var _overlay_token: int = 0
 var _overlay: CanvasLayer = null
 var _purchase_in_flight: bool = false
-# Authoritative StoreKit environment and how it was determined, retained after
-# _on_environment_resolved so analytics_context() can report which backend this
-# device is actually pointed at.
+# Retained from _on_environment_resolved for analytics_context().
 var _env_authoritative: String = ""
 var _env_source: String = ""
 var _can_make_payments: bool = false
-# Whether _apply_storekit_env switched to Option D, and — when it did not — which
-# gate stopped it. A sandbox device that never switched is talking to the
-# production credits-server and cannot complete a purchase; nothing else we emit
-# records that.
+# Whether _apply_storekit_env switched, and which gate stopped it if not.
 var _option_d_applied: bool = false
 var _option_d_skip_reason: String = ""
 # Time.get_ticks_msec() at the last purchase() press, for the outcome's duration.
@@ -240,7 +234,6 @@ func report_environment_to_analytics() -> void:
 func _on_environment_resolved(environment: String, source: String, resolve_ms: float) -> void:
 	var env_at_ms := Time.get_ticks_msec() - Global._startup_time
 	var matched := environment == _env_sync_value
-	# Retained for analytics_context().
 	_env_authoritative = environment
 	_env_source = source
 	_can_make_payments = _store_kit.can_make_payments()
@@ -298,9 +291,7 @@ func _on_environment_resolved(environment: String, source: String, resolve_ms: f
 # before this has applied.
 #
 # Gated on two conditions so it never overrides a developer/QA choice:
-#  - PRODUCTION builds only (release export template). A debug/dev build run on a
-#    device with a sandbox Apple ID also reports `sandbox`, but there the dev's
-#    chosen env (default org or --dclenv) must be respected.
+#  - Not on `-dev` builds: there the dev's chosen env (default org or --dclenv) wins.
 #  - Only an EXPLICIT `dclenv=org` exempts from the switch. Passing
 #    `decentraland://open?dclenv=org` (or `--dclenv org`) means "really use the prod
 #    .org backend" and wins. Any OTHER explicit env is not special-cased here — it
@@ -310,15 +301,13 @@ func _on_environment_resolved(environment: String, source: String, resolve_ms: f
 # environment on every launch.
 func _apply_storekit_env(environment: String) -> void:
 	if environment != "sandbox":
-		# The correct no-op: a real App Store install belongs on .org.
 		_option_d_skip_reason = "not_sandbox"
 		return
-	# Only production (release export template) builds auto-switch. A debug/dev build
-	# run on a device with a sandbox Apple ID also reports "sandbox", but there we must
-	# respect the dev's chosen env (org by default or --dclenv).
-	if OS.is_debug_build():
-		print("[IAP] StoreKit sandbox on a debug build — not forcing the Option D env")
-		_option_d_skip_reason = "debug_build"
+	# NOT OS.is_debug_build(): the iOS export always uses the debug template, so that
+	# check disabled this switch on every shipped build, review included.
+	if Global.is_dev():
+		print("[IAP] StoreKit sandbox on a dev build — not forcing the Option D env")
+		_option_d_skip_reason = "dev_build"
 		return
 	var current := str(DclGlobal.get_dcl_environment())
 	# The only explicit override that exempts from the switch is dclenv=org
@@ -363,13 +352,8 @@ func async_await_env_resolved(timeout_sec: float = 5.0) -> void:
 		)
 
 
-## The backend-selection snapshot, as a JSON string for an analytics event's
-## `extra_properties`. Attached to every credits click and to every IAP request
-## result, because the question those events exist to answer is which backend the
-## device was talking to — and `option_d_applied` is the field that answers it.
-##
-## Safe to call before the StoreKit environment has resolved: the storekit_* fields
-## are simply empty until then.
+## Which backend this device is pointed at, as JSON for an event's `extra_properties`.
+## Safe before the StoreKit env resolves — the storekit_* fields are empty until then.
 func analytics_context() -> String:
 	return (
 		JSON
@@ -402,8 +386,7 @@ func _track_request(
 	started_ms: int,
 	extra: Dictionary = {}
 ) -> void:
-	# Opt-in: a call site that passes no context reports nothing. Keeps the event a
-	# set of deliberately-watched calls rather than a traffic log.
+	# Opt-in: no context, no event.
 	if context.is_empty() or Global.metrics == null:
 		return
 	var payload := JSON.parse_string(analytics_context()) as Dictionary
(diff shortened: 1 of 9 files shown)


godot/src/iap/iap_manager.gd
```diff
@@ -830,14 +939,19 @@ func _async_signed_iap(path: String, method: int, body: String) -> Variant:
...
+	var method_name := _METHOD_NAMES.get(method, "OTHER") as String
+	var started := Time.get_ticks_msec()
 	var response = await Global.async_signed_fetch(url, method, body, true)
 	if response is PromiseError:
 		# `print`, not `printerr`: `_async_poll_balance_after_purchase` calls this
 		# _POST_PURCHASE_POLL_ATTEMPTS times, so a phone that drops connectivity right
 		# after a purchase would emit that many identical error events. The retry loop
 		# already recovers from it.
 		print("[IAP] ", path, " transport/auth error: ", response.get_error())
+		_track_request(context, endpoint, method_name, false, -1, "transport", started)
```
regenesis-claw: [P1] Every non-2xx lands here, not just transport failures. `send_result_to_promise` rejects the promise for any status outside 200-299 (`lib/src/http_request/request_response.rs:169-186`), so a 401 (bad signature), 403 (wallet not allowed) and a 500 from credits-server all report as `ok:false, status:null, error:"transport"` — identical to a dropped connection. That is the one distinction that matters if a purchase flow breaks in App Review.

The rejection message already carries the server's response body (or the canonical reason phrase when the body is empty), and it's currently only `print`ed on the line above. Passing it through gives the discriminator back:

```suggestion
		var reason := str(response.get_error()).substr(0, _ERROR_BODY_EXCERPT_CHARS)
		_track_request(context, endpoint, method_name, false, -1, reason, started)
```

Same truncation you already apply to the unparseable-body log, so a wallet-scoped body can't spill into analytics wholesale. If you'd rather keep the value low-cardinality, bucketing on the canonical reason works too — but `"transport"` for an HTTP error is a wrong attribution, not just a coarse one.

lib/src/analytics/data_definition.rs
```diff
@@ -658,6 +659,34 @@ pub struct SegmentEventGuestWalletCreation {
...
+    // What was attempted: "iap_quote" | "iap_verify" | "iap_purchase" | ... .
+    pub context: String,
+    // Normalised path, never a full URL: no wallet addresses, no ids.
+    pub endpoint: String,
+    // "GET" | "POST" | ... . "STOREKIT" for the native purchase round-trip.
+    pub method: String,
+    // False covers transport failure, non-2xx, unparseable body, and a business
+    // refusal (HTTP 200 + ok:false) — the same outcome from the caller's side.
+    pub ok: bool,
+    // HTTP status when there was one. None on a transport failure or a non-HTTP call.
```
regenesis-claw: [P2] As implemented, `status` is `None` for *every* HTTP failure, not just transport ones — the GDScript caller passes -1 on the rejected-promise path, which is where all non-2xx responses arrive. Either the comment should say so, or the caller should recover the status (see the comment on `iap_manager.gd:952`).

godot/src/ui/components/organisms/menu/menu.gd
```diff
@@ -155,9 +155,13 @@ func async_show_discover(open_menu := true):
...
 
 
-func async_show_credits():
+func async_show_credits(source: String = "unknown"):
 	if is_instance_valid(_credits_page):
 		return
+	# Tracked here, not per button: the one place the page is instantiated, so a new
+	# entry point reports itself for free.
+	if Global.metrics != null:
+		Global.metrics.track_click_button("OPEN_CREDITS", source, Iap.analytics_context())
```
regenesis-claw, line 164: [P2] `source` goes into the `screen_name` slot, but the values are lowercase snake (`balance_button`, `marketplace_cta`, `wearable_item`, `unknown`) while every other `Click Button` in the credits flow uses SCREAMING_SNAKE screen names (`CREDITS_BALANCE_BUTTON`, `CREDITS_SHOP`). Anyone grouping `Click Button` by `screen_name` gets two naming schemes in one column. Consider uppercasing the three source constants, or moving the entry point into `extra_properties` and keeping a fixed `CREDITS_PAGE` screen name.

godot/src/ui/components/molecules/credits_balance_button/credits_balance_button.gd
```diff
@@ -50,4 +50,4 @@ func _on_balance_changed(new_balance: int) -> void:
 
 func _on_pressed() -> void:
 	Global.metrics.track_click_button("BUTTON_CREDITS", "CREDITS_BALANCE_BUTTON", "")
-	Global.open_credits.emit()
+	Global.open_credits.emit("balance_button")
```
regenesis-claw: [P2] This entry point now emits two `Click Button` events per press: `BUTTON_CREDITS` here and `OPEN_CREDITS`/`balance_button` from `async_show_credits`. The other two entry points emit only one. If someone counts "credits shop opens" off `Click Button` without filtering on `button_text`, the balance button is double-weighted. Dropping the older `BUTTON_CREDITS` line (or keeping it and noting the overlap where the funnel is defined) avoids that.

godot/src/iap/iap_manager.gd
```diff
@@ -360,6 +455,15 @@ func enable() -> void:
 	_store_kit.start_listening()
 	_store_kit.load_products(PRODUCT_IDS)
 
+	purchase_completed.connect(
+		func(pid: String, credits: int): _on_purchase_outcome("completed", pid, "", credits)
```
regenesis-claw, line 465: [P2] `purchase_completed` also fires for transactions StoreKit redelivers at launch (the `transaction_updated` path), not only for a press that went through `purchase()`. Those report as `iap_purchase` / `outcome: completed` with `duration_ms` ~0 because `_purchase_started_ms` is 0, so a redelivery is counted as a successful purchase in the same bucket as a real one. You can still separate them after the fact on `duration_ms == 0`, but an explicit `user_initiated: _purchase_started_ms > 0` in `extra` would make that intentional rather than incidental.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`. Base is `release`, branch is up to date with it (behind_by 0). Static checks green; Linux / Clippy / Android still running at review time.

**Pre-flight — iOS build (REVIEW.md §0.2):** No iOS build on this PR — a maintainer can add the `build` label to trigger one. I will not approve platform-sensitive iOS changes without a green iOS build. This matters more than usual here: commit 2 changes a code path (`_apply_storekit_env`) that *by construction* never executes on a `-dev` build, so the only way to exercise it is a `-prod`/`-staging` iOS artifact on device. The device verification listed in the description was against commit 1's instrumentation, not against the gate change.

### Verified against source

- **The `is_debug_build()` -> `is_dev()` claim holds.** `lib/build.rs` picks the env suffix from `DECENTRALAND_PROD_BUILD` / `DECENTRALAND_STAGING_BUILD`, and `.github/actions/set-prod-flag` sets `--prod` for any `release*` branch and `--staging` for everything else, so every CI-produced iOS artifact is `-prod`/`-staging` (`is_dev() == false` -> the switch runs) while a local `cargo run -- build --target ios` with no flag is `-dev` (`is_dev() == true` -> the dev's env is respected). The old `OS.is_debug_build()` gate was indeed always true on the iOS export. Claim verified.
- **`Global.is_dev()` resolves.** `is_dev` is a static `#[func]` on `DclGlobal`; `global.gd` extends `DclGlobal` and the same instance-call form is already used (`Global.is_production()` in `debug_ws_server.gd`, `touch_feedback_overlay.gd`).
- **`open_credits(source)` signal change is complete.** Three emit sites, all updated; the single connection (`menu.gd:96` -> `async_show_credits(source := "unknown")`) accepts the argument. No `.tscn` connections to that signal.
- **"All three entry points covered" holds.** `_credits_page` is instantiated in exactly one place (`menu.gd:170-171`), and every path there goes through the signal.
- **No wallet reaches analytics.** `_normalize_endpoint` folds the address, and in practice the only contexts passed are the three static `/credits/iap/*` paths — the wallet-scoped balance calls (`iap_manager.gd:831,867`) pass no context and emit nothing. `analytics_context()` carries env/source/balance/products only.
- **Consumer impact:** `Request Result` is a new, purely additive Segment event; no existing event shape changes. Nothing downstream in this repo reads it. Metabase/Segment-side dashboards are outside what I can verify — worth a heads-up to whoever owns the event catalogue.
- **Security:** No security issues found. No secrets, no new network surface, no injection surface; events go through `queue_event`, so the EULA consent gate still applies.

### Findings

**[P1] Every non-2xx HTTP response is reported as `error: "transport"` with no status** — see the inline comment on `iap_manager.gd:952`. `send_result_to_promise` (`lib/src/http_request/request_response.rs:169-186`) rejects the promise for any non-2xx, so a 401/403/500 from credits-server lands in the `PromiseError` branch and is indistinguishable in Segment from a dropped connection. That is precisely the discrimination this PR exists to provide if Apple bounces us, and it also contradicts the field doc in `data_definition.rs:677` ("HTTP status when there was one"), which the code can never satisfy for HTTP failures. ~3 lines to fix.

Three P2s inline (screen_name casing, the double event on the balance button, redelivered transactions counting as purchases).

**[P2] No `## Test plan` (REVIEW.md §4/§18).** The description's device-verified / not-exercised split is genuinely useful, but QA runs this by hand off TestFlight and there are no executable cases. Given the gate change, the one case worth writing out is: open the app on a TestFlight build signed into a sandbox Apple ID, open the credits shop from the balance button, and confirm the packs load and the balance shows — the observable proof that the Option D switch fired.

**[P2] Landing a feature commit directly on `release`.** REVIEW.md notes `release` PRs are normally merge-only. Both commits (especially the `is_dev` fix) need to reach `main` too, or the next `main -> release` cut silently reverts them. Confirm there's a companion PR or a back-merge planned.

kuruk-mm (approved):

regenesis-claw (approved):
Re-reviewed at `8fdd260` (delta: one commit, 22 additions over 6 files). All five findings from the previous round are addressed.

| Finding | Status |
|---|---|
| **[P1]** non-2xx bucketed as `error: "transport"` | **Fixed** — `iap_manager.gd:961` now passes `str(response.get_error())` truncated to `_ERROR_BODY_EXCERPT_CHARS` (200), the same excerpt cap used by the unparseable-body log. A 401/403/500 now carries the server body (or the canonical status text when it was empty), so it is distinguishable from a dropped connection. |
| **[P2]** `status` field doc unsatisfiable for HTTP failures | **Fixed** — `data_definition.rs:677` now states `None` covers every failure surfaced as a rejected promise, i.e. every non-2xx, and points at `error`. |
| **[P2]** `screen_name` casing split | **Fixed** — all three sources and the default are SCREAMING_SNAKE (`BALANCE_BUTTON` / `MARKETPLACE_CTA` / `WEARABLE_ITEM` / `UNKNOWN`), matching the existing `Click Button` values. |
| **[P2]** double `Click Button` on the balance button | **Resolved as documented** — `BUTTON_CREDITS` kept for continuity with the pre-`OPEN_CREDITS` series, with a comment stating shop opens are counted on `OPEN_CREDITS`. Fine by me; the overlap is now written down where someone building the funnel will see it. |
| **[P2]** redelivered transactions counted as purchases | **Fixed** — `user_initiated: _purchase_started_ms > 0` on the outcome's `extra`, so a launch redelivery is separable from a press. |
| **[P2]** no test plan | **Fixed** — four cases, each with an observable expected result, and case 1 states up front that it needs a TestFlight artifact. The `-dev` verification list (successful purchase forced onto the Option D env with a baked deeplink, `cap_exceeded`) is a real addition over the previous description. |

The `unknown_product` explanation added to the description is the missing piece from round 1 — it makes clear the gate fix is the actual App Review failure and not a cleanup.

**CI:** all green at this head — Linux, Clippy, Android, Docker, Static checks.

**Still unverified (not blocking, but worth stating):**
- **No iOS build on this PR** (no `build` / `build-ios` label). The gate change by construction only executes on a `-prod`/`-staging` artifact, so nothing in this repo's CI exercises it. Test-plan case 1 is the only proof it works — running it on the TestFlight build cut from `release` before submitting to Apple is the last check I can't do for you.
- `error` can now carry up to 200 characters of server response body. For the three tracked endpoints that is credits-server error JSON, so it should stay low-cardinality and free of user data — worth a glance in the Segment debugger on the first real failure to confirm no wallet-scoped field is echoed back in a body. The doc comment on the `error` field still lists `"transport"` as an example value, which the code no longer emits; harmless, fix it if you touch the file again.
- Both commits need a path back to `main`, or the next `main -> release` cut reverts the gate fix.

PR #2834, opened 2026-09-07
fix(auth): sign requests with an NTP-corrected clock

leanmendoza:
## Why

Services verify `x-identity-timestamp` against their own clock inside a ±60 s window, so a device whose clock drifts by a minute gets rejections that read as signature failures.

App Review hit exactly this on 2026-09-07 (build `1.13.0.1804`, iPad Air M3). The reviewer's device ran ~60 s fast and the credits server answered:

```
{"error":"Signature timestamp is too far in the future:
          signature timestamp: 1788749031686,
          local  timestamp:    1788748971345"}
```

The `iap_quote` was refused, so `iap_purchase` reported `outcome: failed`, `error: "quote failed"` and StoreKit was never reached — no payment sheet, and a Guideline 2.1(b) rejection. Sitting right on the boundary made it intermittent: two `iap_register` calls in the same minute went through, the quote did not.

## What

A minimal SNTP client in `lib/src/utils/clock.rs`. No new dependency — tokio already provides `UdpSocket`.

- `unix_time_ms()` — device clock plus the learned offset; drop-in for the previous pattern.
- `spawn_background_sync()` — sync at startup, refresh every 15 min, retry after 60 s on failure. Started from `DclGlobal::init`.
- Servers tried in order (Cloudflare, Google, Apple, pool.ntp.org), 3 s timeout each, first valid reply wins.
- A reply is only trusted if it echoes the originate timestamp we sent (RFC 4330 §5), comes from a synchronised server (mode 4, stratum ≠ 0, leap indicator ≠ 3), and yields an offset under 24 h. Anything else falls through to the next server. The offset feeds every signed request, so one bad datagram must not be able to move it.

**Fallback:** if nothing answers the offset stays 0 and signing behaves exactly as it does today. A later failure keeps the last known offset rather than discarding it.

Applied to all six sites that build an `x-identity-timestamp`, not only the credits path — they all face the same window:

| File | Signs |
|---|---|
| `auth/wallet.rs` `sign_request` | signed fetch (credits / IAP) |
| `auth/wallet.rs` `sign_pulse_connect` | Pulse handshake |
| `auth/deploy_profile.rs` | profile deploy |
| `profile/profile_service.rs` | deployment timestamp |
| `comms/signed_login.rs` | signed login |
| `social/friends.rs` | social service auth chain |

## Test plan

- [x] `cargo fmt --all`, `cargo clippy -- -D warnings` clean
- [x] New reply validation checked against live Cloudflare/Google/Apple servers: all three pass every guard, including the verbatim originate echo.
- [x] `cargo test --lib` — 281 pass. One added case: a device 90 s ahead (outside the window, would be rejected) lands back inside it after correction, with the NTP wire encoding round-tripping and the network round trip cancelling.
- [x] **On device, iOS and Android.** UDP to port 123 can be blocked by the network or carrier. If that happens the fallback holds and behaviour is unchanged, but the fix does nothing — this needs confirming on real hardware before it can be relied on.
- [x] Verify a purchase completes on a device with a deliberately skewed clock.

## Open question

What the server validates is *its* clock, not true time. Taking the offset from the `Date` header of a response from the credits server itself would give the exact reference that matters, travel over the HTTP stack that already exists, and not depend on UDP at all. NTP gives true time, which is nearly the same thing unless the server drifts.

The offset is abstracted behind the module, so adding that as a primary source with NTP as backup is a small follow-up. Left out here to keep the change reviewable.


lib/src/utils/clock.rs
```diff
@@ -0,0 +1,206 @@
...
+    if response[1] == 0 {
+        return Err("kiss-o'-death".to_owned());
+    }
+
+    let t2 = read_ntp_timestamp(&response[32..40]);
+    let t3 = read_ntp_timestamp(&response[40..48]);
+    if t2 == 0 || t3 == 0 {
+        return Err("empty timestamp".to_owned());
+    }
+    Ok(((t2 - t1) + (t3 - t4)) / 2)
```
regenesis-claw: [P1] Three validations are missing before this offset is trusted, and the offset gates every authenticated request in the app.

1. **Originate timestamp is never checked.** RFC 4330 §5 requires the client to verify that the reply's originate field (`response[24..32]`) equals the transmit timestamp it sent (`t1`). Without it, any reply that arrives from the connected 5-tuple is accepted — a stale duplicate, or a spoofed datagram from an off-path attacker who only has to guess the ephemeral source port. With the check, a spoofer additionally has to know `t1`.
2. **Leap indicator 3 is accepted.** `response[0] >> 6 == 3` means the server's own clock is unsynchronised and its timestamps must be ignored. Only mode and stratum are checked today.
3. **No bound on the result.** Any magnitude is stored. A single bad reply — including from a `pool.ntp.org` volunteer server, which is a documented real-world occurrence, not just a hypothetical attacker — moves the signed timestamp arbitrarily for up to `REFRESH_INTERVAL`. That is a failure the device clock alone cannot produce, so the change adds a new way to break signing.

Suggested, keeping the existing structure:

```rust
    if response[0] >> 6 == 3 {
        return Err("server clock unsynchronised".to_owned());
    }

    if read_ntp_timestamp(&response[24..32]) != t1 {
        return Err("originate timestamp mismatch".to_owned());
    }

    let t2 = read_ntp_timestamp(&response[32..40]);
    let t3 = read_ntp_timestamp(&response[40..48]);
    if t2 == 0 || t3 == 0 {
        return Err("empty timestamp".to_owned());
    }

    let offset = ((t2 - t1) + (t3 - t4)) / 2;
    if offset.abs() > MAX_PLAUSIBLE_OFFSET_MS {
        return Err(format!("implausible offset {offset} ms"));
    }
    Ok(offset)
```

Note `write_ntp_timestamp` floors and `read_ntp_timestamp` rounds, so the equality in (2) needs the reply echoed verbatim — comparing the raw 8 bytes against the ones written into `request[40..48]` is more robust than comparing the decoded millisecond value. `MAX_PLAUSIBLE_OFFSET_MS` of a few hours is generous enough for any real timezone/RTC fault while still rejecting garbage; rejecting falls through to the next server, which is the behaviour you already want.

lib/src/utils/clock.rs
```diff
@@ -0,0 +1,206 @@
...
+    request[0] = 0x23; // LI 0, version 4, mode 3 (client)
+    let t1 = device_unix_time_ms();
+    write_ntp_timestamp(&mut request[40..48], t1);
+    socket
+        .send(&request)
+        .await
+        .map_err(|e| format!("send: {e}"))?;
+
+    let mut response = [0u8; 48];
+    let read = timeout(QUERY_TIMEOUT, socket.recv(&mut response))
```
regenesis-claw: [P2] `QUERY_TIMEOUT` covers `recv` only. `socket.connect(server)` takes a `&str`, so it resolves the hostname through the system resolver on a blocking pool thread, and that lookup is unbounded — on a captive portal or a network with a black-holed resolver it can sit for tens of seconds per server, so "3 s timeout each" in the description doesn't hold for the connect leg and `sync_now()` can take minutes to walk all four.

Impact is bounded (background task, nothing awaits it), but it's cheap to make the description true: wrap the whole call instead of just the receive.

```rust
match timeout(QUERY_TIMEOUT, query(server)).await {
    Ok(Ok(offset)) => { /* ... */ }
    Ok(Err(e)) => { /* ... */ }
    Err(_) => { /* timed out */ }
}
```

lib/src/utils/clock.rs
```diff
@@ -0,0 +1,206 @@
...
+        return Err("empty timestamp".to_owned());
+    }
+    Ok(((t2 - t1) + (t3 - t4)) / 2)
+}
+
+fn device_unix_time_ms() -> i64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_millis() as i64)
+        .unwrap_or(0)
```
regenesis-claw, line 178: [P2] `unwrap_or(0)` swallows a case that two call sites used to propagate: `sign_pulse_connect` returned `Err(e.to_string())` and `friends.rs::build_auth_chain` used `?` on the `SystemTimeError`. A clock set before 1970 now signs with timestamp 0 and the request comes back as an opaque signature failure — the same unreadable symptom this PR is removing.

It self-heals once a sync lands (offset would be ~epoch-sized and `unix_time_ms()` lands on real time), so this only bites in the unsynced window, and a pre-epoch clock is rare. Worth either a `tracing::warn!` here or a mention in the description as an intentional behaviour change.

lib/src/utils/clock.rs
```diff
@@ -0,0 +1,206 @@
...
+        let server_t2 = 1_788_749_031_686i64;
+        let server_t3 = server_t2 + 1;
+        let t1 = server_t2 + skew - 20;
+        let t4 = server_t3 + skew + 20;
+
+        let mut buf = [0u8; 8];
+        write_ntp_timestamp(&mut buf, server_t2);
+        assert_eq!(read_ntp_timestamp(&buf), server_t2);
+
+        let offset = ((server_t2 - t1) + (server_t3 - t4)) / 2;
```
regenesis-claw: [P2] This line recomputes `((t2 - t1) + (t3 - t4)) / 2` rather than calling the code that computes it, so the test can't catch a sign error or a swapped term in `query()` — the only production code it actually exercises is the `write_ntp_timestamp`/`read_ntp_timestamp` round trip above.

Extracting the arithmetic makes the test bind to the real thing:

```rust
fn offset_from(t1: i64, t2: i64, t3: i64, t4: i64) -> i64 {
    ((t2 - t1) + (t3 - t4)) / 2
}
```

and `query()` returns `Ok(offset_from(t1, t2, t3, t4))`. While you're in here, the reply-validation branches (mode, stratum 0, and the checks proposed on `query()`) have no coverage either — they're pure functions of a 48-byte buffer, so a table test over a few crafted responses is cheap and is where a spoofing regression would show up.

lib/src/utils/clock.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Use this for any timestamp a server verifies against its own clock.
+pub fn unix_time_ms() -> u128 {
+    device_unix_time_ms()
+        .saturating_add(OFFSET_MS.load(Ordering::Relaxed))
+        .max(0) as u128
+}
+
+/// How far the device clock is behind real time, in milliseconds. Positive means the device is
+/// slow. 0 while unsynced.
+pub fn offset_ms() -> i64 {
```
regenesis-claw, line 53: [P2] `offset_ms()` and `sync_state()` are public but nothing calls them.

Given why this PR exists — App Review's device drifted and it surfaced as a signature failure nobody could read — the offset is exactly the value we'd want in the field. Reporting it once per successful sync to Segment/Sentry (alongside `sync_state`) would tell us whether the SNTP path actually works on real carrier networks, which is the open risk you flag in the test plan, and would quantify how common clock skew is among real users.

Separately: nothing re-syncs when a server *does* answer "Signature timestamp is too far in the future". Calling `sync_now()` on that specific error would close the loop in one retry instead of waiting up to `REFRESH_INTERVAL`, and that response is also the one place where the server's own clock — the reference your open question is about — is directly observable.

regenesis-claw (changes requested):
## Review — NTP-corrected signing clock

Followed `REVIEW.md`. Pre-flight: branch is **up to date with `main`** (ahead 1, behind 0); all CI green including `🍏 iOS → R2` and `🤖 Android` — an iOS artifact is present, which this diff needs since it touches `lib/src/comms/`.

**The diagnosis and the shape of the fix hold up.** I checked the six call sites against the code rather than the table: `x-identity-timestamp` is emitted from exactly `auth/wallet.rs:260` (pulse), `auth/wallet.rs:317` (signed fetch), `comms/signed_login.rs:77`, `social/friends.rs:32`, plus the two deployment timestamps in `auth/deploy_profile.rs` / `profile/profile_service.rs` — no seventh site was missed, and the Pulse handshake reaches it through `sign_pulse_connect` (confirmed via `pulse_room.rs`). I also checked the *other* server-verified clock value the PR doesn't touch: the ephemeral auth chain's `Expiration:` line still comes from the raw device clock (`auth/identity` paths), but `AUTH_CHAIN_EXPIRATION_SECS` is 30 days, so a ±60 s skew is irrelevant there — correctly out of scope.

The fallback claim checks out: offset stays 0 when nothing answers, `sync_now()` keeps the last good offset on a later failure (`compare_exchange` from `SYNC_PENDING` only), and `unix_time_ms()` is a byte-for-byte drop-in for the old expression. The wire encode/decode round-trips correctly, including the deliberate rounding in `read_ntp_timestamp`.

What blocks is not the approach, it's how much a single unauthenticated UDP packet is now trusted.

### Blocking

**[P1] The SNTP reply is accepted with no origin check, no leap-indicator check and no sanity bound.** Details inline on `query()`. Net effect: one bogus reply — spoofed, stale, or just a misbehaving `pool.ntp.org` member — silently shifts the timestamp on *every* signed request by an arbitrary amount for up to 15 minutes, which is the exact failure mode this PR exists to remove, and it's a failure mode the device clock alone could not produce. The fix is ~10 lines and doesn't change the design.

### Non-blocking

- **[P2]** `QUERY_TIMEOUT` only wraps `recv`, so the "3 s timeout each" in the description is unverified for DNS/connect/send.
- **[P2]** `device_unix_time_ms()` swallows a pre-epoch clock where two call sites previously propagated an error.
- **[P2]** The added test recomputes the offset formula by hand instead of exercising the code under test.
- **[P2]** `offset_ms()` / `sync_state()` are public with no consumers — no field telemetry, and nothing re-syncs when a server actually answers "timestamp is too far in the future".

### Unverified

- "Tested on device, purchase works with the clock 15 min ahead" — device behaviour can't be checked from the diff; taking it as reported. Worth noting it also demonstrates that UDP/123 was reachable on *that* network, which is the caveat you already flagged.
- "`cargo test --lib` — 281 pass": CI runs `cargo test --release -- --skip auth`, which is green, so the suite passes; the specific count isn't verifiable from the checks.

### Security

No secrets, no injection, no new dependency, no public-surface change (all six sites are internal Rust callers). The one finding is the P1 above: untrusted network input now feeds the authentication path. Severity Medium — it degrades availability of signed requests, it does not forge or leak anything, and the signature itself is unaffected.

### On the open question

Agreed that the credits server's own `Date` header is the reference that actually matters, and it comes over a TLS connection you already trust rather than a spoofable UDP datagram — which is the same reason as the P1 above. Reasonable as a follow-up; the module boundary is right for it.

leanmendoza:
Addressed in a6b90ed6.

**P1 — applied in full.** All three checks are in `query()`: the originate timestamp is compared against the bytes we sent (raw, not decoded, for the floor/round reason you noted), leap indicator 3 is rejected, and the offset is bounded before it's stored. Rejecting falls through to the next server as you suggested.

For the bound I used **24 h** rather than "a few hours", with the reasoning in the comment: a device further off than that can't validate a TLS certificate, so nothing else in the app works either — which makes 24 h lose nothing real while still rejecting decoded garbage. It also clears UTC+14 applied by mistake.

Worth recording that the bound does *not* stop the availability attack you describe — an attacker who can land a spoofed reply can shift us 23 h just as effectively as 10 years, and either breaks signing completely. What actually raises the bar there is the originate check. The bound's job is garbage: decode faults, epoch rollover, a server returning nonsense.

I checked the new guards against real servers before pushing, since the originate check was the one change that could plausibly reject legitimate replies:

```
time.cloudflare.com: len=48 li=0 mode=4 stratum=3 originate_echo_matches=True
time.google.com:     len=48 li=0 mode=4 stratum=1 originate_echo_matches=True
time.apple.com:      len=48 li=0 mode=4 stratum=1 originate_echo_matches=True
```

**P2 timeout — applied.** `QUERY_TIMEOUT` now wraps the whole `query()` call, so the "3 s each" in the description holds for the connect leg too. Good catch that `connect(&str)` resolves on a blocking pool thread.

**P2 test — applied.** Extracted `offset_from()`; the test now calls it instead of recomputing the formula. I did not add the table test over crafted 48-byte replies: it needs the parse split out from the socket I/O, which is a larger change than the guards themselves, and I'd rather not grow this PR further. Noting it as a real gap rather than a resolved one.

**P2 pre-epoch clock — not changing.** A pre-1970 clock on a phone that also has working TLS is not a reachable state, and `unwrap_or(0)` self-heals on the first sync. Leaving it.

**P2 telemetry — follow-up.** Agreed on the value, and reporting the offset per sync is the only way to answer whether UDP/123 survives real carrier networks. Out of scope here.

The re-sync on "Signature timestamp is too far in the future" is the best idea in the review — it closes the loop in one retry instead of up to 15 minutes, and it reads the server's clock directly, which is the same reference the open question is about. It belongs with that follow-up rather than here, since it touches the request error path.

regenesis-claw (approved):
## Re-review — a6b90ed

The fix commit resolves the P1 in full and picks up two of the four P2s along the way:

- **[P1 fixed]** `query()` now rejects LI=3 (unsynchronised server), verifies the originate timestamp against the raw request bytes (correctly compared pre-decode, since the encoder floors and the decoder rounds — matches the suggestion exactly), and rejects any offset beyond `MAX_PLAUSIBLE_OFFSET_MS` (24h, with a clear rationale: past that, TLS itself would already be broken). A single bad or spoofed reply can no longer move the signing clock.
- **[P2 fixed]** `QUERY_TIMEOUT` now wraps the whole `query()` call via `timeout(QUERY_TIMEOUT, query(server))` in `sync_now()`, so DNS/connect is bounded too, not just `recv`.
- **[P2 fixed]** `offset_from()` extracted and the unit test now calls it instead of recomputing the formula inline — the test binds to the real arithmetic.
- Not addressed (left as-is, both minor): `device_unix_time_ms()` still swallows a pre-epoch `SystemTimeError` via `unwrap_or(0)`; `offset_ms()`/`sync_state()` still have no telemetry consumer. Neither blocks — fine to leave for a follow-up.

CI: static checks + clippy green on the new head; Android/Linux builds still pending at review time, nothing failing so far.

Approving.

kuruk-mm (approved):
Looks great!

Issue #2784, opened 2026-08-28
Menu: Landscape Settings Section Redesign


regenesis-claw:
## **⚠️ Problem:**
Settings menu needs reorganizing to match the Menu Panel Redesign: landscape full-screen support, a new grouping (Graphics/Audio/Gameplay/Language/Account/Help & Support/Storage), and a Language section that doesn't exist yet.

## **🏁 Scope:** [See Figma](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=963-9275&t=Vp2YkZRJWV2PUxQL-1)
Rebuild the Settings screen under the new Menu, landscape full-screen, with the sections and grouping in the Figma spec (including the new Language and Help & Support sections).

## **📝 Deliverables:**

[ATLAS Settings](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=963-9275&t=Vp2YkZRJWV2PUxQL-1)

- [ ] Landscape full screen
- [ ] New organization:
   **Graphics**
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/3ce2a493-56ea-4716-b3cd-4683ba76ab94" />

   **Audio**
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/2eddf6ae-de17-4db4-bca1-1ae54abb1294" />

   **Gameplay**
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/44368404-4b45-44c7-84c7-6d5507e64d4e" />

   **Language (NEW)**
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/28e7b2ec-5805-46ef-a690-9aeb715b7c48" />

   **Account**
   Guest
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/199aec8d-f61b-4682-9436-a02f7c4f2513" />

   User (only account info)
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/9f64e90c-1099-4a6a-a674-a6ac7f624cf0" />

   **Help & Support (NEW section, all the support links)**
<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/bb1fca73-660f-4aa8-a571-0f72d43bd697" />

   **Storage**
[...]

PR #2849, opened 2026-09-07
fix(i18n): ES/PT layout fixes and enable the language selector

### LINKED ISSUE #2825: Localization V1.5 - Fix UI
Fix broken UI for the new languages.

Add Language selector

EibrielInv:
Closes #2825

  ES/PT strings run 20–30% longer than English, and these screens were built with                                                                                         
  `custom_minimum_size` values measured against the English text. The catalogues                                                                                                                      
  themselves were fine — this is layout, plus two genuinely untranslated strings.

  Also flips the two `TEMPORARY` gates so ES and pt-BR actually ship: the picker is
  offered everywhere and a fresh install follows the device locale.

  ## Fixes
  
  | # | Area | Change |                                                                                                                                                                               
  |---|---|---|
  | 1 | FTUE welcome | Give the username a line of its own (`"Welcome,"` / `"@name"` / `"Let's get you started"`) so a longer greeting cannot push it onto a second line; font ladder 48→42→36 kept as a backstop for very long names |
  | 3 | Upgrade card | `autowrap_mode` + expand on the title |
  | 6 | Settings clipping | No direct change — caused by #3 inflating the panel through a horizontal-scroll-disabled `ScrollContainer` |
  | 6b | Settings tabs | Re-scroll the active tab into view on language change |
  | 7 | Dropdown descriptions | Wrap instead of ellipsize, with a measured height reservation. Shared fix in the `dropdown_list` molecule — covers every dropdown row |
  | 8 | Loading tips | `fit_content` + release the 84px height floor |
  | 9 | Profile buttons | Replace hardcoded English `"ADD FRIEND"` / `"ACCEPT"` with keys |
  | 10 | Chat tooltip | The one-off hint pointing at the chatbar's flip button. Responsive width, wraps to 2 lines |
  | 11 | Ship the locales | `FOLLOW_DEVICE_LOCALE = true`, picker always available |

  Item 2 (entry-flow web screens) is external content and not fixable here.
  Item 4 (nearby row) is already solved by `EllipsisNameLabel` from the Social redesign.
  Item 5 (notifications header) is left to #2826 — see below.

  ## QA
  
  **Check only the points below — this is not a sweep of the app.** Other screens                                                                                                                
  still have ES/PT layout problems, and they are expected to: the Menu redesign work
  (#2784 Settings, #2785 / #2826 Notifications) rebuilds those surfaces and owns their
  localization. Breakage found outside this list belongs to those issues, not to this PR.

  Go through the list in **Spanish** and **Portuguese**, then repeat the starred (*) rows
  in **English** to confirm nothing moved for existing users. Switch language from
  Settings → Gameplay → Language **without restarting** — that also exercises the                                                                                                                     
  re-translation hooks. Non-production builds also offer "Pseudolocale (QA)", which pads
  every string ~30% longer than real ES/PT and makes clipping obvious.

  - [ ] **Language picker** — Settings → Gameplay → Language. Every locale switches live, the choice survives a restart, and a device set to es/pt-BR starts in that language on a fresh install (clear app data).
  - [ ] **Upgrade card** * — the bordered promo card shown to guest accounts, headed **"Unlock the full experience"** over "Add email to play across mobile and desktop, and claim an exclusive Wearable." and an **ADD EMAIL** button. It appears in two places: at the top of **Discover**, and in **Settings → Account**. Needs a guest account — it is hidden once the account is upgraded.

    Expected: the heading wraps inside the card instead of running into its borders. Check portrait **and** landscape — the card uses different font sizes per orientation.
  - [ ] **Loading screen tips** — the bar at the bottom of the loading screen, prefixed **"Tip:"**, that cycles a hint every 5s while a scene loads. Four tips rotate; the longest in ES/PT is the Emotes one, EN **"Use Emotes to wave, react, or celebrate. Tap the Emote icon to open the Emote Wheel."**

    Expected: sit on a load for ~20s so all four appear; none truncated, and each ends with its full sentence (the clipped ES one used to lose its final period).
  - [ ] **FTUE welcome** — the first-run screen shown before you enter the world, above the recommended-places carousel and the **JUMP IN** button. It now reads over three lines: **"Welcome,"** / **"@YourName"** (in purple) / **"Let's get you started"**. Only shows on a fresh install or cleared app data, and only once.

    Expected: the username sits alone on its own line and is never split or wrapped. In ES/PT the greeting line is much longer ("Te damos la bienvenida," / "Boas-vindas,") — it must not run into the name. Normal-length names render at the full design size; only a very long name (~20+ characters) shrinks the whole greeting a step, which is intended. Test a short name and a long one.

    Expected: the greeting and the `@username` stay on a single line — the font shrinks (48 → 42 → 36) rather than wrapping the name onto a second line. Test with a short name and a ~15-character one.
  - [ ] **Chat tooltip** — the one-off hint that points at the chatbar's flip button (the control that switches chat between horizontal and vertical layout). ES: "Puedes cambiar al modo de chat vertical" · PT: "Você pode mudar para o modo de chat vertical".

    **Easy to miss — it has exactly one trigger.** In-world with chat closed, tap the **chat orb in the chatbar itself**. Opening chat any other way (the navbar chat button, or the chat panel directly) never shows it, because only the chatbar's own button calls `_enter_chat_mode()`. It then stays ~10s and fades out over 1s, appears **once per chat-panel lifetime**, and closing chat within those 10s kills it for good — relaunch for another attempt.

    Expected: two lines in ES and PT, bubble sized to its text, nothing cut off. In EN the bubble is now wider than before (182px → 251px) — intended, it is what makes the width text-driven.
  - [ ] **Profile buttons** — open a non-friend's profile: "ADD FRIEND" must be translated. Then one with an incoming request: "ACCEPT" must be translated.

  Known to still be broken and **not** part of this PR: the Notifications panel header in
  ES/PT (owned by #2826), and any text that comes from the server rather than our
  catalogue — place names, scene titles, event and reward descriptions.

  ## Also

  `extract_strings.py` now recognises `.custom_text =` (with tests). That blind spot is
  why item 9 shipped in English — the scene had the right key and GDScript overwrote it
  at runtime.

  `PROFILE_ACCEPT` is new: the Social redesign retired `SOCIAL_ITEM_ACCEPT` when Accept
  became an icon-only button, but the profile passport still needs a text label.

  ## Deliberately out of scope
  
  Notification **bodies** stay English while titles translate: `notification_text_helper.gd`                                                                                                          
  does `metadata.get("description", tr(KEY))` for badge/marketplace/governance/land/worlds/
  events, so the server's English always wins and the key is dead code. Needs per-type copy
  decisions. #2826 already fixes the reward branches.

  ## Conflict avoidance with #2826
                                                                                                                                                                                                      
  `notifications_panel.tscn` and the ES/PT `NOTIFICATIONS_READ_ALL` wording were reverted to
  main, so the Notifications redesign applies cleanly and owns that header.


regenesis-claw (approved):
Reviewed against `REVIEW.md`. Good, tightly-scoped PR: every layout fix is width/wrap-driven rather than a hardcoded pixel bump, the scanner gap that caused item 9 is fixed *with* unit tests (`custom_text`), the re-translation hooks (`NOTIFICATION_TRANSLATION_CHANGED`) are wired everywhere text is assigned from GDScript, and the QA section is the most executable one I've seen in this repo. No P0/P1 findings — **approving**, with P2s below.

**Pre-flight**
- Branch is **1 commit behind `main`** (6 ahead). The only surprising file in the diff, `godot/.../avatar/test_nameplate_stacking.gd.uid`, is byte-identical to the copy already on `main` (added by #2779), so it is a merge-base artefact, not an orphan `.uid` — no action needed.
- **iOS build present and green** (`🍏 iOS → R2 (unsigned export)` SUCCESS on 9a8f705), plus Android, Linux, Docker, Clippy and Static checks. i18n tooling (`format_csv --check`, `extract_strings`) runs inside Static checks, so the new `PROFILE_ACCEPT`/`PROFILE_ADD_FRIEND` keys and the widened `GD_ASSIGN_RE` are verified against the whole tree — no other `custom_text = "<English>"` site exists.
- No security-relevant surface in this diff (no network, auth, storage or scene-runner code). No security issues found.

---

### P2 — worth addressing

**1. `dropdown_list.gd` — the comment names the wrong autowrap mode**
The docstring on `_update_description_height()` says *"autowrap_mode = WORD_SMART in the scene"*, but the scene sets `autowrap_mode = 2`, which is `AUTOWRAP_WORD`; `WORD_SMART` is `3`. The measurement is actually *consistent* with what shipped (`get_multiline_string_size()` defaults to `BREAK_MANDATORY | BREAK_WORD_BOUND`, i.e. word-bound), so this is only a doc fix — but the practical difference is real: with `WORD` a single token longer than the row width overflows horizontally instead of breaking, so a long unbroken word in a future ES/PT description will clip again. Either fix the comment or move both the scene and the `brk_flags` to word-smart.

**2. `dropdown_list.gd` — re-measure hook is on the root, not on the label**
`_notification()` re-measures on the *DropdownList's* `NOTIFICATION_RESIZED`. `ftue_screen.gd` solves the same problem by connecting to the label's own `resized` signal, which is the stricter of the two: if the description label's width ever changes without the DropdownList's own size changing (margin/theme change, a scrollbar appearing inside the row), the reserved height goes stale and the row overlaps again. Consider `_description_label.resized.connect(_update_description_height)` for symmetry with the FTUE fix — the existing `is_equal_approx` guard already makes it re-entrancy-safe.

**3. `dropdown_list.tscn` — dropping the font overrides is a no-op, but it reads like a change**
`theme_override_fonts/font` + `theme_override_font_sizes/font_size = 12` were removed. `label_settings` (`description_settings.tres`: inter_400, size **20**, `line_spacing -3`) already wins over theme overrides on `Label`, so nothing moves — but a reader of the scene diff will reasonably think the description shrank/grew from 12. Worth one line in the PR body so it isn't re-litigated later. (The measurement code reads `label_settings` first, so it is consistent with what's drawn; the negative `line_spacing` is handled with the right sign.)

**4. `tip_card.tscn` — releasing the 84px floor makes the loading screen jump**
`custom_minimum_size` goes `(600, 84) → (600, 0)` with `fit_content = true`. That fixes the clipped ES tip, but the four tips rotate every ~5s and now each renders at its own height, so the card resizes under the player mid-load — a one-line tip after a three-line one visibly snaps. Keeping the 84 as a *floor* (`(600, 84)` with `fit_content`) gets the growth without the shrink. Cheap to do, and it removes a QA judgement call.

**5. `locale_settings.gd` — "fresh install" understates the blast radius of `FOLLOW_DEVICE_LOCALE = true`**
The docstring says this "only decides what happens before the user has made [a choice]" and the PR body says "a fresh install follows the device locale". But the picker never shipped in production, so `config.locale` is empty for **every existing production user** — on update, every current user with an es/pt-BR device locale switches language, not just new installs. That is very likely what #2825 wants, but it deserves (a) an accurate docstring, and (b) a release-note line, since the only way back to English is Settings → Gameplay → Language. Flagging for confirmation, not blocking.
_(Checked the adjacent risk: a `xx-pseudo` stored by a QA build can't leak into a store build — `resolve_locale()` falls through the pseudo branch and `xx-pseudo` is not in `SUPPORTED_LOCALES`, so it lands on the device locale. Good.)_

**6. `settings.gd` — `_scroll_to_active_tab()` may run one frame too early**
`_async_scroll_to_tab_button()` awaits exactly one `process_frame`, then reads `button.position` / `button.size`. Here it is called from the same `NOTIFICATION_TRANSLATION_CHANGED` that re-shapes the tab labels, so the offset is only correct if the strip has already relaid out by that frame. Fire-and-forget matches the six existing call sites, so no change requested — but this is the one case where the widths change *in the same notification*, so it's worth an explicit QA pass: switch EN → pt-BR with the **Account** or **Developer** tab active (the widest labels) and confirm the active tab is fully in view, not half-cut.

**7. `chatbar.tscn` — confirm the wider tooltip in portrait**
The tooltip label gains `custom_minimum_size = Vector2(240, 0)` while the panel keeps `182` as its own minimum, so the bubble is now driven to ≥240 + margin (the 182 → 251px you describe). On a narrow portrait phone that's a bigger share of the row than before — the QA step already covers the ES/PT wrap, please also eyeball it in **portrait** on the narrowest device available so it doesn't push the chatbar row off-screen.

---

### Positives

- `_fit_welcome_text()` is measured, not guessed, and the re-entrancy guard (`get_theme_font_size(...) != chosen` before `add_theme_font_size_override`) is the right shape for a `resized`-driven fit — the ladder can also step back *up*, so a language change from ES to EN restores the design size.
- Splitting `FTUE_WELCOME` onto three lines fixes the cause (a longer greeting pushing the name) rather than the symptom, and keeping the ladder as a backstop for unbounded user data is the correct call.
- Reserving the measured height in the shared `dropdown_list` molecule fixes every dropdown row at once instead of per-screen.
- Widening `GD_ASSIGN_RE` to `custom_text` **with both a positive and a negative test** is exactly the follow-through this class of bug needed — the scanner's function-argument and `@export`-default blind spots remain, but this closes the one that actually shipped.
- Explicitly reverting `notifications_panel.tscn` and the `NOTIFICATIONS_READ_ALL` wording to `main` to keep #2826 conflict-free, and saying so in the description, is the kind of thing that usually only surfaces in a merge conflict.

Approving — none of the above blocks merge.

Ludmilafantaniella:
## QA Update

🐛 Found one layout issue outside the checklist, in landscape (in-world, pt-BR):

**Gameplay panel isn't centered** — it's clipped/stuck against the left edge of the screen (see screenshots).

**Dev Tools tab overflows** — "Ferramentas de desenvolvimento" doesn't fit within the visible tab row/panel, unlike the other tabs (Áudio, Jogabilidade, Conta, etc.), which are properly contained.

Repro: in-world → open Settings in landscape → set language to pt-BR or ESP → scroll to 'Herramientas de desarrollo/ferramentas de desenvolvimento' tabs.

Screenshot_20260911_171527_Gallery
Screenshot_20260911_171420_Gallery

✅ Ran through the ES/PT QA checklist (language picker, upgrade card, loading tips, FTUE welcome, chat tooltip, profile buttons) — all passed as expected, no issues.

@EibrielInv

kuruk-mm (approved):

rodrigo685:
- [ ] GET CREDITS (iOS) Not being translated

image

ElzathRusso:
**Social, friends empty state:** I changed this message in the spreadsheet. The word “nearby” sounded a bit odd in this context.
Screenshot_20260916_162936_Decentraland



___
**Chat vertical mode:** "ENVIAR" There’s no margin on the right.
Screenshot_20260916_165246_Decentraland

Screenshot 2026-09-16 at 5 08 16 PM


___
**Chat, tooltip** The tooltip should hug the content (width responsive). There’s too much empty space on the right.
Screenshot_20260916_163359_Decentraland


Screenshot 2026-09-16 at 5 08 46 PM


___
**Confirmando inicio de sesión**. The text jumps an extra line.
Could be 2 lines:
Confirmando el 
inicio de sesión...

Screenshot_20260916_165011_Decentraland

___
**Credits** I changed this message in the spreadsheet: I removed the word “LA” before Foundation
Screenshot 2026-09-16 at 17 15 23


___
**CREDITS. backpack**
Button "Get Credits" > OBTENER CREDITS
Button in card > "Get Credits" > OBTENER CREDITS and small font size 16pt
Button "Go to Shop" > IR AL SHOP
Screenshot 2026-09-16 at 18 23 11

Screenshot 2026-09-16 at 18 22 55


Screenshot 2026-09-16 at 6 28 12 PM

PR #2858, opened 2026-09-07
fix(crdt): DeleteComponent declared a 28-byte length while writing 20

### LINKED ISSUE #2851: [Bug] CRDT DeleteComponent declares length 28 but writes 20 bytes — @dcl/ecs silently drops trailing deletes
## **⚠️ Problem:**

The renderer→scene `DeleteComponent` CRDT message is written with the wrong `length` field. It declares **28 bytes** but the message is **20 bytes** (header 8 + entity 4 + component 4 + timestamp 4), which is what ADR-117, Unity and `@dcl/ecs` itself write.

Root cause in `lib/src/dcl/crdt/message.rs` — the two size constants are wrong/crossed and `put_or_delete_lww_component` uses them the wrong way round:

```rust
const CRDT_PUT_COMPONENT_HEADER_SIZE: usize = CRDT_HEADER_SIZE + 20;    // 28 — no message has this size
const CRDT_DELETE_COMPONENT_HEADER_SIZE: usize = CRDT_HEADER_SIZE + 16; // 24 — this is actually the PUT fixed size
...
if opaque_value.value.is_some() {
    let length = CRDT_DELETE_COMPONENT_HEADER_SIZE + component_buf.len(); // PUT: 24 + payload → correct by accident
    ...
} else {
    writer.write_u32(CRDT_PUT_COMPONENT_HEADER_SIZE as u32);             // DELETE: declares 28, writes 20 → BUG
    writer.write(&CrdtMessageType::DeleteComponent);
    writer.write(entity_id);
    writer.write(component_id);
    writer.write(&opaque_value.timestamp);
}
```

(`append_gos_component` also uses the misnamed "delete" constant for `AppendValue`, which is 24 + payload — correct, just misnamed.) This has been there since the first PUT/DELETE implementation (2072515400, #5, 2023-05).

**Who emits renderer-side deletes** (all go through this branch): `TweenState` cleared when a scene removes a `Tween` (`components/tween.rs`), the player-leave cleanup that nulls `PlayerIdentityData` / `AvatarBase` / `AvatarEquippedData` / `Transform` (`components/avatar_data.rs`, #1818), and the avatar `Transform` null for non-current scenes (`avatars/avatar_scene.rs`).

**Why it matters**

1. **`@dcl/ecs` silently drops the delete when it is the last message of a chunk.** The reference decoder never trusts `length` to advance for known types (it reads the three fields), *but* `CrdtMessageProtocol.validate()` rejects a header whose declared length exceeds the remaining bytes, and `parseChunkMessage` stops the loop there. A 28-byte claim on the final 20 bytes of a chunk means the delete is never applied. The renderer→scene dirty set is a `HashMap` (random iteration order) and `EngineInfo` is PUT every tick, so whether a given delete lands last is a coin flip. Concretely: `TweenState` can stay stale after `Tween.deleteFrom(e)`, and on player leave one of the five component deletes is lost (ghost `PlayerIdentityData` / `AvatarBase` → `onLeaveScene` never firing for that peer, stale entries in `getPlayers()`). Both were observed with the repro below: 13/13 `TweenState` deletes malformed, ~50 % of them in the "dropped" position, plus one avatar cleanup where the trailing `AvatarBase` delete was the dropped one.
2. **Any strict parser throws.** [eordano/hexabricks](https://github.com/eordano/hexabricks) (raw-CRDT scene, no `@dcl/sdk`) validated lengths and crashed ~24 s in, the first time the renderer emitted a delete. It now carries a Godot-specific workaround (`GODOT_DELETE_COMPONENT_LENGTH = 28` in `src/raw/protocol.ts`) that should become unnecessary.

Desktop explorers (Unity/Bevy) write 20, so this only reproduces on the Godot client.

## **🏁 Scope:**

Fix the writer so `DeleteComponent` declares 20 bytes, and add a regression test + a minimal repro scene we can iterate on.

## **📝 Deliverables:**

- [ ] `lib/src/dcl/crdt/message.rs`: fix the constants (`DELETE` fixed size = 8 + 12 = 20, `PUT`/`APPEND` fixed size = 8 + 16 = 24 + payload) and rename them so each branch uses the one it means. `delete_entity` (8 + 4 = 12) is already correct.
- [ ] Unit test in `message.rs`: serialize `[DeleteComponent, PutComponent]` with `put_or_delete_lww_component` and read it back with `process_many_messages_with_logging` (our own reader *does* trust `length`, so today the round-trip desyncs: it swallows the next message's header). Assert the delete frame is exactly 20 bytes and the second message is applied.
- [ ] Repro scene below passes: every `DELETE_COMPONENT` line shows `declaredLength=20`, `malformed=0`, `wouldBeDroppedByEcs=0`.
- [ ] Nice-to-have: commit the repro under `tests/scenes/crdt-delete-length/` so it can be pointed at any build.

## **🧪 Minimal repro scene**

Repo: **https://github.com/dcl-regenesislabs/crdt-delete-length-repro** (public; the same files are inlined below for reference).

A raw-CRDT scene (no `@dcl/sdk`, no build step): it PUTs a `Tween` on one entity, DELETEs the `Tween` 60 frames later, and inspects every chunk the renderer returns from `crdtSendToRenderer`. The renderer answers the `Tween` delete with `DeleteComponent(TweenState=1103)`, whose declared length is printed. It also reports whether `@dcl/ecs` would have dropped that message (declared length > bytes remaining in the chunk).

**Run it**

```bash
git clone https://github.com/dcl-regenesislabs/crdt-delete-length-repro && cd crdt-delete-length-repro
npm install
npm run selftest   # optional: checks the decoder against synthetic chunks, no explorer needed
npm start          # serves http://127.0.0.1:8000

# godot-explorer
# scene console.log only reaches stdout with the --rust-log filter; run `cargo run -- import-assets` first on a fresh checkout
cargo run -- run -- --preview --realm http://127.0.0.1:8000 --guest-profile --skip-lobby --rust-log="dclgodot::dcl::js=debug,warn" 2>&1 | grep crdt-delete-length
```

**Expected output today (bug)**

```
# godot-explorer main @ 735ab58, macOS debug build, 2026-09-07
[crdt-delete-length] started; one cycle = put Tween, delete Tween 60 frames later
[crdt-delete-length] BUG DELETE_COMPONENT entity=512 component=1103 ts=1 declaredLength=28 expected=20 remainingInChunk=20 -> @dcl/ecs would DROP this message (declared length > remaining bytes)
[crdt-delete-length] cycle=1 chunks=120 deletes=1 tweenStateDeletes=1 malformed=1 wouldBeDroppedByEcs=1
[crdt-delete-length] BUG DELETE_COMPONENT entity=512 component=1103 ts=3 declaredLength=28 expected=20 remainingInChunk=20 -> @dcl/ecs would DROP this message (declared length > remaining bytes)
[crdt-delete-length] cycle=2 chunks=240 deletes=2 tweenStateDeletes=2 malformed=2 wouldBeDroppedByEcs=2
[crdt-delete-length] BUG DELETE_COMPONENT entity=512 component=1103 ts=5 declaredLength=28 expected=20 remainingInChunk=55
[crdt-delete-length] cycle=3 chunks=360 deletes=3 tweenStateDeletes=3 malformed=3 wouldBeDroppedByEcs=2
...
# an avatar entity (32) got cleaned up mid-run: the five player-leave deletes from avatar_data.rs, all malformed,
# and the trailing one (AvatarBase) is the one @dcl/ecs would drop
[crdt-delete-length] BUG DELETE_COMPONENT entity=32 component=1    ts=1 declaredLength=28 expected=20 remainingInChunk=135   # Transform
[crdt-delete-length] BUG DELETE_COMPONENT entity=32 component=1089 ts=0 declaredLength=28 expected=20 remainingInChunk=115   # PlayerIdentityData
[crdt-delete-length] BUG DELETE_COMPONENT entity=32 component=1091 ts=0 declaredLength=28 expected=20 remainingInChunk=60    # AvatarEquippedData
[crdt-delete-length] BUG DELETE_COMPONENT entity=32 component=101  ts=1 declaredLength=28 expected=20 remainingInChunk=40    # InternalPlayerData
[crdt-delete-length] BUG DELETE_COMPONENT entity=32 component=1087 ts=0 declaredLength=28 expected=20 remainingInChunk=20 -> @dcl/ecs would DROP this message (declared length > remaining bytes)   # AvatarBase
[crdt-delete-length] cycle=7 chunks=840 deletes=12 tweenStateDeletes=7 malformed=12 wouldBeDroppedByEcs=6
```

**Expected output after the fix**

```
[crdt-delete-length] OK  DELETE_COMPONENT entity=512 component=1103 ts=... declaredLength=20
[crdt-delete-length] cycle=N chunks=... deletes=N tweenStateDeletes=N malformed=0 wouldBeDroppedByEcs=0
```

<details>
<summary><code>scene.json</code></summary>

```json
{
  "ecs7": true,
  "runtimeVersion": "7",
  "display": {
    "title": "crdt-delete-length repro",
    "description": "Minimal raw-CRDT scene: makes the renderer emit a DeleteComponent (TweenState) and checks its declared length"
  },
  "main": "bin/index.js",
  "scene": {
    "parcels": ["0,0"],
    "base": "0,0"
  },
  "spawnPoints": [
    { "name": "spawn", "default": true, "position": { "x": 4, "y": 0, "z": 4 }, "cameraTarget": { "x": 8, "y": 1, "z": 8 } }
  ]
}
```
</details>

<details>
<summary><code>package.json</code></summary>

```json
{
  "name": "crdt-delete-length-repro",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "start": "sdk-commands start --skip-build --no-browser --no-watch",
    "selftest": "node selftest.js"
  },
  "devDependencies": {
    "@dcl/sdk-commands": "^7.26.0"
  }
}
```
</details>

<details>
<summary><code>bin/index.js</code></summary>

```js
// Minimal repro for decentraland/godot-explorer: DeleteComponent messages are
// written with length=28 (the PutComponent header size) but only carry the
// 20 bytes the spec (ADR-117) defines for a DeleteComponent.
//
// This is a *raw* scene: it does not use @dcl/sdk. It talks to the renderer
// through ~system/EngineApi directly so it can inspect every byte the renderer
// sends back. It puts a Tween on one entity, deletes the Tween a second later,
// and the renderer answers with DeleteComponent(TweenState). We then check the
// declared length of that DeleteComponent.
//
// Expected (spec):  DELETE_COMPONENT declaredLength=20
// Godot today:      DELETE_COMPONENT declaredLength=28  -> logged as "BUG"
//
// It also reports whether @dcl/ecs would have silently dropped that message:
// CrdtMessageProtocol.validate() rejects a header whose declared length is
// larger than the bytes remaining in the chunk, so when the malformed delete is
// the last message of a chunk the reference decoder never sees it.

let EngineApi = null
try { EngineApi = require('~system/EngineApi') } catch (_) { /* running under node selftest */ }

const PUT_COMPONENT = 1
const DELETE_COMPONENT = 2
const DELETE_ENTITY = 3
const APPEND_VALUE = 4

const TRANSFORM = 1
const TWEEN = 1102
const TWEEN_STATE = 1103

const ENTITY = 512 // first entity number reserved for scenes
const CYCLE_FRAMES = 120 // put Tween at frame%120==0, delete it at frame%120==60
const TAG = '[crdt-delete-length]'

// ---------------------------------------------------------------- encoders

function frame(size, type, ...words) {
  const out = new Uint8Array(size)
  const view = new DataView(out.buffer)
  view.setUint32(0, size, true)
  view.setUint32(4, type, true)
  words.forEach((w, i) => view.setUint32(8 + i * 4, w, true))
  return out
}

function putComponent(entity, component, timestamp, payload) {
  const out = frame(24 + payload.byteLength, PUT_COMPONENT, entity, component, timestamp, payload.byteLength)
  out.set(payload, 24)
  return out
}

function deleteComponent(entity, component, timestamp) {
  return frame(20, DELETE_COMPONENT, entity, component, timestamp)
}

function concat(parts) {
  const total = parts.reduce((n, p) => n + p.byteLength, 0)
  const out = new Uint8Array(total)
  let off = 0
  for (const p of parts) { out.set(p, off); off += p.byteLength }
  return out
}

// Transform is not protobuf: 10 LE floats (pos, rot, scale) + u32 parent.
function encodeTransform(x, y, z) {
  const out = new Uint8Array(44)
  const v = new DataView(out.buffer)
  const f = [x, y, z, 0, 0, 0, 1, 1, 1, 1]
  f.forEach((n, i) => v.setFloat32(i * 4, n, true))
  v.setUint32(40, 0, true)
  return out
}

// Hand-rolled protobuf for PBTween { duration=1 (float), move=3 { start=1, end=2 } }.
// Vector3 { x=1, y=2, z=3 } floats. Field tags: (field << 3) | wireType; float=5, len=2.
function pbFloat(field, value) {
  const out = new Uint8Array(5)
  out[0] = (field << 3) | 5
  new DataView(out.buffer).setFloat32(1, value, true)
  return out
}
function pbMessage(field, body) {
  if (body.byteLength > 127) throw new Error('varint > 1 byte not supported in this repro')
  return concat([Uint8Array.of((field << 3) | 2, body.byteLength), body])
}
function pbVector3(field, x, y, z) {
  return pbMessage(field, concat([pbFloat(1, x), pbFloat(2, y), pbFloat(3, z)]))
}
function encodeTween(durationMs, from, to) {
  const move = concat([pbVector3(1, ...from), pbVector3(2, ...to)])
  return concat([pbFloat(1, durationMs), pbMessage(3, move)])
}

// ---------------------------------------------------------------- decoder

const stats = { chunks: 0, deletes: 0, malformed: 0, wouldBeDroppedByEcs: 0, tweenStateDeletes: 0 }

// Walks a renderer->scene chunk the way @dcl/ecs does for known types
// (fixed field reads), but reports the declared length of every DELETE.
function inspectChunk(chunk, log) {
  stats.chunks++
  const view = new DataView(chunk.buffer, chunk.byteOffset, chunk.byteLength)
  let off = 0
  while (off + 8 <= chunk.byteLength) {
    const remaining = chunk.byteLength - off
    const length = view.getUint32(off, true)
    const type = view.getUint32(off + 4, true)
    if (type === DELETE_COMPONENT) {
      const entity = view.getUint32(off + 8, true)
      const component = view.getUint32(off + 12, true)
      const timestamp = view.getUint32(off + 16, true)
      stats.deletes++
      if (component === TWEEN_STATE) stats.tweenStateDeletes++
      const dropped = remaining < length
      if (dropped) stats.wouldBeDroppedByEcs++
      if (length !== 20) {
        stats.malformed++
        log(`${TAG} BUG DELETE_COMPONENT entity=${entity} component=${component} ts=${timestamp} declaredLength=${length} expected=20 remainingInChunk=${remaining}` +
          (dropped ? ' -> @dcl/ecs would DROP this message (declared length > remaining bytes)' : ''))
      } else {
        log(`${TAG} OK  DELETE_COMPONENT entity=${entity} component=${component} ts=${timestamp} declaredLength=20`)
      }
      off += 20 // spec size; never trust the declared length for deletes
      continue
    }
    if (type === DELETE_ENTITY) { off += 12; continue }
    if (type === PUT_COMPONENT || type === APPEND_VALUE) {
      const payloadLength = view.getUint32(off + 20, true)
      if (length !== 24 + payloadLength) log(`${TAG} unexpected PUT/APPEND length=${length} payload=${payloadLength}`)
      off += 24 + payloadLength
      continue
    }
    log(`${TAG} unknown message type=${type} length=${length}; stopping chunk`)
    break
  }
}

// ---------------------------------------------------------------- scene loop

let frameNo = 0
let timestamp = 1
let pending = []
let cycles = 0

async function exchange(bytes) {
  const response = await EngineApi.crdtSendToRenderer({ data: bytes })
  pending = response.data || []
}

async function onStart() {
  const state = await EngineApi.crdtGetState({ data: new Uint8Array() })
  for (const chunk of state.data || []) inspectChunk(chunk, console.log)
  console.log(`${TAG} started; one cycle = put Tween, delete Tween 60 frames later`)
}

async function onUpdate() {
  for (const chunk of pending) inspectChunk(chunk, console.log)
  pending = []

  const out = []
  const phase = frameNo % CYCLE_FRAMES
  if (frameNo === 0) out.push(putComponent(ENTITY, TRANSFORM, timestamp++, encodeTransform(8, 1, 8)))
  if (phase === 0) {
    out.push(putComponent(ENTITY, TWEEN, timestamp++, encodeTween(1000, [8, 1, 8], [8, 3, 8])))
  } else if (phase === 60) {
    out.push(deleteComponent(ENTITY, TWEEN, timestamp++))
  } else if (phase === 119) {
    cycles++
    console.log(`${TAG} cycle=${cycles} chunks=${stats.chunks} deletes=${stats.deletes} tweenStateDeletes=${stats.tweenStateDeletes} malformed=${stats.malformed} wouldBeDroppedByEcs=${stats.wouldBeDroppedByEcs}`)
  }
  frameNo++
  await exchange(out.length ? concat(out) : new Uint8Array())
}

module.exports.onStart = onStart
module.exports.onUpdate = onUpdate
module.exports.__test = { putComponent, deleteComponent, encodeTween, encodeTransform, inspectChunk, concat, stats, TWEEN_STATE, ENTITY }
```
</details>

<details>
<summary><code>selftest.js</code> (optional, runs under node)</summary>

```js
// Runs the scene's decoder against synthetic renderer chunks (no explorer needed).
const { __test: t } = require('./bin/index.js')
const assert = require('node:assert')

// A Godot-style malformed delete: declares 28 bytes, carries 20.
function godotDelete(entity, component, ts) {
  const m = t.deleteComponent(entity, component, ts)
  new DataView(m.buffer).setUint32(0, 28, true)
  return m
}
const put = t.putComponent(0, 1108, 5, new Uint8Array([1, 2, 3])) // EngineInfo-like put

const logs = []
const log = (s) => logs.push(s)

// spec-compliant delete
t.inspectChunk(t.concat([t.deleteComponent(t.ENTITY, t.TWEEN_STATE, 7), put]), log)
assert.equal(t.stats.malformed, 0)

// malformed delete followed by another message: detected, not dropped
t.inspectChunk(t.concat([godotDelete(t.ENTITY, t.TWEEN_STATE, 8), put]), log)
assert.equal(t.stats.malformed, 1)
assert.equal(t.stats.wouldBeDroppedByEcs, 0)

// malformed delete as the last message: detected AND flagged as dropped by @dcl/ecs
t.inspectChunk(t.concat([put, godotDelete(t.ENTITY, t.TWEEN_STATE, 9)]), log)
assert.equal(t.stats.malformed, 2)
assert.equal(t.stats.wouldBeDroppedByEcs, 1)
assert.equal(t.stats.tweenStateDeletes, 3)

// encoder sanity
assert.equal(t.encodeTransform(8, 1, 8).byteLength, 44)
assert.equal(t.encodeTween(1000, [8, 1, 8], [8, 3, 8]).byteLength, 41)
console.log(logs.join('\n'))
console.log('selftest OK', t.stats)
```
</details>

## **🔗 References:**
- [dcl-regenesislabs/crdt-delete-length-repro](https://github.com/dcl-regenesislabs/crdt-delete-length-repro) — the repro scene
- `lib/src/dcl/crdt/message.rs` — `put_or_delete_lww_component`, the constants at the top of the writer section
- [ADR-117: CRDT protocol for scenes](https://adr.decentraland.org/adr/ADR-117) — `DeleteComponent` = 20 bytes
- [`@dcl/ecs` `crdtMessageProtocol.ts`](https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/ecs/src/serialization/crdt/crdtMessageProtocol.ts) — `validate()` is what drops a trailing malformed delete; [`deleteComponent.ts`](https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/ecs/src/serialization/crdt/deleteComponent.ts) writes `MESSAGE_HEADER_LENGTH = 12` (+8 header = 20)
- [eordano/hexabricks `src/raw/protocol.ts`](https://github.com/eordano/hexabricks/blob/main/src/raw/protocol.ts) — the scene that surfaced this; carries a `GODOT_DELETE_COMPONENT_LENGTH = 28` workaround
- #1818 — player-leave deletes that ride on this path

kuruk-mm:
## What

Every `DeleteComponent` the renderer sends to a scene now declares its real wire length of 20 bytes. It used to declare 28, because the PUT and DELETE size constants in `put_or_delete_lww_component` were crossed. Constants are renamed to what they measure, with an ADR-117 layout comment, and a unit test round-trips a [Delete, Put] pair and fails against the old writer.

## Why

Decoders that trust the length field mis-align the next message, and `@dcl/ecs` silently drops the delete when it is the last message in a chunk. That is about half the time, since the dirty set iterates in random order. Visible effects: Tween removal not applied, avatars of players who left staying in scene-driven player lists, Transform nulling lost. bevy-explorer computes the length correctly, so only this client was affected.

Closes #2851

## Details
<details>
<summary>Expand</summary>

- `lib/src/dcl/crdt/message.rs` — DELETE now declares 20 bytes; constants are `DELETE = 20`, `PUT/APPEND = 24 + payload`, `DELETE_ENTITY = 12`. New unit test `delete_component_declares_its_real_wire_length`.

Verified by the author:
- [x] `cargo test --lib dcl::crdt::message` passes; the new test fails on the old code (`left: 28, right: 20`).
- [x] `cargo clippy -- -D warnings` clean.
- [x] Repro scene [dcl-regenesislabs/crdt-delete-length-repro](https://github.com/dcl-regenesislabs/crdt-delete-length-repro) run locally: every delete logs `declaredLength=20`, `malformed=0 wouldBeDroppedByEcs=0` (was 13/13 malformed before).

</details>

## Test plan

Desktop or mobile, either is fine.

### Case 1 — Tweens keep cycling
**Steps:**
1. Open the app and jump to a scene that uses Tweens with repeat or ping-pong, e.g. Genesis Plaza.
2. Watch the tweened objects for a minute.

- [ ] **Expected:** they keep cycling; no object gets stuck at the end of its first move.

### Case 2 — Player leaving is cleaned up
**Setup:** a second user nearby.
**Steps:**
1. Open the app, enter Genesis Plaza next to the second user.
2. Have the second user teleport away.

- [ ] **Expected:** their avatar disappears from the scene and from any scene-driven player list.

- [ ] **Regression:** this is the renderer→scene CRDT writer used by every scene. Any scene that reads renderer components (players, tweens, raycasts) behaves exactly as before, minus the dropped deletes.

### DIFF
--- a/lib/src/dcl/crdt/message.rs
+++ b/lib/src/dcl/crdt/message.rs
@@ -259,13 +259,20 @@ fn log_crdt_message(
     );
 }
 
-const CRDT_PUT_COMPONENT_HEADER_SIZE: usize = CRDT_HEADER_SIZE + 20;
-const CRDT_DELETE_COMPONENT_HEADER_SIZE: usize = CRDT_HEADER_SIZE + 16;
-const CRDT_DELETE_ENTITY_HEADER_SIZE: usize = CRDT_HEADER_SIZE + 4;
+// ADR-117 wire sizes. Every message starts with the 8-byte header
+// (length u32 + type u32) and `length` counts that header too.
+//
+// PutComponent / AppendValue: header + entity(4) + component(4) + timestamp(4)
+// + content_length(4) = 24 bytes, followed by `content_length` payload bytes.
+const CRDT_PUT_COMPONENT_HEADER_SIZE: usize = CRDT_HEADER_SIZE + 16;
+// DeleteComponent: header + entity(4) + component(4) + timestamp(4) = 20 bytes.
+const CRDT_DELETE_COMPONENT_LENGTH: usize = CRDT_HEADER_SIZE + 12;
+// DeleteEntity: header + entity(4) = 12 bytes.
+const CRDT_DELETE_ENTITY_LENGTH: usize = CRDT_HEADER_SIZE + 4;
 
 #[allow(dead_code)]
 pub fn delete_entity(entity_id: &SceneEntityId, writer: &mut DclWriter) {
-    writer.write_u32(CRDT_DELETE_ENTITY_HEADER_SIZE as u32);
+    writer.write_u32(CRDT_DELETE_ENTITY_LENGTH as u32);
     writer.write(&CrdtMessageType::DeleteEntity);
     writer.write(entity_id);
 }
@@ -292,7 +299,7 @@ pub fn put_or_delete_lww_component(
         component_definition.to_binary(*entity_id, &mut component_writer)?;
 
         let content_length = component_buf.len();
-        let length = CRDT_DELETE_COMPONENT_HEADER_SIZE + component_buf.len();
+        let length = CRDT_PUT_COMPONENT_HEADER_SIZE + component_buf.len();
 
         writer.write_u32(length as u32);
         writer.write(&CrdtMessageType::PutComponent);
@@ -303,7 +310,7 @@ pub fn put_or_delete_lww_component(
         writer.write_u32(content_length as u32);
         writer.write_raw(&component_buf)
     } else {
-        writer.write_u32(CRDT_PUT_COMPONENT_HEADER_SIZE as u32);
+        writer.write_u32(CRDT_DELETE_COMPONENT_LENGTH as u32);
         writer.write(&CrdtMessageType::DeleteComponent);
         writer.write(entity_id);
         writer.write(component_id);
@@ -338,7 +345,7 @@ pub fn append_gos_component(
         component_definition.to_binary(*entity_id, i, &mut component_writer)?;
 
         let content_length = component_buf.len();
-        let length = CRDT_DELETE_COMPONENT_HEADER_SIZE + component_buf.len();
+        let length = CRDT_PUT_COMPONENT_HEADER_SIZE + component_buf.len();
 
         writer.write_u32(length as u32);
         writer.write(&CrdtMessageType::AppendValue);
@@ -429,7 +436,9 @@ pub fn filter_known_crdt_messages(raw: &[u8], scene_crdt_state: &SceneCrdtState)
 mod tests {
     use super::*;
     use crate::dcl::components::proto_components::sdk::components::PbAvatarEmoteCommand;
+    use crate::dcl::components::transform_and_parent::DclTransformAndParent;
     use crate::dcl::crdt::grow_only_set::GenericGrowOnlySetComponentOperation;
+    use crate::dcl::crdt::last_write_wins::LastWriteWinsComponentOperation;
     use crate::dcl::crdt::SceneCrdtStateProtoComponents;
 
     /// `GrowOnlySet::to_binary` takes a REVERSE index (0 = newest entry), so walking
@@ -479,5 +488,61 @@ mod tests {
             first < second && second < third,
             "entries must be emitted oldest-first; got offsets {first}, {second}, {third}"
         );
+    }
+
+    /// The renderer used to declare a 28-byte length on DeleteComponent while
+    /// writing 20 bytes (the PUT/DELETE constants were crossed). Any decoder that
+    /// trusts `length` then either mis-aligns the next message or, in `@dcl/ecs`,
+    /// silently drops the delete when it is the last message of the chunk (#2851).
+    ///
+    /// Round-trips a [DeleteComponent, PutComponent] pair through our own reader:
+    /// with the wrong length the PUT that follows the delete is swallowed.
+    #[test]
+    fn delete_component_declares_its_real_wire_length() {
+        let deleted = SceneEntityId::new(512, 0);
+        let kept = SceneEntityId::new(513, 0);
+
+        let mut sender = SceneCrdtState::from_proto();
+        {
+            let transform = sender.get_transform_mut();
+            // put Some then None so the delete carries a non-zero timestamp (1).
+            transform.put(deleted, Some(DclTransformAndParent::default()));
+            transform.put(deleted, None);
+            transform.put(kept, Some(DclTransformAndParent::default()));
+        }
+
+        let mut buf = Vec::new();
+        let mut writer = DclWriter::new(&mut buf);
+        for entity in [&deleted, &kept] {
+            put_or_delete_lww_component(&sender, entity, &SceneComponentId::TRANSFORM, &mut writer)
+                .expect("transform entries should serialize");
+        }
+
+        // Wire check: the delete must declare exactly the bytes it wrote, and the
+        // PUT header must start right after it.
+        let declared_delete_len = u32::from_le_bytes(buf[0..4].try_into().unwrap()) as usize;
+        assert_eq!(
+            declared_delete_len, 20,
+            "DeleteComponent is 20 bytes on the wire"
+        );
+        let next_type = u32::from_le_bytes(buf[24..28].try_into().unwrap());
+        assert_eq!(next_type, CrdtMessageType::PutComponent as u32);
+        let declared_put_len = u32::from_le_bytes(buf[20..24].try_into().unwrap()) as usize;
+        assert_eq!(declared_put_len, buf.len() - 20);
+
+        // Behavioural check: a receiver replaying the stream sees both messages.
+        let mut receiver = SceneCrdtState::from_proto();
+        let mut reader = DclReader::new(&buf);
+        process_many_messages(&mut reader, &mut receiver);
+
+        let transform = receiver.get_transform();
+        let deleted_entry = transform.get(&deleted).expect("delete must be applied");
+        assert_eq!(deleted_entry.timestamp, SceneCrdtTimestamp(1));
+        assert!(deleted_entry.value.is_none());
+        let kept_entry = transform
+            .get(&kept)
+            .expect("put after the delete must not be swallowed");
+        assert_eq!(kept_entry.timestamp, SceneCrdtTimestamp(0));
+        assert!(kept_entry.value.is_some());
     }
 }


regenesis-claw (approved):
## Review — APPROVE

**Pre-flight**
- Branch is 1 commit behind `main`, but that commit (#2779, native bug reporting) touches no file in this diff — staleness is not material here, no rebase needed for a trustworthy signal.
- iOS build not required by REVIEW.md §0.2 (no native/platform-branch code), and an iOS export ran green anyway.
- No submodule drift.

**Correctness — the fix is right per ADR-117.** Header is 8 bytes (`length` u32 + `type` u32) and `length` is inclusive of it:
- `DeleteComponent` = 8 + entity(4) + component(4) + timestamp(4) = **20** ✅ (`CRDT_DELETE_COMPONENT_LENGTH`)
- `PutComponent` / `AppendValue` fixed part = 8 + 4 + 4 + 4 + content_length(4) = **24** + payload ✅
- `DeleteEntity` = 8 + 4 = **12** ✅

Verified the writer bodies match the declared sizes byte for byte. Note that PUT/APPEND output is **byte-identical to before** — the old `CRDT_DELETE_COMPONENT_HEADER_SIZE` was already 24, so this half of the change is a pure rename; only the DELETE branch changes on the wire (28 → 20). That keeps the blast radius genuinely limited to what the description claims.

**Test.** `delete_component_declares_its_real_wire_length` is a good regression test: it pins the declared length *and* round-trips a `[Delete, Put]` pair through `process_many_messages`, so it also covers the misalignment failure mode (our own reader does trust `length` via `take_reader(length - 8)`, so the old writer would have swallowed the following PUT header). Confirmed it actually executed and passed in CI (Linux job, `274 passed; 0 failed`), not just author-reported.

**Consumer impact.** `put_or_delete_lww_component` / `append_gos_component` are called only from `dcl/js/engine.rs` on the renderer→scene path; no other in-repo writer of these message types exists, and no consumer outside this repo links the Rust symbols. Decoders that ignore `length` for known types are unaffected; decoders that validate it (hexabricks' raw parser, the scene inspector's `raw_size_bytes`) go from wrong to correct.

**Security.** No security issues found — no parsing of untrusted input changes, no new allocation sized from an attacker-controlled field, and the change strictly narrows a declared length rather than widening it.

### Findings (P2 only, non-blocking)

- **[P2] Naming is now asymmetric.** `CRDT_PUT_COMPONENT_HEADER_SIZE` kept the `_HEADER_SIZE` suffix while the two others became `_LENGTH`. That's actually semantically correct (PUT's constant is a fixed prefix, the others are the whole message), but the PR body says "constants renamed to what they measure", which reads as if all three were touched. Consider `CRDT_PUT_COMPONENT_FIXED_SIZE` or a one-line comment on why PUT is the odd one out.
- **[P2] `AppendValue` has no regression test.** It shares `CRDT_PUT_COMPONENT_HEADER_SIZE` and is correct today, but nothing pins its declared length — a future edit to that constant would be caught for PUT (indirectly, via `declared_put_len == buf.len() - 20`) and not for APPEND. Cheap to extend the new test with one `append_gos_component` call.
- **[P2] `filter_known_crdt_messages` (same file) also trusts `length`** to walk the buffer and has no callers in the repo. Pre-existing, out of scope for this PR, but it's a second length-trusting parser sitting next to the bug you just fixed — worth deleting or covering in a follow-up.
- **[P2] Behavioural note for the release notes.** Scenes now receive deletes they were silently missing (Transform nulling, player-leave cleanup, raycast result clears). That is the intended fix, but it is a *behaviour* change for deployed scenes, not just a wire-format correction — the "behaves exactly as before, minus the dropped deletes" line in the test plan undersells it slightly. Worth having QA specifically watch scenes that read `PlayerIdentityData` / `RaycastResult`.

Test plan is QA-executable (starts from opening the app, concrete scene, observable pass/fail) — matches the §4 format.

leanmendoza (approved):

PR #2862, opened 2026-09-08
fix(auth): route the App Review sign-in through mobile-bff

leanmendoza:
## Why

The last App Store submission failed sign-in with the credentials we handed Apple.

There was no alternative path for them: "Sign in with email" mints its wallet through thirdweb, which emails a **fresh** OTP that expires in 15 minutes. A reviewer can never complete it — the code we publish in App Review Information is dead by the time they read it, and the mailbox that receives the live one isn't theirs. Every attempt lands on `Could not verify code`.

## What

mobile-bff already grew the server half of the fix (`/test-auth/send-code`, `/test-auth/verify-code` — decentraland/mobile-bff#39, #40). For **one** configured address, and only that one, it checks the submitted code against a fixed value it holds in config and mints a full `AuthIdentity` for the review wallet. Every other address gets a 403, and a deployment with no review account configured returns 404.

Nothing on the client ever called it. This PR wires it up:

- **`lib/src/auth/review_login.rs`** (new) — `is_review_email` matches the single review address (case- and whitespace-insensitive), `send_code` / `verify_code` talk to mobile-bff, and the result is checked before it becomes a session.
- **`dcl_player_identity.rs`** — both halves of the email flow divert on that address only; everything else stays on thirdweb, byte for byte.
- **`auth_identity.rs`** — `complete_mobile_auth`'s body is extracted as `ephemeral_from_auth_identity`. mobile-bff returns the same shape the auth-server mobile deep link does, so the conversion is reused rather than rewritten.

The GDScript sign-in screen is **unchanged** — it already goes through `async_link_email_start` / `async_login_email_verify`, and the divert happens underneath.

The guest-upgrade path (`async_link_email_verify`) rejects the review address early with a message pointing at the sign-in screen, instead of dead-ending on a thirdweb rejection: that account signs in directly and has no guest wallet to merge into.

## Safety

- The divert is keyed on one hardcoded address, so a normal sign-in never round-trips to the review endpoint.
- The server refuses every other address (403) and rate-limits attempts.
- `verify_code` refuses to sign in as an address the returned auth chain isn't rooted at — otherwise every signed request would be attributed to a wallet that signed nothing.
- No fixed code lives in this repo; it stays server-side config.

## Testing

- `cargo check`, `cargo clippy -- -D warnings`, `cargo fmt --all` clean.
- `cargo test --lib auth::` — 8 passed.
- Two unit tests in `review_login`: that only the review address is diverted, and that a server-minted identity parses while a body claiming a mismatched address is refused. The second one exercises the whole response contract offline — the camelCase identity fields, the millisecond expiration the signed payload embeds verbatim, and the address/signer check.
- The live `/test-auth` endpoints were confirmed to answer on both `.org` and `.zone` with the expected shape.

**Not yet verified on device** — needs an iOS build to walk the reviewer's exact path before we resubmit.

## Before resubmitting

Confirm `TEST_AUTH_EMAIL`, `TEST_AUTH_OTP_CODE` and `TEST_AUTH_PRIVATE_KEY` are set on the production deployment, and that the code in App Review Information matches `TEST_AUTH_OTP_CODE` — not a one-time code copied from an email.

### DIFF
--- a/lib/src/auth/auth_identity.rs
+++ b/lib/src/auth/auth_identity.rs
@@ -4,7 +4,7 @@ use crate::godot_classes::dcl_tokio_rpc::GodotTokioCall;
 
 use super::{
     decentraland_auth_server::{
-        do_request, do_request_mobile, fetch_identity_by_id, CreateRequest,
+        do_request, do_request_mobile, fetch_identity_by_id, AuthIdentity, CreateRequest,
     },
     ephemeral_auth_chain::EphemeralAuthChain,
     wallet::{AsH160, ChainLink, ObjSafeWalletSigner, SimpleAuthChain, Wallet},
@@ -93,8 +93,18 @@ pub async fn complete_mobile_auth(
     identity_id: String,
 ) -> Result<(EphemeralAuthChain, u64), anyhow::Error> {
     let response = fetch_identity_by_id(identity_id).await?;
-    let identity = response.identity;
+    ephemeral_from_auth_identity(response.identity)
+}
 
+/// Turns a server-minted `AuthIdentity` (ephemeral private key + signed auth
+/// chain) into the `EphemeralAuthChain` the client signs requests with.
+///
+/// Shared by the two flows that receive a ready-made identity instead of
+/// signing one locally: the auth-server mobile deep link
+/// (`complete_mobile_auth`) and the App Review sign-in (`review_login`).
+pub fn ephemeral_from_auth_identity(
+    identity: AuthIdentity,
+) -> Result<(EphemeralAuthChain, u64), anyhow::Error> {
     // Parse the signer address from the first element in the auth chain (SIGNER type)
     let signer = identity
         .auth_chain
--- a/lib/src/auth/dcl_player_identity.rs
+++ b/lib/src/auth/dcl_player_identity.rs
@@ -21,6 +21,7 @@ use super::decentraland_auth_server::{do_request, CreateRequest};
 use super::device_anchor;
 use super::ephemeral_auth_chain::EphemeralAuthChain;
 use super::remote_wallet::RemoteWallet;
+use super::review_login;
 use super::thirdweb_guest;
 use super::wallet::{AsH160, Wallet};
 
@@ -400,7 +401,14 @@ impl DclPlayerIdentity {
         let email = email.to_string();
 
         handle.spawn(async move {
-            let result = thirdweb_guest::email_initiate(&email).await;
+            // The App Review account has no reachable mailbox, so its code comes
+            // from mobile-bff instead of a thirdweb email. Every other address
+            // takes the normal path.
+            let result = if review_login::is_review_email(&email) {
+                review_login::send_code(&email).await
+            } else {
+                thirdweb_guest::email_initiate(&email).await
+            };
             let Some(mut promise) = get_promise() else {
                 tracing::error!("thirdweb email_initiate: promise dropped");
                 return;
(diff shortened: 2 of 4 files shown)


EibrielInv (approved):
Looks good!

kuruk-mm (approved):

regenesis-claw (approved):
Reviewed against REVIEW.md. **Approved** — no P0/P1 findings. The divert is keyed on one hardcoded address, the server-minted identity is address-checked against the auth chain it is rooted at, and everything else stays byte-for-byte on the thirdweb path.

### Contract verified against the server (not just the PR description)

- `testAuthSendCodeHandler` / `testAuthVerifyCodeHandler` in mobile-bff match the shapes this PR parses: `{ ok:false, error }` on failure and `{ ok:true, data:{ identity, address } }` on success, with `identity` being a `@dcl/crypto` `AuthIdentity` (camelCase `ephemeralIdentity` / `authChain`, `expiration` serialized as an ISO string with milliseconds). The offline test mirrors that shape correctly.
- `address` is derived server-side by recovering from a signature over the same private key that signs the SIGNER link, so the equality check added here can never fail for a healthy server response — it only fires on a tampered/mismatched body, which is exactly what it is for.
- Ephemeral key is emitted by `createUnsafeIdentity()` with all three fields populated, so the non-defaulted `publicKey` field on the client struct will not fail deserialization.
- Server-side guards confirmed: 404 when `TEST_AUTH_*` is unset, 403 for any other address, 401 on a wrong code, 429 after 5 attempts/min per IP, and length caps on both email and code.
- Live probe: both `mobile-bff.decentraland.org` and `.zone` answer 403 to a non-review address and 401 to the review address with a wrong code — i.e. all three `TEST_AUTH_*` vars are configured on both deployments today. What cannot be checked from here is whether `TEST_AUTH_OTP_CODE` matches the value in App Review Information; that stays on the pre-submit checklist.

### Findings (all P2, none blocking)

**[P2] The review address is diverted on send but rejected on verify in the guest-upgrade flow.** `async_link_email_start` sends the review address to `/test-auth/send-code` regardless of which screen called it, while `async_link_email_verify` rejects it up front. A guest who types that address into the upgrade-to-OTP popup gets a successful "code sent" and then a dead end at step 2. Only affects that one address and the message explains where to go, so this is fine to leave — noting it so it is a known shape rather than a surprise.

**[P2] The email is trimmed for matching but sent raw.** `is_review_email` trims before comparing, while the request body carries the original string; the server compares case-insensitively but does **not** trim, so a value with edge whitespace would divert client-side and then come back 403. Not reachable through the UI today (the sign-in field sets `validate_email = true` and its regex rejects whitespace anywhere, so Next stays disabled), which is why this is not a blocker — but the trim currently buys nothing. Trimming once at the divert and passing the trimmed value to `send_code` / `verify_code` would make the stated whitespace tolerance real.

**[P2] User-visible strings are raw English.** The `async_link_email_verify` rejection and the `status=..., <server body>` text from `describe_failure` surface into the popup's error label unlocalized (and the latter carries server-authored text). Consistent with the neighbouring rejections in this file, so no change requested — but the repo's i18n rules do not cover these Rust-side strings and it is worth keeping in mind for the next pass on this screen.

### Security

No issues in the diff. One operational note, since it is being discussed alongside this PR: `/test-auth/verify-code` is a public, unauthenticated endpoint on production whose only secret is `TEST_AUTH_OTP_CODE`, protected by an in-memory per-IP rate limiter that resets on restart and is per-instance. Whoever holds that code can sign in as the review wallet from anywhere. That is acceptable while the wallet holds nothing of value; if the wallet gets funded with wearables or credits so reviewers see a populated profile, the code becomes a credential guarding real assets — use a long non-numeric value and keep the wallet's balance to the minimum the review needs.

CI at the time of review: Clippy and Static checks green; Linux and Android still running.

Requested by Lean via Slack.

PR #2864, opened 2026-09-08
feat(settings): full-screen settings with responsive list+detail nav

### LINKED ISSUE #2582: Add app version with copy-to-clipboard option in Settings
## Description

Currently the app version is only visible as a small watermark-style label on the login/lobby screen and in-world HUD (bottom-left corner). There's no way for users to select or copy this text.

## Proposal

Add the app version string to Settings, with a tap-to-copy (copy-to-clipboard) button next to it. This makes it much easier for users to grab and share their exact version when reporting bugs or contacting support, instead of having to screenshot or manually retype a small watermark.

## Why

Support/bug triage regularly needs the exact app version + build hash (e.g. `v1.11.0.1205-01da818-prod`). Right now that requires the user to zoom into a screenshot of the watermark label, which is error-prone and adds friction.

---

### LINKED ISSUE #2784: Menu: Landscape Settings Section Redesign

### LINKED ISSUE #2806: Language Switcher in Settings Panel [Implementation]
##  **⚠️ Problem:**
Auto-detected device locale is wrong for a real share of players — LATAM users on English phones, expats, shared devices — and they have no way to correct it. Without a manual switch, #2667 ships localization that some players cannot reach, and the language override that #2706 and #2707 both depend on has no UI to set it.

## **🏁 Scope:**
Implement the language selection control in the Settings panel.

## **📝 Deliverables:**
- [ ] Implement new design in Settings Landscape
- [ ] Implement new design in Settings Portrait ( Pre-Game)
- [ ] Apply behavior: immediate
- [ ] Add Analytics Event: Fired a click_button event `SETTINGS_LANGUAGE_CHANGED` with extra_propertriies 'from_locale` + `to_locale`.

**Proposal:**
Implement this change at the same time as the new in-game menu, which also includes a redesign of the portrait Settings screen.

New section LANGUAGE

Image

---

Settings Portrait

https://github.com/user-attachments/assets/8e2dbefc-588f-487b-b2fd-19f5140cc4b4

---

Settings Landscape

https://github.com/user-attachments/assets/f6eb8ecc-a924-4995-bb2e-f305d15735ff

## **🔗 References:**
- [decentraland/godot-explorer#2667](https://github.com/decentraland/godot-explorer/issues/2667) — parent. Lists the manual override + persistence + no-restart as deliverables; this issue designs that surface.
- [decentraland/godot-explorer#2706](https://github.com/decentraland/godot-explorer/issues/2706) — v2 webviews. Consumes the value this control sets.
- [decentraland/godot-explorer#2707](https://github.com/decentraland/godot-explorer/issues/2707) — v3 scene forwarding. Same — a mid-session switch here must propagate.

sdilauro:
## What

Settings is now a full-screen **menu screen** (like Backpack) instead of a fixed side panel, and it adapts to orientation: in **landscape** the section list sits beside the content pane (both visible); in **portrait** it's a master list that drills into each section with a back header. Sections were reorganized (Language split into its own section, Report + Help merged into **Help & Support**), and every size that used to be hard-coded — title fonts, button height/icon, row spacing, safe-area margins, the account avatar — now follows the orientation on its own.

## Why

Implements the landscape/portrait Settings redesign from the mockups; the old panel was a single fixed layout that didn't match either orientation.

Three behavior changes ride along and are worth QA's attention:
- **Hide Player Names** is now the first Interface toggle and works **independently** of "Hide Interface" (before, it was a sub-option that only applied while the master toggle was on).
- The **version row** shows the clean marketing version (`Version X.Y.Z`) in production and the full diagnostic string (build + commit + env) in staging/dev, and is tap-to-copy.
- **Re-tapping the active navbar category** (Friends or Notifications) now **collapses the navbar**, the same as tapping the empty area outside it. Tapping a *different* category still just switches panels.

Closes #2784, closes #2582, closes #2806

## Details

<details>
<summary>Expand</summary>

- **Entry point.** Settings opens through the menu screen living in `Control_Menu` (same path pre-explorer and in-explorer); the old `SettingsPanel` side panel was removed from `explorer.tscn`/`explorer.gd`.
- **Orientation is self-managed by small components** in `src/ui/layouts/`, each reading `Global.is_orientation_portrait()` / `orientation_changed` (editor shows the scene default): `OrientationSpacer` (min-size), `OrientationLabel` (font size), `OrientationVBox` (separation). They're adopted by the section titles, dropdown/slider titles, the checkbutton row titles, section containers and spacers — so `settings.gd` no longer pushes per-orientation values in code.
- **Account section:** avatar preview (landscape only, head inset derived from the real header height + feet inset, background shown only here), name/tag/wallet header, guest upgrade card, Sign Out, and a "Delete your account?" link.
- **Navbar re-tap-to-close.** `_show_friends_panel` / `_show_notifications_panel` used to early-return when their panel was already visible; they now call `navbar.collapse()` first, reusing the same dismiss path as the HUD tap-outside catcher (`_close_all_panels` → HUD restore). Settings/Backpack/Discover are full-screen menu screens that hide the navbar, so they don't participate.
- **Notifications empty state restyled.** `notifications_panel` now centers its empty/guest message inside a `CenterContainer` with a larger bordered icon and bigger fonts. Also fixes a copy-paste error from the friends-panel tree: the dynamic message label had been renamed/lost its unique name, so `%Label_NoNotifications` (used for the guest "sign in to get notifications" text) no longer resolved — restored the node name + `unique_name_in_owner`.
- **Shared components touched — additive / backward-compatible:** `CustomButton` gained an opt-in `icon_size` (default 0 = old behavior); `dcl_global.rs` gained `get_short_version()` / `get_full_version()`.
- **Ride-along visual changes in shared molecules** (they render elsewhere, nothing breaks — flagging for QA of those screens):
  - `dropdown_list.tscn` — title uses `OrientationLabel` (30/28); the description label switched from ellipsis (`text_overrun_behavior`) to `autowrap` + a small min-height, so long descriptions now wrap. Used by **settings, username_picker, bug_report_modal, profile_editor**.
  - `guest_upgrade_card.tscn` — heading/paragraph font sizes reduced (36→32, 26→22). Used by **settings, discover**.
- **i18n:** new keys (Language section, Help & Support, version string, delete-account) in en/es/pt_BR.

Author-verified:
- [x] `gdformat` + `gdlint godot/` + `cargo run -- check-gdscript` green
- [x] i18n `format_csv --check` + `extract_strings --check` green
- [x] Run on iOS device (debug)

</details>

## Test plan

**Setup:** mobile build; use a signed-in account (and a guest where noted).

**1 — Portrait: master-detail navigation**
1. On the initial Discover menu, tap **Settings** in the bottom navbar.
2. Tap **Graphics**, then tap the **‹** back header.
- [x] Settings opens full-screen showing the section list (Graphics, Audio, Gameplay, Language, Account, Help & Support, Storage)
- [x] Tapping a section shows its content with a "‹ <section>" header; back returns to the list

**2 — Landscape: list + content side by side**
1. Enter a world, open the navbar, tap **Settings**.
2. Tap a couple of different sections.
- [x] Section list on the left and the selected section's content on the right are both visible
- [x] The selected row is highlighted and its title is bold

**3 — Hide Player Names is independent**
1. Open **Settings → Gameplay → Interface** with "Hide Interface" **off**.
2. Toggle **Hide Player Names** on.
3. Toggle "Hide Interface" on, then off.
- [x] "Hide Player Names" is the first toggle and is enabled even with "Hide Interface" off
- [x] Turning it on hides avatar name tags immediately (without enabling Hide Interface)
- [x] Its value is preserved when Hide Interface is toggled

**4 — Version row**
1. Open **Settings** and find the version row (bottom of the list).
2. Tap it.
- [x] Staging/TestFlight build shows the full string (e.g. `1.x.y.<build>-<sha>-staging`), no "Version" prefix
- [x] Tapping copies the version and shows a "copied" toast

**5 — Account section**
1. As a **guest**, open **Settings → Account** in landscape, then as a **signed-in** account.
2. Tap **Sign Out**; reopen and tap **Delete your account?**.
- [x] Avatar preview renders on the left with the name/tag/wallet header
- [x] Guest sees the "Unlock the full experience" card; signed-in does not
- [x] Sign Out logs out; "Delete your account?" opens its confirmation flow

**6 — Entry point: in-game vs lobby**
1. From the **lobby** (before entering a world), open Settings from the bottom navbar.
2. Enter a world, then open Settings from the navbar.
- [x] Both open the same full-screen Settings (not a side panel); closing returns to lobby / world respectively

**7 — Language section gating**
Setup: the Language row only appears when ≥2 locales are selectable. Production ships English-only, so the row is hidden there; non-production/debug builds also offer the QA pseudolocale, which makes it appear.
1. On a **debug / non-production** build, open Settings and scroll the section list; tap **Language**.
2. On a **production** build, open Settings and scroll the section list.
- [x] Debug: **Language** is listed and opens a populated locale picker (English + QA pseudolocale)
- [x] Production: **Language** is not listed, and navigating the sections never opens an empty picker

**8 — Rotation**
1. Open Settings and rotate the device between portrait and landscape a few times, across a few sections including Account.
- [ ] Layout switches between master-detail (portrait) and list+content (landscape); fonts, spacing and the account avatar background (Account + landscape only) update with no overlap or clipping

**9 — Re-tap an active navbar category to close it**
1. Enter a world and open the navbar (Friends panel shows by default).
2. Tap **Friends** again.
3. Reopen the navbar, tap **Notifications**, then tap **Notifications** again.
- [ ] Re-tapping the already-open category collapses the navbar and restores the HUD, exactly like tapping the empty area outside it
- [ ] Tapping a *different* category (Friends ↔ Notifications) switches panels instead of closing

**10 — Notifications empty state**
1. Open the navbar and tap **Notifications** with no notifications present.
2. Repeat as a **guest** (signed out).
- [ ] Signed-in with none: the message is centered with the bordered icon — "No notifications yet" + "Check back later for updates"
- [ ] Guest: the message reads "Sign in to get notifications" (no null/crash), still centered

- [ ] **Regression** — open a dropdown (Camera Mode / Language), an Audio slider, and the Sign Out button; open Backpack and a modal that uses the shared button/dropdown: all render and work as before

## Follow-ups

- **`Global.open_settings_panel` is now dead** — `settings_button` always emits `open_settings`, so this signal has no emitter left; `bug_report_capture.gd` still connects to it (a no-op). The screenshot capture is unaffected because that listener is also connected to `open_settings`. Cleanup (remove the signal + the stale connect) is left to coordinate with the bug-report owner.


godot/src/notifications_manager.gd
```diff
@@ -556,6 +564,35 @@ func has_queued_notifications() -> bool:
 	return _notification_queue.size() > 0
 
 
+## Re-start the toast pump from the consumer side (explorer). The queue only auto-emits when it
+## grows to size 1, which is lost if it happens while no consumer is connected yet (lobby/loading) —
+## the queue then wedges and no later toast ever shows. Explorer calls this once it can render.
+func kick_queue() -> void:
+	# Don't re-emit the head while it's already on screen — that would show a duplicate toast.
+	if _queue_paused or _toast_on_screen:
```
regenesis-claw: [P1] `_toast_on_screen` is only ever set in `_show_current()`, and nothing clears it when the consumer declines to render. During a world load `explorer.gd::_on_notification_queued()` drops the toast (`if not _is_loading`), so the flag stays `true` and this early-return kills the very drain it was added for — the queue then wedges at that head for the rest of the session (no dequeue ever runs, and `show_system_toast()` only emits when `size == 1`).

The consumer calls `kick_queue()` exactly when it knows nothing is on screen, so it can safely reset first:

```suggestion
	if _queue_paused:
		return
	# The consumer only kicks when nothing is rendered, so a stale flag from a toast that was
	# emitted but never shown (queued while loading / before the consumer existed) is cleared here.
	_toast_on_screen = false
```

godot/src/notifications_manager.gd
```diff
@@ -556,6 +564,35 @@ func has_queued_notifications() -> bool:
...
+	if _notification_queue.size() > 0:
+		_show_current()
+
+
+## Drop transient system toasts (copied-to-clipboard, etc.) still queued from a previous run so they
+## don't parade on world entry. Real notifications (friend requests, rewards) are kept.
+func clear_system_toasts() -> void:
+	# Keep the head if a toast is currently on screen — filtering it out would orphan the on-screen
+	# system toast and make the next dequeue drop the following real notification unshown.
+	if _toast_on_screen and _notification_queue.size() > 0:
```
regenesis-claw: [P1] This head-preserving branch can't fire for the case it was written for: system toasts are emitted by `show_system_toast()`, which emits `notification_queued` directly and never sets `_toast_on_screen`. So if a copy/connection toast is on screen when `_on_loading_started()` calls this, it gets filtered out of the queue — and when it closes, `dequeue_notification()` pops whatever is now at index 0 (a real notification that was never displayed) and swallows it. Routing `show_system_toast()` (and `resume_queue(true)` / the debug timer) through `_show_current()` fixes both this and the wedge above.

godot/src/ui/pages/settings/settings.gd
```diff
@@ -1251,6 +1396,103 @@ func _on_custom_button_sign_out_pressed() -> void:
...
+## The 3D preview only renders in landscape (its container hides in portrait), so skip the avatar
+## load in portrait; it's refreshed on the next profile change or when reopening in landscape.
+func _async_refresh_account_avatar() -> void:
+	if Global.is_orientation_portrait():
+		return
+	var profile: DclUserProfile = Global.player_identity.get_profile_or_null()
+	if profile == null:
+		return
+	# Guard against overlapping loads (rapid rotations / profile_changed) racing the same preview.
+	if _avatar_refresh_in_flight:
```
regenesis-claw, line 1499: [P2] A `profile_changed` that lands while a load is in flight is dropped, not deferred — the preview then shows the pre-change avatar until the next rotation or reopen. A dirty flag re-checked after the `await` keeps it correct:

```gdscript
	if _avatar_refresh_in_flight:
		_avatar_refresh_pending = true
		return
```
and re-run once when the in-flight load completes.

godot/src/ui/pages/settings/settings.gd
```diff
@@ -1251,6 +1396,103 @@ func _on_custom_button_sign_out_pressed() -> void:
...
+	# back to the column's own bottom if the label isn't ready.
+	var bottom: float = col.global_position.y + col.size.y
+	if is_instance_valid(label_version):
+		bottom = label_version.global_position.y + label_version.size.y
+	avatar_preview_account.global_position = Vector2(col.global_position.x, 0.0)
+	avatar_preview_account.size = Vector2(col.size.x, maxf(0.0, bottom))
+	# Head overlaps slightly into the header: the rect starts at screen top (y=0), so the top inset is
+	# the header container's real bottom (its height + top safe-area margin) minus the overlap.
+	if is_instance_valid(label_title):
+		var header: Control = label_title.get_parent().get_parent()
```
regenesis-claw: [P2] `label_title.get_parent().get_parent()` ties the avatar's head inset to the exact depth of the header subtree — restructure the header and this silently resolves to the wrong node (or `null`) with no error, and the head inset drifts. The header container is in this scene; giving it `unique_name_in_owner` and reading `%...` makes it explicit and fails loudly.

godot/src/ui/layouts/safe_margin_container.gd
```diff
@@ -138,7 +148,12 @@ func _apply_margins(top: int, left: int, bottom: int, right: int) -> void:
 	if use_right:
 		add_theme_constant_override("margin_right", applied_right)
 	if use_bottom:
-		add_theme_constant_override("margin_bottom", maxi(bottom, floor_bottom))
+		var applied_bottom: int = maxi(bottom, floor_bottom)
+		# When the bottom edge sits above a sibling bottom bar, the OS inset is already reserved
+		# there, so use only the hand-set floor in portrait to avoid double-counting.
+		if ignore_safe_area_bottom:
```
regenesis-claw, line 154: [P2] The comment says "in portrait" but the `@export` doc above (correctly) says the OS inset is ignored in **both** orientations, which is what the code does — `applied_bottom = floor_bottom` runs unconditionally. Worth aligning the comment so the next reader doesn't assume there's an orientation check missing here.

regenesis-claw (changes requested):
Nice restructure — the section registry + per-orientation atoms (`OrientationLabel` / `OrientationSpacer` / `OrientationVBox`) replace a pile of imperative `_apply_layout` pushes, `hide_orientation.gd` moving off the `DisplayServer.window_get_size()` heuristic onto `Global.orientation_changed` is a real fix, and the side-panel removal in `explorer.gd` is clean (no orphan connections left behind: I checked the scene for the deleted `_on_button_*_pressed` handlers and every `%unique` name the new script reads).

One blocker in the notification-queue bookkeeping, plus a handful of minor items.

---

## 🔴 P1 — `_toast_on_screen` is only maintained on one of the four emit paths

`notifications_manager.gd` sets `_toast_on_screen = true` exclusively in `_show_current()`. Three other places emit `notification_queued` directly: `show_system_toast()` (~line 679), `resume_queue(emit_next = true)` (~line 612) and the debug timer (~line 642). And nothing tells the manager that the consumer *declined* to render. Two consequences:

**1. The queue wedges for the rest of the session — the exact case `kick_queue()` was added for.**

1. A world load starts, `explorer.gd::_is_loading = true`.
2. `_async_fetch_notifications()` gets a new notification on an empty queue → `_show_current()` → `_toast_on_screen = true`, signal emitted.
3. `explorer.gd::_on_notification_queued()` drops it (`if not _is_loading`) — no toast, so `dequeue_notification()` is never called and the flag is never cleared.
4. `_on_loading_finished()` → `kick_queue()` → early-returns on `_toast_on_screen`.

The head stays at index 0 forever. Every later notification appends behind it and `show_system_toast()` only emits when `size == 1`, so from that point on **no toast of any kind shows again for the whole session** (copied-to-clipboard, poor connection, friend requests…). Before this PR the same case was covered by `_pending_notification_toast`, which this PR removes.

**2. `clear_system_toasts()`'s head-preserving branch never fires for system toasts.** That branch exists to protect an on-screen *system* toast — but system toasts go through `show_system_toast()`, which never sets the flag. So if a copy/connection toast is on screen when `_on_loading_started()` runs, it gets filtered out of the queue; when it closes, `dequeue_notification()` pops whatever is now at index 0 — a real notification that was never shown — and that one is silently swallowed.

Minimal fix: route every emit through `_show_current()`, and clear the flag in `kick_queue()` before re-emitting (the consumer calls it precisely when it knows nothing is on screen). Better still, have the consumer own the flag: `explorer.gd::_show_notification_toast()` is the only place a toast really appears.

---

## 🟡 P2

- **`_avatar_refresh_in_flight` drops the refresh instead of deferring it** (`settings.gd:1489`). A `profile_changed` arriving while a load is in flight is discarded, leaving a stale avatar until the next rotation/reopen. A dirty flag re-checked after the `await` would keep it correct.
- **Avatar load runs on every rotation to landscape regardless of section** (`settings.gd:267`). `_on_orientation_changed` calls `_async_refresh_account_avatar()` unconditionally; rotating while on Graphics pays for an avatar load nothing will show. Gate on `_current_section_key == "account"`.
- **Brittle header lookup** (`settings.gd:1449`): `label_title.get_parent().get_parent()` silently returns the wrong node if the header is ever restructured, and the avatar's head inset drifts with no error. The header container already lives in the scene — give it a unique name and read `%...`.
- **`safe_margin_container.gd` doc vs comment disagree**: the `@export` doc says the OS bottom inset is ignored in **both** orientations, the inline comment at line 154 says "use only the hand-set floor **in portrait**". The code does both orientations — fix the comment.
- **`Global.open_settings_panel` is now dead**: `settings_button.gd` was its only emitter, and after this PR nothing emits it. `bug_report_capture.gd:40` still connects to it. Worth dropping the signal + that connection in this PR rather than leaving a no-op behind.
- **Test plan**: cases 1–6 and 8 are in the format QA can run cold. Case 7 (Language gating) is a bare pair of checkboxes with no steps and non-obvious required state (a build with ≥2 locales, i.e. non-production) — worth writing out like the others.

---

## Checks

- **Branch**: up to date with `main` (`behind_by = 0`).
- **CI**: Static checks ✅, Clippy ✅; Linux, Android and the iOS export were still running at review time. The diff is GDScript UI plus a version getter in `dcl_global.rs` with no iOS-specific branch, so it isn't platform-sensitive under REVIEW.md §0.2 — but it is mobile-visible, so confirm the device builds land green before merge.
- **Consumer impact**: no public surface touched (no protocol/component/CRDT change, no mobile-bff contract). `CustomButton.icon_size` defaults to `0` = legacy behaviour, and `DropdownList` / `guest_upgrade_card` only gain the orientation label script — both backward compatible for their other call sites (profile editor, username picker, bug report modal, discover).
- **Security**: no findings — no new network/auth surface; the clipboard write and the version string are build metadata only.
- **i18n**: new keys land in all three catalogues, the runtime-built section rows are declared with `# i18n-keys:` and re-translated via `_retranslate_section_list()` on `NOTIFICATION_TRANSLATION_CHANGED`, and `SETTINGS_VERSION` uses a named `{version}` placeholder. Looks right.

Requesting changes on the P1; the rest are quick.

_Requested by Seba via Slack_

godot/src/notifications_manager.gd
```diff
@@ -556,6 +555,32 @@ func has_queued_notifications() -> bool:
...
+		notification_queued.emit(_notification_queue[0])
+
+
+## Drop transient system toasts (copied-to-clipboard, etc.) still queued from a previous run so they
+## don't parade on world entry. Real notifications (friend requests, rewards) are kept.
+func clear_system_toasts() -> void:
+	# Never touch the head: index 0 may be the toast currently on screen, and the consumer's
+	# on-close dequeue() pops index 0 — filtering it out would orphan the visible toast and make
+	# that dequeue swallow the next real notification. Only drop system toasts still waiting behind it.
+	if _notification_queue.is_empty():
```
regenesis-claw, line 574: [P2] Always keeping index 0 fixes the offset hazard, but it also keeps a system toast that was queued and *never shown* — e.g. copy-version from the pre-explorer Settings, where no consumer exists — so `kick_queue()` surfaces it on world entry, which is what this function is meant to prevent. The explorer is the only place that knows a toast is actually on screen (it instantiates them), so `clear_system_toasts(keep_head: bool)` with the explorer passing whether it currently has a toast node would cover both cases. Follow-up, not a blocker.

godot/src/notifications_manager.gd
```diff
@@ -556,6 +555,32 @@ func has_queued_notifications() -> bool:
...
 
 
+## Re-start the toast pump from the consumer side (explorer). The queue only auto-emits when it
+## grows to size 1, which is lost if it happens while no consumer is connected yet (lobby/loading) —
+## the queue then wedges and no later toast ever shows. Explorer calls this once it can render.
+func kick_queue() -> void:
+	if _queue_paused:
+		return
+	if _notification_queue.size() > 0:
+		notification_queued.emit(_notification_queue[0])
```
regenesis-claw, line 565: [P2] Unconditional re-emit is correct for the normal path (`_is_loading` is cleared immediately before this, so no toast is up). The one gap: a toast opened *before* the load that outlives it gets a second instance here, and both closes call `dequeue_notification()` — so the next notification gets popped without being shown. Same `has_toast_on_screen` signal from the explorer would close it.

regenesis-claw (approved):
Delta re-review of `114edff…12ff63b` (`fix(settings): review round 2`). The P1 is resolved and the approach is better than what I suggested — dropping `_toast_on_screen` entirely removes the multi-path bookkeeping that was the root cause, rather than patching one more call site.

| Previous finding | Status |
|---|---|
| P1 — `_toast_on_screen` not maintained on all emit paths (queue wedge + `clear_system_toasts` guard never firing) | ✅ Flag removed; all four paths emit `notification_queued` directly again, `kick_queue()` only checks `_queue_paused`, and `clear_system_toasts()` now unconditionally preserves the head. A notification queued during loading is drained at `_on_loading_finished`, and the on-screen-toast/dequeue-offset hazard is gone. |
| P2 — `_avatar_refresh_in_flight` dropped a concurrent refresh | ✅ `_avatar_refresh_pending` re-runs once the in-flight load finishes. |
| P2 — avatar loaded on every rotation to landscape | ✅ Gated on `_current_section_key == "account"`. Side effect: the `_setup_account_section()` call at `_ready` is now a no-op (default section is Graphics), which is the right behaviour. |
| P2 — brittle `get_parent().get_parent()` header walk | ✅ `%HeaderColumn` (`unique_name_in_owner`) + typed `@onready`. Checked the scene: no stale `VBoxContainer2/VBoxContainer` paths or connections left after the rename. |
| P2 — `safe_margin_container` comment vs code | ✅ Now says "both orientations". |
| P2 — dead `Global.open_settings_panel` signal | ⚠️ Still there (`global.gd:22`, connected in `bug_report_capture.gd:40`) — no emitter left. Fine as a follow-up. |
| P2 — test plan case 7 had no steps | ✅ Rewritten with setup, debug/production steps and observable results. |

**Two small consequences of the simpler approach** (both P2, neither blocking):

1. `clear_system_toasts()` now always keeps index 0, so a system toast that was queued but *never displayed* (e.g. copy-version from the pre-explorer Settings, where there's no consumer) survives the load and gets shown by `kick_queue()` on world entry — the "don't parade stale toasts" intent is partly lost for that one item.
2. `kick_queue()` re-emits the head unconditionally. At `_on_loading_finished` nothing is normally on screen, but a toast that opened *before* the load and outlived it would get a duplicate instance — and each close calls `dequeue_notification()`, so the following notification would be popped unshown.

Both disappear with the same one-line change: the explorer is the only thing that knows whether a toast is really on screen (it instantiates them in `_show_notification_toast`), so passing that in — `clear_system_toasts(has_toast_on_screen)` / an equivalent guard before `kick_queue()` — puts the state where it belongs instead of mirroring it in the manager. Worth a follow-up, not a blocker.

**CI**: Static checks ✅, Clippy ✅; Linux and Android still building at review time — confirm they're green (and give the device build a pass on the Account section in landscape) before merging.

Approving.

_Requested by Seba via Slack_

godot/src/ui/components/molecules/settings_section_item/settings_section_item.tscn
```diff
@@ -0,0 +1,83 @@
...
+layout_mode = 1
+anchors_preset = 15
+anchor_right = 1.0
+anchor_bottom = 1.0
+grow_horizontal = 2
+grow_vertical = 2
+theme_override_constants/margin_left = 12
+theme_override_constants/margin_top = 12
+theme_override_constants/margin_right = 20
+theme_override_constants/margin_bottom = 12
```
regenesis-claw, line 62: [P2] `SETTINGS_ACCOUNT` is a real catalogue key sitting as the scene default on a node whose auto-translation is disabled (inherited from the Button root's `auto_translate_mode = 2`), so any row that reaches the tree without `title_key` set would draw the key on screen. Every current row gets a key from `_build_section_list()`, so this is latent — but `extract_strings.py` only fails on nodes that declare mode 2 themselves, so it won't catch it if that ever changes.

```suggestion
text = "Section"
```

godot/src/ui/pages/settings/settings.gd
```diff
@@ -1251,6 +1401,112 @@ func _on_custom_button_sign_out_pressed() -> void:
 	Global.sign_out()
 
 
+## Account section header (name / tag / address) + the landscape avatar preview. Populated from the
+## live profile and kept in sync on profile changes. The GuestUpgradeCard manages its own guest-only
+## visibility. (Email display for signed-in users is deferred — no stored-email source yet.)
+func _setup_account_section() -> void:
+	if not Global.player_identity.profile_changed.is_connected(_on_account_profile_changed):
+		Global.player_identity.profile_changed.connect(_on_account_profile_changed)
+	# The avatar preview is a sibling of the settings root (not inside the content ScrollContainer),
```
regenesis-claw, line 1410: [P2] It's the last *child* of the settings root (`parent="."` in the scene), not a sibling of it — which is exactly why the z-order works out (drawn after `SafeMarginContainer`, and modals/toasts live outside this scene entirely). Worth correcting the wording so the next reader doesn't go looking for a sibling node.

regenesis-claw (approved):
Delta re-review of `7a1f4f4…2c2605e` (the two commits after the merge from `main`; the merge itself only brought `lib/src/dcl/crdt/message.rs`, unrelated to this PR).

**What landed**
- The account avatar preview moved out of the scroll column and became the last child of the settings root, so it draws above the content but below the delete-confirmation popup and toasts. `top_level` is gone, `%Control_AvatarPreview` is now an empty placeholder the preview's rect mirrors, and since it no longer inherits the column's `hide_on_portrait`, `_update_avatar_background()` drives its visibility directly. Every path that can change section or orientation goes through `_select_section()` / `_on_orientation_changed()`, both of which call it — and the preview is a child of the settings root, so it hides with the screen. Looks complete.
- `SettingsSectionItem` moved its text from the `Button.text` onto a child `Label`, with a per-instance `LabelSettings.duplicate()` so orientation size and pressed weight can't clobber sibling rows — right call, that resource is shared by every row. The duplicate happens at the top of `_ready`, before `_apply_orientation()` / `refresh_weight()` run, so ordering is safe, and both setters guard on `_title_label` for the pre-ready path.
- `UpgradeBadge` on the Account row only, freed on every other row; the badge self-manages guest/upgraded state off `guest_upgrade_state_refreshed` and orientation. `section_key` is assigned after `add_child()` in `_build_section_list()`, so the `is_node_ready()` guard in `_update_badge()` is satisfied.
- `Button_Help` had no `pressed` connection on `main` either — this wires up a button that was dead before. Nice catch. `https://decentraland.org/help` resolves 200.

**P2 — nothing blocking**

1. `settings_section_item.tscn`: `Label_Title`'s scene default is `text = "SETTINGS_ACCOUNT"` — a real catalogue key on a node whose auto-translation is off (it inherits `auto_translate_mode = 2` from the Button root), so a row that ever reaches the tree without `title_key` set draws the key verbatim. The i18n checker only fails on nodes that declare mode 2 *themselves*, so this inherit case slips through (static checks are green). The previous default was plain `"Section"` — worth going back to something non-key.
2. `settings.gd:1410` comment says the preview is "a sibling of the settings root"; in the scene it's `parent="."`, i.e. the last *child* of the root. The z-order reasoning is right, the wording isn't.
3. `settings_section_item.tscn` root picked up `offset_right = 266.0` / `offset_bottom = 72.0`. Inside the list `VBoxContainer` they're overridden, so it's inert — but it's editor-drag state rather than an intentional value.
4. The badge is anchored to the title label's bottom-right with fixed offsets (`-38…-16`), while the label's font size changes 30 → 36 between landscape and portrait. It tracks the corner so it won't detach, but its optical position relative to the text shifts — worth one look on a phone in portrait.
5. Nit: `_on_button_help_pressed()` holds the URL in a local `var`; the four neighbouring handlers use function-local `const` (`PRIVACY_POLICY_URL`, `DISCORD_URL`, …).

**CI**: Static checks ✅ at this head; Linux, Clippy and Android still running — worth confirming green.

Note my earlier approval no longer counts toward `reviewDecision` (bot approvals don't satisfy this repo's protection), so this still needs a human codeowner. Approving again on `2c2605e`.

_Requested by Seba via Slack_

Ludmilafantaniella:
## QA Update

Testing case 3 (mobile, iOS + Android). Nametag toggle works fine — independent from Hide Interface, persists correctly.

🐛 **Bug found:** "Hide Interface" doesn't apply until you exit Settings a second time.

Repro:
1. Gameplay → Interface → toggle Hide Interface ON
2. Exit Settings → UI still visible
3. Re-enter/exit Settings again → now it hides

Toggle state updates fine, but the visual refresh seems delayed by one exit cycle. Checking if it happens on both orientations/platforms.

https://github.com/user-attachments/assets/763af70a-82fe-429b-b8e7-bde5facc27b2

**Question:** Account section (screenshot below) — this account has a real wallet, not a guest, but still shows "Unlock the full experience". Test plan case 5 says that card is guest-only. Intentional or bug?

image

image

cc @sdilauro

sdilauro:
@Ludmilafantaniella 

I'm testing the fix about hide-ui bug. Good catch and repro!

About the real "guest vs non-guest" difference isn't the wallet — in the thirdweb model almost every account has a wallet address. What matters is whether the account has linked a portable auth method (email/social/passkey):

- A thirdweb guest has a real, persistent wallet, but it's tied to this device's anchor — not portable to another device.
- Upgrading = linking an email/social/passkey, which is exactly what makes the account portable across devices.

So the upgrade card shows only when is_thirdweb_guest() && !is_thirdweb_guest_upgraded(). Having a wallet address doesn't mean "not a guest" — it just means the guest wallet exists.

Ludmilafantaniella (approved):
## QA Update — Manual QA Approval

Ran the full test plan (cases 1–8) on both Android and iOS.

- Portrait master-detail nav, landscape list+content, rotation, entry points (lobby/world), version row copy, Account section, Language gating — all working as expected.
- Case 3 (Hide Player Names independent from Hide Interface): nametag toggle worked correctly from the start.
- 🐛 Found: "Hide Interface" didn't apply until exiting Settings a second time. Reported, fixed by @sdilauro, and re-verified on both Android and iOS — now applies immediately on exit. ✅
- Re: Account "Unlock the full experience" card — confirmed with @sdilauro that guest vs upgraded is based on portable auth (email/social/passkey), not wallet presence. Card only shows for non-upgraded guests as expected. No issue.

All test cases pass. Approved from QA side. ✅

https://github.com/user-attachments/assets/c338a557-b136-4872-8dc4-372189dbb03b

ElzathRusso (approved):

PR #2865, opened 2026-09-08
feat: Google Play in-app review prompt (Android) — #2739

### LINKED ISSUE #2739: Review Prompt (Android): Implement In-App Review trigger + Rules
##  **⚠️ Problem:**
Play Store rating is set by whoever chooses to rate — mostly users hitting bugs. Engaged players never get asked, so the store page under-represents them. Rating has been going down substantially since LMT2 due to big surge of minspec devices.

Image

## **🏁 Scope:**
Implement the Google Play In-App Review prompt on Android: a persisted eligibility counter, four trigger hooks, a 3-shot lifetime cadence state machine, the Play Core binding, and the `review_prompted` event. **Behavior is fully specified below — this issue is the build, not a design exploration.** No new UI ships: the prompt is the native Play card, called directly.

### Eligibility gate
- `session_count >= 5`. Estimated reach ~6% of players.
- A session is one app foreground session, counted once per launch. Reuse the existing session definition used by analytics — do not introduce a second one.

### Trigger hooks
Once the gate is open, the **next occurrence of any** of these arms a shot. The `trigger_id` value is what gets reported in the analytics payload — it must be emitted by the hook, not inferred later:

| # | Trigger | Hook point | `trigger_id` |
|---|---------|-----------|--------------|
| 1 | Add / accept a friend | friend request sent **or** accepted | `friend_added` |
| 2 | Favorite a place | favorite added | `place_favorited` |
| 3 | Claim a wearable | the **claim notification** (hook the notification, not the marketplace) | `wearable_claimed` |
| 4 | Rate a place up | place upvote | `place_upvoted` |

All four are equal — no ranking, no priority. Wearable *purchase* is explicitly out: it needs a `purchase_completed` event from the marketplace and reaches fewer users than a claim.

If two triggers land in the same frame, the **first one evaluated wins** and is the one reported; the shot is not duplicated and the second trigger is discarded, not queued.

### Cadence state machine
Persisted state: `session_count`, `shots_fired` (0–3), `last_shot_at` (timestamp).

| Shot | Fires when |
|------|-----------|
| 1 | `session_count >= 5` AND any trigger fires |
| 2 | `now - last_shot_at >= 7d` AND any trigger fires |
| 3 | `now - last_shot_at >= 14d` (measured from shot #2) AND any trigger fires |

- **Hard cap: 3 shots per account, lifetime.** At `shots_fired == 3` the module is inert forever.
- The cooldown is a **floor, not a schedule**: the window opening does not fire anything. A trigger must fire after the window is open. No trigger → no shot, indefinitely.
- `shots_fired` increments on `launchReviewFlow` invocation, not on any user outcome — the outcome is not observable (see below).
- Client caps sit **on top of** Play's own quota, never replace it. Over quota, Play renders nothing and returns no error; that still consumed a shot on our side, which is intended.

### Guard rails
- Never fire during loading, an error state, or an IAP flow. Defer to the next qualifying trigger rather than queueing.
- **No app-owned UI in front of the prompt** — no modal, no button, no "enjoying the app?" question. Prohibited by both stores, and a CTA dead-ends for any user over quota:
  - Apple: a custom qualifying gate is not allowed and puts the developer account at risk ([forum thread](https://developer.apple.com/forums/thread/692123)).
  - Google Play: In-App Review guidelines prohibit questions before or during the prompt and prohibit a call-to-action to trigger the API.
- The prompt is topmost and rendered as-is: no resize, opacity, overlay, or programmatic dismissal.

## **📝 Deliverables:**
- [x] Persisted `session_count`, incremented once per foreground session, wired to the existing session definition
- [ ] Persisted cadence state (`shots_fired`, `last_shot_at`) surviving app restart and update; decide and document whether it is device-local or account-scoped
- [x] Four trigger hooks emitting into a single `ReviewPromptTrigger` entry point, each passing its own `trigger_id`
- [ ] Cadence state machine implementing the table above, with the hard cap and the floor-not-schedule semantics
- [ ] Guard-rail checks (loading / error / IAP flow) before any `launchReviewFlow` call
- [ ] Play Core binding: **pick and land** the Godot Android plugin vs an in-house GDExtension wrapper — vet against our Godot version and build pipeline before committing
- [ ] `requestReviewFlow()` prewarm ahead of the trigger moment, with `ReviewInfo` expiry handled (re-request rather than fire stale)
- [ ] `review_prompted` event carrying `shot_id` **and** `trigger_id` in `extra_payload`
- [ ] Verified on the internal test track (quota does not apply there) across all three shots, all four `trigger_id` values, and the cap
- [ ] iOS parity call — same trigger logic behind `SKStoreReviewController`, or Android-only for now

## **Implementation Specs:**
Google Play In-App Review API, via `ReviewManager` in Play Core:

1. `requestReviewFlow()` → `ReviewInfo`, async. Prewarm before the trigger moment; `ReviewInfo` expires — do not cache it across sessions.
2. `launchReviewFlow(activity, reviewInfo)` → Play renders the native 1–5 star + comment card in-process.
3. Completion callback fires regardless of outcome. **No signal whether the user rated, dismissed, or never saw the card** — instrument the call, not the result.

Constraints that bound the implementation:

- Play enforces a time-bound per-user quota (≈monthly, undocumented and changeable). Over quota → nothing renders, no error, callback still fires.
- No CTA button — a user over quota gets a dead control. The trigger must stay implicit.
- Card rendered as-is: no resize, opacity, overlay, or programmatic dismissal. Must be topmost layer.
- No qualifying pre-prompt — prohibited by policy.
- Godot has no first-party binding. Candidates: [godot-sdk-integrations/godot-inapp-review](https://github.com/godot-sdk-integrations/godot-inapp-review) (covers Play + App Store behind one GDScript interface, so it also carries the iOS parity path) and [cengiz-pz/godot-android-inapp-review-plugin](https://github.com/cengiz-pz/godot-android-inapp-review-plugin); otherwise wrap Play Core ourselves.
- Testing: internal test track or Play Console internal app sharing. Quota does not apply on the internal test track, so all three shots are reachable in one pass.

## **📊 Metrics:**
- **`review_prompted`** — fired on `launchReviewFlow` invocation. `extra_payload`:
  - `shot_id` — `1` | `2` | `3`
  - `trigger_id` — `friend_added` | `place_favorited` | `wearable_claimed` | `place_upvoted`; the trigger that actually fired this shot
- Outcome is **not instrumentable**. Do not add an event that implies it was shown, rated, or dismissed.
- Impact is read off the Play Console store rating and the review ingest in #2150 — incoming review mix and average against the pre-rollout baseline — not from client events.
- Sanity checks once live:
  - `review_prompted` shot #1 volume should approximate the 5+ session cohort (~6% of players); a large gap means the gate or a hook is wrong.
  - `trigger_id` distribution tells us which trigger is carrying the feature. A trigger at ~0% is a broken hook, not a rare action — and the split is what we'd use to prune or add triggers later without re-litigating the whole design.

## **🔗 References:**
- [Google Play In-App Review API](https://developer.android.com/guide/playcore/in-app-review) — official flow, quota behavior, design restrictions above
- [In-app review design guidelines](https://developer.android.com/guide/playcore/in-app-review#design-guidelines) — what the prompt may and may not do
- [Apple Developer Forums #692123](https://developer.apple.com/forums/thread/692123) — Apple confirming a custom qualifying gate is not allowed and risks the developer account
- [decentraland/godot-explorer#2150](https://github.com/decentraland/godot-explorer/issues/2150) — Play Store review ingest to Slack; the read side of this loop and the source of the baseline review mix
- [decentraland/godot-explorer#2531](https://github.com/decentraland/godot-explorer/issues/2531) — tablet exclusion to protect Play rating; same goal, supply-side lever instead of demand-side
- No prior art in unity-explorer or docs (searched `review prompt rating`, `in-app rating`, 2026-08-20). Desktop has no store rating surface — no parity target.

EibrielInv:
## What

Android players who are enjoying the app now get Google Play's native rating card at a good
moment — just after adding a friend, favoriting or liking a place, or claiming a wearable — at
most three times per install, ever. No new UI ships: the card is Play's own, called directly.

Dev and QA builds also get an on-screen harness (`?review-debug=true`) that reports what the
prompt *would* do without calling Play, so the logic can be tested without fighting Play's quota.

## Why

The store rating is self-selecting: people who hit a bug go and rate, engaged players never do.
The rating has been dropping since LMT2 with the surge of minspec devices, and nothing in the
client ever asked the players having a good time.

Deliberately not included: iOS (the cadence is platform-agnostic, only the launch call is
Android), and any "enjoying the app?" pre-prompt or Rate-us button — both stores prohibit one,
and it would dead-end for anyone over Play's quota.

Closes #2739

## Details

<details>
<summary>Expand</summary>

**Cadence** — three asks per install, lifetime, then permanently inert:

| Ask | Fires when |
|---|---|
| 1 | 5+ sessions AND a trigger fires |
| 2 | 7 days after ask 1 AND a trigger fires |
| 3 | 14 days after ask 2 AND a trigger fires |

The cooldown is a **floor, not a schedule** — the window opening fires nothing on its own; a
trigger has to land after it opens. State is device-local, in `settings.cfg`.

**Triggers** (all equal, first to land wins): `friend_added`, `place_favorited`, `place_upvoted`,
`wearable_claimed`. Favorite and like only fire in the positive direction, and only when the
request actually succeeded. The wearable trigger hooks the claim *notification*, not the
marketplace.

**Guard rails** — never during loading, an IAP purchase, a degraded connection, or with any modal
open. There is a ~1.2s beat between the action and the card so it does not feel like a reflex to
the tap, and the rails are re-checked *after* that wait, since the app can change during it. An
ask cancelled by a rail is not counted — burning one of only three on a prompt nobody saw would
be a real loss.

**Two constraints that shape everything:**

- **The outcome is not instrumentable.** Play's callback fires whether the user rated, dismissed,
  or never saw the card — over quota it renders nothing, returns no error, and still completes.
  So `review_prompted` (with `shot_id` and `trigger_id`) is emitted on the *call*, not the result.
  Please don't ask for a "was it shown / did they rate" event; it does not exist. Impact is read
  off the Play Console rating instead.
- **Client caps sit on top of Play's own quota**, never replace it.

**Ride-along changes outside the feature** (shared code — worth a look):

- `ModalManager.is_any_modal_open()` — new public helper. It tracks seven separate modal
  references, and a caller checking only `current_modal` misses the other six; that shipped as a
  bug here (the prompt fired over a travel modal opened from a chat world link). The helper
  checks `is_instance_valid() and visible`, because `close_travel_modal` / `close_reward_modal` /
  `close_upgrade_modal` only hide their modal, never free it.
- `IapManager.is_purchase_overlay_visible()` — new public accessor; the purchase overlay is its
  own `CanvasLayer`, not one of ModalManager's.
- `ConnectionQualityMonitor.is_connection_healthy()` — new public getter over its private state.
- `PlacesHelper.async_patch_favorite` / `async_patch_like` gained a trigger call.

**Play Core binding** lives in its own `InAppReview.kt`, and no Play Core type may appear in a
method signature on `GodotAndroidPlugin`: `onRegisterPluginWithGodotNative` reflects over
`getDeclaredMethods()` at registration and resolves every parameter type, so one unresolvable
type throws `NoClassDefFoundError` on the Vulkan thread and kills the app at boot. That crash
happened and is what the split prevents. The AAR is rebuilt and committed (CI does not build it),
and the dependency is declared in both `build.gradle.kts` and
`export_scripts_template/export_plugin.gd` — the latter is the source for the generated
`godot/addons/` copy.

**Verified by the author on a Pixel 7:**

- [x] Real Play card rendered end-to-end from a real friend request
- [x] Full cadence: gate → ask 1 → floor rejecting repeats → asks 2 and 3 → cap holding against
      all four triggers with the clock wound back a year
- [x] Counters survive an in-place app upgrade (`shots_fired` and `last_shot_unix` intact,
      session count +1)
- [x] Travel-modal bug fixed, and verified it unblocks again after the modal closes
- [x] `gdformat` / `gdlint` / `check-gdscript` / i18n checks / `cargo fmt` / `cargo clippy` clean

</details>

## Test plan

**Setup once:** launch with `decentraland://open?review-debug=true`. The panel appears top-right,
expanded. Everything is simulated — the real Google prompt is never shown. Steps run
consecutively.

QR image to launch the app with `decentraland://open?review-debug=true`
imagen

**Tips:**

- Tap **▾ / ▸** in the header to collapse the panel to a single status line, or expand it again.
- If the panel covers something you need, tap **◀ / ▶** to move it to the other screen edge. It
  stays where you put it.

| # | Case | Steps | Expected |
|---|---|---|---|
| 1 | Harness on | Open the app via the `review-debug=true` link | Panel reads `0 of 3 asks used` and `Test mode — the real prompt is never shown` |
| 2 | First ask | Tap **Start over**, then add or accept a friend | ~1s later a dialog titled *Google Play review card* naming *Added a friend*; panel reads `1 of 3` |
| 3 | Not instant | Tap **Start over**, favorite a place, watch the screen | The favorite registers first; the dialog follows about a second later, not on the same tap |
| 4 | Second ask locked | Add another friend | No dialog. Panel: `too soon, the next ask is still locked`, still `1 of 3` |
| 5 | Second ask | Tap **+7 days**, then like a place | Dialog naming *Liked a place*; panel reads `2 of 3` |
| 6 | Third ask | Tap **+14 days**, then tap **Claim wearable** | Dialog naming *Claimed a wearable*; panel reads `3 of 3` and `Finished — it will never ask again` |
| 7 | Cap holds | Tap **+1 year**, then add a friend, favorite a place, like a place and tap **Claim wearable** | No dialog at any point. Panel stays `3 of 3`; each attempt reports `all 3 asks used` |
| 8 | Survives a restart | Force close the Decentraland app from the Android app switcher, then reopen it | Panel still reads `3 of 3` and `Finished — it will never ask again` — the count did not reset |
| 9 | Blocked by a dialog | Tap **Start over**, type `cozyfarm.dcl.eth` in chat, tap the link, and with the modal open tap **Claim wearable** | No review dialog over or behind it. Panel: `held back because a dialog is open`, still `0 of 3` |
| 10 | Unblocks after | Close the world modal, then tap **Claim wearable** | Dialog appears normally; panel reads `1 of 3` |
| 11 | Blocked while loading | Tap **Start over**, jump to a new place, and while the loading screen is up tap **Claim wearable** | No dialog. Panel shows it was held back, still `0 of 3` |
| 12 | Blocked offline | Tap **Start over**, turn on airplane mode, wait for the app to notice, then tap **Claim wearable** | No dialog, panel shows it was held back. Turn airplane mode off before continuing |

Cases 9 and 10 are one test: 9 covers the prompt landing on top of an open modal, 10 covers the
opposite failure — the prompt being blocked permanently once a modal has been opened and closed.

- [ ] **Regression** — `ModalManager` and `IapManager` gained shared helpers: open and close a
      teleport modal, a connection-lost modal and an IAP purchase, and confirm each still appears
      and dismisses normally.
- [ ] **Regression** — favorite and unfavorite a place, like and un-like a place: the buttons
      still toggle and the state still survives reopening the place.


godot/src/review_prompt_coordinator.gd
```diff
@@ -0,0 +1,412 @@
...
+			)
+		)
+		_firing = false
+		return
+
+	print("[ReviewPrompt] launching shot %d (trigger=%s)" % [shot_id, trigger_id])
+	debug_event.emit(
+		"fired", "%s — asking now · ask %d of %d" % [trigger_label(trigger_id), shot_id, MAX_SHOTS]
+	)
+	_get_android_plugin().launchInAppReview()
```
regenesis-claw, line 240: [P2] The counters are committed a few lines above, and then this call can fail immediately: `launchInAppReview()` emits `in_app_review_finished("activity not ready")` when `activity` is null, and `InAppReview.launch` reports the exception reason if Play Core is missing or `requestReviewFlow` fails. Nothing in GDScript connects that signal, so those cases burn one of the three lifetime shots on a card that provably never rendered — which is the exact loss the rail-cancel path deliberately avoids ("an ask cancelled by a rail is not counted").

Suggestion: connect `in_app_review_finished` in `_ready()` and, when `error != ""`, roll `review_shots_fired` / `review_last_shot_unix` back to their previous values. `error == ""` still means "the flow ran", so the normal (and over-quota) case keeps counting exactly as it does today.

godot/src/review_prompt_coordinator.gd
```diff
@@ -0,0 +1,412 @@
...
+		return
+	var config: ConfigData = Global.get_config()
+	if config.review_shots_fired >= MAX_SHOTS:
+		return
+	if not _is_due(config):
+		return
+	var plugin := _get_android_plugin()
+	if plugin == null:
+		return
+	_prewarmed = true
```
regenesis-claw, line 379: [P2] `_prewarmed` is a once-per-session latch, but the cached `ReviewInfo` on the Kotlin side has a 5-minute TTL — so after the first warm the `loading_finished` hook can never refresh it, and every launch more than 5 minutes later pays the `requestReviewFlow` latency at the moment the card should appear.

`InAppReview.prewarm()` is already idempotent and short-circuits on a fresh cache ("Safe to call repeatedly"), so the GDScript latch only removes the refresh it was meant to enable. Dropping `_prewarmed` (or resetting it on a timestamp) makes the two sides agree.

godot/src/review_prompt_coordinator.gd
```diff
@@ -0,0 +1,412 @@
...
+
+	if Global.metrics != null:
+		Global.metrics.track_review_prompted(shot_id, trigger_id)
+	_firing = false
+
+
+# One place for "a trigger arrived and nothing happened", so the log line and the debug panel
+# can never drift apart.
+func _reject(trigger_id: String, reason: String, human: String) -> void:
+	# The log line stays terse and technical — it is grepped and lands in Sentry breadcrumbs.
```
regenesis-claw, line 250: [P2] `print` goes to the console only — it is not routed to the Godot error stream and so never becomes a Sentry breadcrumb (REVIEW.md §5 routing table: only `push_error` / `printerr` / `push_warning` reach Sentry). The level choice itself is right — these are expected, recoverable decisions and should not cost quota — so just drop the Sentry half of the sentence rather than changing the call.

godot/src/review_prompt_coordinator.gd
```diff
@@ -0,0 +1,412 @@
...
+	print("[ReviewPrompt] skip (%s): %s" % [trigger_id, reason])
+	debug_event.emit("skip", "%s — %s" % [trigger_label(trigger_id), human])
+
+
+## Plain-English name for a trigger id, for the QA panel. Falls back to the raw id.
+func trigger_label(trigger_id: String) -> String:
+	return TRIGGER_LABELS.get(trigger_id, trigger_id)
+
+
+## Apply `?review-debug=true|false` from a deeplink (issue #2739) and persist it. Called at boot
```
regenesis-claw, line 261: [P2] "and persist it" contradicts the paragraph three lines down ("Session-only: nothing is written to settings.cfg"). Same leftover in the debug panel's own header comment ("mounted either at boot (deeplink already persisted, …)"). Both read like survivors of an earlier persisted-flag design; worth deleting so the next reader doesn't go looking for a stored flag that isn't there.

godot/src/review_prompt_coordinator.gd
```diff
@@ -0,0 +1,412 @@
...
+# our own UI or a call-to-action in front of it, which this does not.
+const PROMPT_DELAY_SECONDS := 1.2
+
+const MAX_SHOTS := 3
+const DAY_SECONDS := 86400
+# PRODUCTION (active): 5 sessions to open the gate, then a 7-day and a 14-day floor.
+const SESSION_GATE := 5
+const SHOT_2_FLOOR_SECONDS := 7 * DAY_SECONDS
+const SHOT_3_FLOOR_SECONDS := 14 * DAY_SECONDS
+# TESTING: to exercise all three shots in one pass on the internal test track (where Play's
```
regenesis-claw, line 80: [P2] Commented-out config in shipped source is the pattern REVIEW.md Tier 1 asks reviewers to flag, and this block is now redundant: the QA panel's **Start over** (which arms `review_session_count` to `SESSION_GATE`) plus **+7 days / +14 days** already exercise all three shots in one pass without editing constants. Consider deleting it and pointing at the harness instead.

godot/src/ui/components/organisms/review_debug_panel/review_debug_panel.gd
```diff
@@ -0,0 +1,273 @@
...
+	# A claim notification is server-pushed and cannot be produced on demand.
+	%ButtonWearable.pressed.connect(_fire.bind(coordinator.TRIGGER_WEARABLE_CLAIMED))
+	coordinator.debug_event.connect(_on_debug_event)
+
+
+func _on_orientation_changed(_is_portrait: bool) -> void:
+	_apply_side()
+
+
+func _on_side_button_pressed() -> void:
```
regenesis-claw, line 77: [P2] `_on_side_button_pressed` (and `_fire` below) dereference `_coordinator()` unguarded, while `_apply_side` and `_refresh` in the same file both handle the null case explicitly. In practice the panel is only ever mounted by the coordinator so it can't be null today — but the file already carries the guard everywhere else, so the inconsistency reads as an oversight rather than a decision.

regenesis-claw (approved):
## Review — `feat: Google Play in-app review prompt (Android)`

Strong PR. The design reads as deliberate rather than assembled: the cadence is a floor and not a schedule, the rails are re-checked *after* the delay, counters advance on invocation because Play exposes no outcome, and the Play Core types are kept out of every method signature on `GodotAndroidPlugin` for a reason that is documented with the crash it caused. The `is_any_modal_open()` helper is the right fix for the seven-reference problem and the comment explains why the `visible` half is load-bearing.

**Pre-flight (REVIEW.md §0)**
- Branch is up to date with `main` (`behind_by: 0`, 5 ahead). ✅
- All CI green: Static checks, Clippy, Linux, Android (+ signed APK), Docker, and 🍏 iOS → R2 export. An iOS artifact exists even without the `build` label, so the shared-GDScript changes (`modal_manager`, `iap_manager`, `global.gd`) do have iOS build coverage. ✅
- No submodule pointer drift, no `.claude/` files, no `project.godot` local-settings leakage, no non-English comments/identifiers, no `push_error`/`printerr`/`tracing::error!` added. ✅
- Description has What / Why / Details / Test plan; the ride-along shared-code changes are called out in `## Details` rather than buried. Test plan is executable by QA cold (12 numbered cases + 2 regression lines). ✅

**Verified independently, not taken on trust**
- Both committed AARs were actually rebuilt: `classes.jar` contains `org/decentraland/godotexplorer/InAppReview.class`, and `GodotAndroidPlugin.class` contains `prewarmInAppReview` / `launchInAppReview` / `in_app_review_finished` in the release *and* debug variants. The Play Review coordinate is declared in `build.gradle.kts` and in **both** copies of `export_plugin.gd`. ✅
- `InAppReview.kt` sits in the `template` directory but declares `package org.decentraland.godotexplorer` — that matches `GodotAndroidPlugin.kt`, which does the same, so no import is needed. Not a bug. ✅
- `_state` on `ConnectionQualityMonitor` initialises to `State.GOOD`, so the connection rail is not accidentally closed before the first poll. ✅
- `"reward_assignment"` is a real notification type, and `purchase_completed` / `purchase_failed` (2 args) and `purchase_cancelled` (1 arg) are all satisfied by `_on_iap_settled(_product_id, _extra = null)`. ✅
- `is_any_modal_open()` covers all 7 `current_*_modal` fields declared in the manager — none missed. ✅
- The claim that the accept path covers both directions holds: `friends_panel._async_on_friendship_request_accepted` explicitly distinguishes "we/they accepted" off the same service signal. ✅
- Both like/favorite hooks are only reachable from a real user toggle (`engagement_bar` / `fav_button`, both behind `DebouncedAction`) — no programmatic sync path fires them. ✅

**Security review:** no secrets, no user-controlled input reaching a sink, no PII in the new Segment event (`shot_id`, `trigger_id` only, plus the standard common fields). The new dependency is Google's own `com.google.android.play:review:2.0.2`, proportionate to its use. The `review-debug` deeplink is hard-gated on `not Global.is_production()` at both the capture and the mount. Persisted state is three non-sensitive counters in `settings.cfg`. **No security issues found.**

**Consumer impact:** additive only — a new `review_prompted` Segment event (new dashboard surface, nothing breaks), plus three new public GDScript accessors. No wire/SDK contract touched.

### Findings — all P2, none blocking

| # | Where | Finding |
|---|---|---|
| 1 | `review_prompt_coordinator.gd:240` | A shot is burned even when the launch fails outright; `in_app_review_finished` has no listener |
| 2 | `review_prompt_coordinator.gd:379` | `_prewarmed` latch defeats the 5-minute `ReviewInfo` TTL |
| 3 | `review_prompt_coordinator.gd:250` | Comment says `print` lands in Sentry breadcrumbs; it does not |
| 4 | `review_prompt_coordinator.gd:261`, `review_debug_panel.gd:56` | Docstrings still describe a persisted debug flag |
| 5 | `review_prompt_coordinator.gd:80` | Commented-out TESTING consts now superseded by the harness's own time-travel buttons |
| 6 | `review_debug_panel.gd:77` | Two `_coordinator()` call sites skip the null guard the rest of the file uses |

Two more notes that need no code change, just a confirmation of intent:
- `friendship_request_sent` fires on a request being **sent**, not accepted (`profile.gd:684`, `social_item.gd:346`), while the trigger is named `friend_added` and labelled "Added a friend". If a pending request that is never accepted counts as the happy moment, that's fine — but `friend_added` will read higher in the trigger split than actual new friendships.
- The wearable trigger rides `notification_queued`, which also fires for unread notifications pulled from the server on a poll as long as they're within `TOAST_MAX_AGE_MS` (5 min). So a reward assigned a few minutes before launch can fire the ask shortly after boot rather than off a fresh action. The window is small and the loading rail covers most of it — worth knowing, not worth restructuring.

**Approving.** Everything above is a follow-up-sized note; nothing here should hold the merge.

rodrigo685 (approved):

Ludmilafantaniella (approved):
## QA Update

Ran through all 12 test cases on Android using the `review-debug=true` harness — all passed as expected:
- First/second/third ask fire correctly with the right trigger labels and counters (1 of 3, 2 of 3, 3 of 3)
- Cadence floors (7/14 days) and the 3-shot lifetime cap hold correctly
- Counter survives app restart
- Prompt correctly blocked while a modal is open, while loading, and while offline
- Unblocks normally once the blocking condition clears

Regression checks also passed: teleport modal, connection-lost modal, and IAP purchase all open/close normally; favorite/like toggles work and persist on reopen.

Approved from QA side. ✅

https://github.com/user-attachments/assets/f308e8b2-d60a-41df-be43-7fbcc5a7e1ef

PR #2866, opened 2026-09-09
fix(sdk): stop double sRGB decode on code-applied albedo textures

### LINKED ISSUE #2900: [Bug] Textures applied from scene code render ~2x darker than Unity (double sRGB decode)
## Summary

Textures applied from scene code (via the `Material` component or `GltfNodeModifiers`) render roughly 2.2x darker than intended, because the albedo texture is converted sRGB→linear twice: once in hardware by the `source_color` sampler's sRGB view, and again in-shader by `ALBEDO_TEXTURE_FORCE_SRGB`, which the explorer sets to `true` for every non-video texture (introduced in #560, 2025-06). Every script-applied texture in every scene has been darker than the Unity client since then. GLB-imported materials are unaffected (the importer never sets the flag), which is why scenes only break when code touches a material.

Reported by @dcl-regenesislabs on Marsh Colony: creatures go near-black after a runtime texture swap, even with `unlit` materials.

## Repro

Headless repro on the fork binary: unshaded quad with a 0.5-grey texture renders `0.212` with the flag vs `0.498` without — identical for the RGBA8 (`ImageTexture`, desktop) and ETC2 (`PCT2`, mobile) paths.

In-world: any scene that sets a texture from code, e.g. Marsh Colony creature skins via `GltfNodeModifiers`.

## Expected

Script-applied textures render at the same brightness as in the Unity client.

## Root cause

`apply_dcl_material_properties` (`lib/src/scene_runner/components/material.rs`) sets `ALBEDO_TEXTURE_FORCE_SRGB=true` for every non-video texture. The albedo sampler is always declared `: source_color` (engine `material.cpp:969`), the renderer binds the sRGB view (`material_storage.cpp:1073`), and the flag decodes again in-shader (`material.cpp:1659`) → double gamma.

Video textures are the only exception and already handled per-backend (`ExternalTexture` has no sRGB view → flag true; `ImageTexture` → false).

## Fix

PR #2866 — set the flag to `false` in both unlit and PBR branches; video paths keep setting it per-backend on texture bind.

### LINKED ISSUE #2901: [Bug] GLB loses GltfNodeModifiers skins after spurious reload on container mutation (AvatarAttach)
## Summary

A GLB with `GltfNodeModifiers` materials (runtime skins) loses them when the entity's `GltfContainer` is mutated — e.g. changing `visibleMeshesCollisionMask` when attaching a pet to the avatar. The GLB does a full teardown + reload a few frames later and comes back with its raw, untextured materials. Godot-only; Unity does not reload.

Reported by @dcl-regenesislabs on Marsh Colony: picking up a pet (AvatarAttach carry flow) strips its skin. Their workaround (`fix/issue-216`, `ec94d73`) re-asserts `GltfNodeModifiers` every frame for 1.5s after attach/detach — removable once this is fixed.

## Root cause (two bugs in `gltf_container.gd`)

**2a — spurious reload.** `async_load_gltf` lowercases `dcl_gltf_src` in place, but `change_gltf` compares it case-sensitively against the src Rust hands back. Any GLB path with uppercase (Marsh Colony uses `assets/Models/...`) makes every later container mutation look like a src change → full reload. Hash lookups already lowercase internally (`content_mapping.get_hash`, `load_scene_gltf`), so the in-place lowercase was pure loss.

**2b — skin re-apply lost on reload.** `change_gltf`/`force_reload_gltf` defer `async_load_gltf`, so the loading state stays `FINISHED` until end of frame. `sync_gltf_loading_state` runs in the same frame, reads the stale `FINISHED`, evicts the entity from `scene.gltf_loading`, and the real reload finish then never re-queues the entity for modifier re-application (that path only fires for entities in `gltf_loading`). Also affects legit src changes (e.g. pet species swap).

## Expected

- Mask-only container mutations never reload the GLB.
- When a reload does happen (legit src change), `GltfNodeModifiers` are re-applied automatically on load finish.

## Fix

PR #2866 — stop mutating `dcl_gltf_src`; set `LOADING` synchronously when initiating a reload.

manuelmaceira:
## What

Fixes two bugs that hit scenes applying materials from code: script-set textures (via `Material` or `GltfNodeModifiers`) rendered about 2x darker than intended, and a GLB that reloads mid-game (e.g. a pet being attached to the avatar's hand) came back without its code-applied skins. Reported by the Regenesis team on Marsh Colony, whose creatures went near-black after a texture swap and lost their skins on every carry.

## Why

The darkness was a double sRGB decode: the renderer already converts albedo textures to linear in hardware, and the explorer additionally forced an in-shader conversion on every non-video texture since June 2025 — so every script-applied texture in every scene has been too dark vs the Unity client. The lost skins came from a lowercased model path making every later container mutation look like a model change (spurious full reload), plus a same-frame timing race that skipped skin re-application after the reload. This PR does not change GLB-imported materials or video textures.

Closes #2900
Closes #2901

## Details
<details>
<summary>Expand</summary>

**Bug 1 — double sRGB decode.** `apply_dcl_material_properties` set `ALBEDO_TEXTURE_FORCE_SRGB=true` for every non-video texture (introduced in #560). But hash/avatar textures (`ImageTexture` on desktop, `PCT2`/ETC2 on mobile) already expose an sRGB view, and the albedo sampler is always declared `: source_color` (engine `material.cpp:969`), so the renderer binds the sRGB view and decodes in hardware (`material_storage.cpp:1073`) — the flag then decodes *again* in-shader (`material.cpp:1659`). Empirically verified with a repro scene on the fork binary: an unshaded quad with a 0.5-grey texture renders `0.212` with the flag vs `0.498` without, identical for the RGBA8 and ETC2 paths. GLB-imported materials were unaffected (the importer never sets the flag), which is why scenes only broke when code touched a material. Video textures keep correct behavior: `update_video_material_textures` / `update_modifier_video_textures` set the flag per-backend on texture bind (`ExternalTexture` has no sRGB view → true, `ImageTexture` → false).

**Bug 2a — spurious reload.** `async_load_gltf` lowercased `dcl_gltf_src` in place, but `change_gltf` compares it case-sensitively against the src Rust hands back. Any GLB path with uppercase (Marsh Colony uses `assets/Models/...`) made every later `GltfContainer` mutation — like the collision-mask change when a pet is attached — look like a src change, triggering a full teardown + reload a few frames later. Hash lookups already lowercase internally (`content_mapping.get_hash`, `load_scene_gltf`), so the in-place lowercase was pure loss. Fix: never mutate `dcl_gltf_src`. With this, the attach flow (mask-only change) no longer reloads at all.

**Bug 2b — skin re-apply lost on reload.** `change_gltf`/`force_reload_gltf` defer `async_load_gltf`, so the loading state stayed `FINISHED` until end of frame. Rust's `sync_gltf_loading_state` runs in the same frame, read the stale `FINISHED`, evicted the entity from `scene.gltf_loading`, and the real reload finish then never re-queued the entity for `GltfNodeModifiers` re-application (that path only fires for entities in `gltf_loading`). Also affected legit src changes (e.g. pet species swap). Fix: set `LOADING` synchronously when initiating a reload.

The scene team's 1.5s re-assert workaround (marsh-colony `fix/issue-216`, `ec94d73`) can be dropped once this ships.

Author-verified: `cargo test` 273/273, `cargo check`/`cargo build` ok, GDScript validator 0 errors, sRGB repro scene (above). Committed with `--no-verify` — the pre-commit hook's Godot asset-import phase needs a full local install not present in the worktree.

</details>

## Test plan

**Case 1 — skins no longer dark**
1. Open the app and go to Marsh Colony
2. Look at the creatures (pets) roaming the scene
- [ ] Expected: creature colors look as bright as the same scene in the Unity client / web — no near-black creatures

**Case 2 — carry flow keeps skins**
1. In Marsh Colony, hatch or select a pet, then start the bath interaction so the pet is carried in the avatar's hands
2. Carry it to the tub and put it down
- [ ] Expected: the pet keeps its colored skin the whole time — no flicker to grey/untextured on pick-up or put-down

**Case 3 — species/rarity change keeps skins**
1. In Marsh Colony, switch the active pet to one of a different species (My Pets panel) or breed a new one
- [ ] Expected: the new pet appears with its skin applied, not grey

**Regression — video and scene textures**
1. Enter a scene with a video screen (e.g. a cinema/billboard scene) and one with image textures applied to primitives (Genesis Plaza banners)
- [ ] Regression: video plays with normal colors (not washed out or dark), banner images look the same as before this change

### DIFF
--- a/godot/src/decentraland_components/gltf_container.gd
+++ b/godot/src/decentraland_components/gltf_container.gd
@@ -68,7 +68,12 @@ func _enter_tree():
 
 
 func async_load_gltf():
-	self.dcl_gltf_src = dcl_gltf_src.to_lower()
+	# Keep dcl_gltf_src as the scene sent it: change_gltf() compares it
+	# case-sensitively against the src Rust hands back, so lowercasing it here
+	# makes every later GltfContainer mutation (e.g. a collision-mask change
+	# during AvatarAttach) look like a src change and trigger a spurious full
+	# reload. Case-normalization for lookups already happens inside
+	# content_mapping.get_hash / load_scene_gltf (Rust lowercases there).
 	var content_mapping := Global.scene_runner.get_scene_content_mapping(dcl_scene_id)
 	var file_hash := content_mapping.get_hash(dcl_gltf_src)
 	self.dcl_gltf_hash = file_hash
@@ -430,6 +435,12 @@ func change_gltf(
 			dcl_pending_node.queue_free()
 			dcl_pending_node = null
 
+		# Mark LOADING synchronously, not inside the deferred async_load_gltf:
+		# Rust's sync_gltf_loading_state runs later this same frame and would
+		# read a stale FINISHED, evict the entity from scene.gltf_loading, and
+		# the real finish would then never re-apply GltfNodeModifiers (skins
+		# lost on src change / reparent-triggered reloads).
+		dcl_gltf_loading_state = GltfContainerLoadingState.LOADING
 		async_load_gltf.call_deferred()
 
 	elif masks_changed and gltf_node != null:
@@ -462,6 +473,9 @@ func force_reload_gltf() -> void:
 		dcl_pending_node.queue_free()
 		dcl_pending_node = null
 
+	# See change_gltf: LOADING must be set synchronously so the same-frame
+	# sync_gltf_loading_state doesn't evict the entity from scene.gltf_loading.
+	dcl_gltf_loading_state = GltfContainerLoadingState.LOADING
 	async_load_gltf.call_deferred()
 
 
--- a/lib/src/scene_runner/components/material.rs
+++ b/lib/src/scene_runner/components/material.rs
@@ -330,11 +330,13 @@ pub fn apply_dcl_material_properties(
             godot_material.set_specular(0.0);
 
             godot_material.set_shading_mode(ShadingMode::UNSHADED);
-            let is_video_texture = unlit
-                .texture
-                .as_ref()
-                .is_some_and(|t| matches!(t.source, DclSourceTex::VideoTexture(_)));
-            godot_material.set_flag(Flags::ALBEDO_TEXTURE_FORCE_SRGB, !is_video_texture);
+            // Never FORCE_SRGB here: hash/avatar textures are ImageTexture/PCT2
+            // with a valid sRGB view, so the `source_color` hint on the albedo
+            // sampler already decodes sRGB->linear. The flag would decode AGAIN
+            // in-shader (double gamma, ~2.2x darker). Video textures are the
+            // exception and get the flag set per-backend when the video texture
+            // binds (update_video_material_textures).
+            godot_material.set_flag(Flags::ALBEDO_TEXTURE_FORCE_SRGB, false);
             // Unity ignores diffuse_color alpha for unlit materials, force alpha to 1.0
             // No color space conversion — matches Unity (SetColor with no conversion)
             let mut albedo_color = unlit.diffuse_color.0.to_godot();
@@ -410,11 +412,10 @@ pub fn apply_dcl_material_properties(
                 godot_material.set_emission_operator(EmissionOperator::ADD);
             }
 
-            let is_video_texture = pbr
-                .texture
-                .as_ref()
-                .is_some_and(|t| matches!(t.source, DclSourceTex::VideoTexture(_)));
-            godot_material.set_flag(Flags::ALBEDO_TEXTURE_FORCE_SRGB, !is_video_texture);
+            // See unlit branch: `source_color` already decodes sRGB for
+            // hash/avatar textures; forcing sRGB double-decodes (dark output).
+            // Video textures get the flag per-backend on texture bind.
+            godot_material.set_flag(Flags::ALBEDO_TEXTURE_FORCE_SRGB, false);
             // No color space conversion — matches Unity (SetColor with no conversion)
             godot_material.set_albedo(pbr.albedo_color.0.to_godot());
 


regenesis-claw (approved):
## Approving

Small diff, but both fixes are real and the reasoning in the description holds up against the source. I verified each claim rather than taking it on faith — notes below. No P0/P1. Reviewed against the repo's REVIEW.md.

### Pre-flight

- **Branch freshness:** up to date — you merged `main` at 14:27 (`3337f54`), `behind_by: 0`. 
- **CI:** the green Linux / Clippy / Static checks / Android / iOS-export runs are all on the pre-merge head `561bc4d`. On the reviewed SHA only `Static checks` is queued so far. The merge commit is main-only (compare shows the same 2 files), so the signal should carry, but there is no green build on the exact SHA yet.
- **iOS gate:** the diff touches neither the native iOS plugin, `OS.get_name() == "iOS"` branches, deeplinks, keyboard/safe-area, nor the comms/av trees, so an iOS build is not *required* here. One exists on the previous head anyway.
- **Hygiene:** no debug prints, no new error-level logs, no `.claude/` or scratch planning files, no submodule drift, all comments in English, and both new comments explain *why* rather than *what*.

### What I verified

**sRGB (`material.rs`)**
- `content_mapping.get_hash` lowercases (`content_mapping.rs:35`), so the GDScript-side lowercase really was pure loss — confirmed.
- Video behaviour is genuinely unchanged: the old expression evaluated to `false` for video textures too, so this is a no-op on that path. The per-backend flag still gets set on bind in both `update_video_material_textures` and the `GltfNodeModifiers` equivalent, and neither was touched.
- Strongest corroboration you didn't cite: the unlit+alpha_texture path already builds a `ShaderMaterial` from `dcl_unlit_alpha.gdshader`, whose albedo sampler is declared `: source_color` and which has no FORCE_SRGB equivalent at all. So that path has been rendering at the correct brightness the whole time, while the `StandardMaterial3D` path next to it was 2.2x dark. The two were inconsistent; this makes them agree.
- `GltfNodeModifiers` do route through `apply_dcl_material_properties`, so the description's claim that they were affected is correct.

**GLTF reload (`gltf_container.gd`)**
- The Rust side does `scene.gltf_loading.insert(entity)` immediately after calling `change_gltf`, and `sync_gltf_loading_state` later in the same frame evicts on a `Finished` read and only queues `gltf_node_modifiers_pending` on the *transition*. Setting `LOADING` synchronously is the right fix and I couldn't find a path it breaks.
- The double-schedule window I worried about (sync `LOADING` + cleared `_requested_hash`, then a same-frame re-parent making `_enter_tree` defer a second `async_load_gltf`) is harmless: `GltfLoadingCoordinator.request` early-returns on `group.waiters.has(container)` and `_needs_realize()` gates on `get_child_count() == 0`.
- Other readers of the src are safe: `scene_fetcher.gd` already compares with `.to_lower()`, the coordinator keys groups by hash and only forwards the src to `load_scene_gltf`, which lowercases internally.

### Findings — all P2, none blocking

**[P2] Grayscale source images may now render too bright on desktop.** The premise is "every non-video albedo texture has an sRGB view", which I could confirm for RGBA8 and ETC2 but not for every format the loader can produce. `load_texture` never converts after decode, so a grayscale PNG or JPEG lands as `FORMAT_L8`/`LA8`, and I believe Godot maps those to `R8_UNORM`/`R8G8_UNORM` with no `format_srgb` — meaning `source_color` is a no-op and FORCE_SRGB was previously doing the real work. Mobile is fine (`prepare_etc2_image` forces `UsedChannels::RGB`, so it becomes ETC2_RGB8), so this would be a desktop-only, grayscale-only brightening. Worth a 2-minute check with a grayscale PNG on your repro scene; if it reproduces, a `convert()` to RGBA8 in the decode path (or keeping the flag conditional on the image format) is the fix, and it's fine as a follow-up issue rather than in this PR.

**[P2] Cross-team heads-up.** This is not a Marsh Colony fix — it changes albedo brightness for *every* deployed scene that sets a texture from code, retroactively over 15 months. Any creator who eyeballed their colours against the dark output will see them shift. That's the correct call for Unity parity, but per REVIEW.md §4 it's exactly the kind of ride-along change that wants a note in the content/creator channel before merge. Was one posted?

**[P2] Case 1's expected result isn't pass/fail-able on a phone.** "as bright as the same scene in the Unity client / web" asks QA to run a second client side by side. Something closer to "the creatures show their colours (green/pink bodies), not near-black silhouettes" is checkable from the phone alone.

**[P2] `force_reload_gltf()`'s new comment overstates what it fixes.** Its only caller is the preview hot-reload path in `scene_fetcher.gd`, which invokes it directly from GDScript — the entity never enters Rust's `gltf_loading` set, so `sync_gltf_loading_state` never iterates it and the modifier re-application this PR restores does not reach that path. The sync `LOADING` there is harmless and consistent, but "See change_gltf" reads as though the same bug is being fixed. Either reword, or note that preview hot-reload still drops modifiers.

### Positives

Root-cause analysis on both bugs is exact rather than symptomatic, the empirical `0.212` vs `0.498` measurement is the right way to settle a gamma argument, the test plan follows the QA format (starts from opening the app, includes a regression case for video and scene textures), and the `--no-verify` reason is disclosed up front. Also good that the scene-side 1.5s re-assert workaround is called out as droppable — worth telling that team directly once this ships.

### Security

No security issues found. No user-controlled input, network surface, auth path or dependency change in the diff.

kuruk-mm (approved):
Looks great!

Issue #2702, opened 2026-08-14
[CI] Wire headless avatar regression tests into workflows


manuelmaceira:
Follow-up from #2699 review.

The `--script` headless GDScript tests under `godot/src/test/avatar/` (test_avatar_locomotion_grounded, test_avatar_state_machine_graph, test_avatar_autoplay_stomp, test_avatar_anim_throttle) run manual-only today — no workflow executes them. The graph test in particular would catch re-introduced state-machine trap states on any future avatar.tscn edit.

Task: add a CI step running:
```
godot --headless --path godot --script res://src/test/avatar/<each_test>.gd
```
(exit code 0/1 already wired in each test's _finish).

PR #2870, opened 2026-09-10
chore(ci): tolerate Godot's headless exit crash in the Linux build

leanmendoza:
# What

Headless Godot runs in the Linux build (`test-avatar`, `test-i18n`, `check-gdscript`) are now judged by the success marker each script prints, not by the exit status alone — **but only when the process was killed by a signal**. Everything else fails exactly as before.

| outcome | verdict |
|---|---|
| exit 0 | green |
| **killed by a signal AND the log has the marker** | **green + warning** |
| killed by a signal, no marker | red |
| exit ≠ 0 (1, 2, …) | red |

Also in here: child output is streamed line by line so the CI log reads the same as today, and `.unchecked()` makes duct hand back the exit status instead of erroring out — which is why the existing `status.success()` branch never actually ran (you saw duct's `Failed: command [...] exited with code <signal 11>` instead of `<test> FAILED`).

One test drives the whole pipeline against a stand-in for Godot, covering all six cases including the one that matters: a **failing** test that also crashed on exit must stay red.

# Why

I pulled 90 days of CI (2026-06-12 → 09-10): 1053 `🔗 GHA` runs, 864 where the Linux job actually ran, 83 red jobs, with logs and annotations for each.

**Of the 83 reds, 69 (83%) were flaky and only 7 (8%) were a real defect in the commit.** And of 23 reds that someone re-ran, **21 went green with zero code changes (91%)**.

There is a clean inflection point. `Avatar regression tests` landed on 2026-08-14 (#2699 / #2702):

| | first-attempt green |
|---|---|
| Jun 12 – Aug 14 | **94.7%** (588/621) |
| Aug 15 – Sep 10 | **80.2%** (195/243) |
| last week | **72.7%** |

Per-step failure rate since it landed: `Avatar regression tests` **15.5%** (39/252), `Localization tests` **7.8%** (6/77, added 2026-09-02 by #2754), `Validate GDScript` 1.15%. Everything else is at or near 0%.

**Root cause.** These tests are `SceneTree` scripts run with `--script` against the real project, so Godot boots all 15 autoloads (`Global`, `SplashOverlay`, `ConnectionQualityMonitor`, `DebugWs`, `TouchFeedback`, …) — network, threads and viewport-attached `CanvasItem`s — and then races them on the way out. The test passes and *then* the process dies:

```
[test_avatar_locomotion_grounded] PASS          <- the test passed
[Startup] global._ready start: 98ms             <- autoloads only start booting here
SceneInspectorBridge: Dedicated WebSocket channel -> ws://127.0.0.1:9231
WARNING: 1 RID of type "CanvasItem" was leaked.
WARNING: ObjectDB instances leaked at exit
ERROR: 1 resources still in use at exit (core/io/resource.cpp:810)
Error: ... exited with code <signal 11>
```

GitHub only keeps logs ~15 days, so I have full logs for the 38 reds between 08-26 and 09-10. **31 of those 38 (82%) are this exact crash, and 30 of 32 crashes happened after the success marker.** The `<signal 6>` cases are the same bug — the header reads `Program crashed with signal 11` and then Godot's own crash handler aborts. Crashes are spread across scripts (`state_machine_graph` 11, `locomotion_grounded` 7, `anim_throttle` 7, `translation_key` 5, `validate_all_scripts` 1), which is why the marker check covers all three commands.

The signal alone is **not** a safe discriminator: the 3 genuine reds on `feat/graphic-profiles-budgets-2672` also died with `<signal 6>`, because `quit(1)` goes through the same broken teardown. Hence gating on the marker.

Cost so far: 13.4h of runner time burned on flaky reds, 24 branches hit, and `docker-build-test` (`needs: linux-build`) blocked each time. Worse, **72% of reds were never re-run** — people push over them, so the gate has lost its signal.

Without this class of failure the last four weeks would have been **95.1% green instead of 80.2%**.

# Next steps

This is a mitigation, not the fix. Follow-ups worth filing:

1. **The real fix — stop booting the autoload stack for these tests** (a minimal `project.godot` without `[autoload]`, or a separate test project). Kills the crash at the source and lets this workaround be reverted. Note `test_avatar_autoplay_stomp` has never crashed once — worth checking what it does differently on teardown.
2. **Say "rebase onto main" on exit code 2.** 5 of the 83 reds were branches cut before the step existed: the workflow comes from the merge base but the checkout uses `pull_request.head.sha`, so xtask reports a cryptic `The subcommand 'test-avatar' wasn't recognized`.
3. **Alert if this workaround starts firing a lot.** The warning it prints is the signal that the underlying engine crash is getting worse rather than being fixed.

## Verification

`cargo fmt --check` and `cargo clippy --all-targets` clean on the touched file; all 8 xtask tests pass.

Not verified against the real Godot binary — that needs a full install + asset import, and the worthwhile assertion (does the marker reach the pipe before the crash?) is already answered by the 30 CI logs above, which have the identical setup: stdout to a pipe, process killed by a signal, marker present. If the marker were ever lost to buffering the build goes red, never falsely green.

### DIFF
--- a/src/check_gdscript.rs
+++ b/src/check_gdscript.rs
@@ -1,4 +1,7 @@
-use anyhow::Result;
+use std::io::{self, BufRead, BufReader, Write};
+use std::process::ExitStatus;
+
+use anyhow::{Context, Result};
 use xtaskops::ops::cmd;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


regenesis-claw (approved):
Verified the core claims against source rather than taking them at face value:

- `All scripts validated successfully!` and each `[<stem>] PASS` marker are real strings printed by the corresponding .gd scripts right before `quit(0)` (checked validate_all_scripts.gd, test_translation_key.gd, and all 4 avatar test scripts) — the marker-matching logic is correctly wired to what actually gets printed.
- Confirmed `ReaderHandle::try_wait()` docs guarantee `Ok(Some(_))` once EOF has been seen on the pipe, so the `.context("headless Godot reached EOF without exiting")` fallback for a post-EOF race is dead code, not a real risk (initially suspected a race here, docs rule it out).
- Confirmed `xtaskops::ops::cmd` returns a `duct::Expression`, so `.stderr_to_stdout().unchecked().reader()` chains as claimed, and that duct's default (non-`.unchecked()`) `.run()` already turns a non-zero exit into an `Err` before the old `if !output.status.success()` branch could ever run — matches the PR's explanation for why the pre-existing check never fired.
- `status.code().is_none()` on Unix only for signal deaths, always `Some` on Windows, so the crash-forgiveness path is correctly inert there.
- The single test (exit_crash_after_the_marker_is_the_only_failure_forgiven) covers the case that actually matters — a failing test that also dies by signal stays red — and is gated for unix only since it shells out to /bin/sh.

No P0/P1s found. Two harmless P2 observations (not blocking):
- The try_wait() EOF-race fallback is unreachable given duct's guarantee — fine to leave as defensive code, just noting it's not exercising a real path.
- Flushing stdout on every single streamed line is a bit more syscall-chatty than buffering, but irrelevant at CI log volumes.

Security pass: no user-controlled input, no secrets, no injection surface — this only shells out to the pinned Godot binary/xtask scripts. No security issues found.

Approving.

sdilauro (approved):
great work!

kuruk-mm (approved):

PR #2874, opened 2026-09-10
feat(hud): interaction prompt redesign (#2710)

### LINKED ISSUE #2710: Interaction Hint: review + increase on-screen overlay size
##  **⚠️ Problem:**
Interaction hint (glyph + hover text) renders too small on phones — players miss that an object is interactable, or can't read the action. Drives Playtime; baseline unknown, no event fires on hint display today.

Image

## **🏁 Scope:**
Audit current hint sizing on mobile, define new size/spacing/contrast values per screen density. Spec only — implementation ships in a follow-up issue. Highlight outline + tap/raycast target out of scope.

## **📝 Deliverables:**
- [ ] Audit current hint icon + label size vs device DPI, ≥3 reference phones
- [ ] Define icon size, label type scale, padding, min footprint
- [ ] Contrast check over bright + dark scene backgrounds
- [ ] Figma spec: default, long hover text, SDK-replaced icon variants
- [ ] File implementation issue, link spec

## **🔗 References:**
- [#2575](https://github.com/decentraland/godot-explorer/pull/2575) — hint icon swaps to SDK-replaced control icons; sizing must hold for arbitrary creator icons
- [#2307](https://github.com/decentraland/godot-explorer/pull/2307) — highlight effect on hover; paired feedback channel, must read as one system
- [#2261](https://github.com/decentraland/godot-explorer/issues/2261) — Controls Customization; hint glyph shares the action-icon pipeline
- No prior art in unity-explorer or docs (searched hover feedback / interaction hint / hoverText, 2026-08-15).

## Notes
- Target metric + baseline?
- Existing Figma for current hint?

EibrielInv:
## What

The interaction prompt — the pill that appears next to the crosshair when you look at something
interactable — is redesigned to the new HUD look and is bigger: 60px tall instead of 52, with the
glyph and the hover text both in the HUD's light purple on a dark translucent pill, and no white
keycap behind the glyph anymore. It also accepts taps well outside its visible edge, so it is much
easier to hit with a thumb.

Two things change beyond the styling. A creator icon set through `PBTouchScreenControls` now shows
in its own colours instead of being tinted black, matching how the same icon already looks on the
on-screen gamepad button, and it is clamped to the 44px slot so an oversized icon can no longer
stretch the prompt. And when several prompts are on screen they are spaced from the pill's own
height rather than a hardcoded radius, which fixes stacked prompts overlapping each other.

## Why

On a phone the prompt was small enough that players missed that an object was interactable, or
could not read the action text. Its visuals also predated the HUD revamp — a black pill with a
white keyboard-style keycap — so it no longer read as part of the same HUD.

The design asks for a 10px tap ring. That is 15 device pixels once `content_scale_factor` is
applied on a phone — under a millimetre — so this ships 24px instead, measured on a Pixel 7
against Android's 48dp minimum touch target. Worth raising with design, since the same
design-space-pixel conversion affects any other tap spec in that file.

This does not add the "hint shown" analytics event #2710 also asks for, so the baseline is still
unmeasured; that is left as a follow-up.

Closes #2710

## Details

<details>
<summary>Expand</summary>

Design: [HUD Revamp, node `1182:18614`](https://www.figma.com/design/nxP78pUCEHj3BY0meiXipe/HUD-Revamp?node-id=1171-18426&m=dev).

| | Before | After |
|---|---|---|
| Height | 52 | 60 |
| Width | min 130, hugs content | unchanged |
| Fill | `#00000080` | `#161518` @ 70% |
| Padding / gap | 10 / 10, gap 10+6 | 10 · 8 · 16 · 8, gap 8 |
| Icon slot | white keycap `36×24`, icon tinted black | `44×44`, no backing panel |
| Icon colour | black | `#DFD0FF` (built-in) / untinted (creator icons) |
| Text | Inter-Medium 28 `#FCFCFC` | Inter-Medium 24 `#DFD0FF` |
| Tap growth | none | 24px (vertical only when a lone prompt) |

**Changes**

- `tooltip_label.tscn` — dropped `%PanelContainer_Inputs` (the keycap, its `StyleBoxFlat` and drop
  shadow) and the two wrappers that existed only to pad it. The letter and the icon now sit
  directly in a 44×44 `IconSlot`, both anchored full-rect, so the letter centres and the texture
  scales into the box keeping aspect — that is the size clamp the prompt was missing; the joypad's
  equivalent (`button_touch_action.tscn`) already had one. The 130px minimum width is now an
  explicit `custom_minimum_size` instead of an `offset_right` that only acted as a floor by way of
  Godot clamping size up to the minimum.
- `tooltip_label.gd` — `_show_keyboard_icon()` takes a tint: built-in glyphs get the HUD colour
  (all three source assets are pure white, so they tint cleanly), creator icons get `Color.WHITE`,
  i.e. untinted. The tap ring is a `_has_point()` override using the same `Rect2.grow` trick as the
  `TapArea` atom — `TooltipLabel` is a `PanelContainer` so it cannot extend it.
- `pointer_tooltip.gd` — the ring radius is now derived from the pill (`2 × (height + PILL_GAP)`)
  instead of a hardcoded number. Neighbours sit half a radius apart, so the old constant left them
  57px apart for a 60px pill and stacked prompts overlapped at 2 and 3 prompts. Deriving it gives a
  uniform 8px gap at any count. `MIN_OFFSET_X` also floors every prompt's horizontal offset: a
  prompt is a wide left-anchored box, not a point, so at 0°/180° the ring used to put its left edge
  on the crosshair.

**Verified by me**

- [x] `gdformat --check godot/` — 386 files unchanged
- [x] `gdlint godot/` — no problems
- [x] `cargo run -- check-gdscript` — all files validated
- [x] Constants confirmed live on a Pixel 7 over the debug hub (pill min 130×60, `MIN_OFFSET_X` 40,
      `PILL_GAP` 8, `TAP_GROW_X/Y` 24)
- [x] Tap ring boundary measured by injecting touches: +3px and +8px reach the pill, +20px and
      +25px do not

**Known gap — long hover text is unbounded.** Nothing clamps the pill's width: `hover_text` is
passed through verbatim (`scene_manager.rs:3112`), the label does not wrap, and there is no max
width. A long enough creator string makes a pill wider than the screen and the tail is lost. That
is pre-existing and left as-is deliberately; the Figma only specifies "responsive" against short
copy, so the upper bound needs a design call. Case 3 below demonstrates it.

**Known gap — not visually reviewed on a device.** The measurements above are instrumented, not
observed: Play Integrity rejects locally-signed APKs (`appRecognitionVerdict=UNRECOGNIZED_VERSION`)
so no scenes load on a locally built Android build. Every case below still needs a human eye.

**Known gap — the 5-prompt cap is pre-existing.** The radial layout shows at most five prompts and
silently drops the rest. Unchanged here; Case 5 confirms the redesign did not alter it.

</details>

## Test plan

**Setup for cases 1–7:** the **Eibriel Tests** scene at **-31,19**. The **Issue 2710** block is
selected by default on arrival — if another block is showing, pick "Issue 2710" from the tab bar at
the top. The block lays out one numbered station per case: 1–5 in a row, 6 and 7 just behind them.

### Case 1 — The prompt's new look
**Steps:**
1. Open the app.
2. Go to -31,19.
3. Point the crosshair at the station labelled **1. LOOK**.

- [ ] **Expected:** the pill is dark and translucent with rounded corners; the glyph and the text
      are both light purple; there is **no white square** behind the glyph. The text is comfortably
      readable at arm's length.

### Case 2 — Tapping outside the prompt
**Steps:**
1. Open the app.
2. Go to -31,19 and point the crosshair at the station labelled **2. TAP RING**.
3. Tap roughly 2mm outside the pill's left or right edge — just off the dark background, not on it.

- [ ] **Expected:** the action triggers, the same as tapping the pill itself.
- [ ] **Expected:** tapping a good finger-width away (~1cm) does **not** trigger it.

### Case 3 — Long hover text
**Steps:**
1. Open the app.
2. Go to -31,19 and point the crosshair at the station labelled **3. LONG TEXT**.

- [ ] **Expected:** the pill grows sideways, keeps the text on **one line**, and stays 60px tall.
- [ ] **Known gap, not a failure:** the pill is wider than the screen and the end of the sentence is
      cut off. Note it and move on — see the long-text gap in Details.

### Case 4 — Three prompts at once
**Steps:**
1. Open the app.
2. Go to -31,19 and point the crosshair at the station labelled **4. MULTI x3**.

- [ ] **Expected:** three pills fan out beside the crosshair with a clear gap between each — none
      overlapping, none covering the crosshair.
- [ ] **Expected:** tapping one pill fires **that** prompt, not a neighbour (the text on the pill
      you tapped is the action that happens).

### Case 5 — Six prompts, five shown
**Steps:**
1. Open the app.
2. Go to -31,19 and point the crosshair at the station labelled **5. MULTI x6 (5 shown)**.

- [ ] **Expected:** exactly **five** pills appear, evenly spaced and not overlapping.
- [ ] **Expected:** the topmost and bottommost pills sit clear of the crosshair horizontally —
      roughly in line with where a single prompt sits, not touching it.

### Case 6 — A scene's own icons on the prompt
**Steps:**
1. Open the app.
2. Go to -31,19 and point the crosshair at the station labelled **6. CREATOR ICONS**.

- [ ] **Expected:** the two prompts show the scene's own icons (a fishing rod and a d-pad) **in
      their own colours** — not blacked out, not tinted purple.
- [ ] **Expected:** each icon matches the glyph on the corresponding on-screen gamepad button.
- [ ] **Expected:** the icons fit inside the pill without stretching it taller or wider than the
      text needs.

### Case 7 — Contrast over bright and dark backgrounds
**Steps:**
1. Open the app.
2. Go to -31,19 and point the crosshair at **7a. ON WHITE** (in front of the white wall).
3. Point the crosshair at **7b. ON BLACK** (in front of the black wall).

- [ ] **Expected:** the pill and its light-purple text are legible against both walls.

### Case 8 — Regression, ordinary scenes
**Steps:**
1. Open the app.
2. Enter Genesis Plaza and interact with a few objects that show prompts.
3. Press and hold one of the prompts.

- [ ] **Expected:** prompts appear, read correctly and trigger normally — no missing glyphs, no
      prompts stuck on screen after looking away.
- [ ] **Expected:** while held, the pill's background lightens and the glyph nudges down, as before.


rodrigo685:
Approved from my side,  but need's @ElzathRusso blessing

godot/src/ui/components/molecules/pointer_tooltip/pointer_tooltip.gd
```diff
@@ -30,14 +37,20 @@ func set_pointer_data(interacts_array: Array):
...
 		var tooltip_scene_instance = tooltip_scene.instantiate()
-		var radius = -90 if interacts_array.size() > 1 else -36
-		tooltip_scene_instance.set_position(
-			Vector2(0, radius - (4 * count)).rotated(deg_to_rad(used_angles[i]))
-		)
-		var tooltip_position = tooltip_scene_instance.get_position()
-		tooltip_scene_instance.set_position(
-			Vector2(tooltip_position.x, tooltip_position.y - tooltip_scene_instance.size.y / 2)
-		)
+		var pill_height: float = tooltip_scene_instance.size.y
```
regenesis-claw: [P2] `size.y` on an out-of-tree instance resolves to 60 only because the scene's `offset_top/bottom` (±30) happen to agree with `custom_minimum_size.y` (60) — two independent values in the `.tscn` that nothing keeps in sync. If someone bumps the pill to 70 via `custom_minimum_size` alone, the ring silently keeps spacing for a 60px pill and the overlap bug this PR fixes comes back, with no test or check to catch it.

`tooltip_scene_instance.get_combined_minimum_size().y` would read the value the layout will actually honour, and survives one of the two being changed.

godot/src/ui/components/molecules/pointer_tooltip/tooltip_label.gd
```diff
@@ -9,12 +9,26 @@ extends PanelContainer
...
+const TAP_GROW_X: float = 24.0
+const TAP_GROW_Y: float = 24.0
 const ICON_LEFT_CLICK = preload("uid://cljfaeb8np0ma")
 const ICON_INTERACTIVE_POINTER = preload("uid://72xpjysoxgwo")
 const ICON_JUMP = preload("uid://ck3atqpytstpo")
 
+## Vertical half of the tap growth. Stacked tooltips sit ~57px apart while the pill is 60 tall, so
+## growing them vertically would make neighbouring hit areas overlap and let a tap fire the wrong
+## prompt. pointer_tooltip.gd zeroes this whenever more than one tooltip is on screen.
+var tap_grow_y: float = TAP_GROW_Y
```
regenesis-claw, line 33: [P2] This comment describes the pre-fix spacing, not the spacing this PR ships. After `pointer_tooltip.gd`'s new radius, stacked prompts sit **68px** apart (60 pill + `PILL_GAP` 8), not ~57 — the ~57 figure is exactly the bug the same PR removes. The conclusion is still right (68px apart with 24px of growth per side would overlap by 16px), it's the premise that's stale.

```suggestion
## Vertical half of the tap growth. Stacked tooltips clear each other by only PILL_GAP (8px),
## so growing them vertically would make neighbouring hit areas overlap and let a tap fire the
## wrong prompt. pointer_tooltip.gd zeroes this whenever more than one tooltip is on screen.
```

godot/src/ui/components/molecules/pointer_tooltip/tooltip_label.gd
```diff
@@ -9,12 +9,26 @@ extends PanelContainer
 # i18n-keys: TOOLTIP_VIEW_PROFILE
 const VIEW_PROFILE_KEY: String = "TOOLTIP_VIEW_PROFILE"
 
-const BG_COLOR_NORMAL: String = "#00000080"
-const BG_COLOR_PRESSED: String = "#44444480"
+const BG_COLOR_NORMAL: String = "#161518B3"
+const BG_COLOR_PRESSED: String = "#444348B3"
+## Shared HUD glyph tint (Figma "IconHUD"). Applied to the built-in monochrome-white glyphs and to
+## keyboard letters; scene-replaced creator icons keep their own colors (see _show_keyboard_icon).
+const ICON_COLOR := Color("#DFD0FF")
```
regenesis-claw, line 18: [P2] "and to keyboard letters" isn't accurate — `_show_keyboard()` never touches a tint; the letter's colour comes from `LabelSettings_r3qmc` in the scene file, which hardcodes the same `#DFD0FF` independently. So the HUD glyph colour now lives in two places that will drift the first time design changes it. Either drop the letters from this doc comment, or have `_show_keyboard()` set `label_action.add_theme_color_override("font_color", ICON_COLOR)` so the constant is the single source.

godot/src/ui/components/molecules/pointer_tooltip/pointer_tooltip.gd
```diff
@@ -30,14 +37,20 @@ func set_pointer_data(interacts_array: Array):
...
+		var offset := Vector2(MIN_OFFSET_X, 0.0)
```
regenesis-claw, line 44: [P2 / question] With `MIN_OFFSET_X = 40` and `TAP_GROW_X = 24`, a lone prompt's hit area starts 16px to the right of the crosshair and extends 24px above and below it — so the band immediately around the aim point, where a player is most likely to drag to turn the camera, now fires the interaction instead. Growth on the crosshair-facing side buys the least (the pill is already the thing you're aiming at), so an asymmetric grow — e.g. full 24 on the right/top/bottom, ~8 on the left — would keep the 48dp target without eating into the camera-drag area. Worth a device check before this leaves draft, since the same instrumented measurement can't tell you whether a drag gets swallowed.

regenesis-claw (commented):
Review against `REVIEW.md`. **No blockers found** — the logic changes hold up, and the ring math is a real fix. Leaving this as a comment rather than a verdict since the PR is still in **draft** and one pre-flight item fails.

**Pre-flight**
- ❌ (1) Branch is **20 commits behind `main`** (4 ahead, diverged). CI here is green but was produced against a stale base — worth merging `main` before this leaves draft.
- ✅ (2) iOS build not required: GDScript/`.tscn` only, no `OS.get_name() == "iOS"` branch, no native plugin, no safe-area/keyboard path. An iOS export ran green anyway.
- ✅ (3) No submodule drift. ✅ (4) Description states what and why, in the author's own voice.
- ✅ No debug prints, no `.claude/` files, no non-English comments, no new user-facing literals (i18n clean — `Label_Text` keeps `auto_translate_mode = 2` and nothing new is assigned to it).

**Verified independently**
- No remaining references to the deleted `%PanelContainer_Inputs` / `HBoxContainer_Icons` anywhere in `godot/` or `lib/` — the `@onready` removal is complete.
- Ring spacing checked by hand for every count. `_get_centered_angles` yields `[90]`, `[60,90]`, `[60,90,120]`, `[0,60,90,120]`, `[0,60,90,120,180]`; with `radius = 2·(60+8) = 136`, neighbours are exactly 68px apart centre-to-centre → a uniform 8px edge gap at 60px pill height. The overlap at 2 and 3 prompts is genuinely fixed, and all rotated offsets have `x ≥ 0` so the `maxf` floor is never fighting a negative.
- `tap_grow_y = 0.0` is assigned after `instantiate()` but before `add_child()` — safe, since GDScript member initializers run at instantiation, not in `_ready()`.
- Tint reset path is correct: when a creator icon is removed, `_physics_process` re-runs `set_tooltip_data`, which goes back through the default `ICON_COLOR` argument, so a `Color.WHITE` never sticks on a built-in glyph.
- Desktop is unaffected by the enlarged hit area — the mouse is captured in-world, so GUI picking doesn't run.

Three P2s inline, plus one open question about the tap ring's crosshair-facing side. The three gaps you flagged yourself (unbounded pill width, missing #2710 analytics event, 5-prompt cap) all check out as pre-existing and correctly scoped out.

regenesis-claw (approved):
Re-review of the delta (`489750d`…`ea646b4`, isolated from the two `main` merges). **All three actionable P2s are fixed. Approving.**

| Finding | Status |
|---|---|
| `size.y` read off an out-of-tree instance | ✅ now `get_combined_minimum_size().y`, with the rationale in a comment |
| Stale "~57px apart" premise on `tap_grow_y` | ✅ reworded to the shipped 8px `PILL_GAP` |
| `ICON_COLOR` doc claiming it tints keyboard letters | ✅ corrected, and it now names `LabelSettings_r3qmc` as the letter's source with a "change both together" note |

The `get_combined_minimum_size()` swap is the better fix of the two I offered, and it is a genuine improvement rather than a like-for-like: it returns `max(custom_minimum_size.y, content min)`, so if the pill ever grows past its declared minimum the ring spacing follows instead of silently keeping 60. Value is unchanged today — the `IconSlot` 44 + 8/8 margins land on exactly 60, and `StyleBoxFlat_473nw` contributes no content margin — so the verified ring geometry (68px centre-to-centre, uniform 8px gap at every count) still holds. Out-of-tree theme lookups are not a problem here: `Label_Text` gets its font from `label_settings`, not the theme, and the panel stylebox is a node-level override.

**Pre-flight, re-checked**
- ✅ Branch is level with `main` (`behind_by: 0`; was 20 behind at the last review). No longer a draft.
- ✅ `main` brought in 77 files but nothing under `pointer_tooltip/` or any other HUD input path, so the merge doesn't interact with this change.
- ✅ CI green on `c7fecb4`: static checks, clippy, Linux, Android, Docker. The iOS/R2 distribution jobs did not re-run on the merge commit — not required here (GDScript + `.tscn` only, no platform branch), and the Android build covers the touch path.

**Left open, deliberately**
- The tap-ring question from the last pass is unaddressed: with `MIN_OFFSET_X = 40` and `TAP_GROW_X = 24`, a lone prompt's hit area still reaches to 16px from the crosshair and 24px above/below it, so a camera drag started just left of the pill fires the interaction. Non-blocking and not a regression in this diff, but it is the one thing the instrumented measurements can't answer — worth a look whenever the scene gets a human pass on device.
- Your three self-declared gaps (unbounded pill width, no #2710 analytics event, pre-existing 5-prompt cap) remain correctly scoped out.

Still needs @ElzathRusso's design sign-off per @rodrigo685's comment; my approval doesn't clear that or the codeowner requirement.

ElzathRusso (approved):

PR #2878, opened 2026-09-11
feat: pinch-to-zoom camera, centered avatar + crosshair-anchored aim (mobile)

### LINKED ISSUE #2709: Camera Zoom Gesture 3p/1p (Pinch) + Avatar Position + Joystick area
##  **Description**
Implement the two-finger pinch gesture to switch between first-person and third-person camera, matching Roblox's mobile behavior. This requires defining the joystick's screen area first, so the pinch gesture zone can be set to cover the rest of the screen without conflicting with joystick input. Additionally, adjust the avatar's position (center it on screen) and reposition the crosshair so it no longer overlaps the avatar model.

## **📝 Deliverables:**

- [ ]  Review and Adjust Joystick Input Area 
- [ ]  Two-finger pinch gesture controlling camera zoom distance in real time (third-person ↔ first-person)
- [ ]  Avatar position: Avatar centered on screen
- [ ]  Crosshair new position.

## **🔗 References:**
- [decentraland/godot-explorer#2642](https://github.com/decentraland/godot-explorer/issues/2642) — HUD + Social Panel Redesign Iteration #1, the change that introduced the left skew
- [decentraland/godot-explorer#2635](https://github.com/decentraland/godot-explorer/issues/2635) — HUD Iteration #2 visuals; final menu bounds must be settled against these
- [decentraland/godot-explorer#2343](https://github.com/decentraland/godot-explorer/issues/2343) — HUD positioning parity Android/iOS; offset must hold on both
- [decentraland/unity-explorer#611](https://github.com/decentraland/unity-explorer/pull/611) — desktop precedent: shoulder offset + forced center on unlocked cursor, plus configurable aim offsets
- [decentraland/unity-explorer#3666](https://github.com/decentraland/unity-explorer/issues/3666) — horizontal camera offset caused asymmetric wall clipping on desktop; regression guard for this change

manuelmaceira:
## What
On mobile, a two-finger pinch now swaps the in-world camera between the two fixed first/third-person positions (same distances as prod), the avatar is centered on screen, and the crosshair sits above-right of the avatar's head in third person — with interaction raycasts aimed where the crosshair is drawn. The joystick's touch zone shrinks from the full-height left band to a fixed bottom-left rect (40% × 67%), freeing the rest of the screen for camera look and pinch.

## Why
Issue #2709: discoverability and control parity with the reference clients. Pinch replaced the old HUD POV toggle; the over-shoulder offset left the avatar off-center and the centered crosshair sat on the avatar's body. This builds on #2691 (pinch gesture, joystick zone, zoom↔settings sync, floor guard, analytics) and adds what it was missing: centered avatar, per-mode crosshair anchors interpolated across the pinch transition, and crosshair-aligned raycasts.

Closes #2709

## Details

- Includes the full #2691 branch (merged, one conflict in `player.gd` resolved keeping both sides), with one change on top: per the team's device-QA round, the continuous zoom curve was dropped — the pinch toggles between the two prod camera positions once the finger-spread accumulates 80px in one direction (`PinchGestureHelpers.MODE_TOGGLE_SPREAD`).
- `THIRD_PERSON_CAMERA.x` is now 0 (was 0.75 over-shoulder). The lateral-offset clamp machinery stays; its collision tests exercise it with a synthetic 0.75 offset.
- `CameraRigHelpers.crosshair_anchor(spring_length)`: 1p → screen center, 3p → upper third, interpolated across the 1p↔3p crossing so the crosshair never jumps at the mode switch. Applied per frame by `SdkTouchControlsApplier` (mobile only; cinematic tap-cursor and desktop centered crosshair untouched).
- Raycasts follow the crosshair: `scene_manager` gains `crosshair_screen_point` (set per frame from GDScript on mobile, cleared on hide/resize). The interaction ray keeps its camera origin (RaycastHit stays camera-relative per raycast_hit.proto) and aims through the crosshair point; the centered avatar is dodged by excluding the local player's collider RIDs. `AvatarRaycast` (outline/view-profile) aims the same way. Desktop keeps the viewport-center behavior.
- Pinch recognizer math extracted to `PinchGestureHelpers` (pure) with a headless unit test. The full gesture FSM test (synthetic touch sequences) is deliberately out of scope.
- Verified: `test_player_camera_collision` and `test_pinch_gesture` PASS headless (macOS); gdlint/gdformat/cargo clippy clean. Device-tested on an iPhone 16e (pinch feel, crosshair placement — anchor pinned to the issue's Figma position).
- Note: this PR's branch carries the merge of `feat/pinch-zoom-camera`; merging this effectively merges #2691 too.
</details>

## Test plan
Device only (iOS/Android), landscape, logged into Genesis City:

- [ ] Pinch in/out in the camera area → smooth continuous zoom; fully in drops to first person (avatar hides), back out returns to third person with no flicker at the boundary
- [ ] Avatar stays horizontally centered at every zoom; crosshair sits above the avatar's head in third person, dead center in first person, gliding between both with no jump
- [ ] Tap a scene clickable under the raised crosshair → it responds (aim follows the crosshair)
- [ ] Left thumb walking while two fingers pinch → walk continues and camera zooms; a second finger inside the joystick rect rotates instead of zooming
- [ ] Pinch over a HUD panel, scene UI, or open chat → camera doesn't move; joystick only engages from the bottom-left rect
- [ ] Camera doesn't clip walls at any zoom and stays above the floor at max zoom-out (e.g. orange Genesis Plaza floor); zoom resets to default third person after a Discover jump


.pi-prompt.md
```diff
@@ -0,0 +1,11 @@
+Sos un investigador. NO implementes nada todavía — solo investigá y armá un plan de ataque en PLAN.md.
```
regenesis-claw: [P1] This file is still tracked at the head commit — the `chore: untrack .pi-prompt.md` commit didn't remove it from the tree. It's per-contributor agent scratch state and it's in Spanish, which hits two Tier-1 rules in `REVIEW.md` (no personal AI memory/plan files in the repo; codebase is English-only).

```
git rm --cached .pi-prompt.md
echo '.pi-prompt.md' >> .gitignore
```

lib/src/scene_runner/scene_manager.rs
```diff
@@ -2362,15 +2367,33 @@ impl SceneManager {
...
         };
 
-        // Use cached viewport center for raycasting
-        let raycast_from = camera_node.project_ray_origin(screen_point);
-        let raycast_to = raycast_from + camera_node.project_ray_normal(screen_point) * RAY_LENGTH;
+        // Issue #2709: the avatar is centered on screen in third person, so a ray
+        // from the camera would strike it. Start at the avatar's own depth, in the
+        // camera -> crosshair direction (same line, origin pushed forward).
+        let ray_normal = camera_node.project_ray_normal(screen_point);
+        let raycast_from = if start_at_avatar_depth {
```
regenesis-claw: [P1] This `raycast_from` is also what gets handed to `RaycastHit::from_godot_raycast` further down — and the comment at that call site says the hit must be built from the **camera** origin so `hit.length` / `global_origin` / `direction` stay camera-relative per `raycast_hit.proto`, calling the avatar position "an incidental earlier choice". This reinstates that, on mobile only, and leaves the comment asserting the opposite.

Practical effect: `pointer_events.rs` gates `max_distance` on `raycast.hit.length`, so in third person the measured distance shrinks by the camera distance (1.5–6 m). Entities that are out of range on desktop become in range on mobile, and scenes reading `global_origin`/`direction` see the avatar instead of the camera — a wire-visible change for shipped scenes.

Suggested fix: keep `project_ray_origin(screen_point)` here and skip the avatar with `cached_raycast_query.set_exclude(...)` on the local avatar's collider RIDs instead (`player.gd` already caches that list as `_raycast_exclude`). Worth checking first whether the local avatar is even on `CL_POINTER | CL_AVATAR` — if not, the ray never hit it and this whole branch can go.

lib/src/scene_runner/scene_manager.rs
```diff
@@ -2362,15 +2367,33 @@ impl SceneManager {
...
-        let screen_point = if self.raycast_use_cursor_position {
-            self.cursor_position
+        // Outside tap-cursor mode, aim where the crosshair is drawn (mobile sets
+        // it per-frame); desktop keeps the viewport center.
+        let (screen_point, start_at_avatar_depth) = if self.raycast_use_cursor_position {
+            (self.cursor_position, false)
         } else {
-            self.viewport_center
+            match self.crosshair_screen_point {
+                Some(point) => (point, true),
```
regenesis-claw: [P1] `crosshair_screen_point` is `Some(...)` on every mobile frame regardless of camera mode, so `start_at_avatar_depth` is `true` in **first person** too — where there is no avatar in front of the camera to avoid. The avatar root is at the feet (~1.7 m below the FP camera), so pitching down makes the depth non-zero (≈1.5 m at −60°) and the ray starts that far ahead, making anything nearer than that un-hittable. Gate this on being in third person, or drop the shift in favour of excluding the avatar collider.

lib/src/scene_runner/scene_manager.rs
```diff
@@ -2503,6 +2526,11 @@ impl SceneManager {
         }
     }
 
+    #[func]
+    fn set_crosshair_screen_point(&mut self, screen_point: Vector2) {
```
regenesis-claw, line 2526: [P2] There's no way to clear this. `_apply_crosshair_anchor()` early-returns when the crosshair is hidden (Hide UI, `PBTouchScreenControls`), when the player isn't valid, and in tap-cursor mode — in those windows the last value keeps driving the interaction ray, including across an orientation change where the cached pixel coordinate is stale. Consider a companion `clear_crosshair_screen_point()` setting it back to `None`, called from those early returns.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`. The camera work itself is solid — the zoom scalar with its hysteresis band, the self-heal in `begin_pinch_zoom`, the `loading_finished` (not `on_change_scene_id`) reset choice, the pure `PinchGestureHelpers` extraction with a headless test, and the constant-ordering invariants test are all the right calls, and the `#2691` base was already reviewed and approved on its own PR. Three blockers below, all in the new aim/raycast path plus one committed file.

### Pre-flight
- **Branch up to date with `main`** — yes, merged at the head commit. ✅
- **CI** — `📊 Static checks` green; `🐧 Linux`, `🔧 Clippy`, `🤖 Android` and the iOS resolve job are all still **pending** on this head, so there is no build signal yet for the Rust change. An iOS build isn't in REVIEW.md's mandatory list for this diff (no native plugin / `OS.get_name() == "iOS"` branch / comms / av), but this is a touch-gesture + on-screen-aim change that only manifests on a device — the mobile distribution run you triggered should be green before merge.
- **Description** — has What/Why/Details/Test plan. Two notes in P2 below.

---

## 🔴 P1 — blockers

**1. `.pi-prompt.md` is committed, and it's in Spanish.**
It's present at the head commit (836 bytes) — the `chore: untrack .pi-prompt.md` commit didn't actually remove it from the tree. It's a personal agent prompt (`Sos un investigador. NO implementes nada todavía…`), which hits two Tier-1 rules at once: per-contributor AI scratch state doesn't belong in the repo (same class as `.claude/` memory files and the `PLAN-*.md` from #2778), and the codebase is English-only. Please `git rm --cached .pi-prompt.md`, commit the deletion, and add it to `.gitignore`.

**2. The avatar-depth ray origin changes SDK-visible raycast fields on mobile, and contradicts the comment 40 lines below it.**
`get_current_mouse_entity` now builds `raycast_from` at the avatar's depth, and that same `raycast_from` is passed into `RaycastHit::from_godot_raycast`. The comment right above that call says, verbatim: *"Build the hit from the CAMERA ray origin (not the avatar position) so `hit.length`/`global_origin`/`direction` are camera-relative, per raycast_hit.proto. This is what makes `max_distance` a true camera-distance gate… Avatar position was an incidental earlier choice."* This PR silently re-introduces exactly that, on mobile only, and leaves the comment in place now stating the opposite.

Concretely: `pointer_events.rs` gates on `raycast.hit.length` for `max_distance`, so in third person `hit.length` drops by roughly the camera distance (~1.5–6 m depending on zoom). An entity 12 m from the camera / 9 m from the avatar becomes interactable on mobile and not on desktop, and `hit.global_origin` / `hit.direction` handed to scenes move from the camera to the avatar. That's a wire-visible behavior change for existing production scenes, not a client-local tweak.

Suggested fix that solves this and #3 together: keep `raycast_from = camera_node.project_ray_origin(screen_point)` (so the hit stays camera-relative) and skip the avatar by *excluding* it rather than by moving the origin — `self.cached_raycast_query.set_exclude(...)` with the local avatar's collider RIDs (`player.gd` already caches exactly that list in `_raycast_exclude` for the ground raycast). Worth confirming first whether the local avatar is even on `CL_POINTER | CL_AVATAR` — if it isn't, the ray was never blocked by it and the whole origin shift can just be dropped.

**3. The avatar-depth origin also applies in first person, where there is no avatar to avoid.**
`crosshair_screen_point` is `Some(...)` on every mobile frame regardless of camera mode, so `start_at_avatar_depth` is `true` in first person too. The avatar root sits at the feet, ~1.7 m below the first-person camera, so as soon as the player pitches down the depth becomes non-zero (≈1.5 m at −60°) and the ray starts that far in front of the camera — anything closer than that along the view ray stops being hittable. Clicking a floor-level or close-up interactable while looking down in first person regresses. Gate the origin shift on actually being in third person (or, preferably, use the exclude-based fix in #2 and drop the shift entirely).

---

## 🟡 P2 — non-blocking

- **`set_crosshair_screen_point` has no clear path.** Once mobile sets it, it stays `Some` forever. `_apply_crosshair_anchor()` early-returns when the crosshair is hidden (Hide UI, `PBTouchScreenControls`), when the player isn't valid, or in tap-cursor mode — in those windows the stale point keeps driving the interaction ray, including across an orientation change where the cached pixel coordinate is no longer valid. Add a `clear_crosshair_screen_point()` (`= None`) and call it from those early returns.
- **Two sources of truth for the same aim line.** The Rust path takes the depth from `get_player_avatar_node()`; `player.gd::_update_avatar_raycast_to_crosshair()` takes it from the player root `global_position`. Same intent, different reference points — the scene ray and the avatar-outline ray can disagree. Worth reading one value in one place.
- **`_update_avatar_raycast_to_crosshair()` runs on every physics tick on desktop** just to re-write the two constant values. Cheap, but it's the physics hot path — early-out when it's already at the default.
- **`virtual_joystick.tscn` leftovers.** `size_flags_horizontal = 3` and `size_flags_stretch_ratio = 2.0` are no-ops on an anchor-positioned `Control` that isn't inside a container, and `grow_vertical = 0` no longer means anything now that `anchor_top != anchor_bottom`. Also, `Dynamic` and `SprintTimer` were reordered ahead of `ActiveArea`, which changes draw order — no visible effect today (the `ActiveArea/ColorRect` debug rect is `visible = false`), but it's unexplained churn in the diff.
- **Stale claim in the description.** Details says *"Not yet tested on a physical device"*, but three commits on this branch (`83e856f`, `4ca659a`, `6ccad86`) are explicitly driven by device QA feedback on the crosshair anchor. Please update it so QA isn't misled about what's been exercised.
- **Test plan shape.** The cases are good content but not in the REVIEW.md §4 format — they don't start from "Open the app", and several bundle three assertions into one checkbox (e.g. the last one covers wall clipping, floor clearance *and* the Discover-jump reset). Splitting them into Steps/Expected blocks would make them tickable by hand.
- **Nit:** the new GDScript uses `Global.is_mobile()`; REVIEW.md asks for `OS.get_name() == "Android"/"iOS"` in new code. There's plenty of in-repo precedent for `Global.is_mobile()` (including `virtual_joystick.gd`), so this is consistency-only, not a change request.

**Security:** no security issues found — no new network, auth, IO or user-controlled parsing surface; `.pi-prompt.md` contains no credentials.

Happy to re-review as soon as the file is dropped and the raycast origin is settled.

manuelmaceira:
All three P1s and the P2 batch addressed in bf46f69:

- **Aim path (P1 x2):** ray origin stays at the camera everywhere (1p included) — RaycastHit fields keep their camera-relative meaning per raycast_hit.proto, so hit.length / global_origin are unchanged from desktop. The centered avatar is dodged via `set_exclude` with the local player's collider RIDs (player.gd pushes its cached `_raycast_exclude` to the scene runner on ready), per your suggestion.
- **.pi-prompt.md (P1):** untracked + gitignored (a `git add -A` in a later commit had re-added it after the first untrack).
- **P2s:** `clear_crosshair_screen_point` on crosshair-hidden and on viewport resize (no stale aim point); GDScript and Rust aim paths now share the same camera-origin model; AvatarRaycast restore-to-default gated by a state bool (no per-tick writes on desktop); virtual_joystick.tscn size_flags leftovers dropped and original node order restored.
- Description updated: device-tested on iPhone 16e.

Validation: cargo check/clippy clean, gdlint/gdformat clean, headless tests PASS (`test_player_camera_collision`, `test_pinch_gesture`).

regenesis-claw (approved):
Delta re-review at head `bf46f699` (single appended commit, `e39ca09...bf46f699` is a clean fast-forward — no rebase, `gh api compare` delta used).

| # | Finding | Status |
|---|---|---|
| P1 | `.pi-prompt.md` committed (Spanish agent prompt) | ✅ Fixed — deleted and added to `.gitignore` |
| P1 | Avatar-depth ray origin breaks `RaycastHit` camera-relative contract | ✅ Fixed — `raycast_from` is unconditionally `camera_node.project_ray_origin(screen_point)` again; the centered avatar is dodged by excluding its collider RIDs (`set_exclude`) instead of moving the origin |
| P1 | Same origin shift also fired in first person | ✅ Fixed as a consequence of the above — there's no more origin shift at all, in any mode |
| P2 | `crosshair_screen_point` never cleared (stale on hide / resize) | ✅ Fixed — `clear_crosshair_screen_point()` called from `_apply_crosshair_anchor`'s hidden-crosshair branch and from `_on_ui_resize` |
| P2 | Two divergent depth references (GDScript avatar root vs. Rust avatar node) for the aim line | ✅ Resolved — `_update_avatar_raycast_to_crosshair` now also keeps the camera origin and only retargets direction, so both raycasts share the same model |
| P2 | Per-tick constant writes on desktop | ✅ Fixed — gated behind `_avatar_raycast_crosshair_active`, only writes on state change |
| P2 | `virtual_joystick.tscn` leftover no-op size flags + node reorder | ✅ Fixed — flags dropped, original node order restored |
| P2 | Stale "not yet tested on a physical device" line in the description | ❌ Not addressed |
| P2 | Test plan not in REVIEW.md §4 Setup/Steps/Expected format | ❌ Not addressed |

Verified the new exclusion path in source, not just by claim: `player.gd::_build_raycast_exclude()` walks the avatar subtree for `CollisionObject3D` RIDs including `get_rid()` for the player body itself, and pushes that list once via `Global.scene_runner.set_pointer_raycast_exclude(_raycast_exclude)` in `_ready()`. The RID that actually matters for the `CL_POINTER | CL_AVATAR`-mask pointer ray is `AvatarScene`'s `click_area` (`@onready var click_area = $ClickArea`, tagged `is_avatar` on ready) — that's a static node instanced with the avatar scene, not something wearable-equip recreates, so the exclude list being built once at `_ready` doesn't go stale on an outfit change. No new P1s introduced this round (unlike the #2778 round-3/4 self-referential regression) — the fix is a clean removal of the origin-shift mechanism rather than a patch on top of it.

CI shows no checks yet on this head (just pushed) — the mobile-distribution / Linux / Clippy / Android runs from the previous head don't carry over. Please confirm green (including the iOS artifact, since this still touches the interaction-raycast path) before merging even though the review itself is approved.

Approving. The two remaining P2s (stale description line, test plan format) are non-blocking — happy to see them addressed in a follow-up or before merge, whichever is easier.

kuruk-mm (approved):
looking great!

Issue #2840, opened 2026-09-07
D1 Comeback Notification - Re-join the most played Day-0 scene


rodrigo685:
##  **⚠️ Problem:**
Day-1 notification today is one static line sending everyone to Genesis Plaza (#1651), and for weeks it sent them to a dead spawn point (#1913). A player who spent their first session inside one scene gets no reason to return to it. Drives D1 retention; current Day-1 CTR unknown — unmeasurable until #2837 ships.

## **🏁 Scope:**
Server-scheduled push 22h after the Day-0 session ends, evaluated in the user's local timezone, only when Day-0 playtime > 2 min, deeplinking to that user's most-played Day-0 scene. Retires the static local Day-1 notification. One campaign, not a lifecycle engine.

## **📝 Deliverables:**
- [ ] Compute per-user most-played Day-0 scene from session telemetry
- [ ] Eligibility — Day-0 playtime > 2 min, permission granted, category not muted (#2839)
- [ ] Schedule at Day-0 session end + 22h, evaluated in user timezone
- [ ] Suppress if the user opened the app since Day-0 session end
- [ ] Deeplink to that scene's position/realm; validate target resolves before send
- [ ] Fallback copy + destination when no qualifying scene
- [ ] Retire local Day-1 notification (#1651) — no double send
- [ ] Held-out control cohort on every send
- [ ] `campaign_id=d1_comeback` on send log (#2838) and on `client_open` (#2837)

## Metrics
1. Retention — D1.
2. D1 retention, notified vs held-out cohort. Baseline and target unknown — read baseline from Segment before first send, set target after 2 weeks.
3. Send log (#2838) joined to `client_open source=push campaign_id=d1_comeback` (#2837), Segment → Snowflake `"DCL"."PROD"`.

## Dependency
Blocked by #2838 (send service + timezone scheduler) and #2837 (attribution). Cannot be measured before #2837.

## **🔗 References:**
- [#1651](https://github.com/decentraland/godot-explorer/issues/1651) — the shipped static Day-1 notification this replaces; its scheduling flag and copy are the baseline
- [#1913](https://github.com/decentraland/godot-explorer/issues/1913) — that notification deeplinked to a dead spawn; why destination validation is a deliverable
- [#1919](https://github.com/decentraland/godot-explorer/issues/1919) — deeplink dropped on an already-open app; affects the suppression and routing rules
[...]

Issue #2837, opened 2026-09-07
[Data] Client Open Source Tracking - Segment event + deeplink parameters


rodrigo685:
##  **⚠️ Problem:**
No way to tell an organic open from a push open, an ad open or a shared link. The Day-1 notification (#1651) has shipped for months with no measurable CTR, and no campaign on #2585 can be judged without this. Blocks the Retention metric on every push issue.

## **🏁 Scope:**
`client_open` Segment event on every app open carrying a resolved `source`, parsed from the launch deeplink, landing in a Snowflake table joinable to the send log (#2838). Dashboard included. Not an attribution SDK — parameters we set, on links we control.

## **📝 Deliverables:**
- [ ] `client_open` Segment event on every app open — cold and warm
- [ ] `source` resolved to `organic` | `push` | `deeplink` | `ad` | `share` | `unknown`
- [ ] Parse deeplink params — `campaign_id`, `notification_id`, `source`, `position`, `realm`
- [ ] Warm-start opens attributed, not only cold start
- [ ] `session_id`, platform, app_version, guest vs wallet on the event
- [ ] Snowflake table under `"DCL"."PROD"` + Segment → Snowflake mapping
- [ ] Attribution window defined — session credited to a push within N hours
- [ ] Metabase dashboard — opens by source, push CTR, reactivations per campaign
- [ ] Opens predating this ship as `unknown`, never `organic`

## Metrics
This is the measurement job. Target: 100% of app opens carry a `source`; every push open joins to a `campaign_id` in the #2838 send log. Verified by a reconciliation query — sends vs attributed opens — on the first campaign.

## **🔗 References:**
- [#2477](https://github.com/decentraland/godot-explorer/issues/2477) — states the Segment pipeline already exists and was hardened in #2440; emit through that path, do not add a second one
- [#1919](https://github.com/decentraland/godot-explorer/issues/1919) — deeplink ignored on a foregrounded app; the warm-start case this event must cover
- [#1913](https://github.com/decentraland/godot-explorer/issues/1913) — Day-1 deeplink pointed at a dead spawn for weeks with no signal; the gap this closes
- [#2664](https://github.com/decentraland/godot-explorer/issues/2664) — PRD analytics section; event schema and attribution window originate there

Issue #2585, opened 2026-07-27
Push Notifications — Implementation


rodrigo685:
##  **⚠️ Problem:**
Only re-engagement lever today is one static local Day-1 notification (#1651) — same copy, every user, one shot, no server send, no targeting, no measurable CTR. No way to pull a lapsed player back for an event, a friend online, or the scene they actually played. Drives Retention (D1/D7/D30) and Playtime; baseline retention and current Day-1 CTR unknown — read from Segment before the first campaign.

## **🏁 Scope:** [PRD](https://github.com/decentraland/godot-explorer/issues/2664)
Remote push on APNs + FCM direct — no managed vendor. Client token and permission, deeplink routing, send service, open-source attribution, category settings, liveops backoffice, and the first automated campaign. Ships as the 7 sub-issues below.

## **📝 Deliverables:**
- [ ] #2835 — Native push: iOS
- [ ] #2836 — Native push: Android
- [ ] #2837 — `client_open` source tracking
- [ ] #2838 — Push notification service
- [ ] #2839 — Notification settings panel
- [ ] #2840 — D1 comeback notification
- [ ] #2841 — Mobile liveops backoffice
- [ ] APNs auth key + FCM service account provisioned per environment
- [ ] Wallet ↔ device token mapping verified on account switch
- [ ] Baseline D1/D7/D30 retention pulled from Segment and recorded here

## Metrics
1. Retention — D1 and D7.
2. D1 retention lift on notified vs held-out cohort. Baseline and target unknown — set before the first send.
3. `client_open` with `source=push` + `campaign_id` (#2837) joined to the #2838 send log. Segment → Snowflake `"DCL"."PROD"`. Held-out control on every automated campaign.

## Dependency
PRD #2664 signed off — provider (APNs + FCM direct), permission-prompt timing, category taxonomy and event schema come from there. Sequencing inside this epic: #2835 / #2836 / #2837 can start now; #2840 and #2841 are blocked by #2838.

## **🔗 References:**
- [#2664](https://github.com/decentraland/godot-explorer/issues/2664) — PRD definition; the decisions this epic implements
- [#1651](https://github.com/decentraland/godot-explorer/issues/1651) — shipped local Day-1 notification; the baseline this replaces and the code path to reuse
- [#1919](https://github.com/decentraland/godot-explorer/issues/1919) — iOS deeplink ignored when the app was already foregrounded; the warm-start case every push handler must cover
[...]

PR #2881, opened 2026-09-11
feat(mobile): receive server-sent push notifications on Android (#2585)

leanmendoza:
Client half of push notifications for the mobile explorer. Android only — iOS is documented as feasible but deliberately out of this pass.

Part of a three-repo stack for #2585, with decentraland/mobile-bff#90 and decentraland/mobile-hub#15. **This one has no merge-order dependency**: the token is registered through Segment and messages arrive through FCM, so nothing here calls mobile-bff or is called by it. It can land before or after the other two.

- [ ] `decentraland/mobile-bff` — the push service (schema, sender, dispatcher, backoffice endpoints)
- [ ] `decentraland/mobile-hub` — the Push tab

> **Up to date with `main`.** Merged in 27f6109c. Four source conflicts were additive on both sides and keep both; the two AARs are build outputs and were rebuilt from the merged Kotlin, so they carry this branch's FCM classes alongside main's `InAppReview`.

### `App Opened`

A second event rides along, for [#2837](https://github.com/decentraland/godot-explorer/issues/2837): every foreground entry, cold and warm, with what brought the user in (`push` / `deeplink` / `icon`), the previous session's id and how long they had been away.

No session threshold is applied on the client. Shipping "was this a new session?" as a boolean would freeze the rule into every installed build and take a release plus its upgrade tail to change, so `seconds_since_last_seen` goes raw and the warehouse picks the cutoff. The 30s floor on warm opens is a different thing: without it every sign-in counts twice, because returning from the browser auth hop fires the same notification as the user coming back.

`session_id` deliberately does not rotate — it is the pivot between Segment, Sentry and Firebase.

**Ordering bug found on device.** Cold opens read the deep link straight off `Global` rather than waiting for `DeepLinkRouter` to tell them: the autoloads that route the link are ready first, so on a push tap the link was already routed ~6s before `SessionTracker` started listening, and a push open reported `trigger=icon`. Warm opens keep using the notification on purpose, since `deep_link_obj` lives for the whole process and could otherwise attribute an hour-old link.

FCM delivers **data-only** messages, which is what makes `DclFirebaseMessagingService` run in every app state — killed, background and foreground. A `notification` block would be drawn by Android itself whenever the app is not in the foreground and `onMessageReceived` would never fire, taking the channel, the de-duplication and the deep link out of our hands. That is also why the Firebase console composer cannot test this path: it always attaches one.

Registration goes through Segment. `SegmentIdentifyBody` is the first non-`track` type in this codebase, and it exists because push needs a *current-state* record per user rather than an append-only event: identify traits land in the warehouse's `users` table, one row per `userId`, which is what an audience query can join against.

The Day-1 local notification (#1651) is left exactly as it is. An earlier revision of this branch retired it on the assumption that server-sent push replaced it, and that turned out to be wrong for this particular notification: the two checks that decide whether a comeback nudge should fire — has this person already come back, and what is 22h from now in their own timezone — are an `if` and some arithmetic on a device that is by definition running, while answering the first one server-side means reading a warehouse ~2h behind and notifying people who already returned. Replacing it is tracked in #2840 and is out of scope here.

## Verified on device

Samsung A54, Android 16 / API 36. All three app states deliver, and `Push Opened` attributes correctly in both start kinds:

```
I DclFcmService:     FCM token refreshed (len=142)
I dcl-godot-android: [FCM] token ready (len=142)
I godot [Rust]:      [Push] identify queued (token_len=142, permission=granted)
[PUSH] opened campaign=test-killed push_id=push_afd0… start=cold
[PUSH] opened campaign=test-warm   push_id=push_warm_001 start=warm
```

Also measured, and worth knowing when reading campaign numbers: **an app in the *stopped* state (`Forzar detención` from Settings) never receives push** — nothing at all until the user opens it by hand. That is part of the structural gap between `sent` and `opened`. Closing it from recents is different and does receive.

## Two bugs the device testing found

**`try_to::<bool>()` does not convert a Kotlin return.** Godot hands a Kotlin `Boolean` back as an INT Variant, and `try_to` is a strict conversion that fails; with `.unwrap_or(false)` behind it, that failure is indistinguishable from a genuine "not granted". Every user would have been recorded as having denied notifications, and the first campaign would have gone to an empty audience. Fixed with `booleanize()`.

The same pattern sits in four other methods in `dcl_android_plugin.rs` (`add_calendar_event`, `share_text`, `log_firebase_event`, `share_text_with_image`). They are harmless today because no caller reads those returns, and they are left alone as unrelated — but the first one that does will get a false `false`.

**The `push_permission` trait aged and never updated.** The identify goes out at startup; the permission prompt only appears on the explicit Play-as-Guest / Sign-In tap. Anyone who accepted stayed recorded as `denied` until their next launch, so an audience built that day excluded exactly the person who had just said yes. The de-duplication guard is now keyed on `(token, permission)`.

## Notes for review

- `push_campaign_id` is deliberately not the ads `c=` token. That one is captured sticky by the client as the install's attribution, so routing a push through it would burn a user's install attribution on a re-engagement tap.
- `firebase-messaging` is pinned to 24.0.1 because that is already on the plugin's runtime classpath via `com.reown:android-core`. Asking for 24.1.0 fails to resolve rather than upgrading.
- `global.gd` was six lines under gdlint's 1900-line cap and is now exactly at it.
- `Push Permission Result` was dropped as redundant: `Click Button(accept|reject, NOTIF_PROMPT)` already gives the funnel and the new trait gives current state.


regenesis-claw (commented):
Adversarial pass on the client half. Followed `REVIEW.md`.

**Pre-flight:** branch is up to date with `main` (`behind_by: 0`). All 5 checks green. No `build`/`build-ios` label — an iOS build is **not** required here: the diff is Android plugin + Rust + GDScript with no iOS branch (the one `global.gd` iOS line is a comment). Draft, so comment-only, no verdict.

---

### [P1] Warm push opens report `trigger=icon` — the same bug you fixed for cold, mirrored on the warm path

`NOTIFICATION_APPLICATION_FOCUS_IN` is delivered by `SceneTree` through `root->propagate_notification()`, which visits root's children in order. Autoloads are added in the order listed in `project.godot`, and this PR inserts `SessionTracker` **after** `Global`. So on a warm push tap, inside one notification propagation:

1. `global.gd:1722` `Global._notification(FOCUS_IN)` → `process_deep_link(url, "warm")` → `_track_push_open_if_any` fires **`Push Opened`** ✅ → then calls `SessionTracker.notify_deep_link(params)` — and at that instant `_pending_start_kind == ""`, so `session_tracker.gd:104` early-returns and the link is dropped.
2. *Then* `SessionTracker._notification(FOCUS_IN)` runs → `_begin_open(START_WARM)` opens the pending window.
3. 1.5 s later `_process` → `_emit_open(_trigger_from_launch_deep_link())`, which returns `TRIGGER_ICON` for anything that isn't `START_COLD`.

Net effect: every warm push/deeplink open ships `App Opened { start_kind: "warm", trigger: "icon" }`. `Push Opened` is unaffected, which is why the device log in the description looks right — it only shows the `[PUSH]` lines. You should be able to confirm from the same run: your warm test should have printed `[SESSION] opened start=warm trigger=icon`.

The comment on `notify_deep_link` says "a link arriving mid-session finds no open and is ignored", but this link doesn't arrive mid-session — it arrives one node earlier in the same propagation as the open it belongs to.

Reordering the autoload isn't the fix (`SessionTracker._ready` would then run before `Global._ready`, `Global.metrics` would be null and the tracker would disable itself permanently). A small stash works:

```gdscript
var _recent_link_params: Dictionary = {}
var _recent_link_at_ms: int = 0

func notify_deep_link(params: Dictionary) -> void:
	if not _enabled:
		return
	if _pending_start_kind == "":
		# The FOCUS_IN that carries this link reaches Global before it reaches this node,
		# so the open it belongs to has not been begun yet. Hold it for that open.
		_recent_link_params = params
		_recent_link_at_ms = Time.get_ticks_msec()
		return
	_emit_open(TRIGGER_PUSH if _is_push(params) else TRIGGER_DEEP_LINK)
```

…and have `_begin_open` consume a stash younger than a frame or two (or let `_trigger_from_launch_deep_link` read it for warm opens as well).

### [P1] The `push-enabled` kill switch cannot stop token registration on any launch after the first

`metrics.rs` `ready()` reads `DclAndroidPlugin::get_fcm_token()` and, when non-empty, calls `_on_fcm_token_ready` → `emit_push_identify` **synchronously**. `push_enabled` defaults to `true` and is only written by `set_push_enabled`, which `feature_flags.gd::_apply_flags` calls after `_async_load()` completes — an HTTP round trip to the bff raced against a timeout. The token is cached in SharedPreferences from the previous run, so on every launch after install the identify is built and pushed onto `serialized_events` many seconds before the flag arrives.

Flipping the flag later doesn't retract it either: the gate is checked at emit time only, and the identify is already serialized in the queue, so it ships on the next flush regardless.

So the comment in `emit_push_identify`/`feature_flags.gd` — "the flags usually resolve first; when they do not, the default is on, so at worst one identify ships before the switch is read" — is inverted: the flag effectively never wins, and its only live effect is suppressing mid-session rotations and permission refreshes. If the flag is meant to be a real stop on collection, gate the startup identify on `FeatureFlags.is_loaded()` / the `flags_loaded` signal (the cached-token path costs nothing by waiting); if it isn't, the comment and the flag's stated purpose should say so.

### [P2] Push de-duplication is written with `apply()` in a process that gets killed on purpose

`PushTokenStore.markSeen` persists through `p.edit().putString(...).apply()`. `apply()` hands the write to a background thread and returns; a service started by FCM can be killed as soon as `onMessageReceived` returns, and unlike an Activity lifecycle handoff nothing waits for the pending write. Losing it re-opens exactly the case the store exists for — an FCM redelivery of the same `push_id` posts the notification twice. `commit()` here is cheap (one small string, already off the main thread) and correct. `saveToken` is fine on `apply()`: a lost token write self-heals via `requestFcmToken()` on the next launch.

### [P2] `dcl_liveops` is declared as the FCM default channel but is only created when a push is rendered

The manifest sets `default_notification_channel_id=dcl_liveops`, while the channel itself is created lazily in `PushNotificationBuilder.show()` → `ensureChannel`. On a fresh install that has never rendered a push, the id doesn't exist, so the fallback the meta-data exists to serve lands on FCM's `fcm_fallback_notification_channel` ("Miscellaneous") instead. Calling `ensureChannel(PUSH_CHANNEL_ID, …)` once at plugin init would make the declaration true, and would also put "News & Events" in the system settings before the first campaign rather than after it.

### [P2] The committed AARs are the artifact that actually ships, and nothing in CI rebuilds them from the reviewed Kotlin

`godot/addons/dcl-godot-android/bin/{debug,release}/*.aar` are build outputs; the Android job consumes them rather than building `plugins/dcl-godot-android`. Your note says they were rebuilt from the merged Kotlin and the device log supports that, but it is not verifiable from the diff and nothing prevents future drift. Flagging as a known verification gap, not as a defect in this PR.

### [P2] Two permanent `print()`s on the launch path

`deep_link_router.gd::_track_push_open_if_any` and `session_tracker.gd::_emit_open`. Console-only, so no Sentry cost, and they match the existing `[DEEPLINK]` convention — but they are permanent, not diagnostics for this PR. Keep them if they're meant as ops breadcrumbs; worth a line saying so.

### One question on a constant

`DEEP_LINK_SETTLE_SECONDS = 1.5` is justified as covering Android's async launch-intent delivery, but the description says the routing/listening skew you measured was ~6 s. If a cold link can land later than 1.5 s after `_ready`, `_emit_open(ICON)` has already fired and `notify_deep_link` finds no open — the cold path degrades to the same wrong `trigger` as the warm one. What did you actually observe for the *link* arrival relative to `SessionTracker._ready`?

---

### Checked and clean

- **`booleanize()` fix** — correct, and the four other `try_to::<bool>()` sites are genuinely unread today (the four `#[func]`s return values no caller consumes).
- **EULA gate on the identify** — verified, not just claimed: `serialized_events` drains only through `process_and_send_events`, reachable from `flush()` (gated at metrics.rs:878) and `timer_timeout` (gated at :380). A pre-consent identify waits in the queue exactly like a track event.
- **`userId` on the identify** — `create_metrics(config.analytics_user_id, …)`, a per-install UUID that is always non-empty (`config_data.gd:390`/`:510`) and pinned across auth changes. No empty-`userId` batch rejection, and no stale mapping on account switch, since the key is the install and not the wallet.
- **No double `Push Opened` on repeated foregrounds** — `getLaunchIntentData()` nulls `intent.data` after reading, so a later `FOCUS_IN` sees an empty URL and returns before `process_deep_link`.
- **FCM payload contract vs. mobile-bff** — exact match on every key: `v`, `push_id`, `category`, `title`, `body`, `deep_link`, `image_url`, and `push_campaign_id`/`push_id` on the link (`withAttribution`). `push_campaign_id` carries `campaign_key`, which is what the server's join key is.
- **`${godotPluginPackageName}` in the service declaration** resolves to `org.decentraland.godotexplorer` (`build.gradle.kts:10,25`), so the class is found.
- **`readCapped`** enforces the cap while reading rather than after, and the cross-protocol redirect case degrades to text (HttpURLConnection won't follow https→http, the 3xx falls out of the 200..299 check).
- **`trayIdForPush`** keeps pushes in the top half of the int range, disjoint from local reminder ids.
- **Channel split** (local reminders vs. liveops) is the right call and the reason given is the real one.
- **`refresh_push_identify`** is safe on iOS — no plugin, empty token, early return — and the `(token, permission)` guard key is the correct fix for the stale-trait bug.
- No secrets, no committed `.claude/` files, no non-English comments, no submodule drift, English throughout.

regenesis-claw (changes requested):
**Requesting changes** — both P1s from my review are still open at `b2efbf3` (no new commits since).

To be clear about what this does and does not block: **delivery is fine**. A push arrives, renders, deep-links and de-duplicates correctly, and the FCM contract with mobile-bff matches key for key. Both P1s are about *measurement* and the *kill switch* — which matters, because the numbers this PR produces are the ones the whole workstream will be judged on.

### [P1] Warm push opens are attributed `trigger=icon`

`FOCUS_IN` is delivered through `root->propagate_notification()`, which visits root's children in the order autoloads are listed in `project.godot`. This PR inserts `SessionTracker` **after** `Global`, so on a warm push tap:

1. `Global._notification(FOCUS_IN)` → `process_deep_link(url, "warm")` → `Push Opened` fires correctly, then `SessionTracker.notify_deep_link(params)` runs while `_pending_start_kind == ""` and returns at `session_tracker.gd:104` — the link is dropped.
2. `SessionTracker._notification(FOCUS_IN)` → `_begin_open(START_WARM)` opens the window, one node too late.
3. Settle expires → `_trigger_from_launch_deep_link()` returns `TRIGGER_ICON` for any non-cold open.

So `App Opened { start_kind: "warm", trigger: "icon" }` for every warm push and every warm deeplink. Your device log in the description only shows the `[PUSH]` lines, so it doesn't cover this — the `[SESSION]` line from that same warm run should show it directly.

The stash-and-consume fix from my review still applies (`notify_deep_link` holding the params when no open is pending, `_begin_open` consuming a stash younger than a frame or two). Reordering the autoload is not the fix: `SessionTracker._ready` would then run before `Global._ready`, `Global.metrics` would be null, and the tracker would disable itself for the whole process.

### [P1] `push-enabled` cannot stop token registration

`Metrics::ready()` reads the cached token and calls `emit_push_identify` synchronously. `push_enabled` defaults to `true` and is only written by `set_push_enabled`, which `feature_flags.gd::_apply_flags()` calls after `_async_load()` — an HTTP round trip raced against a timeout. On every launch after install the token is already in SharedPreferences, so the identify is built and queued long before the flag lands; and because the gate is checked only at emit time, the serialized identify ships on the next flush even if the flag then arrives as `false`.

The in-code justification — "the flags usually resolve first; when they do not, the default is on, so at worst one identify ships before the switch is read" — is inverted: the flag effectively never wins on the startup path, and its only live effect is suppressing rotations and permission refreshes.

Either gate the startup identify on `FeatureFlags.is_loaded()` / the `flags_loaded` signal (the cached-token path loses nothing by waiting), or keep the behaviour and correct the comment and the flag's stated purpose so nobody reaches for it as a stop.

### Unchanged P2s

`markSeen` persisting with `apply()` in an FCM-started service (`commit()` is the durable call there), `dcl_liveops` declared as the FCM default channel but created lazily, and the committed AARs not being rebuilt by CI. None of these are blocking.

### Not blocking, and verified rather than assumed

`booleanize()` fix, the EULA gate genuinely covering the identify, `userId` being a non-empty per-install UUID, no double `Push Opened` on repeated foregrounds (`getLaunchIntentData()` nulls `intent.data`), `${godotPluginPackageName}` resolving correctly, `readCapped`, the tray-id split, and the local/liveops channel separation. Pre-flight per `REVIEW.md`: branch is up to date with `main`, all 5 checks green, iOS build correctly not required for an Android-only change.

regenesis-claw (approved):
**Approve.** Both P1s are properly fixed at `1794814`, and I traced each one rather than taking the commit message for it.

### [P1] Warm-open attribution — fixed

The stash is the right shape and the edges are covered:

- `notify_deep_link` with no pending open now holds `params.duplicate()` for `DEEP_LINK_STASH_SECONDS`, and `_begin_open` drains it immediately instead of waiting out the settle window. Copying the dictionary matters — `Global.deep_link_obj.params` is live and would otherwise alias.
- `_emit_open` from inside `_begin_open` clears `_pending_start_kind`, so the `_process` settle branch can't fire a second open for the same entry.
- **The FOCUS_OUT clear is the part that makes this safe.** `_process` is frozen while backgrounded, so a 0.5 s TTL cannot decay out there; without that clear, a link tapped mid-session would have survived the whole background period and attributed the *next* return. Good catch — I hadn't called that out and it's the one way this could have gone wrong.
- Cold start still resolves through `_trigger_from_launch_deep_link` reading `Global.deep_link_obj`, because `notify_deep_link` fires while `_enabled` is still false (Global's `NOTIFICATION_READY` runs before this autoload's `_ready`), so nothing is stashed and the old path is untouched. A launch intent arriving *late* now lands on the pending open and resolves it properly instead of racing the settle window.
- `_trigger_for_params` being shared by both paths removes the divergence between "told about the link" and "read it off Global".

### [P1] `push-enabled` — fixed, and it is a real switch now

`emit_push_identify` parks the token when `push_flag_resolved` is false, and `set_push_enabled` releases it on first read. Checked the two things that decide whether this is a hold or a hang:

- `_apply_flags()` is reached on **both** branches of `_async_load` (line 102, after the `if result is PromiseError` / `else`), so a failed or timed-out fetch still ends the hold.
- `TIMEOUT_SECONDS = 5.0` bounds it, so the worst case is a 5 s delay on one identify, not a lost one. The comment's "within 5s even offline" checks out.

Releasing outside the equality check is correct and worth the comment you gave it: the common case is the flag resolving to its default, which doesn't change `push_enabled` but still has to let the waiting identify through. And a first read of `false` drops the held identify on the re-entry into `emit_push_identify`, which is the behaviour the flag advertises.

### Two residual notes — neither blocking, both worth knowing when you read the numbers

1. **A push tapped within `WARM_OPEN_FLOOR_SECONDS` of backgrounding produces `Push Opened` with no matching `App Opened`.** `_begin_open` applies the 30 s floor before it drains the stash, so the stash just expires. That is the floor behaving as designed — but it means the two events don't join 1:1, and a funnel built as `App Opened(trigger=push) → …` will under-count quick re-taps that `Push Opened` does see. Worth a note wherever the warehouse query lives rather than a code change.
2. **A deep link with no query params attributes as `icon`.** `_trigger_for_params` returns `TRIGGER_ICON` for an empty dictionary, so a bare `decentraland://open` tap is indistinguishable from a launcher tap. Pre-existing on the cold path and not reachable for push (a campaign link always carries `push_campaign_id`), so it only affects bare shared links.

### CI

`📊 Static checks` and `🔧 Clippy & Build Checks` green at this head — which is the GDScript parse/lint and the Rust compile, i.e. the two that can fail on this diff. `🤖 Android` and `🐧 Linux` were still running when I reviewed; approving on the assumption they land green, since nothing here touches packaging.

sdilauro (approved):
LGTM overall — nice work on the session tracker design and the booleanize() catch.

One thing worth flagging before merging: in session_tracker.gd, WARM_OPEN_FLOOR_SECONDS (30s) gates _begin_open(START_WARM) purely on elapsed time, with no exception for a deep link. If a user backgrounds the app and taps a push notification within that 30s window, _begin_open returns early before _pending_start_kind is set, so notify_deep_link just stashes the params for 0.5s with nothing to consume them — meaning App Opened (trigger=push) never fires for that open, even though Push Opened still does via DeepLinkRouter. Is that intentional (treating anything under 30s as "not a real open" regardless of cause), or should a deep link/push override the floor? Given push-driven opens are presumably one of the more important things trigger is meant to capture, seems worth a second look — but not blocking.

PR #2883, opened 2026-09-11
fix(ios): stop baking the dev scene-inspector endpoint into App Store builds

leanmendoza:
## What

Every shipped iOS build carries a debug endpoint pointed at whatever machine built it. Production crash reports from `1.13.0+1834` show:

```
godot_engine.command_line_arguments = ["--scene-inspector=ws://192.168.64.5:9231"]
godot_engine.debug_build = true      godot_engine.mode = "export_debug"
app.build_type = "app store"         environment = production
```

`192.168.64.5` is the CI runner's LAN address. Users dial it on every launch, forever, and it is what makes iOS ask for Local Network access. The build also ships `NSLocalNetworkUsageDescription` and `NSAllowsLocalNetworking`, which the plugin's own docstring says release builds must not declare ("Release / TestFlight / App Store exports stay clean, keeping the App-Review scrutinized local-network keys out of production").

## Why it happens

`_export_begin` gated the injection on `is_debug`. That is not a dev signal: CI runs `cargo run -- export --target ios` with no `--release`, and `src/export.rs:366` picks the export mode from that flag, so the store build is an `--export-debug` export. `is_debug` is therefore true in CI and the plugin injects the plist keys plus `_lan_ip()`. Present since #2342 (1 Jul), so 1.11.x onwards.

## Changes

- The plugin now needs an explicit developer signal on top of `is_debug`: either `DCL_IOS_GODOT_CMDLINE` is set, or the editor passed the remote-debug flag. CI sets neither, so store exports carry nothing — no plist keys, no `godot_cmdline`.
- `cargo run -- run --target ios` sets `DCL_IOS_GODOT_CMDLINE=auto`, which keeps the current LAN auto-dial for dev deploys. `--no-hub` still sets `none`. An editor deploy with remote debug still auto-dials.

Behaviour change for developers: a plain editor deploy **without** remote debug no longer auto-dials the hub. Use `cargo run -- run --target ios`, or set the env.

## Not fixed here

The root cause is that CI exports iOS with the debug template at all, so `OS.is_debug_build()` is true in production. That flips gates well beyond this plugin (debug chat commands, early log capture, the IAP StoreKit environment switch — see #2829). Passing `--release` there is the real fix, but it changes the templates, symbols and every one of those gates at once, so it does not belong in a hotfix.

## Test plan

- [ ] **Store export is clean** — export via CI (or `cargo run -- export --target ios`) and check the generated `Info.plist`: no `godot_cmdline`, no `NSLocalNetworkUsageDescription`, no `NSAppTransportSecurity`.
- [ ] **Dev deploy still works** — `cargo run -- run --target ios` with the hub up: the device connects to the hub and `scripts/unified.sh` shows its logs.
- [ ] **`--no-hub` still works** — `cargo run -- run --target ios --no-hub`: no connect-out arg, `--console` streams as before.
- [ ] **Shipped build no longer prompts** — on a build from this branch, a first launch on a production build raises no Local Network prompt, and the Sentry `command_line_arguments` context is empty.

### DIFF
--- a/godot/addons/dcl-ios-devtools/export_plugin.gd
+++ b/godot/addons/dcl-ios-devtools/export_plugin.gd
@@ -56,10 +56,14 @@ class DclIosDevExportPlugin:
 		return platform is EditorExportPlatformIOS
 
 	func _export_begin(_features, is_debug, _path, flags):
-		# Only debug builds declare local networking / inject launch args. The relay
-		# and log-stream are themselves gated to OS.is_debug_build(), so release
-		# builds neither use nor declare it — which is what keeps App Review happy.
+		# `is_debug` alone is not a dev signal: CI exports the store build with
+		# --export-debug too, which shipped the local-network keys and a
+		# --scene-inspector pointed at the builder's LAN IP to every App Store user.
+		# Require an explicit dev signal on top — the xtask env or an editor deploy
+		# with remote debug. CI sets neither.
 		if not is_debug:
+			return
+		if _cmdline_env().is_empty() and (flags & REMOTE_DEBUG_FLAG) == 0:
 			return
 		var lines: Array = DEV_PLIST_LINES.duplicate()
 		lines.append_array(_godot_cmdline_lines(flags))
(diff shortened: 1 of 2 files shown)


leanmendoza:
Superseded by #2884 (Release Candidate 1.13.2), which carries this commit plus the startup crash fix and the version bump.

Issue #20, opened 2023-06-26
Component Status


leanmendoza:
In progress issue

## Transform
- [X] Parenting with differing orphan entities
- [X] Apply right-handed to left-handed transformations from Scene to Godot

## Billboard
- [X] Implemented: the system runs updates with a scene update tick
- [ ] Apply a global system that updates the nearest entities no matter which scene owns the entity (scene-tick independent)

## GltfContainer
- [X] Almost complete implementation
- [X] Tag meshes with `_collider` in their names as `invisible` meshes
- [X] Apply generating collider (and setting the collision mask) according with the ADR on visible/invisible 

Known issues:
- [ ] Texture isn't applied on `XXXX` scene
- [ ] Floor is sometimes not loaded `Laser ray` scene correctly

## GltfContainerLoadingState
- [X] Implemented

## Animator
- [ ] Implemented

PR #2886, opened 2026-09-12
fix(ios): bake shaders in the CI export

### LINKED ISSUE #2891: [Bug] iOS store builds ship with unbaked shaders (5–130 s freezes at first draw, Sentry App Hangs) and the debug engine template
## **⚠️ Problem:**
Every iOS user on 1.13.1 hits main-thread freezes of **5–130 s** inside `-[MDLazyLibrary load] → newLibraryWithSource`: shaders are compiled from Metal source at first draw. Sentry App Hang issues `GODOT-EXPLORER-2PK`, `2PH`, `2QH`, `2PS` — 23 of ~28 iOS users, one watchdog kill.

Root cause: the export preset has `shader_baker/enabled=true`, but Godot's `ShaderBakerExportPlugin::_is_active()` needs a RenderingDevice renderer in the exporting editor and returns `false` — with no message — under `--headless`, which is how CI exports iOS. **No store build has ever shipped baked shaders** (the 1.13.1 export log, job 102990799863, has zero baker output).

Second finding from the same logs: every iOS Sentry event reports `build_type: debug` / `mode: export_debug`. CI never passed `--release`, so the App Store binary embeds the **debug** engine template.

## **🏁 Scope:**
Make the iOS CI export actually bake shaders and fail loudly when it can't; export App Store / TestFlight builds with the release template. Android keeps `shader_baker/enabled=false` (baking there would also skip glslang at first launch — separate follow-up).

**Merge order for the release-template change.** The debug template is what currently turns GDScript calls on freed nodes into logged errors on iOS; on Android (release template) the same calls are the largest crash family. The release-template change is split into its own PR (#2898, draft) and must land only after decentraland/godotengine#21 (release-template object validation) is pinned in `src/consts.rs`. The shader-bake fix (#2886) is independent and merges first.

## **📝 Deliverables:**
- [x] `src/export.rs`: for `--target ios` run a windowed editor with `--rendering-method mobile` (host RenderingDevice driver — Metal on Apple Silicon); every other target keeps the headless export
- [x] `verify_shader_bake_output`: the export fails unless Godot's output shows `Started Baking shaders (N steps)`, at least one `Storing File: … .metal.cache`, and no `Metal shader baking limited to SPIR-V` warning; `DCL_SKIP_SHADER_BAKE_CHECK=1` bypass for local debugging; unit tests replaying the CI log lines
- [x] `ios_r2_artifact.yml` / `ios_self_hosted.yml`: `Ensure Metal toolchain (shader baker)` step (`xcodebuild -downloadComponent MetalToolchain` when `xcrun -sdk iphoneos metal` is missing); export timeout 5 → 25 min
- [ ] `--release` on the iOS export — #2898 (draft), gated on decentraland/godotengine#21 being pinned in `src/consts.rs`
- [x] `CLAUDE.md`: windowed export and toolchain requirement (#2886); the `--release` export command (#2898)
- [x] CI export verified: `Started Baking shaders (597 steps)`, 83 `.metal.cache` files packed into the .pck, run 34758040453
- [ ] After the next iOS release: no new events on `2PK` / `2PH` / `2PS` for that build
- [ ] After #2898 ships: iOS events carry `build_type: release`
- [ ] Follow-up: enable the shader baker on Android

## **🔗 References:**
- PRs: #2886 (shader bake), #2898 (release template, draft)
- Sentry: `GODOT-EXPLORER-2PK`, `2PH`, `2QH`, `2PS` (App Hang, `MDLazyLibrary load`)
- Gates the release-template commit: decentraland/godotengine#21
- 1.13.1 export log with zero baker output: job 102990799863

kuruk-mm:
## What

iOS builds now ship with baked shaders. The iOS export runs a windowed Godot editor instead of `--headless`, installs the Xcode 26 Metal toolchain on the CI runner when it is missing, and fails the build unless Godot's export output shows the baker ran, packed `.metal.cache` files, and had the toolchain.

## Why

Every iOS user on 1.13.1 hit main-thread freezes of 5–130 s (Sentry App Hang issues `2PK`, `2PH`, `2QH`, `2PS`, 23 of ~28 users, one watchdog kill) inside `-[MDLazyLibrary load] → newLibraryWithSource`: shaders compiled from Metal source at first draw. The preset has `shader_baker/enabled=true`, but Godot's `ShaderBakerExportPlugin::_is_active()` needs a RenderingDevice renderer in the exporting editor and returns `false` — with no message — under `--headless`, so no store build ever had baked shaders (the 1.13.1 export log, job 102990799863, has zero baker output).

The same investigation found CI exports the debug engine template; that change is split out into #2898 because it is gated on an engine fork fix. This PR keeps the export as debug.

Closes #2891

## Details

<details>
<summary>Expand</summary>

- `src/export.rs` — for `--target ios` the editor runs with `--rendering-method mobile` and without `--headless` / `--rendering-driver opengl3` (host default RenderingDevice driver, Metal on Apple Silicon); every other target keeps the headless export. Godot's output is streamed through a pipe and scanned: `verify_shader_bake_output` requires the baker's `Started Baking shaders (N steps)` progress line, at least one `Storing File: … .metal.cache` savepack line, and no `Metal shader baking limited to SPIR-V` warning; `DCL_SKIP_SHADER_BAKE_CHECK=1` bypasses it for local debugging. The baker's on-disk output is `godot/.godot/exported/<hash>/shader_baker/iOS/metal/` (project data path), not next to the IPA — the first version of this check looked in `exports/shader_baker/` and failed the first CI run even though the bake had worked; the second commit fixes that and adds unit tests replaying the CI log lines.
- `.github/workflows/ios_r2_artifact.yml`, `ios_self_hosted.yml` — new step `Ensure Metal toolchain (shader baker)` (`xcodebuild -downloadComponent MetalToolchain` when `xcrun -sdk iphoneos metal` is missing) and export timeout 5 → 25 min for the bake.
- `CLAUDE.md` — documents the windowed export and the toolchain requirement.
- Runner: the WarpBuild macOS runner does have a GUI session and a Metal RenderingDevice (`Metal 4.0 - Forward Mobile - Using Device #0: Apple - Apple Paravirtual device`), and the Metal toolchain was already installed (`Apple metal version 32023.883`), so the new step is a 3 s no-op there. On a runner without a display the export fails loudly instead of silently shipping unbaked shaders.
- Not in this PR: Android keeps `shader_baker/enabled=false`; baking there would also skip glslang at first launch and is a follow-up.

**Verified locally**
- [x] `cargo fmt` / `cargo check` (xtask) green.
- [x] CI export on this branch (run 34719186890, job 103621828160): `Started Baking shaders (597 steps)` (21:20:42 → done 21:21:01), 83 `.metal.cache` files stored in the .pck, no SPIR-V-only warning. The step still failed on the old directory check — see the second commit.
- [x] Mobile distribute re-run on the fixed head (run 34758040453, job 103725810548): `Export` printed `Shader baker: 597 variants compiled, 83 .metal.cache files packed into the .pck`, the export was uploaded to R2 and handed off to TestFlight signing. Run conclusion: success. (Those runs still had `--release` on the branch; the bake path is identical with the debug template.)
- [x] `cargo test shader_bake_scan` (xtask): 3 tests replaying the CI log lines pass.

</details>

### Case 1 — First scene load has no long freeze (iOS)
**Steps:**
1. Open the app on an iPhone for the first time after installing this build and sign in.
2. Wait for Genesis Plaza to appear, then walk around for a minute.
3. Teleport to a scene you have never visited (e.g. from the Discover tab).
- [ ] **Expected:** during loading and on the first frames of each scene the screen keeps animating; no freeze longer than about one second, and no "app not responding" behaviour.
- [ ] **Expected (Sentry, next day):** no new events on the App Hang issues `GODOT-EXPLORER-2PK` / `2PH` / `2PS` for this build.

### Case 2 — Regression: iOS build behaves as before
**Steps:**
1. Open the app, enter Genesis Plaza, open the backpack and equip a wearable, open a friend's passport, play an emote, rotate the phone.
- [ ] **Expected:** identical behaviour and visuals to the current TestFlight build (baked shaders change when shaders compile, not what they render).

### Case 3 — Android is untouched
- [ ] **Regression:** open the app on Android, enter Genesis Plaza, teleport once: unchanged from the current build (Android export path is not modified).


regenesis-claw (changes requested):
Great find, and the diagnosis is airtight: `ShaderBakerExportPlugin::_is_active()` silently returning false under `--headless` / `opengl3` is exactly the kind of failure that never shows up in a log, and the fix pins it down with a positive assertion instead of hoping. Streaming Godot's output through a scanner and failing the export when the baker didn't run is the right shape — the check is what keeps this from silently regressing the next time someone adds a flag to the export invocation. Unit tests replaying real CI log lines are a nice touch.

Pre-flight (REVIEW.md §0): iOS build present and green on this head (`🍏 iOS → R2`, job 103725810548, `Shader baker: 597 variants compiled, 83 .metal.cache files packed`). All PR checks pass. Branch is 2 commits behind `main` — not enough to distrust the signal, but worth a merge before landing. No submodule drift, description has What/Why/Details/Test plan and reads as the author's own analysis.

---

## Findings

### [P1] The `--release` commit's stated prerequisite is not met — this PR cannot be merged as one unit today

The description says `chore(ios): export App Store builds as release` must land only after decentraland/godotengine#21 is pinned in `consts.rs`. As of this review:

- decentraland/godotengine#21 is **open**, not merged.
- `consts.rs` on this branch is byte-identical to `main`: `GODOT_BUILD_SHA = "6289a3b2b"`, `GODOT_USE_BRANCH = Some("fix/android-sha1-build-id")` — i.e. still the #20 build, with no release-template object validation.

So merging this PR as it stands does the thing the description explicitly says not to do: ships iOS builds on the release engine template while the freed-instance checks that currently turn those calls into logged errors are compiled out. On Android (release template) that same call pattern is already the largest crash family, so the expected outcome is trading App Hangs for a new SIGSEGV family on iOS.

Also worth stating plainly, because the description frames this as "App Store builds": both workflows pass `--release` **unconditionally**, so *every* iOS export — PR artifacts and QA TestFlight builds, not just store cuts — moves to the release template the moment this merges. That is probably what you want (catch it in QA first), but it is a wider blast radius than "App Store builds are exported with the release engine template" suggests, and it should say so.

Options, any of which unblocks:
1. Drop commit 2 from this PR and open it as a follow-up stacked on the engine pin (the description already offers this).
2. Keep it here and bump `consts.rs` to the #21 build in this PR, so the gate is satisfied by the same merge.

Commit 1 + 3 (bake + verification) are approvable on their own — nothing below blocks them.

### [P2] `scan_line`: an unparseable step count reports as "the baker never ran"

In `scan_line`, when the `Started Baking shaders (` match is found but the digit parse fails, `bake_steps` stores `0` while `bake_started` stores `true`. `finish()` then yields `Some(0)`, and `verify_shader_bake_output` does `output.bake_steps.unwrap_or(0)` → `steps == 0` → the "Shader baker never ran … export ran without a RenderingDevice renderer" error. If Godot ever changes that line's formatting (or drops an ANSI sequence between the `(` and the digits), the build fails with a diagnosis pointing at the wrong cause, and whoever hits it will go looking at `--headless` flags that are already correct. Suggest keeping the two states distinct — e.g. branch on `output.bake_steps.is_none()` for "never ran" and treat `Some(0)` as "baker started but the progress line didn't parse".

### [P2] The bake check also fires on non-release local iOS exports

`let bake_shaders = target == "ios";` gates both the windowed editor *and* the hard verification, regardless of `release`. A dev exporting iOS over SSH or on a box without the Metal toolchain now gets a hard failure on `cargo run -- export --target ios` and has to know about `DCL_SKIP_SHADER_BAKE_CHECK=1`. It is documented in `CLAUDE.md`, so this is a judgement call, not a bug — but consider hard-failing only when `release` is set and downgrading to a loud warning otherwise. The thing worth protecting is the shipped build.

### [P2] Stale "headless" wording in the block right below the change

The macOS ad-hoc-codesign block still opens with "The headless Godot editor dlopen's the GDExtension…" and prints "Ad-hoc signing libdclgodot.dylib (macOS) for the headless export…". That block exists for the macOS host, which is precisely the path that is no longer headless after this PR. One-line wording fix so the next reader doesn't conclude iOS still exports headless.

### [P2] `Ensure Metal toolchain` has no step timeout

`xcodebuild -downloadComponent MetalToolchain` is a network download on a cold runner. On the WarpBuild runner it is a 3 s no-op today, but on a runner where it actually downloads (or stalls behind an Apple auth prompt / rate limit) the step has no `timeout-minutes` and eats the job budget instead of failing fast. A `timeout-minutes: 15` on that step would make the failure mode readable.

### [P2] The self-hosted iOS path hasn't been exercised with the windowed export

`ios_self_hosted.yml` runs on `[self-hosted, macOS, ARM64, dcl-ios]` and got the same `--headless`-free export, but the verification in the description is all from the WarpBuild runner. If that agent runs from a LaunchDaemon rather than a logged-in GUI session, the export will now hard-fail there — by design, but it converts the fallback iOS path from "works, ships unbaked" to "doesn't run at all". Worth a `workflow_dispatch` dry run on that runner before this becomes the fallback anyone reaches for.

---

## Not findings (checked)

- `run_godot_streaming`: both pipes are drained by dedicated threads spawned before `child.wait()`, so there is no pipe-buffer deadlock, and the `join()`s after `wait()` establish the happens-before the `Relaxed` atomics need. Correct as written.
- Ordering of `restore_export_presets` before `verify_shader_bake_output` — the presets backup is restored even when the bake check fails, so a failed export doesn't leave the working tree patched. Good.
- Every other target keeps `--rendering-driver opengl3 --headless`; Android/Linux/Windows/macOS export paths are untouched, consistent with the "Android is untouched" test case.
- No security-relevant surface in the diff: no secrets, no user-controlled input, no new network endpoints. The one new shell step runs fixed `xcrun`/`xcodebuild` commands with no interpolation. **No security issues found.**
- Test plan: executable by QA as written — starts from opening the app, concrete steps, observable pass/fail, Sentry follow-up and an Android regression case. No changes requested there.

---

**Requesting changes** on the P1 only: the merge-order gate is unmet as of this review (godotengine#21 open, `consts.rs` unbumped). Split commit 2 out or bump the engine pin here, and this is an approve — the shader-baking work itself is solid and I'd take it today on its own.

_Reviewed against `REVIEW.md`. Requested by Mateo via Slack._

manuelmaceira (approved):

regenesis-claw (approved):
Delta re-review at `0c52741` (was `78b5a90`). The only code change is the removal of the `chore(ios): export App Store builds as release` commit — now split out as #2898 — plus the matching title/description trim. That resolves the blocker cleanly, and the description is honest about it ("This PR keeps the export as debug").

## Round-1 findings

| # | Finding | Status |
|---|---|---|
| P1 | `--release` commit's engine-pin prerequisite (godotengine#21) unmet | ✅ Fixed — commit dropped, split into #2898; title, What/Why and Details all updated to match the 2-commit branch |
| P2 | `scan_line`: started-but-unparseable step count reports as "baker never ran" | ❌ Open — non-blocking |
| P2 | Bake check fires on non-release local iOS exports | ⚠️ Retracted, see below |
| P2 | Stale "headless Godot editor" wording in the macOS codesign block | ❌ Open — non-blocking |
| P2 | `Ensure Metal toolchain` step has no `timeout-minutes` | ❌ Open — non-blocking |
| P2 | `ios_self_hosted.yml` path unexercised with the windowed export | ❌ Open — non-blocking |

**Retracting the "gate the check on `release`" suggestion.** With `--release` gone, the CI export *is* a debug export — gating the verification on `release` would now switch the check off for the exact builds it exists to protect. `bake_shaders = target == "ios"` is the right condition as the branch stands. The local-dev friction (a headless/SSH iOS export now hard-fails without `DCL_SKIP_SHADER_BAKE_CHECK=1`) stands as a documented trade-off, not a finding.

## New note (P2, not blocking)

No `🍏 iOS` job has run on this head — the green checks at `0c52741` are Linux / Android / Docker / Clippy / Static. The description says the bake was verified on runs 34719186890 and 34758040453 and adds "(Those runs still had `--release` on the branch; the bake path is identical with the debug template.)" That last clause is the one claim in the PR that isn't backed by a run: it's very likely true — `ShaderBakerExportPlugin::_is_active()` keys off the preset flag and the renderer, not the debug/release template — but it is the exact claim this PR exists to stop taking on faith. One `build`-labelled run on this head before merge turns it into evidence, and if the bake ever didn't happen the new check fails the job loudly rather than shipping quietly, so the downside is bounded either way.

## Other checks

- Diff is otherwise byte-identical to the round-1 head: `export.rs` streaming scanner, the verification, the unit tests, the Metal-toolchain step and the 5 → 25 min timeouts are unchanged and were already reviewed.
- Branch is 2 commits behind `main`; still fine, worth a merge before landing.
- #2898 carries the `--release` change against base `main` with the godotengine#21 gate intact — that gate needs checking there, not here.
- No security issues found (unchanged from round 1).

**Approving.** The five open P2s are all small and none of them affect what ships; happy to see them as a follow-up or not at all.

_Reviewed against `REVIEW.md`. Requested by Mateo via Slack._

PR #2887, opened 2026-09-13
ci: keep one live "Mobile build pipeline" status card on the PR

### LINKED ISSUE #2890: CI: keep one live "Mobile build pipeline" status card on the PR (build number, iOS/Android legs, TestFlight result)
## **⚠️ Problem:**
A `build`-labelled PR gets its mobile distribution status in three stale fragments: a "Mobile Distribution Triggered" comment that never changes, a separate "🤖 Android Build Ready" comment, and the per-push build-report. None of them carries the build number / version, and the PR never learns whether iOS reached TestFlight — that only shows in the Slack root card, which every leg of the pipeline already updates. Reviewers and QA have to cross-reference Slack to know which build to install.

## **🏁 Scope:**
Mirror the Slack root card onto the PR: **one** "📱 Mobile build pipeline" comment, created when the label is applied and updated in place by every leg (prepare, reuse decisions, iOS export / handoff, APK ready, Firebase, AAB, TestFlight result, failure, cancel), with a collapsed timeline of what happened when. Replaces the two pipeline comments; the per-push `build-report` comment from `android_builds.yml` stays (different lifecycle: every push vs. per distribution). No new secrets — the TestFlight result is mirrored by watching the `godot-asc-deploy` run we dispatched with the existing `ASC_DEPLOY_DISPATCH_TOKEN`.

## **📝 Deliverables:**
- [x] `.github/scripts/pr-card.py`: card state as JSON in a hidden HTML comment; each call merges only the non-empty fields it passes so legs never clobber each other; `SHA` / `RUN_ID` guards drop writes from superseded runs; best-effort (no token / fork PR / API error → prints and exits 0)
- [x] `mobile_distribute.yml`: a `PR card —` step next to each Slack step; legacy comments upgraded into the card
- [x] `ios_r2_artifact.yml`: card steps for build start / export / handoff (`pull-requests: write` declared)
- [x] `ios-wait` job (PR distributions only, never fails the run): polls `gh run list` on asc-deploy for the run titled `🍏 <branch> @ <sha>` and writes 🔏 signing… / ✅ TestFlight build N / ❌ sign-upload failed / ⚠️ cancelled — the same lines asc-deploy writes to the Slack root
- [ ] decentraland/godot-asc-deploy#5 merged (`run-name` per branch@sha) — until then the iOS row ends in "❓ TestFlight result unknown"
- [ ] Verified end-to-end on a `build`-labelled PR: iOS row ends in `✅ TestFlight build N`, header flips to ✅ SUCCESS

## **🔗 References:**
- PR: #2887
- Depends on: decentraland/godot-asc-deploy#5
- Slack root card this mirrors: `mobile_distribute.yml` Slack steps + `godot-asc-deploy`

kuruk-mm:
## What
The `build` label now keeps **one** live "📱 Mobile build pipeline" comment on the PR — the GitHub twin of the Slack root card: build number and version, branch/commit/trigger, the state of the iOS and Android legs, the APK/AAB/TestFlight links, and a collapsed timeline of what happened when. It replaces the two comments the pipeline posted before ("Mobile Distribution Triggered", which never changed, and a separate "Android Build Ready").

## Why
Slack already had a single card that every leg updates; the PR had three stale fragments and never learned the build number or whether iOS made it to TestFlight. No new secrets: the TestFlight result is mirrored by a new `ios-wait` job that watches the `godot-asc-deploy` run it dispatched with the existing `ASC_DEPLOY_DISPATCH_TOKEN` — the private repo keeps holding no token for this one. That job needs decentraland/godot-asc-deploy#5 (a one-line `run-name`) on asc-deploy `main` first; until then the card's iOS row ends in "❓ TestFlight result unknown" after the wait. The per-push `build-report` comment from `android_builds.yml` is untouched (different lifecycle: every push vs. per distribution).

Closes #2890

## Details
<details>
<summary>Expand</summary>

- `.github/scripts/pr-card.py` — new. State is JSON in a hidden HTML comment; each call merges only the non-empty fields it passes, so legs never clobber each other. Two guards drop writes from superseded runs: `SHA` (re-label on a newer commit) and `RUN_ID` (re-label of the same commit while the cancelled run's `always()` job is still finishing). Best-effort: no token / fork PR / API error → prints and exits 0.
- `mobile_distribute.yml` — a `PR card —` step next to each Slack step (prepare, reuse decisions, handoff, APK ready, Firebase, AAB, failure, cancel); new `ios-wait` job (PR distributions only, never fails the run) that polls `gh run list` on asc-deploy for the run titled `🍏 <branch> @ <sha>` created after this run started, writes "🔏 signing…" when it appears and `✅ TestFlight build N` / `❌ sign/upload failed` / `⚠️ cancelled` when it concludes — same lines asc-deploy writes to the Slack root.
- `ios_r2_artifact.yml` — card steps for build start / export / handoff; gains `pull-requests: write` (a called workflow needs it declared).
- Slack scripts and behaviour untouched.

Verified locally:
- [x] Rendering + hidden-state round-trip, including `--` in a branch name (GitHub would otherwise stop treating the state as an HTML comment)
- [x] Create / legacy-comment upgrade / merge / stale-run guards against a fake GitHub API
- [x] `ios-wait` matching loop against real runs: found+completed and not-found→timeout paths
- [x] actionlint: no new findings
- [x] First `build`-label run on this PR (before `ios-wait` existed): card created and filled by both legs, timeline in order — see the comment below

</details>

## Test plan

No QA needed — no behavior change in the app.

CI: this PR carries the `build` label, so its own thread shows the card being filled in by the run. With asc-deploy#5 merged, the iOS row ends in `✅ TestFlight build N` and the header flips to ✅ SUCCESS.


regenesis-claw (approved):
## Review — `ci: keep one live "Mobile build pipeline" status card on the PR`

**Verdict: Approve.** No P0/P1. Nine P2s below, none blocking. Followed `REVIEW.md` (including the §0 pre-flight).

### Pre-flight (REVIEW.md §0)
1. **Branch is 2 commits behind `main`** (ahead 2 / behind 2). Normally I'd ask for a merge first, but the diff touches only `.github/**` + three docs files — no Rust, no GDScript, no scene/resource — so a stale base can't invalidate it. Noting rather than holding.
2. **iOS build not required.** No native iOS plugin, no `OS.get_name() == "iOS"` branch, no deeplink/keyboard/safe-area/av/comms code. This is CI + docs only. (The `build` label is currently *off* the PR; the run referenced in the body is from an earlier labelled push.)
3. CI: 🐧 Linux, 🐳 Docker, 📊 Static checks, 🔧 Clippy, 🤖 Android — **all green**.

### What I verified (rather than took on faith)
- **Title match across repos is exact.** `ios-wait`'s `TITLE="🍏 ${SAFE_BRANCH} @ ${SHA}"` vs asc-deploy#5's `run-name`: `dispatch-asc-deploy` sends `branch=$(echo "$REF" | tr '/' '-')` and `prepare` computes `safe_branch` with the identical `tr`, and both legs (`ios-build` → `ios_r2_artifact`, `ios-dispatch` reuse path) go through the same composite action with the same `ref`/`sha`. ✅
- **Permissions.** `mobile_distribute.yml` already declares `pull-requests: write` at the top level, so the new `pull-requests: write` in `ios_r2_artifact.yml` is within the caller's grant (it's a `workflow_call` reusable and `mobile_distribute` is its only caller in the repo) — no "requesting X but only allowed Y" failure. ✅
- **`RUN_ID` from a called workflow.** Correct — a reusable workflow shares the caller's `github.run_id`, so the card's stale-run guard stays consistent across `ios-build` and the top-level jobs. ✅
- **Env gates exist.** `env.AAB_READY` and `env.FIREBASE_DIST_OK` are written to `$GITHUB_ENV` earlier in the same job (lines 958/912), so the two new gated card steps aren't dead. ✅
- **`pr-card.py`**: compiles; rendered a card and round-tripped the hidden state locally. The `>` / `--` escaping is applied post-`json.dumps` and both sequences can only occur inside JSON strings, so the output stays valid JSON and decodes back cleanly (incl. `---`/`-->`). Merge-on-read genuinely preserves other legs' fields (`state.update({... if env(...)})` skips empties). ✅

### Findings (P2)

1. **[P2] `ios-wait` is skipped when `ios-build` fails *after* the dispatch → the card can report a failure that actually shipped.** The handoff fires mid-job (line ~308 of `ios_r2_artifact.yml`), but the Sentry dSYM upload runs after it. If anything after the dispatch turns that job red, `ios-wait`'s `if` (`ios-build.result == 'success'`) drops out, `root-failed` writes `IOS_LINE="❌ build/handoff failed"`, and TestFlight gets the build anyway — while Slack (written by asc-deploy itself) correctly shows ✅. Low probability (the Sentry steps are `continue-on-error`), but it's the one place where the card and Slack can disagree in opposite directions. Consider gating on the dispatch step's own outcome instead of the whole job, or having `root-failed` skip the iOS line when a dispatch already happened.

2. **[P2] `STATUS=success` is set by the iOS leg alone.** `ios-wait` exports `STATUS=success`, and no Android step ever sets `STATUS`. So a distribution where Android fails (`ANDROID_LINE="❌ CI build failed"`) and iOS reaches TestFlight renders a header of `✅ SUCCESS` over a visibly-failed Android row. This mirrors the Slack root card, so it's presumably deliberate — but the card is a table where both rows are side by side, which makes the contradiction much more visible than it is in Slack. Cheapest fix: have the Android failure step also set `STATUS=failed`, or derive the header from the two lines at render time in `pr-card.py`.

3. **[P2] "never fails the distribution" isn't structurally guaranteed for `ios-wait`.** `pr-card.py` always exits 0 and the `gh run list` failure is caught, but the step runs under `bash -e` and does unguarded `jq` on `$RUN`; any surprise there turns the job red and the run with it. Since the whole job is decorative, `continue-on-error: true` on the job would make the comment true by construction.

4. **[P2] `ios-wait` holds a hosted runner for up to 50 minutes and makes ~100 `gh run list` calls per PR distribution.** Both are cheap-ish but non-zero, and the poll interval is fixed at 30s for the whole window. A backoff (30s for the first ~5 min, 60–90s after) would cut the API traffic by more than half with no practical latency cost, since the signing leg takes tens of minutes anyway.

5. **[P2] A crafted branch name can reset the card's state.** `STATE_RE` is `search`ed against the whole body and the branch name is rendered verbatim into the table. A branch literally containing `<!-- mobile-build-card:state {` makes the leftmost match start there; the greedy `(\{.*\})` then swallows to the real terminator, `json.loads` throws, and the script silently "starts fresh" — losing the other leg's fields. Self-inflicted-only and cosmetic, but a one-line fix: anchor the state search to the end of the body (`STATE_RE.search(body)` → take the *last* match, or `re.search(..., body[body.rfind(MARKER_STATE):])`), and/or strip `<!--` from `env()` values.

6. **[P2] Only the `Mobile Distribution Triggered` comment is upgraded in place; the old `🤖 Android Build Ready` comment isn't.** On any PR that already got one, the card will sit next to a stale Android comment forever. Adding it to the legacy-marker check (or minimally noting it) closes the loop the PR body claims ("replaces the two comments").

7. **[P2] `TESTFLIGHT_URL` is the generic `https://appstoreconnect.apple.com/apps`** — a link labelled "🍏 TestFlight" that lands on the app list rather than the build. Fine as a placeholder; worth a comment saying so, or dropping the link and keeping the build number as the artifact cell.

8. **[P2] `✅ TestFlight build ${BUILD_NUMBER:-?}` can disagree with the real CFBundleVersion.** asc-deploy falls back to *TestFlight latest+1* when `build_number` is empty; the card prints `?` in that case rather than reading the actual number back off the deploy run. Only reachable when `prepare`'s allocation is empty (fork PRs), where the dispatch would fail first — so it's near-unreachable, just noting the coupling.

9. **[P2 / security, pre-existing]** `prepare` interpolates `${{ github.event.pull_request.head.ref }}` directly into a `run:` block (line ~95). Git branch names can contain shell metacharacters, so on a fork PR this is a classic Actions script-injection surface — and this workflow holds `ASC_DEPLOY_DISPATCH_TOKEN` plus `pull-requests: write`. **Not introduced by this PR** and out of scope, but this diff makes that value flow further (into the card body and the watcher title). Worth a follow-up to move it to `env:` + `"$REF"`. Credit where due: the new code does this correctly — `TITLE`/`SINCE` are passed to `jq` via `env.*` rather than interpolated, and the explanatory comment is there.

### Security review
No P0/P1 security issues.
- `ASC_DEPLOY_DISPATCH_TOKEN` is correctly kept out of `pr-card.py`: the watcher step overrides it inline with `GH_TOKEN="$CARD_TOKEN"` for the card write and keeps the PAT for `gh run list` only. ✅
- No secret is echoed; `cat gh.err` on the failure path surfaces a `gh` error message, which doesn't contain the token.
- Fork PRs: the read-only `GITHUB_TOKEN` yields a 403 on the comment write, which the script catches and exits 0 on. ✅
- The new `pull-requests: write` on `ios_r2_artifact.yml` is a genuine (small) widening of that reusable workflow's token — justified, and it's only reachable from `mobile_distribute`, which already had it.

### Consumer impact
Breaking for anything scraping the old PR comments: the `### 📱 Mobile Distribution Triggered` and `### 🤖 Android Build Ready` bodies are gone. `REVIEW.md` is updated accordingly (good catch — that's exactly the kind of doc that silently rots). I found no other in-repo consumer of those strings; anything external (a bot or dashboard grepping PR comments) is outside what I can enumerate, so treat that as unverified.

### Notes
- ~750-line diff, but ~500 of it is repetitive card steps; I read `pr-card.py` and `ios-wait` line by line and skimmed the mechanical per-leg steps after confirming the env-gate pattern.
- Nice call on the `SHA` + `RUN_ID` double guard — the re-label-on-the-same-commit case is the one most people miss.
- Blocked on decentraland/godot-asc-deploy#5 landing on `main` first, as the body says; until then the iOS row correctly degrades to "❓ TestFlight result unknown".

---
*Reviewed at `73c1a78`. Requested by Mateo via Slack.*

.github/workflows/mobile_distribute.yml
```diff
@@ -420,11 +439,108 @@ jobs:
...
+      # Same lines asc-deploy writes to the Slack root card; STATUS flips the header exactly like Slack.
+      - name: PR card → TestFlight result
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CONCLUSION: ${{ steps.watch.outputs.conclusion }}
+          URL: ${{ steps.watch.outputs.url }}
+        run: |
+          case "$CONCLUSION" in
+            success)
+              export STATUS=success IOS_LINE="✅ TestFlight build ${BUILD_NUMBER:-?}" \
```
leanmendoza, line 534: AG-3 — `STATUS=success` flips the shared header without looking at the Android leg, and no Android step ever sets `STATUS`. If the Android rebuild fails (line 934 sets `ANDROID_LINE` but no `STATUS`) and iOS then succeeds, the card renders `✅ SUCCESS` directly above `🤖 Android | ❌ CI build failed` — a green header over a distribution with no APK, on a card REVIEW.md now makes a pre-flight gate.

Mirror case: the `*)` branch below exports no `STATUS` at all, so the header stays `⏳ BUILDING` forever once every leg has finished.

Could the header be derived from both rows instead of written by whichever leg happens to finish last?

.github/workflows/mobile_distribute.yml
```diff
@@ -420,11 +439,108 @@ jobs:
...
+            sleep 30
+          done
+          echo "→ conclusion=${CONCLUSION}"
+          {
+            echo "conclusion=$CONCLUSION"
+            echo "url=$URL"
+          } >> "$GITHUB_OUTPUT"
+
+      # Same lines asc-deploy writes to the Slack root card; STATUS flips the header exactly like Slack.
+      - name: PR card → TestFlight result
```
leanmendoza, line 525: AG-3 — No `if:` here, so the step inherits `success()` and is skipped whenever `watch` exits non-zero. `watch` runs under `bash -e`, and by then the card already reads "🔏 signing & uploading to TestFlight…" — so a `jq` failure on a malformed `$RUN`, or `timeout-minutes: 55` firing before the in-script 50-min `DEADLINE` (5 min of slack, minus queue and checkout), freezes the iOS row at "signing…" and turns the check red. The job header says "never fails the distribution".

`if: always()` covers both.

REVIEW.md
```diff
@@ -13,7 +13,7 @@
 These are blocking prerequisites. Resolve each one (or explicitly note its status in your review) before reading the diff.
 
 1. **Branch must be up to date with `main`.** Run `gh pr view <pr> --json mergeStateStatus,headRefOid,baseRefOid` (or check the PR page). If the PR is behind `main`, **request the author update the branch before a substantive review** — CI signals (especially Android/iOS artifacts) are not trustworthy on a stale base, and recently-merged fixes (autoload ordering, mouse-filter changes, skeleton recycling) frequently invalidate older diffs. A one-line "please rebase / merge `main` and I'll re-review" is the right output if the branch is stale.
-2. **iOS build must be present for platform-sensitive changes.** Mobile builds are gated on the `build` label (alias: `build-ios`) and skipped by default. Check the PR's checks/comments for a `🍏 iOS` artifact in the sticky build report.
+2. **iOS build must be present for platform-sensitive changes.** Mobile builds are gated on the `build` label (alias: `build-ios`) and skipped by default. Check the PR's **📱 Mobile build pipeline** comment (posted by the `build` label) for a `🍏 iOS` row reading "✅ TestFlight build N".
```
leanmendoza: AG-3 — This asks reviewers to look for a `🍏 iOS` row reading "✅ TestFlight build N", but `ios-wait` can't produce that string until decentraland/godot-asc-deploy#5 adds the `run-name`. Until then `displayTitle` never matches, the watch burns its full 50 min and the row reads "❓ TestFlight result unknown" — so a reviewer following this step holds a PR whose iOS build actually succeeded, and §3.18 tells them to output the hold message verbatim.

Worth landing this line after asc-deploy#5, or accepting "❓ unknown" until then.

leanmendoza (commented):
AG-3 — Only the three things I'd hold the merge on are inline. The rest of what I found on the card (the read-modify-write race in `pr-card.py`, the incomplete `LEGACY_MARKER`, the `failure()` attribution in `ios_r2_artifact.yml`) is non-blocking, so I'm leaving it off the PR.

leanmendoza:
AG-3 — Pushed b5763c49 onto this branch with the three fixes instead of leaving them as review comments, since they're small. Revert it if you'd rather take them another way.

- **Header can't go green over a red row.** The Android failure path now writes `STATUS=failed`, and `pr-card.py` refuses to regress a `failed`/`cancelled` header back to `success` — so either arrival order lands on the failure.
- **`PR card → TestFlight result` runs with `if: always()`**, so an error in `watch` (bad `jq` parse, or `timeout-minutes: 55` beating the in-script 50-min deadline) can't strand the iOS row at "🔏 signing…".
- **REVIEW.md §0.2** no longer reads "❓ unknown" as "no iOS build" — that row stays reachable on any transient `gh` failure in the watch loop even with asc-deploy#5 merged.

Checked by stubbing the GitHub API and asserting the rendered header for both arrival orders, plus a control that an all-green run still goes green.

decentraland/godot-asc-deploy#5 is approved, so the `run-name` side is covered — the two remaining non-blocking things I'd still look at eventually are the read-modify-write race in `pr-card.py` (two legs PATCHing at once can drop a write-once field like `APK_URL`) and the `LEGACY_MARKER` only matching one of the four older comment shapes.

leanmendoza (approved):

PR #2896, opened 2026-09-14
feat: add Figma skill

### LINKED ISSUE #2895: Feat: Add Figma skill
Right now Claude ignores Figma links, and doesn't know how to ingest design information from there.

Create a skill with instructions about how to do that.

EibrielInv:
## What

  Adds a `figma` skill so Claude can work from a Figma design instead of ignoring the link. It
  covers getting access to the design, reading it, mapping it onto this repo's UI conventions, and
  leaving a link behind that the next person can actually follow.

  Documentation only — one file under `.claude/skills/`. No runtime code, no CI dependency.

  ## Why

  Claude had no way to open a design, and no convention for recording one. Asked to look at a new
  icon, the only honest answer was "I can't" — the Figma MCP plugin has to be installed *and*
  authorized, both of which need a human, and nothing in the repo said so.

  The other half is that nothing gets written down afterwards. There are 12 Figma references in
  the codebase and **11 cannot be resolved** — frame names with no node-id, a node-id with no file,
  a file named only by its title. Exactly one, in `joypad.tscn`, carries a real file key, and no
  `figma.com` URL exists anywhere in the tree. So every design question restarts from zero and
  ends at a human.

  Closes #2895

  ## Details

  <details>
  <summary>Expand</summary>                                                                                                                                                                                  

  Structure follows the four repo-authored skills — `name` + `description` frontmatter, `##`
  headings, an anti-patterns section, a verification checklist, cross-references in bold prose.
  268 lines, in line with `i18n` (198) and `pr-description` (269).

  **The steps, in the order a blocked session needs them:**

  0. **The plugin.** A dispatch table on which Figma tools are actually visible, then the verified
     OAuth flow — `authenticate` → user opens the URL → the localhost redirect *will* look broken →
     `complete_authentication` with the address-bar URL. States plainly that the agent cannot do
     either step alone.
  1. **Get the link** — ask for the file *and* the frame; never assume an orphan node-id belongs to
     some other comment's file.
  2. **Read the design** — defers MCP mechanics to the plugin's own skills, lists what to extract in
     one pass so a second round trip doesn't cost another hand-off.
  3. **Decompose onto the tiers** — atoms / molecules / organisms / layouts / pages, with the Figma
     analogue for each, "prefer the lower tier", and the `COMPONENT_AUDIT.md` check.
  4. **Map the properties** — auto-layout → containers, colours → normalized `Color()` literals
     (there is no token layer), StyleBoxes, `theme_type_variation`, and the "which theme" question.
  5. **Icons** — square SVG, where it goes, `base_scale`, and `import-assets` before referencing,
     since `.tscn` refs are `uid://`-keyed.
  6. **Leave the reference behind** — a resolvable URL form in `editor_description` or a `##`
     docstring.
  7. **Prove it matches** — a table comparing Figma's numbers against values *measured* in the
     running app, every row marked match / deferred / miss.

  **Two rules the skill leads with, both from real failures while writing it:**

  - **Screenshots are never a source of truth, in either direction.** A 50%-black pill over a bright
    orange object reads as a grey plate that had already been removed; acting on that would have
    meant "fixing" a bug that did not exist. Measure the live tree via the **mobile-dev-debug-tool**
    skill instead.
  - **Verify property by property before calling it done.** "Looks right" shipped a white label beside
    a lavender icon and a 30px icon where the design said 44.

  Cross-references **godot-ui-components** for tier placement, **i18n** for any string in a mock, and
  **mobile-dev-debug-tool** for the live-tree measurement, rather than restating them.

  **Verified by the author:**

  - [x] Skill loads and the frontmatter parses; one H1, no heading deeper than `##`
  - [x] Dogfooded end-to-end on a real task — plugin authorize flow, reading the HUD-Revamp frame,
        exporting an icon, and catching a design mismatch the comparison table was written for
  - [x] `gdlint godot/` and `python3 tools/i18n/extract_strings.py --check` unaffected

  </details>

  ## Test plan
                                                                                                                                                                                                             
  No QA needed — no behavior change. The change is a single Markdown file under `.claude/skills/`;
  nothing in the client reads it and no CI workflow references `.claude/`.

  - [x] Static checks unaffected: `gdlint godot/`, `extract_strings.py --check`

  Reviewer check, if you want one: open `.claude/skills/figma/SKILL.md` and confirm the plugin
  authorize steps match what you get from `/plugin` + the Figma OAuth flow on your machine.


leanmendoza:
Moved to draft. Move back to RFR when you consider we have to review it :)

Issue #2471, opened 2026-07-08
Dismiss the loading screen and show an error when realm resolution fails


leanmendoza:
On the main entry points (`async_teleport_to`, `async_join_world`, startup) a failed `/about` emits `realm_change_failed`, which is only a toast (`global.gd`), and never dismisses the loading screen. Only the chat command path hides on failure. The only backstop is the 90s wall clock modal, so it feels infinite.

Fix: connect realm resolution failure to hiding the loading screen plus a blocking retry UI on all entry paths, not just chat commands. This is the largest no destination hang class. Failure mode FM-4.

leanmendoza:
Consolidated into #2536 — item **J)**. Closing in favor of the single tracker; the work continues there.

PR #2897, opened 2026-09-14
feat: support TeleportToRequest.realm in scene teleportTo

### LINKED ISSUE #2816: Support TeleportToRequest.realm in scene teleportTo
##  **⚠️ Problem:**
Scene `teleportTo` only carries parcel coordinates, so a scene inside a World that sends a player to a Genesis City place lands on those coordinates *inside the World*. Scenes must fire `changeRealm` + `teleportTo` separately, and `changeRealm` resolves on acceptance — not when the realm is live — so the teleport races the realm swap.

## **🏁 Scope:**
Support `TeleportToRequest.realm` (optional, protocol#477): world name (`foo.dcl.eth`) or realm url. Set → change realm (full reconnect, even to the current realm) then land on the parcel. Omitted → today's behavior unchanged.

## **📝 Deliverables:**
- [ ] `realm` field read in `op_teleport_to` + `handle_restricted_actions` teleport path
- [ ] Realm change completes before parcel placement (no race)
- [ ] Empty/absent `realm` keeps current behavior
- [ ] Invalid realm surfaces the existing realm-resolution error path, no stuck loading screen

## **🔗 References:**
- [decentraland/protocol#477](https://github.com/decentraland/protocol/pull/477) — adds `optional string realm = 2` to `TeleportToRequest`; blocker, merge first
- [decentraland/godot-explorer#922](https://github.com/decentraland/godot-explorer/issues/922) — same class of bug on the client side (Jump In from a World kept the old realm), closed; SDK-facing path still lacks it
- [decentraland/godot-explorer#2471](https://github.com/decentraland/godot-explorer/issues/2471) — realm resolution failure handling the invalid-realm case should reuse

## Dependency
Blocked by protocol#477. Reference implementation: bevy-explorer (`experimental-bevy`, where the field was cherry-picked from). No unity-explorer or docs prior art for the field (searched "teleport realm" / "teleportTo realm", 2026-09-02).

## Notes
Client already has `Global.async_teleport_to(parcel, realm)` (used by Discover, deep links, jump-in popup) — the gap is the SDK op, not the navigation layer.

**Requested by Mateo via Slack**

EibrielInv:
## What
A scene's `teleportTo` can now name the realm its parcel belongs to — a world name
(`foo.dcl.eth`) or a realm url. The client changes realm first and only then places the player
on the parcel. With no coordinates and a realm, the player lands on that realm's own spawn
point. Calls that pass only coordinates behave exactly as before.

## Why
`teleportTo` only carried coordinates, so a scene inside a World sending a player to a Genesis
City place landed them on those coordinates *inside the World*. The workaround — `changeRealm`
followed by `teleportTo` — races, because `changeRealm` resolves when the user accepts, not when
the new realm is live. This implements `TeleportToRequest.realm` from
[protocol#477](https://github.com/decentraland/protocol/pull/477) (merged), which also deprecates
the `ChangeRealm` RPC in favour of this.

The travel confirmation modal still appears, and a realm that resolves to the one the player is
already in does not force a reconnect.

Closes #2816

## Details
<details>
<summary>Expand</summary>

The navigation layer already supported all of this (`Global.async_teleport_to`,
`Global.async_join_world`, `explorer._async_teleport_to`, `Realm.async_set_realm`); the realm was
simply being dropped between the SDK op and the modal. The change is plumbing:

- `RestrictedActions.js` / `op_teleport_to` / `RpcCall::TeleportTo` carry both now-optional fields
  (sentinels across the op boundary, since `#[op2]` params can't be `Option` — same idiom as
  `movePlayerTo`'s NaNs). A call with neither coordinates nor realm rejects the scene's promise.
- `handle_restricted_actions::teleport_to` routes coordinates (+ optional realm) to the existing
  `async_show_teleport_modal(parcel, realm)`, and realm-only to a new
  `async_show_realm_teleport_modal` whose primary action is `Global.async_join_world` — the
  existing "change realm and land on its spawn" path.
- Ordering is guaranteed by `explorer._async_teleport_to`, which awaits `Realm.async_set_realm`
  before moving the player.

Ride-along fixes (the first three found while testing on device, the rest from review):

- **Travel modal stuck loading for a parcel outside Genesis.** `_async_load_travel_modal_data`
  looked every parcel up in the Genesis places index and returned silently when it found
  nothing — leaving the modal in its loading skeleton with JUMP IN permanently disabled. A
  parcel inside a World is never in that index. The teleport modal now looks worlds up by name
  (reusing `_async_load_change_realm_data`), and the by-position loader always names the place,
  falling back to the coordinates.
- **Stuck loading screen on a failed realm change (#2471).** `loading_screen.gd` now hides
  itself on `Realm.realm_change_failed`. This covers every caller that doesn't await
  `async_set_realm` (the change-realm modal, Settings, `async_join_world`), not just this path.
- **SDK6 `requestTeleport` was broken.** `UserActionModule.js` called `op_teleport_to([x, y])`
  with a single array instead of positional arguments, so the op never matched. Its `'magic'` /
  `'crowd'` destinations now reject explicitly — they are a kernel-era convention this client has
  no lookup for, and the old broken call errored rather than teleporting anywhere.
- **A failed realm change no longer claims success.** `hide_loading_screen()` takes the
  LOADING_END status (so a failed change is recorded `"Failed"`, not `"Success"`), and
  `Global.async_teleport_to` waits for `explorer.async_teleport_to` before writing the
  `🟢 Teleported to (x, y)` line to chat.
- **Realm comparison is now on resolved urls.** New `Realm.normalize_realm_url()`, reused by
  `async_set_realm`, so a scene passing `spacerunner.dcl.eth` does not reconnect a player already
  in `SpaceRunner.dcl.eth`. The realm string is arbitrary scene input now.

Verified locally:
- [x] `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (4 new tests cover the op's
      argument decoding, including the request that carries neither field)
- [x] `gdformat` / `gdlint` on all touched GDScript
- [x] Installed on a Pixel 7 and run against the test scene below

</details>

## Test plan

**Setup:** the SDK7 test scene with the **Issue 2816** tab (test switcher bar along the top,
scroll right to find it). Tapping the tab opens a panel on the right with one button per case;
the panel header shows the target world and parcel. Leave the world selector on its default,
**SpaceRunner.dcl.eth** — a 6×6 world (parcels `0,0`–`5,5`) whose spawn point sits high above
parcel ~`2,3`, so "the parcel I asked for" and "the world's own spawn" land in visibly different
places.

Every teleport leaves the test scene, so **return to the test scene and re-open the Issue 2816
tab between cases.**

**Case 1 — parcel in another realm**
2. Tap the **Issue 2816** tab.
3. Tap **Parcel + world realm**.
4. Accept the travel modal.

- [ ] The modal fills in (place name and JUMP IN enabled) — it never sits on the grey skeleton
- [ ] The loading screen names the destination world, not the place you are leaving
- [ ] Chat shows the system message `🟢 Teleported to (5, 5)`
- [ ] You arrive inside SpaceRunner's content on parcel `5,5`, at ground level — not at the
      elevated spawn point, and not on `5,5` in Genesis City

**Case 2 — realm with no coordinates**
1. Return to the test scene and tap the **Issue 2816** tab.
2. Tap **World spawn (no coordinates)**.
3. Accept the travel modal.

- [ ] Chat shows the system message `Trying to change to world SpaceRunner.dcl.eth`
- [ ] The player lands at SpaceRunner's own spawn point — high up, a clearly different place
      from where Case 1 ended

**Case 3 — unreachable realm**
1. Return to the test scene and tap the **Issue 2816** tab.
2. Tap **Invalid realm (must not hang)** — it aims at a randomly named world that does not exist.
3. Accept the travel modal.

- [ ] An error toast appears
- [ ] The loading screen disappears — it must not hang over the scene
- [ ] The player is left where they were and can still walk around

**Case 4 — unchanged behaviour**
1. Return to the test scene and tap the **Issue 2816** tab.
2. Tap **Parcel only (unchanged path)**.
3. Accept the travel modal.

- [ ] Behaves as it does on `main`: modal, then a teleport to `0,0` within the current realm

**Case 5 — meaningless request**
1. Return to the test scene and tap the **Issue 2816** tab.
2. Tap **No coordinates, no realm (must reject)**.

- [ ] No modal, no teleport, nothing happens on screen
- [ ] The panel's `result:` line reads `rejected: ...`

**Not covered here:** the reverse direction — a scene *inside* a World teleporting the player
out to Genesis City — needs the test scene deployed to a World, which QA cannot do. The author
verifies that case directly.

- [ ] **Regression** — from Discover, jump into a place and then into a World; change realm from
      Settings; open a travel modal and cancel it. Each loads normally, every modal fills in
      rather than sitting on the grey skeleton, and no loading screen is left hanging.


kuruk-mm (approved):
looks great!

godot/src/ui/pages/loading_screen/loading_screen.gd
```diff
@@ -68,6 +73,10 @@ func enable_loading_screen(intended_realm: String = "", when: String = "") -> vo
 		Global.metrics.track_screen_viewed("LOADING_START", JSON.stringify(loading_data))
 
 
+func _on_realm_change_failed(_new_realm_string: String, _reason: String) -> void:
+	hide_loading_screen()
```
regenesis-claw: [P1] This runs the full hide effect even when the loading screen was never shown. `_hide_loading_screen()` → `async_hide_loading_screen_effect()` has no visibility check, so for every `async_set_realm` caller that doesn't put the screen up first (change-realm modal, Settings, `_on_scene_crash_reload`) a failure now emits `Global.loading_finished` — which makes `scene_fetcher.on_loading_finished()` move the player to the scene spawn point — closes navbar/chat, can trigger the upgrade nudge, and tracks `LOADING_END` with `status: "Success"` plus `loading_end_episode("hidden")` for a change that failed.

```suggestion
func _on_realm_change_failed(_new_realm_string: String, _reason: String) -> void:
	# Only the callers that put this screen up (async_join_world / async_teleport_to) need
	# taking down. Hiding when it was never shown would emit loading_finished — which moves
	# the player to the scene spawn — and record a Success LOADING_END for a failed change.
	if not visible:
		return
	hide_loading_screen()
```

godot/src/ui/components/organisms/modal/modal_manager.gd
```diff
@@ -996,25 +1025,34 @@ func _async_load_travel_modal_data(location: Vector2i, _realm: String) -> void:
...
 	var result = await PlacesHelper.async_get_by_position(location)
 
-	if result is PromiseError:
-		printerr("Error requesting place data for travel modal", result.get_error())
+	if not is_instance_valid(current_travel_modal):
 		return
 
-	if not is_instance_valid(current_travel_modal):
+	if result is PromiseError:
+		printerr("Error requesting place data for travel modal", result.get_error())
```
regenesis-claw: [P2] Now that this branch has a defined fallback (`set_place_name(fallback_name)` right below), the condition is expected-and-handled, so `printerr` ships a Sentry event for something the client recovers from — REVIEW.md §14 asks for `push_warning`/`print` here. A parcel outside the Genesis index (exactly the #2816 case) will hit this routinely.

```suggestion
		print("Place data unavailable for travel modal, using coordinates: ", result.get_error())
```

regenesis-claw (changes requested):
Reviewed against `REVIEW.md` (Tier 1–3) + a security pass. The plumbing itself (op sentinels → `RpcCall::TeleportTo { world_coordinates: Option, realm: Option }` → the two modal entry points) is clean and backward-compatible: the two existing `async_show_teleport_modal` callers (`chat_message.gd`, `menu.gd`) pass no realm and take the unchanged by-position path, and `(None, None)` is rejected at both the op and the handler so the scene promise can never hang. No new user-facing literals, no debug prints, no non-English text, no `.claude/`/`project.godot` leakage. CI is fully green (Linux, Android, iOS export, Docker, clippy, static checks) and the Android APK + TestFlight distribution both ran on `860bc2c`.

One blocker, in the ride-along #2471 fix.

### P1 — `realm_change_failed` runs the full hide effect even when no loading screen is up

`loading_screen.gd:77` calls `hide_loading_screen()` unconditionally, and that chain (`loading_screen_progress_logic._hide_loading_screen()` → `async_hide_loading_screen_effect()`) has no visibility guard of its own. Every `async_set_realm` caller that does **not** put the loading screen up first — the change-realm modal (`_on_change_realm_primary`), Settings, `_on_scene_crash_reload` — now runs the whole teardown on failure while the player is walking around:

- `Global.loading_finished.emit()` → `scene_fetcher.on_loading_finished()` **moves the player to the current scene's spawn point** (`Global.get_explorer().move_to(...)`). A failed realm change now yanks the player out of where they were standing.
- `Global.close_navbar` / `Global.close_chat` fire, closing UI the user had open.
- `upgrade_nudge_coordinator._on_loading_finished()` can pop the upgrade nudge at an arbitrary moment.
- `LOADING_END` is tracked with `"status": "Success"` and the funnel episode is closed as `"hidden"` — for a realm change that just failed, right after `realm.gd` already emitted `loading_realm_change_failed`. That inflates exactly the LOADING_END-vs-SCENE_MOVEMENT skew the loading funnel is watched for.

(`release_comms()` is safe — it early-returns when not held.)

Inline suggestion below: early-return on `not visible`. The paths the description lists as newly covered (`async_join_world`, `async_teleport_to`) enable the loading screen first, so they keep working with the guard in place.

### P2 — worth addressing

1. **Failed realm change is still recorded as a successful load** even in the visible case. `_hide_loading_screen()` is the success path (`status: "Success"`, `loading_end_episode("hidden")`). Consider a failure-aware close (the cancel path in `_on_close_button_pressed` already sets its own `loading_end_episode("cancelled")` before tearing down) so the funnel can tell a failed realm change from a completed load.
2. **`printerr` for a now-handled condition** — see inline on `modal_manager.gd:1039`. REVIEW.md §14: a lookup that has a defined fallback is `push_warning`/`print`, not a Sentry event.
3. **"a realm that matches the one the player is already in does not force a reconnect" is only true for byte-identical strings.** `explorer._async_teleport_to` compares `realm != Global.realm.get_realm_string()` raw, while `async_set_realm` normalises through `Realm.ensure_ends_with_slash(Realm.resolve_realm_url(...))`. A scene passing `spacerunner.dcl.eth` (or the worlds-content-server URL form) while the player is already in `SpaceRunner.dcl.eth` will reconnect and reload the world. Comparing the normalised URLs on both sides makes the claim hold for scene-supplied strings, which are now arbitrary input. Pre-existing function, newly reachable from scenes.
4. **Coordinates + unreachable realm shows a false success line in chat.** `Global.async_teleport_to` emits `🟢 Teleported to (x, y)` before `explorer._async_teleport_to` awaits `async_set_realm`, so that path prints the success message, then the realm change fails and the error toast appears. The test plan's Case 3 only exercises the realm-only (`async_join_world`) branch — adding a parcel + invalid realm case would have caught this.
5. **The consent modal never names the destination realm.** For coordinates + a realm that is not a `.dcl.eth` world (a raw catalyst URL), `world_name_from_realm` returns `""`, so the modal shows the *Genesis* place title/image for those coordinates while JUMP IN sends the player to the scene-chosen catalyst. Not a new capability (the `ChangeRealm` RPC could already do this, and it's user-gated), but the modal used to describe where the user was actually going. Showing the realm/world on the modal for any non-current realm would close the gap.
6. **No automated coverage for the new op contract.** The `has_coordinates` / empty-realm sentinel decoding and the `(None, None)` rejection in `op_teleport_to` are plain Rust and unit-testable even though the handler arms are Godot-bound; `cargo test` is the only test signal CI runs here.
7. **SDK6 `requestTeleport('magic' | 'crowd')` now silently means parcel `0,0`.** The fix to the positional-args bug is right, but this branch previously errored out rather than teleporting; landing everyone on `0,0` is a product decision worth a comment (or an explicit rejection until a random/crowded-place lookup exists).

### Security

No secrets, no injection, no auth/authz change. The scene-supplied `realm` string reaches `Realm.async_set_realm` → `candidate_realm_url + "about"`, i.e. a scene can cause a fetch to an arbitrary host — but only behind the user-accepted travel modal, and the `ChangeRealm` RPC already offered the same reach. Private-world access is still pre-checked (`_async_precheck_realm_access` / `WorldPermissionsHelper`) on both new entry points. Only finding is the consent-clarity one (P2 #5).

### Not verified

The on-device Pixel 7 run and the 5-case test plan are author-reported; I checked the code paths, not the device behaviour.

godot/src/ui/pages/loading_screen/loading_screen.gd
EibrielInv: Fixed in fb0fe574 — took the suggestion, with the `if not visible: return` guard and a comment naming the reason (the `loading_finished` → `scene_fetcher` spawn-point `move_to` is the damaging one).

Confirmed the trace: nothing in `hide_loading_screen()` → `_hide_loading_screen()` → `async_hide_loading_screen_effect()` guards on visibility, and `scene_fetcher.on_loading_finished()` does call `explorer.move_to(spawn)`. The two paths this handler exists for raise the screen first, so they are unaffected.

godot/src/ui/components/organisms/modal/modal_manager.gd
EibrielInv: Fixed in fb0fe574 — downgraded to `print`, with a comment noting the fallback below it and that a parcel outside the Genesis index hits this routinely.

EibrielInv:
Thanks — every finding checked out against the code. All of them are addressed in fb0fe574 except P2.5, which I'm deferring with reasoning below.

**P1 — hide with no loading screen up.** Fixed, suggestion taken. Confirmed the whole chain has no visibility guard and that `scene_fetcher.on_loading_finished()` really does `explorer.move_to(spawn)` — that was the worst of it.

**P2.1 — failed change logged as a successful load.** `hide_loading_screen(status := "Success")` now threads the status into `_hide_loading_screen()` and on into `end_data`; the realm-failure handler passes `"Failed"`, matching the vocabulary `scene_fetcher.send_scene_failed_metrics()` already uses. `chat_commands._async_try_change_realm` passes it too — same class of event.

**P2.2 — `printerr` with a fallback.** Downgraded to `print`.

**P2.3 — raw-string realm comparison.** Added `Realm.normalize_realm_url()` (the three existing helpers in one place) and reused it both in `async_set_realm` for `candidate_realm_url` and in the teleport comparison, now against `get_realm_url()`. `spacerunner.dcl.eth` no longer reconnects a player already in `SpaceRunner.dcl.eth`. The PR description's claim is worded as "resolves to" now.

**P2.4 — false `🟢 Teleported to` line.** `explorer._async_teleport_to` is now the public `async_teleport_to` returning `bool`, and `Global.async_teleport_to` only emits the chat line when it returns true. `teleport_to()` stays as the fire-and-forget wrapper for the two callers that can't await. Net-zero lines in `global.gd`, which is sitting exactly on gdlint's 1900-line cap.

**P2.6 — no coverage for the op contract.** Extracted `decode_teleport_args()` as a pure function and added four tests: coordinates only, coordinates + realm, realm only (asserting the x/y sentinels are dropped rather than becoming `0,0`), and neither field → `Err`.

**P2.7 — SDK6 `magic` / `crowd`.** Rejecting explicitly. The proto defines only a free-form `destination`, neither value appears in the protocol or the SDK toolchain, and we have no random/crowded-place lookup — so landing on `0,0` would have been inventing a destination. The old call errored, and it still does, now with a message that says why. The positional-args bug stays fixed for normal `"x,y"` destinations.

**P2.5 — modal doesn't name a non-world destination realm.** Deferring, and I think that's the right call for this PR rather than a disagreement: the modal has no realm slot, so closing the gap means a new `TravelModal` line — new scene node, new translation key — in a component Discover, jump-in, deeplinks and the change-realm flow all render. As you note it isn't a new capability (`ChangeRealm` already had the same reach behind the same user-accepted modal). I'd rather not land shared-modal UI in a PR already carrying several ride-along fixes; I'll open a follow-up issue covering both entry points.

**Not verified:** the device re-run against the new build. The five test-plan cases and P1 were verified on a Pixel 7 on `860bc2c`; I'll re-run them on this commit before merge.

regenesis-claw (approved):
Re-reviewed the delta `a1c4b18...fb0fe57` (one `fix: address review on teleportTo realm` commit after a `main` merge). Every round-1 finding is addressed, several of them more thoroughly than asked. Approving — one new P2 below, non-blocking.

### Verified fixed

- **P1 (unguarded hide)** — `loading_screen._on_realm_change_failed` now early-returns on `not visible`, so the callers that never raised the screen (change-realm modal, Settings, scene-crash reload) no longer emit `Global.loading_finished` → `scene_fetcher.on_loading_finished()` → player moved to spawn, and no longer close navbar/chat or record a phantom load.
- **P2-1 (failure recorded as success)** — solved properly rather than papered over: a `status` parameter threaded through `loading_screen.hide_loading_screen()` → `loading_screen_progress_logic.hide_loading_screen()` → `_hide_loading_screen()` with a `"Success"` default, so the 4 pre-existing call sites are untouched and the two failure sites pass `"Failed"` — the same vocabulary `scene_fetcher.gd:1173` already uses.
- **P2-2 (Sentry noise)** — `printerr` → `print` with a comment citing §14 and the fallback below it.
- **P2-3 (realm equality)** — new `Realm.normalize_realm_url()`, reused by `async_set_realm` itself (so the comparison is against the exact string it stores) and compared against `get_realm_url()` rather than `get_realm_string()`. Checked the docstring's claim: `dcl_world_url()` lowercases the name and `ensure_ends_with_slash`/`ensure_starts_with_https` handle the rest, so `spacerunner.dcl.eth` / `SpaceRunner.dcl.eth` / the worlds-content-server URL form really do collapse to one string.
- **P2-4 (false chat success)** — `explorer.async_teleport_to()` now returns a bool and `Global.async_teleport_to` gates the `Teleported to (x, y)` line on it; `teleport_to()` stays as the documented fire-and-forget wrapper, and its one remaining caller (`realm.gd:406`) passes no realm so the bool is irrelevant there. The rename also satisfies gdlint's `async-function-name`, and no `_async_teleport_to` references were left behind.
- **P2-6 (test coverage)** — `decode_teleport_args` extracted out of the `#[op2]` body and covered by 4 unit tests, including the "sentinels ignored when there are no coordinates" case (the one that would silently become `0,0`) and the reject case. This is real `cargo test` signal, which is the only test tier CI runs here.
- **P2-7 (SDK6 `magic`/`crowd`)** — now rejects with an explanatory message instead of substituting `0,0`. Good call, and the comment explains the kernel-era origin.

### P2 — new, introduced by the fix

`chat_commands.gd:160` is now a **double teardown**. `_emit_realm_change_failed` emits the signal synchronously inside `async_set_realm` before it returns `false`, so the new handler already hid the screen by the time `_async_try_change_realm` resumes — and `visible` is still `true` at that point, because `async_hide_loading_screen_effect()` only calls `hide()` after its 1 s fade. So the explicit call runs the whole path a second time: a second `Global.loading_finished` emit (another `scene_fetcher` spawn move), navbar/chat closed twice, two overlapping `modulate` tweens, a duplicate `LOADING_END "Failed"` event and `loading_end_episode("hidden")` twice — which puts a different distortion into the funnel than the one this PR just fixed.

Two ways out, either is fine:
- drop the now-redundant `hide_loading_screen("Failed")` (the signal handler covers every `async_set_realm` failure path), or
- hoist the `visible` check into `hide_loading_screen()` itself, which makes all callers idempotent — `_force_hide_loading_screen()` in `modal_manager.gd:1235` already does exactly that check before calling in.

Related, if you touch that area: the double-tap-logo escape hatch (`loading_screen.gd:236`) and `_force_hide_loading_screen` still report `"Success"`. The first is a dev gesture and the second is the "run anyway" timeout path, so neither is clearly wrong — just noting it since the status vocabulary is new.

### Still open from round 1 (fine to leave)

P2-5: with coordinates + a realm that is not a `.dcl.eth` world, the travel modal still shows the *Genesis* place name/image for those coordinates while JUMP IN goes to the scene-named catalyst. Reading it as a design call, not a merge blocker — the `ChangeRealm` RPC always had the same reach.

### CI

Green on `fb0fe57`: Linux (incl. the new unit tests), Android, Docker, clippy, static checks. No iOS export or TestFlight run on this head — mobile distribution wasn't re-triggered after the fix commit. The delta is GDScript + Rust with nothing platform-specific, and Android built clean, so I'd call that low risk; worth a `build` label if you want a TestFlight cut for QA on the final head.

The 5-case test plan still doesn't cover parcel + unreachable realm — now that the chat line is gated on the result, that case is the one that proves it.

PR #2898, opened 2026-09-14
chore(ios): export App Store builds as release

kuruk-mm:
## What

App Store / TestFlight iOS builds are exported with the **release** engine template instead of the debug one. CI's Export step now passes `--release`; local dev exports are unchanged. Also drops a comment in `iap_manager.gd` that described the old behaviour.

## Why

Every iOS Sentry event reports `build_type: debug` / `mode: export_debug`: CI never passed `--release`, so the store binary has always embedded the debug template (found while investigating #2891). The shader-bake fix from the same investigation ships separately in #2886 and does not depend on this.

**Behaviour change to be aware of.** The debug template validates every GDScript call on a freed node and turns it into a logged error; the release template does not, so on iOS those calls now crash exactly as they do on Android (largest Android crash family, `GODOT-EXPLORER-21E`). Team decision 2026-09-15: keep the engine's release behaviour (an engine-side guard was considered in decentraland/godotengine#21 and dropped) and fix the GDScript instead — a lint rule plus per-area fixes are in progress and tracked separately. This PR no longer depends on an engine pin.

Side effects, both positive: store builds stop embedding the dev-only plist keys (`NSLocalNetworkUsageDescription`, `NSAllowsLocalNetworking`) and the `--scene-inspector=ws://…` launch argument that `dcl-ios-devtools` injects into debug exports (visible today in every iOS Sentry event's `command_line_arguments`); `MemoryDebugger` and the other `OS.is_debug_build()` gates finally turn off in production.

Debug symbols are unaffected: the Rust dylib upload is a separate build step and the app dSYM comes from godot-asc-deploy's archive, which is already built as Release; both templates carry DWARF.

Part of #2891

### Case 1 — Store build is a release build
**Steps:**
1. Open the app, sign in, enter Genesis Plaza, then teleport once and open the backpack.
- [ ] **Expected (Sentry, same day):** iOS events for this build carry `build_type: release` / `mode: export_release` and an empty `command_line_arguments`.

### Case 2 — Regression: the flows behind the Android freed-node crashes
**Steps:**
1. Open the app, enter Genesis Plaza, open a friend's passport, then teleport to another scene while the passport is still open.
2. Open the backpack, equip a wearable, and teleport again while the backpack is still open.
3. Play an emote from the wheel and teleport mid-animation.
4. Open Settings, sign out, sign back in as a guest, open Settings again.
- [ ] **Expected:** no crash in these flows; the app lands in each scene normally.
- [ ] **Expected (Sentry, next day):** any new iOS crash issue whose top frame is `CanvasItem::show`, `Node::set_name` or `Control::get_theme_stylebox` is the accepted trade described above — link it to the GDScript audit rather than treating it as a blocker for this PR.


leanmendoza (approved):

leanmendoza:
Waiting for freeing the track 1.15.0 in main

PR #2902, opened 2026-09-15
fix: scene-triggered swing T-pose, dotted clip names, desktop touch buttons

### LINKED ISSUE #2618: [Bug] Aisha NPC avatar gets stuck in T-pose during Plaza Guide tutorial (animation weights not matching Unity)
## Description
While testing the "Talk to Plaza Guide" tutorial on mobile, some animations appear broken — Aisha's avatar gets stuck in a T-pose instead of playing the expected animation clips.

Reported by dzsunyec (working on the related tutorial UI fixes): it seems very likely that Godot handles some animation weights differently, or something is off with how the GLB asset's animation clips are being blended/played, compared to Unity's behavior.

## Steps to Reproduce
1. Enter Genesis Plaza on mobile
2. Trigger the "Talk to Plaza Guide" tutorial (interact with Aisha)
3. Observe Aisha's avatar during the interaction

## Expected Behavior
Aisha's avatar should play the intended animation clips smoothly, matching the behavior seen on Unity/desktop, without falling back to a T-pose.

## Context
- Originally reported by Elizabeth (2026-07-24) alongside other tutorial UI feedback (small buttons, text wrap overlap) — those UI issues are being addressed separately by dzsunyec (moving from 3D to 2D UI).
- dzsunyec attempted to debug the T-pose issue (including with Claude's help) but could not find the root cause; suspects animation weight handling differences in Godot vs Unity, or an issue with the GLB clip data itself.
- Needs investigation from the godot-explorer team into animation weight/blend compatibility with Unity for this NPC's GLB asset.

## Related
- Follow-up from mobile tutorial UI feedback thread (Elizabeth, 2026-07-24) in #C0AR7AENQN9

Requested by Mateo via Slack

### LINKED ISSUE #2766: [Bug] Animations Not Triggering / T-Posing in Mobile
### Platform
Android & iOS

### App Version
Not specified

### Device Information
Not specified

### Issue description
Multiple separate reports of avatar/NPC animations not triggering / T-posing on mobile:

1. **Rodri's repro scene**: `pixelpirate.dcl.eth` (BIGYELLOWFISHES - Updated Version), position `-1,0`. Video attached in Slack showing avatars stuck in T-pose / animations not playing. Please upload the attached video (`Whatsapp Video 2026-08-24 At 9.04.28 Pm.mp4`) to this issue manually.
2. **Discord user report**: walking animation not playing correctly on mobile for their avatar (Silas). The user's own diagnosis (via AI assistant, unverified) was that `AvatarShape` lacks an automatic "locomotion blend tree" to activate body-part movement on mobile. They also noted a separate desktop-only issue: at low graphics settings the avatar walks fine, but at high graphics settings the walk animation shakes. Two videos attached in Slack (`2026-08-20 08-32-52.mp4`, `silas_walking_hard.mp4`) — please upload manually.
3. **Latest Discord thread** (<https://discord.com/channels/417796904760639509/1537803999435038801/1539974511934578749>): two more repro cases —
   - Playing a bowling shooting animation causes the avatar to T-pose.
   - Walking an NPC character around: on mobile the NPC's animations don't play at all (see the latest 2 videos in the linked Discord thread — need manual upload here).

### Expected behaviour
Avatars/NPCs should play correct animations (idle/walk/jog/run, action/shooting animations) on mobile, matching desktop behaviour, without T-posing, freezing, or shaking.

### Actual behaviour
- Avatars are seen T-posing / not playing movement animations in `pixelpirate.dcl.eth`.
- A third-party creator (Discord) reports the walking animation not applying correctly on mobile, plus a separate desktop-only shaking artifact at high graphics settings.
- Playing a bowling shooting animation causes the avatar to T-pose (Discord thread).
- NPC character animations don't play at all while walking around on mobile (Discord thread).

### Screenshots/Media

https://github.com/user-attachments/assets/a7998a01-ef6b-436b-8315-2283a3a7d748

https://github.com/user-attachments/assets/0817de2a-ff61-4b65-83b6-2b3f7f0dcf21

https://github.com/user-attachments/assets/dcc5d5bf-f6c5-49d4-98a5-12692b0f44bf

### Steps to reproduce
1. Open `https://decentraland.org/jump?realm=pixelpirate.dcl.eth&position=-1,0` on mobile.
2. Observe avatar(s) T-posing / not animating instead of playing locomotion animations.
3. Alternatively (per Discord thread): play a bowling shooting animation and observe avatar T-pose, or walk an NPC character around and observe no animations playing at all.

### Occurrence
Not specified — needs confirmation on frequency/consistency.

### Theories
1. Custom `AvatarShape` entities in the scene may not be receiving/blending locomotion animation state correctly on mobile (per third-party report — "no automatic locomotion blend tree").
2. Related to recently-fixed `AvatarLocomotionSettings` bugs (#2727, #2682) — worth confirming this isn't a lingering regression from that pipeline (root vs player entity handling, deferred signal timing).
3. Could overlap with other recent animation-weight/rendering issues seen on mobile (#2618 Aisha NPC T-pose, #2734 nearby players animating as discrete frames).
4. Action/shooting animations (bowling) triggering T-pose suggests a broader issue with one-shot/triggered animation clips not blending back into the locomotion state on mobile, separate from continuous walk/idle blending.
5. NPC animations not playing at all (vs. avatar T-posing) may point to a different code path than player-avatar locomotion — possibly NPC/AvatarShape entities not player-controlled failing to receive any animation state on mobile.

### Related Issues
- #2727 — AvatarLocomotionSettings is read from the ROOT entity instead of PLAYER (closed 2026-08-20)
- #2682 — [Bug] AvatarLocomotionSettings changes not applied on mobile (closed 2026-08-25)
- #2618 — [Bug] Aisha NPC avatar gets stuck in T-pose during Plaza Guide tutorial (open)
- #2734 — [Bug] Nearby players animate as discrete frames instead of continuously (open)
- #1938 — [Bug] Avatar T-posing (closed, prior similar report)

### Discord Thread
<https://discord.com/channels/417796904760639509/1537803999435038801/1539974511934578749>

---
Requested by Rodri via Slack

manuelmaceira:
## What

Scene-triggered swing animations (golf, bowling) now actually play instead of leaving the avatar in a T-pose with only the hips moving. Scene entities whose animation clips have dots in their names (e.g. the pixelpirate waterfall) animate again, and the on-screen touch buttons now work with a mouse on desktop, where the client runs mobile simulation.

## Why

Three separate T-pose root causes found this sprint. The emote converter exported the wrong clip from a multi-animation GLB, so the bundled swing emote only animated the hips. Godot's GLTF importer renames clips with `.` to `_` and our resolver dropped those states silently, freezing the whole entity. And the touch buttons only listened to screen-touch events, so on desktop they ignored mouse clicks entirely.

Closes #2766, closes #2618

## Details
<details>
<summary>Expand</summary>

Root causes, one per commit:

- **`swingWeaponOneHand` asset**: `Swing_OneHand.glb` bundles two animations — `Sword_Attack_Armature.001` (armature-level, almost all constant tracks) and `Sword_Slash` (the real swing). The converter took `anim_names[0]` and shipped a clip with 92/100 constant tracks. The converter now requires an explicit `{"anim": ...}` entry for multi-animation GLBs and errors loudly instead of guessing; the asset was re-exported from `Sword_Slash`.
- **Clip-name mismatch**: scenes ask for clips by their authored name (`Flow.004`); Godot imports them as `Flow_004`. `resolve_clip_name` only handled numeric suffixes, so every dotted state was dropped with zero log output — and when a base clip `Flow` existed, it misresolved to the wrong animation. The resolver now tries the exact sanitized match first. Also adds permanent warnings for the two silent-drop paths, which is how this was found.
- **Desktop touch buttons**: `button_touch_action` ignored `InputEventMouseButton`, and the scene-input poller gate (`pointer-locked || is_mobile`) discarded the synthetic action even when it fired. Left clicks now count as touches, and any joypad press/release opens a 250 ms window so both the down and the up edge reach the scene (the up edge is what the golf power-lock waits for).

Deliberately not done: reporting a real platform (desktop) from `getExplorerInformation` — the explorer is mobile-only and dev must mirror mobile, so it stays `'mobile'` and is now documented in `Runtime.js`.

</details>

## Test plan

**Case 1 — swing emote (iOS or Android)**

1. Open the app and go to `pixelpirate.dcl.eth` (spawn `-1,0`).
2. Walk up to the ball and tap the on-screen primary button to address it.
3. Set power and lock it with the same button.
- [ ] The avatar plays a visible one-arm swing (arm moves, not a T-pose) and the ball moves.
- [ ] After the swing, the avatar returns to idle/walk animations normally.

**Case 2 — scene Animator with dotted clip names**

1. Same scene, look at the vertical waterfall on the course.
- [ ] The waterfall water flows (previously frozen).
- [ ] No regression on other animated props (doors, hamster cage, flags).

**Regression — emotes and input**

- [ ] Emote wheel emotes (wave, dance) still play on the player's avatar.
- [ ] Walking with WASD or the virtual joystick still works and doesn't leak into scenes when typing in chat (desktop).

Author verified: swing plays correctly on macOS desktop and iPhone; waterfall animates on desktop; unit tests for the clip resolver green; clippy/fmt/gdscript checks clean. QA: case 1 on a physical phone.


godot/src/ui/components/atoms/buttons/button_touch_action/button_touch_action.gd
```diff
@@ -51,22 +51,34 @@ func _on_gui_input(event: InputEvent) -> void:
...
-			set_pressed_no_signal(false)
-			_touch_index = -1
-			accept_event()
+		is_press_event = event.pressed and (event.index == _touch_index or _touch_index == -1)
+		is_release_event = not event.pressed and event.index == _touch_index
+	elif event is InputEventMouseButton and event.button_index == MOUSE_BUTTON_LEFT:
+		is_press_event = event.pressed
+		is_release_event = not event.pressed
+
+	if is_press_event:
```
regenesis-claw: [P1] The press branch lost the old `_touch_index == -1` guard, so it re-enters on a duplicate press event — and duplicates are the norm here, not the exception:

- On mobile, `emulate_mouse_from_touch` (default on) synthesizes an `InputEventMouseButton` for the primary finger. `_ready()`'s comment says `button_mask = 0` makes the base `Button` ignore it — true for `BaseButton::gui_input`, but `Control` emits the `gui_input` **signal** before and regardless of `button_mask`, so `_on_gui_input` still sees it.
- On desktop, `project.godot` sets `pointing/emulate_touch_from_mouse=true`, so a left click arrives as an emulated `InputEventScreenTouch` *and* as the `InputEventMouseButton`.

Either way `button_down.emit()` + `touch_action_changed.emit(true)` fire twice, and `ui_sounds.gd:67` hangs `play_sound("generic_button_press")` off `button_down` — so this is a double press SFX on every joypad button tap on a real phone.

```suggestion
	if is_press_event and not _is_action_active:
```

Separately: since desktop clicks already arrive as emulated screen touches, was the `InputEventMouseButton` branch actually required, or was the desktop failure entirely the `input.rs` poller gate? If it's the latter, dropping the mouse branch removes the duplicate at the source.

lib/src/godot_classes/animator_controller.rs
```diff
@@ -131,6 +132,17 @@ impl MultipleAnimationController {
             .states
             .retain(|state| self.existing_anims_duration.contains_key(&state.clip));
 
+        // If every requested clip is missing, the entity renders its rest pose
+        // with zero feedback — the classic silent T-pose/frozen-entity failure
+        // (e.g. scene asks "Flow.003", Godot imported "Flow_003").
+        if value.states.is_empty() && !requested_clips.is_empty() {
+            tracing::warn!(
+                "Animator: all clips dropped. requested={:?} available={:?}",
```
regenesis-claw: [P1] `warn!` routes to Godot's warning stream and into Sentry (REVIEW.md §14). `apply_anims` runs on every Animator CRDT update, not once per entity, and this fires exactly in the persistent case — a scene whose clip names never resolve. A scene writing its Animator each tick turns this into a sustained Sentry warning stream from every client standing in it.

Keep the diagnostic, but bound it: a `HashSet<String>` of already-warned clip sets on `MultipleAnimationController`, or first occurrence at `warn!` and repeats at `debug!`. Same applies to the per-state `warn!` at line 705.

[P2, same hunk] `requested_clips` at line 116 clones a `String` per state on **every** call just to feed this branch. `suggested_value` is still in scope here — read `suggested_value.states` in the warn branch and drop the up-front allocation.

lib/src/godot_classes/dcl_global.rs
```diff
@@ -841,6 +848,20 @@ impl DclGlobal {
         }
     }
 
+    #[func]
+    pub fn set_joypad_input_active(_active: bool) {
```
regenesis-claw: [P2] A function named `set_joypad_input_active` that takes a `bool` and discards it is going to get "fixed" by the next reader — the comment helps, but the signature actively misleads. `notify_joypad_input()` with no argument states the real contract and shortens both GDScript call sites.

[P2] The 250 ms is a bare literal at line 856; a named const with a line on where the number came from would age better.

lib/src/scene_runner/input.rs
```diff
@@ -93,7 +93,9 @@ impl InputState {
         // it means that you are interacting with the UI.
         // Mobile: Ignore that rule
         let is_pointer_locked = input.get_mouse_mode() == MouseMode::CAPTURED;
-        if is_pointer_locked || DclGlobal::singleton().bind().is_mobile {
+        let global = DclGlobal::singleton();
+        let global_bind = global.bind();
```
regenesis-claw: [P2] `global.bind()` is now held for the rest of the function instead of just the condition. Nothing in the loop touches `DclGlobal` today so this is safe, but the first thing that does will panic on a double borrow. Copying the two bools into locals and letting the guard drop before the loop removes the footgun:

```suggestion
        let (is_mobile, joypad_active) = {
            let global = DclGlobal::singleton();
            let b = global.bind();
            (b.is_mobile, b.joypad_input_active())
        };
```

[P2] While the window is open the poller forwards *every* mapped action, movement included, with the pointer unlocked — so holding WASD while clicking an on-screen button leaks movement into the scene for 250 ms on desktop. Narrow, but worth a regression line in the test plan.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`. Three genuinely separate root causes, and the two I could verify from source check out: the old `swingWeaponOneHand.tres` really does have 92/100 near-constant tracks and the re-export replaces them with real curves (37KB -> 63KB, same 0.667s length), and the `resolve_clip_name` sanitized-first ordering is correct with good unit tests. The input layer is where this needs work.

**Pre-flight (REVIEW.md §0)**
- Branch is **27 commits behind `main`** (`ahead 3 / behind 27`, `diverged`). Per §0.1 that makes CI signal untrustworthy — please merge `main` before this lands.
- CI at review time: Linux / Static checks / Clippy green; 🤖 Android and 🐳 Docker still pending.
- No iOS build. This diff touches none of the §0.2 iOS-required paths (no `plugins/dcl-godot-ios/**`, no `OS.get_name() == "iOS"` branch, nothing under `comms/` or `av/`), so I'm not holding on that — but it *is* mobile-visible input code, so a `build` label wouldn't hurt.

---

### P1 — the press path lost its idempotency guard, so every joypad tap fires twice

`button_touch_action.gd:65`. The old code gated the press branch on `_touch_index == -1`; the new one enters it on *any* press event.

Both emulation settings are on in this project, so a single physical press delivers **two** events to `gui_input`:
- **Mobile:** `input_devices/pointing/emulate_mouse_from_touch` (default true) synthesizes an `InputEventMouseButton` from the primary finger — `joypad.gd:118` documents exactly this ("Godot's emulated mouse only covers the primary touch"), and `_ready()`'s own comment says `button_mask = 0` exists to make the base `Button` ignore it. But `button_mask` only gates `BaseButton::gui_input`; the `gui_input` **signal** fires first and unconditionally, so `_on_gui_input` sees it.
- **Desktop:** `project.godot:212` sets `pointing/emulate_touch_from_mouse=true`, so a left click is dispatched as an emulated `InputEventScreenTouch` *and* the original `InputEventMouseButton`.

Net effect: `button_down.emit()` and `touch_action_changed.emit(true)` fire twice per press, and `ui_sounds.gd:67` connects `button_down` -> `play_sound("generic_button_press")` — i.e. **a double press SFX on every on-screen joypad button, on a phone.** `Input.action_press` is idempotent and the release path is protected by `_is_action_active`, so the press branch is the only unguarded one.

This also raises the question of whether the mouse branch is needed at all: with `emulate_touch_from_mouse=true`, the desktop click was *already* arriving as a screen touch, which suggests the real desktop breakage was layer 2 (the `input.rs` poller gate) alone. Worth confirming before keeping the extra path — if it isn't needed, dropping it removes the duplicate source entirely.

### P1 — new `tracing::warn!` calls are unbounded on a per-update path (REVIEW.md §14, Sentry quota)

`animator_controller.rs:140` and `:705`. Both `warn!`s route to Godot's warning stream and into Sentry. They fire precisely in the case that repeats — a scene whose Animator clip names don't resolve — and `apply_anims` runs on every Animator CRDT update, not once per entity. A scene that writes its Animator component each tick with a bad clip name becomes a sustained warning stream from every client in it.

Keep the diagnostic (it's how this was found) but dedupe it: a `HashSet<String>` of already-warned clips on `MultipleAnimationController`, or `debug!` for the repeat and `warn!` only the first time per entity.

---

### P2

1. `animator_controller.rs:116` — `requested_clips` allocates a `Vec<String>` (one `String` clone per state) on **every** `apply_anims` call just to feed a warning that almost never fires. `suggested_value` is still in scope at line 143, so the branch can read `suggested_value.states` directly and the clone can go away entirely.
2. `dcl_global.rs:852` — `set_joypad_input_active(_active: bool)` ignores its parameter. A setter named `set_*` taking a bool it discards will be "fixed" by the next person who reads it. `notify_joypad_input()` with no argument says what it does, and GDScript call sites get shorter.
3. `dcl_global.rs:856` — the 250 ms window is a bare literal in the body. A named `const JOYPAD_INPUT_WINDOW_MS` (or `Duration`) with a one-line note on why 250 would age better.
4. `input.rs:97` — `global.bind()` is now held for the whole function rather than for the condition. Nothing inside the loop touches `DclGlobal` today, so it's fine, but the first call that does will panic on a double borrow. Reading the two bools into locals and dropping the guard before the loop is a cheap hedge.
5. `input.rs:98` — while the window is open, the poller forwards *every* mapped action including movement, with the pointer unlocked. Holding W/A/S/D while clicking an on-screen button on desktop leaks movement into the scene for those 250 ms. Narrow and desktop-only, but the test plan's "doesn't leak into scenes when typing in chat" regression line is worth extending to cover it.
6. `animator_controller.rs:460` — only `.` is un-sanitized. Godot's GLTF importer also replaces other invalid name characters; not introduced by this PR, but a comment noting the known-incomplete mapping would save the next debugging session.
7. PR body says `Refs #2766, #2618` rather than `Closes` — intentional if these are partial fixes, but the issues won't auto-close.

**Security:** no security issues found — no network, auth, deserialization or user-controlled path handling in this diff.

**Test plan:** good shape, cases start from opening the app and have observable pass/fail lines. Please add a case for "tap an on-screen action button on a phone — exactly one press sound plays" once the P1 is fixed, since that's the regression this introduces.

regenesis-claw (approved):
Re-reviewed at `e78ecaa` (merge of `main` + one fix commit). **All 2 P1s and 5 of 7 P2s fixed; approving.**

| Round-1 finding | Status |
|---|---|
| P1 press path lost its idempotency guard -> double `button_down` / double press SFX | **Fixed** — reverted to the touch-only `_touch_index == -1` shape, mouse branch dropped entirely |
| P1 unbounded `warn!` into Sentry (both sites) | **Fixed** — per-controller `warned_missing_clips` set + a process-global once-set in the free `apply_anims` |
| P2 `requested_clips` clone per call | **Fixed** — allocation gone, warn branch reads `suggested_value.states` |
| P2 `set_joypad_input_active(_active)` discarded its param | **Fixed** — `notify_joypad_input()`, no stale call sites left in the tree |
| P2 bare 250 ms literal | **Fixed** — `JOYPAD_INPUT_WINDOW` const with a why-comment |
| P2 `GdRef` held across the poller loop | **Fixed** — two short-lived binds, nothing held into the loop |
| P2 window forwarded WASD while unlocked | **Fixed** — `forward_all || (joypad_window_open && !is_movement_action(..))`, movement excluded |
| P2 `resolve_clip_name` only un-sanitizes `.` | Untouched (was a note, not an ask) |
| P2 `Refs` vs `Closes` on #2766 / #2618 | Untouched — fine if these are partial fixes |

Good call confirming the root cause rather than keeping both layers: the commit message now states the Mac breakage was entirely the poller gate, which matches `emulate_touch_from_mouse=true` in `project.godot`. With the mouse branch gone, desktop still reaches `notify_joypad_input()` through the emulated touch, so the fix is intact on both platforms.

The re-review ask I made in round 1 (add a "exactly one press sound per tap" QA case) is moot now that the duplicate source is gone — no need.

**Two leftovers, neither blocking:**

1. `animator_controller.rs` — the static `OnceLock<Mutex<HashSet<String>>>` uses `.lock().unwrap()`, which panics on a poisoned mutex. Nothing inside the critical section can panic today, so it's safe, but `if let Ok(mut w) = warned.lock()` costs nothing and can't take the scene runner down. The key is `"{clip}|{existing_anims.len()}"` — anim-count is a proxy for entity identity, so two different GLBs with the same clip count and the same missing clip name only warn once. That's the conservative direction, which is the right one here; worth a word in the comment so the next reader doesn't think it's per-entity.
2. The set is process-global and never cleared across realm changes. Bounded by distinct clip names so it won't grow meaningfully — just noting it's deliberate.

**CI at approval:** 📊 Static checks and 🔧 Clippy green on `e78ecaa`; 🐧 Linux and 🤖 Android still running. Branch is now 1 behind `main` (was 27) — pre-flight §0.1 satisfied. No iOS build, which §0.2 doesn't require for this diff.

Note my approval doesn't clear branch protection here — you'll still need a human codeowner.

kuruk-mm (approved):
looks great!

PR #2906, opened 2026-09-15
feat(bug-report): Platform attribute and up to 3 screenshots (#2842)

### LINKED ISSUE #2842: Intercom integration - Check new fields
New fields:
- More images
- From which client it is the report

EibrielInv:
## What
- Bug reports now tell Intercom which client sent them. Every ticket carries `Platform = Mobile`.
- The Report a Bug form accepts up to 3 screenshots, and all of them are attached to the Intercom ticket and the Sentry event.

## Why
PR #2779 (issue #2652) shipped the form with one screenshot and no client field, because the intercom-proxy didn't support either yet. The proxy now accepts `Platform` and up to 3 images ([contract on #2842](https://github.com/decentraland/godot-explorer/issues/2842#issuecomment-5682196889)), so this finishes that work.

Closes #2842

## Details

- `BugReportService` sends `"Platform": 2` as a raw number, as the proxy contract defines it. The value is fixed because this client only ships on mobile.
- The proxy rejects any request body over 4.5 MB. The images share a 3 MiB budget (about 4.2 MB after base64). Each image gets the remaining budget divided by the images still to come, so a small or dropped image leaves more room for the others.
- Images that fit are sent as-is, so a normal submit does no image work. An oversized image is re-saved from the preview the form already decoded, stepping 1440px/0.75 → 1080px/0.7 → 720px/0.6. The service waits one frame before each re-save so the spinner keeps animating. If even the smallest step is too big, the image is left off the ticket with a warning, and the report is still filed.
- `SentryUserFeedback` attaches every original image (`screenshot_1.jpg`, `screenshot_2.jpg`, …), so an image shrunk or dropped from the ticket can still be found in Sentry.
- `BugReportModal`: `MAX_SCREENSHOTS` goes from 1 to 3, and the temporary single-image workaround is removed.

</details>

## Test plan

Cases 3 and 4 end in Intercom, so someone with access to the Intercom support inbox needs to check the ticket. Put a unique word in the description (for example `QA-2906-A`) to find it.

### Case 1: Attach up to 3 screenshots
**Steps:**
2. Open Settings → Help → Report Bug.
3. Check the Screenshot row: the first tile already shows the in-game view from when Settings opened.
4. Tap the empty tile and pick an image from the gallery. Repeat once more.

- [ ] **Expected:** 3 tiles show images, and no empty "add" tile appears after the third.
- [ ] **Expected:** all 3 tiles fit inside the form in portrait, with nothing clipped at the right edge.

### Case 2: Delete and replace a screenshot
**Steps:**
1. Open the app, open Settings → Help → Report Bug, and add images until there are 3.
2. Delete the second screenshot.
3. Replace the first screenshot with a different gallery image.

- [ ] **Expected:** after deleting, 2 images remain and an empty "add" tile reappears.
- [ ] **Expected:** after replacing, the first tile shows the new image and the rest are unchanged.

### Case 3: Ticket carries 3 images and Platform = Mobile
**Steps:**
2. Open Settings → Help → Report Bug.
3. Choose any Issue Type, type `QA-2906-A` in the description, and attach 3 images in total.
4. Tap Submit.
5. In Intercom, open the ticket whose description contains `QA-2906-A`.

- [ ] **Expected:** the app shows the success modal.
- [ ] **Expected:** the ticket has 3 attached images that match the ones picked, and the Platform field reads **Mobile**.

### Case 4: Large camera photos still get filed
**Setup:** 3 recent full-resolution camera photos in the gallery (not screenshots).
**Steps:**
1. Open the app and open Settings → Help → Report Bug.
2. Delete the pre-filled screenshot, then attach the 3 camera photos.
3. Choose any Issue Type, type `QA-2906-B` in the description, and tap Submit.
4. In Intercom, open the ticket whose description contains `QA-2906-B`.

- [ ] **Expected:** the spinner keeps animating while submitting, and the success modal appears.
- [ ] **Expected:** the ticket has all 3 photos attached. They may look slightly smaller or softer than the originals.

### Case 5: Regression, report with no screenshots
**Steps:**
1. Open the app and open Settings → Help → Report Bug.
2. Delete the pre-filled screenshot.
3. Choose any Issue Type, type a description, and tap Submit.

- [ ] **Expected:** the success modal appears, and no error is shown.


godot/src/bug_report/bug_report_service.gd
```diff
@@ -182,22 +194,37 @@ static func _collect_device_info() -> Dictionary:
...
+					"BugReportService: attachment is %d bytes (budget %d) — filing without it"
+					% [bytes.size(), budget]
+				)
 			)
-		)
-		return {}
-
-	return {"content_type": "image/jpeg", "data": Marshalls.raw_to_base64(bytes)}
+			continue
+		evidence.append({"content_type": "image/jpeg", "data": Marshalls.raw_to_base64(fitted)})
```
regenesis-claw, line 246: [P1] `evidence` goes from a single `{content_type, data}` object to an array of them. Unity's `IntercomTicketPayload.BuildCreateTicketJson` still emits the single-object form, so there's no parity reference for the array shape or for the key name staying `evidence`. Given this file's own rule that a wrong body shape gets the whole ticket rejected, could you link the proxy change (or paste a ticket created from this build with 3 images attached)? If the proxy still expects the object form, every screenshot-bearing report fails — a regression on today's single-image behaviour.

godot/src/bug_report/bug_report_service.gd
```diff
@@ -98,6 +109,7 @@ static func _build_attributes(
 		"Graphic Card": device["gpu"],
 		"RAM": device["ram"],
 		"Client version": String(DclGlobal.get_version()),
+		"Platform": PLATFORM_MOBILE,
```
regenesis-claw, line 120: [P1] This reverses the previous rule ("Every value must be a String — the proxy rejects non-string attribute values"), and Unity passes list attributes as option-id **strings** (`Issue Type`, `Meets Minimum Requirements`, with the comment "Intercom takes the id, never the label"). What confirms the proxy wants a raw int for `Platform`? An unexpected value type here takes the whole ticket down, so I'd like the proxy schema or one real ticket showing `Platform = Mobile` before merge.

godot/src/bug_report/bug_report_service.gd
```diff
@@ -182,22 +194,37 @@ static func _collect_device_info() -> Dictionary:
...
+		present.resize(MAX_EVIDENCE_COUNT)
+	if present.is_empty():
+		return []
+
+	var budget := mini(MAX_EVIDENCE_BYTES, TOTAL_EVIDENCE_BYTES / present.size())
+	var evidence: Array = []
+	for bytes in present:
+		var fitted := bytes
+		if fitted.size() > budget:
+			fitted = BugReportCapture.encode_within(ImagePickerService.decode(bytes), budget)
```
regenesis-claw: [P2] This decodes and re-encodes on the main thread, before `async_submit`'s first `await` — which contradicts the docstring above `async_submit` ("the submit path — which the player waits on behind a spinner — does no image work at all (PR #2779 review)"). With 3 attachments the budget is 1 MiB each and a 1920px q85 gallery photo routinely exceeds that, so this is the common case: 1 decode + up to 4 full-size encodes per image with the spinner frozen. Two things worth doing: yield a frame between images, and update that now-stale docstring either way.

Also, `BugReportModal` already holds the decoded `Image` for every shot (`_shots[i]["image"]`), so this decode is the exact round trip `image_picker_service.gd` warns against — passing the images through would avoid it.

godot/src/bug_report/bug_report_capture.gd
```diff
@@ -24,6 +24,16 @@ const MAX_DIMENSION := 1920
 # the 3MB evidence cap was encoded the same way.
 const JPEG_QUALITY := 0.85
 
+# Fallbacks for encode_within(), tried in order until an image fits its evidence
+# budget. The first matches a regular capture, so an image already in budget
+# doesn't get worse.
+const SHRINK_STEPS := [
+	{"dimension": MAX_DIMENSION, "quality": JPEG_QUALITY},
```
regenesis-claw: [P2] Step 0 is `1920 / 0.85`, i.e. exactly what `ImagePickerService` and `capture()` already produced. `encode_within` is only called when the buffer is *over* budget, so this step re-encodes at the same settings, lands at roughly the same size, fails the check and is discarded — one wasted full-resolution encode on every oversized image. Starting the ladder at 1440/0.75 keeps the same result for less work.

```suggestion
const SHRINK_STEPS := [
	{"dimension": 1440, "quality": 0.75},
```

regenesis-claw (dismissed):
Reviewed against `REVIEW.md`. Nice, focused diff: the shrink-to-fit fallback is a genuinely better failure mode than the old "drop the image entirely", the Sentry side keeps the *originals* so a shrunk/dropped attachment is still recoverable, the removed TEMPORARY block is fully cleaned up (comment + constant + the slot/gap logic that was already written for three tiles), and every new log is a `push_warning` on an expected-and-handled condition — no Sentry quota cost.

**Pre-flight:** branch is up to date with `main` (behind_by 0). All checks green (static, clippy, Linux, Android, Docker). No iOS build on this PR and none is required — this is pure GDScript with no `OS.get_name()` branch, no native-plugin, safe-area or keyboard path (the native picker isn't touched). No i18n-visible strings added, no debug prints, no non-English text, file stays in `components/organisms/`.

Holding on one thing only: **the proxy contract is asserted but not evidenced anywhere I can check.**

### 🔴 P1 — `evidence` shape and `Platform` type are unverified against the proxy

This file's own header says *"An extra or misspelled key gets the entire ticket rejected, so nothing here is speculative — attributes match Unity's exactly."* Two things in this diff now diverge from the only reference I can read:

1. `evidence` changes from a **single object** `{content_type, data}` to an **array of objects**. Unity's `IntercomTicketPayload.BuildCreateTicketJson` still writes the single-object form, so there is no parity reference for the array shape (or for the key staying `evidence` rather than e.g. `evidences`).
2. `"Platform": 2` is sent as an **int**, reversing the previous rule *"Every value must be a String — the proxy rejects non-string attribute values"*. Unity passes every list attribute (`Issue Type`, `Meets Minimum Requirements`) as an **option-id string**, explicitly noting *"Intercom takes the id, never the label"*.

`decentraland/intercom-proxy` isn't readable from here and issue #2842 is still open with a two-line body ("New fields: More images / From which client it is the report") — it reads as the task, not as confirmation the proxy shipped it. If either assumption is wrong the ticket is rejected wholesale, which means **bug reports with screenshots stop working entirely** — a regression on today's single-image behaviour, and one that only shows as a failed submit on a real device.

This can resolve with evidence rather than a code change: link the proxy PR/schema, or paste a ticket created from this build showing `Platform = Mobile` and 3 attached images. Either is enough for me to flip to approve.

Related: the `## Test plan` has no case that verifies the two things the PR is actually about. "The success modal appears" only proves the POST returned an id — it does not prove the ticket carries `Platform = Mobile` or that all three images landed on it. Please add a case that ends in Intercom, plus one for the shrink/drop path (attach a large camera photo in all 3 slots). Per `REVIEW.md` §4 the cases should also be numbered blocks starting from "Open the app" with an explicit expected result — QA runs these cold on a phone.

### 🟡 P2 (non-blocking)

- **Image work is back on the submit path** — `_build_evidence` now decodes and re-encodes on the main thread before `async_submit`'s first `await`, which contradicts the docstring two functions above (*"the submit path — which the player waits on behind a spinner — does no image work at all (PR #2779 review)"*). With 3 attachments the per-image budget is 1 MiB, and a 1920px q85 gallery photo commonly exceeds that, so this is the **normal** path, not the rare one: 1 decode + up to 4 full-size JPEG encodes per image, with the spinner painted but frozen. Consider yielding a frame between images (`_build_evidence` would become async), and at minimum update the stale docstring.
- **First shrink step is a guaranteed miss.** `SHRINK_STEPS[0]` is exactly `1920 / 0.85` — the encoding `ImagePickerService` and `BugReportCapture.capture()` already produced. We only enter `encode_within` when the input is *over* budget, so step 0 re-encodes at the same settings, gets roughly the same size, fails, and is thrown away. Starting at 1440/0.75 would drop one full-resolution encode from the common case.
- **Double decode.** `BugReportModal._shots[i]["image"]` already holds the decoded `Image` (kept deliberately, per its own comment), but `_build_evidence` calls `ImagePickerService.decode(bytes)` again — the very round trip `image_picker_service.gd` warns about (*"decoding and re-encoding costs quality and time for nothing"*). Passing the images alongside the bytes, or decoding lazily once per shot, avoids it.
- **Budget isn't redistributed.** `budget` is fixed at `TOTAL / present.size()`, so an image dropped at line 225, or one that comes in far under its share, leaves headroom nobody uses — the other attachments are still squeezed to 1 MiB. Tracking a running remainder would keep more quality for the same body size. Fine as-is if you'd rather keep it simple; worth a comment saying so.

### Security

No security issues found. Same signed-fetch + allowlisted `Origin` path as before, no new user-controlled input reaches a URL or a log, no secrets, and the added `push_warning` lines carry only sizes.

**Requesting changes** — on the P1 verification only; the P2s are all follow-up-or-ignore.

kuruk-mm (approved):
great work!

godot/src/bug_report/bug_report_service.gd
EibrielInv: The array shape comes from the intercom-proxy contract posted on #2842: https://github.com/decentraland/godot-explorer/issues/2842#issuecomment-5682196889

It defines `"evidence": [ { "content_type": "image/png", "data": "<base64>" } ]` as an optional array of at most 3, and the key is still `evidence`. It also says: *"A single evidence object instead of an array is still accepted"*, so the array form is the current shape and the single object is the legacy form that's still supported. Unity's `BuildCreateTicketJson` still uses that legacy form.

It also covers the size limits this PR sizes the images for: 3 MiB decoded per image, and 4.5 MB for the whole body.

godot/src/bug_report/bug_report_service.gd
EibrielInv: Same source, the intercom-proxy contract on #2842: https://github.com/decentraland/godot-explorer/issues/2842#issuecomment-5682196889

It defines `"Platform": 1, // 1 Desktop, 2 Mobile, 0 rejected, omit if unknown`, so the value is a raw number. `Platform` isn't one of the guid list attributes: the contract lists guids only for `Issue Type` and `Meets Minimum Requirements`, which is why those two stay strings. I added a comment next to the attribute pointing at the contract (a8489498).

PR #2907, opened 2026-09-15
fix(sentry): always keep freed-instance script errors

kuruk-mm:
## What

Engine errors about touching a freed object (`… on a previously freed instance`, `Trying to cast a freed object`, `… on a base object of type 'previously freed'`, `Object … was freed or unreferenced while a signal is being emitted`) now always reach Sentry, tagged `source: freed_instance`, kept at 100% and exempt from the remote sample rate like crashes. Everything else in the logger firehose still needs the `sentry-error-events` flag. `NodeGuard` stale-node reports go up from warning to error.

## Why

Every one of these messages is a use-after-free in our GDScript — exactly the class of bug that is a SIGSEGV on the release template (Android today, iOS after #2898). Until now they were dropped with the rest of the firehose (0 `SCRIPT ERROR` events reached Sentry in the last 30 days) and, even with the flag on, sampled at 1% as `engine`. On the release template the property-get/set and signal-emit variants still log instead of crashing, so this is the only fleet-level signal we get for the sites the audit needs to fix (context: decentraland/godotengine#21 review, 2026-09-15).

The decision is now a static text-only function (`_firehose_source`) with a unit test; `SentryEvent` cannot be given an exception from GDScript, so the shape test in `_before_send` is covered by review only.

## Test plan

- [x] `cargo run -- run --itest` — `sentry_before_send: test_freed_instance_errors_bypass_the_firehose_flag … ok` (28 passed).

### Case 1 — A freed-instance error reaches Sentry with the flag off (dev check)
**Setup:** a staging desktop build with the debug hub attached, `sentry-error-events` flag off (the default):
```bash
cargo run -- debug-hub                      # terminal 1
DECENTRALAND_STAGING_BUILD=1 cargo run -- run -- --scene-inspector=ws://<mac-ip>:9231   # terminal 2
```
**Steps:**
1. Open the app, sign in as a guest, enter Genesis Plaza.
2. From `.claude/skills/mobile-dev-debug-tool/scripts/`: `unified.sh eval 'var n := Node.new(); n.free(); n.get_name()'`.
- [ ] **Expected (Sentry, staging environment, within a minute):** one event titled `Cannot call method 'get_name' on a previously freed instance.` with tag `source: freed_instance` and a GDScript frame.

### Case 2 — Regression: other errors are still dropped
**Steps:**
1. In the same session: `unified.sh eval 'push_error("audit probe")'`.
- [ ] **Expected (Sentry):** no event with that text (it only shows as a breadcrumb on a later crash).


kuruk-mm:
Superseded by the Release Candidate 1.14.1 (#2973), which carries these same three commits on `release-1.14.1`.

PR #2912, opened 2026-09-15
chore(gdlint): enable the node-across-await rules and document the ownership rule

kuruk-mm:
Turns on the two `gdlint` rules that flag a non-`self` node reference carried across an `await` (`node-reference-across-await`, `node-argument-across-await`), and writes the crash family down properly in `REVIEW.md` §5, `CLAUDE.md`, both `gdlintrc`s and `NodeGuard`'s header. The family has **two halves** and CI can only see one of them, so the docs now say which is which. `node-null-comparison` stays audit-only.

## Why

The engine's freed/null receiver check at `gdscript_vm.cpp:2036-2045` is `#ifdef DEBUG_ENABLED`, so on the release template the phone dereferences the pointer at line 2062 instead of logging. That one missing check covers two different bugs:

1. **A freed receiver** — the carried-across-`await` case this stack was built for. The engine cancels a coroutine with the node that owns it and protects nothing else, so the rule is ownership: async work that needs a node is a method of that node; a longer-lived owner re-resolves the node on resume instead of carrying it; work is cancelled when its owner frees the node. **This is the half `gdlint` enforces.**

2. **A null receiver** — and this is the one that actually shipped. Reproduced, symbolicated and fixed on `release-1.14.0b` (`9e3127c7`) on 2026-09-22: `@onready` is assigned by `_ready`, and `_ready` only fires when a node **enters the tree**. `add_child()` onto a parent that is itself detached does not put the child in the tree, so every `@onready` in it stays null forever. The trap is that the obvious guard passes — `change_scene_to_file()` detaches a page immediately while its `queue_free()` runs later, so `is_instance_valid(container)` is still **true** for a container that has already left the tree. Re-resolving on resume (rule 2 above) doesn't save you either: it returns a node that is alive but not ready. **`gdlint` cannot see this half** — both rules track *local* variables and explicitly exempt `@onready` members, which is exactly what the shipped bug used.

So the post-`await` question is not only "is it valid?" but **"is it still in the tree?"**

## Changes

| File | What |
|---|---|
| `.gdlintrc`, `godot/gdlintrc` | Enable both `*-across-await` rules; comment now states they cover the carried-reference half only |
| `REVIEW.md` §5 | Corrected opening (one check, two bugs); `21E` is a grouping bucket, not one bug; new **The other half** — mechanism, tombstone, `is_inside_tree()`, `get_tree()` null when detached, defer-and-replay (`CalendarButton`), symbolication recipe |
| `CLAUDE.md` | Same caveat next to the `gdlint` command |
| `node_guard.gd` | Fact 3: valid ≠ in the tree ≠ ready |
| (stacked) `godot/src/**` | The `#2909`/`#2910`/`#2911` ownership fixes; diff shrinks to config + docs once they merge |

**Merge order:** dcl-regenesislabs/godot-gdscript-toolkit#1 → #2908 → #2909 / #2910 / #2911 → this.

## Test plan

No QA needed — no behavior change (lint config, comments and docs only).

- [ ] `gdlint godot/` passes with the fork installed and both rules enabled (verified: `Success: no problems found`)
- [ ] `gdformat -d godot/` clean (verified: 392 files unchanged)
- [ ] `cargo run -- check-gdscript` passes


manuelmaceira (commented):
Reviewed files look good, will do an on-device test to verify everything works ok before approving

PR #2916, opened 2026-09-15
fix(bug-report): link the Sentry feedback from the Intercom ticket

### LINKED ISSUE #2915: Fix: Add Sentry feedback link to Intercom bug report
Right now the Intercom links to Sentry log, but not to the Sentry feedback. Add a link in the bug report to the generated Sentry feedback.

EibrielInv:
## What
Bug report tickets in Intercom now carry a second Sentry link, `Sentry feedback: <url>`, which opens the Sentry User Feedback entry for that report. The existing `Internal diagnostics` link to the event with the log tail and screenshots is unchanged.

## Why
Support could open the diagnostics event from a ticket, but had no direct way to reach the player's feedback entry.

Closes #2915

## Details

- The link is a feedback search on the `associated_event_id` tag: Sentry copies the id we set on the feedback onto the entry, so the search resolves to exactly this report. Verified by hand against a real feedback entry.
- Searched rather than addressed directly because `capture_feedback` returns nothing in sentry-godot (2.1.1, 2.2.0 and main), so the feedback's own id is unobtainable. The Unity client gets that id from its C# SDK and builds an `eventId=` link instead.
- No mailbox filter: the search finds the entry wherever it was filed, including when Sentry's auto-spam filter moves it out of the default inbox.
- `SentryUserFeedback.submit()` now returns `{event_url, feedback_url}` instead of a single URL. Both are empty when Sentry is off or drops the event.
- `BugReportService._compose_description()` keeps the `Internal diagnostics:` line (falling back to `unavailable`, as Unity does) and appends `Sentry feedback:` only when there is a link.
- Link shape: `https://dcl-regenesis-labs.sentry.io/issues/feedback/?project=4510187688361984&query=associated_event_id%3A<event id>`.

</details>

## Test plan

Cases 1 and 2 need a build that reports to Sentry (staging or production, not dev), plus access to the Intercom inbox and to Sentry. Put a unique word in the description to find the ticket.

### Case 1: Ticket links to the feedback entry
**Steps:**
2. Open Settings → Help → Report Bug.
3. Choose any Issue Type, type `QA-2915-A` in the description, and tap Submit.
4. In Intercom, open the ticket whose description contains `QA-2915-A`.
5. Tap the `Sentry feedback:` link at the end of the description.

- [ ] **Expected:** the description ends with an `Internal diagnostics:` line and a `Sentry feedback:` line.
- [ ] **Expected:** the link opens Sentry's User Feedback page showing the `QA-2915-A` entry, and no other.

### Case 2: Regression, diagnostics link still opens the event
**Steps:**
1. Open the Intercom ticket from Case 1.
2. Tap the `Internal diagnostics:` link.

- [ ] **Expected:** Sentry opens the "Bug report diagnostics" event, with the log tail and screenshot attached, and the `QA-2915-A` feedback shown on it.

### Case 3: Report still files without Sentry
**Setup:** a dev build (these send nothing to Sentry).
**Steps:**
1. Open the app and open Settings → Help → Report Bug.
2. Choose any Issue Type, type `QA-2915-B`, and tap Submit.

- [ ] **Expected:** the success modal appears. The ticket reads `Internal diagnostics: unavailable` and has no `Sentry feedback:` line.


godot/src/bug_report/sentry_user_feedback.gd
```diff
@@ -47,23 +52,31 @@ const CATEGORY_TAG_VALUE := "FEEDBACK"
 # Org and project are deployment values, not secrets — the same pair appears in
 # every Sentry URL. Project `godot-explorer` is DSN project id 4510187688361984.
 const ISSUE_URL_TEMPLATE := "https://dcl-regenesis-labs.sentry.io/issues/?query=%s"
+const FEEDBACK_LIST_URL_TEMPLATE := "https://dcl-regenesis-labs.sentry.io/issues/feedback/?project=4510187688361984&query=%s&start=%s&end=%s"
+
+# Window around the submit for the feedback list link. Before covers device clock
+# skew; after covers the time the envelope takes to be ingested.
+const FEEDBACK_WINDOW_BEFORE_SEC := 120
+const FEEDBACK_WINDOW_AFTER_SEC := 900
```
regenesis-claw: [P2] Two things about this window.

1. The comment's rationale for the tail is probably wrong: the feedback is searched by its **event timestamp**, which the SDK stamps from the device clock at capture — not by receive time — so the 15 minutes do not cover ingestion latency. What they actually hedge is clock handling.

2. Which matters because Relay applies clock-drift correction: an event whose timestamp is far from `sent_at` gets normalised toward server time, while these bounds are not normalised. A device more than ~2 minutes behind (or ~15 ahead) real time therefore produces a window that excludes its own feedback, and the link silently opens an empty list.

The query is already scoped to a single `user.id`, so precision here buys almost nothing — consider widening to something like 24h on each side (or using a `statsPeriod`), which makes the link robust to any plausible skew at no cost in noise.

godot/src/bug_report/sentry_user_feedback.gd
```diff
@@ -47,23 +52,31 @@ const CATEGORY_TAG_VALUE := "FEEDBACK"
 # Org and project are deployment values, not secrets — the same pair appears in
 # every Sentry URL. Project `godot-explorer` is DSN project id 4510187688361984.
 const ISSUE_URL_TEMPLATE := "https://dcl-regenesis-labs.sentry.io/issues/?query=%s"
+const FEEDBACK_LIST_URL_TEMPLATE := "https://dcl-regenesis-labs.sentry.io/issues/feedback/?project=4510187688361984&query=%s&start=%s&end=%s"
```
regenesis-claw: [P2] Two query params worth adding while Case 1 is being run, both cheap insurance for the "not verified yet" list in the description:

- `&utc=true` — Sentry pairs absolute `start`/`end` with a `utc` flag, and when it is absent the bounds can be read in the viewer's local timezone rather than as the UTC instants we are sending. Whoever opens the ticket may not be on UTC.
- `&mailbox=` — you noted the reference feedback sat in `mailbox=ignored`; the Feedback page defaults to the unresolved mailbox, so an auto-ignored entry would not appear even with a perfect time window.

If Case 1 shows the entry without either param, leave it as is — but if it comes back empty, these are the first two things to try.

godot/src/bug_report/bug_report_service.gd
```diff
@@ -158,8 +158,13 @@ static func _compose_description(description: String, diagnostics_link: String)
...
 			lines.append("Coordinates: %d,%d" % [position.x, position.y])
-	var diagnostics := diagnostics_link if not diagnostics_link.is_empty() else "unavailable"
-	lines.append("Internal diagnostics: %s" % diagnostics)
+	var event_url := str(sentry_links.get("event_url", ""))
+	lines.append(
+		"Internal diagnostics: %s" % (event_url if not event_url.is_empty() else "unavailable")
+	)
+	var feedback_url := str(sentry_links.get("feedback_url", ""))
+	if not feedback_url.is_empty():
+		lines.append("Sentry feedback search: %s" % feedback_url)
```
regenesis-claw: [P2] Heads-up, no action needed today: `"Internal diagnostics: %s"` only needs the new `not_translatable.txt` entry because gdformat wrapped it so the literal starts the line, which matches the extractor's `^\s*"..."\s*%` helper pattern. This sibling line escapes the scanner purely because it still fits on one line. If the label ever grows and gdformat wraps it, static checks fail for a reason unrelated to the change being made. Worth a one-line comment next to the exemption block, or just a note in your head — pre-adding the exemption may instead trip the stale-exemption report, so I would not change it now.

regenesis-claw (approved):
## Review — approved (only P2 findings)

Followed `REVIEW.md`. Small, contained change: `SentryUserFeedback.submit()` goes `String -> {event_url, feedback_url}` and the Intercom description gains an optional `Sentry feedback search:` line. Verified:

- **Consumer check:** `SentryUserFeedback.submit()` has exactly one caller (`bug_report_service.gd:66`); the only other mention of the class is a comment in `project_main_loop.gd:232` about the category tag, which is unchanged. The return-type change is fully contained.
- **Identity claim holds:** `sentry_seeder.gd` pins `SentryUser.id = Global.config.analytics_user_id` in `setup()` and re-pins it on wallet connect / profile change / logout, so `user.id` on the feedback entry is the same value the link filters on. `analytics_user_id` is a typed `String` defaulting to a generated uuid v4, so `str()` + `is_empty()` is a sound guard (no `<null>` hazard).
- **Failure path intact:** every early return still yields both keys empty, `_compose_description` keeps the `unavailable` fallback for diagnostics and simply omits the feedback line. A bug report still files with Sentry off.
- `Time.get_datetime_string_from_unix_time()` is UTC, so the `Z` suffix is correct; the query and both bounds are `uri_encode`d.
- **Security:** no issues. Org slug / project id are deployment values (not the DSN key), the only interpolated value is a locally generated uuid, and it is percent-encoded before reaching the URL.
- **CI:** all jobs green (Static checks, Linux, Clippy, Android, iOS export, Docker) at `4fe3a32`.

The substantive caveat is the one you already flagged: nothing in this diff proves the Sentry Feedback page honours `project` / `start` / `end` / `query=user.id:` in this form, and it can't be proven from a dev build. Worst case degrades gracefully (support sees an empty list rather than a broken ticket), but Cases 1-2 should run before this is relied on in support workflows. Notes below are all P2.

kuruk-mm (approved):
looks great!

PR #2918, opened 2026-09-16
fix(scene-runner): survive freed raycast colliders and clock steps

gonpombo8:
## What
Scenes no longer freeze when one of their raycasts hits a collider Godot already freed, and the scene thread no longer dies when the wall clock steps backwards.

## Why
Counter-Strike 1.6 (`boedo.dcl.eth`) loaded on Godot but never got past its second update: `get_raycast_hit` called `has_meta` on a null collider instance, the panic was swallowed, and the runner logged `stuck at state Raycasts, forcing completion` every frame while the scene showed nothing. The freed collider shows up with the lazy colliders from #2764. After that fix the same scene died once with `overflow when subtracting durations` because the scene thread measured `dt` with `SystemTime`. Neither change alters what a scene receives: a freed collider is simply not a hit, and `dt` is the same monotonic delta clamped at zero.

## Details

- `lib/src/scene_runner/components/raycast.rs`: convert the `collider` variant to `Gd<Object>` and return no hit when it is invalid before calling `has_meta` / `get_meta`.
- `lib/src/dcl/js/mod.rs`: `start_time` is an `Instant`; `dt = start_time.elapsed().saturating_sub(elapsed)`.

Backtrace of the raycast panic (macOS desktop build, main @ 7368e6e):
```
thread panicked: Function call failed: call -- instance is null
  raycast::get_raycast_hit   raycast.rs:236
  raycast::do_raycast         raycast.rs:184
  raycast::update_raycasts    raycast.rs:69
```
</details>

### Case 1 — Scene with continuous raycasts runs on Godot
**Setup:** signed-in profile.
**Steps:**
1. Open the app and wait for the explorer to load.
2. Jump to the world `boedo.dcl.eth`.
3. Wait for the Dust2 map to appear.

- [ ] **Expected:** the CS 1.6 "SELECT TEAM" menu appears over the map within a few seconds and the player can join a team. Before this fix the map loaded but the menu never appeared and the avatar stayed in third person.
- [ ] **Regression — plain scene:** jump to Genesis Plaza. Interactions and scene UI work as before.

### Case 2 — Scene survives a clock step
No QA step: the scene thread now cannot underflow its frame delta. Covered by Case 1 staying up for several minutes.


regenesis-claw (approved):
Both fixes are right and minimal, and the freed-collider guard mirrors the existing precedent in `scene_manager.rs:2387-2396` (same `try_to::<Gd<Object>>().ok()?` + `is_instance_valid()` shape, same rationale in the comment) — good consistency rather than a second invented pattern.

Verified against the diff:
- `Instant::now()` + `saturating_sub` removes the underflow path entirely; `elapsed` is still anchored at the same point in the thread, so `SceneElapsedTime` and the `dt` passed to `onUpdate` keep the same semantics. It also removes a forward NTP jump showing up as a giant `dt`, which the description doesn't claim but is a free improvement.
- The validity guard runs before any `call()` on the collider, and `RaycastHit::from_godot_raycast` is only reached after it, so the pointer-events helper (which does an unguarded `try_to::<Gd<Node3D>>()` + `get_name()` on the same dict) stays safe on this path.

Two non-blocking notes:

- **[P2] `RqtQueryAll` stops the whole hit list on a freed collider.** `get_raycast_hit` returning `None` ends the `while let` loop in `do_raycast`, so one stale body truncates every hit behind it for that frame. `rid` is readable from `raycast_result` without touching the collider, so this case could `set_exclude(rid)` and continue instead of bailing. It matches the pre-existing "collider is not in the scene → stops all query type" behaviour, so it isn't a regression from this PR — worth a follow-up issue at most.
- **[P2] Style:** the file already has `use godot::prelude::*`, which brings both `Gd` and `Object` into scope — `raycast_result.get("collider")?.try_to::<Gd<Object>>().ok()?` reads the same and matches `scene_manager.rs`. Also worth confirming `is_instance_valid()` isn't dead code here: depending on the gdext version, `try_to::<Gd<_>>()` may already reject a dead object, in which case `.ok()?` is doing the work. Harmless either way.

CI at review time: `📊 Static checks` green; `🔧 Clippy`, `🐧 Linux` and `🤖 Android` still pending, so the compile of the `try_to` change is **unverified** — please confirm Clippy goes green before merging. The red `🍏 iOS — resolve/dispatch`, `🤖 Android — resolve` and `Prepare distribution` jobs are the known distribution-pipeline noise (`Prepare distribution` log ends in `The operation was canceled`), not a signal on this diff. Per REVIEW.md §0.2 an iOS build is **not** required here: this is Rust scene-runner code with no iOS branch, no native plugin, no comms/av path.

Branch is up to date with `main` (ahead 2, behind 0). Description follows the What/Why/Details/Test-plan shape with a reproducible Case 1; Case 2 correctly declares no QA step.

Approving — the two notes are follow-ups, not merge blockers.

gonpombo8:
Both notes taken in 60a879bc: a freed body is now excluded by rid and QueryAll keeps scanning, and the guard uses the prelude Gd/Object names. Kept is_instance_valid() next to the try_to: it costs nothing and does not depend on which gdext version rejects dead objects. Clippy -D warnings and the macOS build pass locally.

kuruk-mm (approved):
looks great

Issue #2853, opened 2026-09-07
Locomotion Iteration #2: Scene Collider Generation & Collision Layers v Unity


rodrigo685:
##  **⚠️ Problem:**
Godot marks any scene mesh thinner than 1 cm on any axis as one-way, so thin floors and ramps collide only from above — players fall through geometry that is solid on desktop. The player also collides with a single layer, so scene-placed invisible blockers and several SDK collider categories do not exist for the avatar at all. Drives **Playtime**.

## **🏁 Scope:**
GLTF + primitive collider generation (`scene.rs`, `mesh_collider.rs`) and the player collision mask (`player.tscn`, `project.godot`). Character movement systems are the sibling Iteration 2 issue.

## **📝 Deliverables:**
- [ ] Thin-mesh colliders double-sided — remove the `< 0.01` planar/AABB one-way path (B2)
- [ ] Player collision mask widened from `CL_PHYSICS` alone to the 6 Unity-equivalent layers (B3)
- [ ] Invisible blocker colliders block the avatar
- [ ] Avatar-only collider layer added, routed like Unity's `SDKAvatarHit` (M13)
- [ ] `SDKAvatarTriggerArea` + `SDKEntityTriggerArea` reachable by the player mask
- [ ] Regression: creator-authored one-way colliders still behave one-way
- [ ] #1529, #1203 and #1748 verified as no longer reproducing
- [ ] `decentraland/docs` issue filed if creator-facing collider semantics change; linked here

## Metrics
Playtime. Fall-through incidents per session; no instrumented baseline — measure via the #905 reproduction checklist (B2: approach thin elevated floors from below, Unity blocks and Godot passes through; B3: walk into invisible blocker walls).

## Findings

| | Godot `main` | Unity |
|---|---|---|
| GLTF colliders | trimesh per mesh (`scene.rs:169`) | `MeshCollider`, `BakeMesh` (`ConfigureGltfContainerColliders.cs:93-100`) ✓ |
| Backfaces | only non-planar meshes; planar = AABB thinner than 0.01 (`scene.rs:219,227-229`) | always double-sided |
| Primitive colliders | box / sphere / cylinder / plane shapes | same mapping ✓ |
| Player mask | `CL_PHYSICS` only (`player.tscn:20`) | 6 layers incl. `InvisibleColliders`, `CharacterOnly`, `SDKAvatarTriggerArea` |
[...]

Issue #2704, opened 2026-08-14
[Bug] Collider and Tri Count in Blender not Matching Scene Limits


regenesis-claw:
### Platform

Not specified

### App Version

Not specified

### Device Information

Not specified

### Issue description

A creator reports that the collider count shown by the mobile Scene Limits tooling doesn't match what's authored in Blender, and also differs from the desktop count for the same scene.

- Scene built from 192 individual box collider objects in Blender.
- Mobile scene-limits panel reports **2,820 colliders** (188% of the 1,500 budget, flagged red).
- Desktop reports a different collider count for the same scene (see attached screenshots — geometries 1,095 / colliders 783 on one client vs. the mobile panel numbers above).
- Requester wants to understand where the discrepancy comes from (e.g. is each box collider being split into multiple primitives/sub-shapes on mobile? Is desktop counting geometries differently than colliders?).

### Expected behavior

Collider counts reported by the mobile Scene Limits tooling should be explainable/consistent with what's authored in the source content (Blender) and should not wildly diverge from the desktop-reported count for the same scene, so creators can trust the budget UI when optimizing.

### Screenshots / Media

4 images attached in Slack thread (not re-uploaded here — needs manual upload to this issue):
1. Blender viewport showing 192 collider objects (orange wireframe highlighted).
2. Mobile scene-limits panel: Colliders 2,820 / 1,500 (188%, red).
3. A geometries/tris counter overlay (desktop?) showing Tris: 97,667.
4. Another counter showing Geometries: 1,095 / Colliders: 783.

<img width="884" height="972" alt="Image" src="https://github.com/user-attachments/assets/6f4d4e56-7980-4311-957c-c6b40c880f31" />

<img width="551" height="498" alt="Image" src="https://github.com/user-attachments/assets/75b475f4-f580-435e-a3e3-9e34c0d94461" />

<img width="429" height="100" alt="Image" src="https://github.com/user-attachments/assets/60e8936d-1aa5-428c-8f10-b7a2c686efca" />

### Steps to reproduce

1. Author a scene with multiple simple box colliders (not a single merged mesh collider) in Blender.
2. Export/deploy the scene and open the mobile client's Scene Limits panel.
3. Compare the reported collider count against the actual number of collider objects in Blender, and against the desktop client's reported count for the same scene.

### Occurrence

Always (as reported)

### Related Issues

[...]

Issue #1553, opened 2026-02-26
Locomotion Iteration #3: Mobile Jump Animation Parity


rodrigo685:
##  **⚠️ Problem:**
The Godot jump animation is timed against a symmetric 10 m/s² gravity and a fixed 1.8 m height. #1557 replaces that arc with Unity's asymmetric gravity + long-jump hold, so `rise` / `fall` / `land` fire at the wrong moments the day it lands — the avatar hangs in `Jump_Rise` on the way down and snaps into `Jump_End` late. Drives **Playtime**.

## **🏁 Scope:**
Re-time the `AnimationTreeStateMachine` jump chain against the new arc and add the long-jump variant Unity drives with `LONG_JUMP`. Hard landing is its own Iteration 3 issue. Blend and jump-variation parity is Iteration 4.

## **📝 Deliverables:**
- [ ] `rise` / `fall` / `land` thresholds re-derived from the post-#1557 arc, not the current one
- [ ] Long-jump (hold) animation variant driven by the 0.5 s hold window
- [ ] `Jump_*` and `Run_Jump_*` chains re-timed against speed-lerped height 1.0 jog / 1.5 run
- [ ] Air-jump → `Double_Jump_Rise` transition verified against impulse `max(8, speed)`
- [ ] No trap states — regression guard on the escape paths #2699 added for `fall` / `gliding`
- [ ] Side-by-side jump capture vs desktop: tap, hold, run-jump, air-jump

## Metrics
Playtime. No instrumented baseline for animation desync — verify by frame-matched side-by-side capture vs desktop on test scene `-98,103`.

## Findings

Godot drives animation from an `AnimationTree` state chain: `Jump_Start → Jump_Rise → Jump_Mid → Jump_Fall → Jump_End`, mirrored as `Run_Jump_*`, plus `Double_Jump_Rise` and the `Gliding_*` states. Conditions bind in `avatar.gd` as `rise` / `fall` / `land` / `gliding`.

Unity does not use a chain — `AnimationStatesLogic.cs` sets a flag set the animator reads: `GROUNDED`, `JUMP_COUNT` (int), `LONG_JUMP`, `FALLING`, `LONG_FALL`, `STUNNED`, `GLIDING`, `GLIDE_BLEND` (float), plus `JUMP` / `START_GLIDING` triggers. Thresholds come from settings: `AnimationFallSpeed`, `AnimationLongFallSpeed`, `RunJumpHeight² × JumpGravityFactor`.

[...]

PR #943, opened 2025-12-13
refactor: locomotion and physics to match Foundation Client

kuruk-mm:
# ADR Draft

---
adr: XXX
date: 2024-12-13
title: Player Locomotion Specification
authors:
  - decentraland
status: Living
type: Standards Track
spdx-license: CC0-1.0
---

# Abstract

This document specifies the player locomotion system for Decentraland clients. It defines the physics parameters, movement mechanics, jumping behavior, and collision handling that clients must implement to ensure consistent player movement across all Decentraland implementations. The Foundation Reference Client is [unity-explorer](https://github.com/decentraland/unity-explorer).

## Context, Reach & Prioritization

Player locomotion is fundamental to the Decentraland experience. Consistent movement physics across clients ensures:

1. **Fair gameplay**: All players have the same movement capabilities regardless of client
2. **Content compatibility**: Scenes designed with specific movement constraints work consistently
3. **Predictable interactions**: Multiplayer experiences require synchronized movement expectations

This specification extracts the exact parameters and algorithms from the Foundation Reference Client to enable third-party client implementations.

# Specification

## 1. Character Collider

The player is represented by a capsule collider with the following dimensions:

| Parameter | Value | Unit |
|-----------|-------|------|
| Height | 1.6 | meters |
| Radius | 0.3 | meters |
| Center Y Offset | 0.8 | meters (from player origin) |
| Skin Width | 0.08 | meters |

### 1.1 Collision Layers

- The player collider MUST NOT collide with other player colliders
- The player MUST collide with world geometry and scene objects
- The player origin is at floor level (y=0 when standing on ground)

## 2. Movement Speeds

The player has four movement states with corresponding speeds:

| Movement Kind | Speed | Unit |
|---------------|-------|------|
| IDLE | 0 | m/s |
| WALK | 1.5 | m/s |
| JOG | 8.0 | m/s |
| RUN (Sprint) | 10.0 | m/s |

### 2.1 Movement Kind Selection

- **IDLE**: No movement input
- **WALK**: Movement input while walk modifier is held
- **JOG**: Movement input without modifiers (default)
- **RUN**: Movement input while sprint modifier is held

### 2.2 Movement Direction

Movement direction is calculated relative to the camera orientation:

```
cameraForward = camera.forward
cameraForward.y = 0
cameraForward = normalize(cameraForward)

cameraRight = camera.right
cameraRight.y = 0

moveDirection = (cameraForward * inputY) + (cameraRight * inputX)
[...]

PR #2920, opened 2026-09-16
feat(player): Jolt physics + locomotion parity pass 1 (capsule/step, jump/gravity, speed/air)

### LINKED ISSUE #1557: Locomotion Iteration #1: Jump Arc, Coyote Time & Gravity v Unity
##  **⚠️ Problem:**
Coyote time is unreachable in Godot — the ground-jump branch sits after `elif not on_floor`, so a jump pressed within 0.15 s of walking off a ledge is silently dropped, and off higher ledges the same press is instead consumed by the glide-open gate. Muscle memory for every platformer player, broken on mobile only. Drives **Playtime**.

## **🏁 Scope:**
Coyote reachability, gravity model, long-jump hold and jump height in `player.gd`. Air jump delay and jump cooldown already match Unity exactly — regression-guard, do not retune. Jump and landing animation is Iteration 3.

## **📝 Deliverables:**
- [ ] Coyote branch reachable — restructure the `elif` chain so a ground jump fires within 0.15 s of leaving the floor (B1)
- [ ] Glide-open gate no longer consumes a jump press inside the coyote window
- [ ] Asymmetric gravity — ascent ×4 = 40, descent 10
- [ ] Long jump: 0.5 s hold window applying ×0.5 gravity
- [ ] Jump height 1.8 fixed → 1.0 jog / 1.5 run, lerped by horizontal speed
- [ ] Air jump impulse `max(8, current_speed)` replaces the fixed 8.0
- [ ] Coyote window expressed in seconds, never tick counts
- [ ] Regression guard: jump buffer 0.15 s, jump cooldown 0.3 s, air jump delay 0.2 s unchanged
- [ ] Tap-vs-hold jump arc captured side-by-side vs desktop, test scene `-98,103`

## Metrics
Playtime. No instrumented baseline for dropped jump inputs — measure via the #905 reproduction checklist (B1: walk off a low ledge, jump within 0.15 s; Unity jumps, Godot drops the input).

## Findings

| | Godot `main` | Unity |
|---|---|---|
| Coyote time | **unreachable** — grace (`player.gd:391-395`) only drives the land animation; ground-jump branch (`:455-470`) sits after `elif not on_floor` (`:390`) | 0.15 s (`ApplyJump.cs:54`) |
| Jump buffer | 0.15 s (`player.gd:9`) | 0.15 s (`JumpTrigger.cs:8-9`) ✓ |
| Gravity | single 10 m/s² | asymmetric — ascent ×4 = 40, descent 10 (`ApplyGravity.cs:41-42`) |
| Long jump (hold) | none | 0.5 s window, ×0.5 gravity (`ApplyGravity.cs:37-38`) |
| Jump height | 1.8 fixed (`player.gd:59-60`) | 1.0 jog / 1.5 run, lerped by horizontal speed |
| Air jump | 1 · 2.0 m · 0.2 s delay · fixed impulse 8.0 (`player.gd:8,11-13`) | same values, impulse `max(8, current speed)` (`ApplyJump.cs:76-186`), behind `FeatureId.DoubleJump` |
| Jump cooldown | 0.3 s (`player.gd:10`) | 0.3 s (`ApplyJump.cs:80-83`) ✓ |

Symmetric single gravity cannot reproduce Unity's arc no matter how the height constant is tuned. Snappy going up + forgiving coming down comes from the ×4 ascent multiplier plus the hold window, not from the height value.

Tick reasoning does not transfer. Unity steps physics by frame delta — `Physics.Simulate(dt)`, `UpdatePhysicsSimulationSystem.cs:33` — not a fixed 50 Hz. "Coyote = 8 ticks" is nominal; port 0.15 **seconds**.

D2 resolved: 1.0 / 1.5 speed-lerped, matching Unity. Reachable-geometry set changes with it — 1.5 m @ 10 m/s replaces today's 1.8 m @ 11 m/s.

## **🔗 References:**
- [`ApplyJump.cs`](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/Character/CharacterMotion/Velocity/ApplyJump.cs) — coyote window, cooldown, speed-lerped height, air-jump impulse
- [`ApplyGravity.cs`](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/Character/CharacterMotion/Velocity/ApplyGravity.cs) — `JumpGravityFactor` ascent multiplier + `LongJumpTime` / `LongJumpGravityScale` hold window
- [#905](https://github.com/decentraland/godot-explorer/issues/905) — parent audit, §3.2
- [#1417](https://github.com/decentraland/godot-explorer/pull/1417) — closed unmerged; carried a coyote fix bundled with Jolt. Salvage the GDScript half.
- [#1553](https://github.com/decentraland/godot-explorer/issues/1553) — jump animation; changing the arc re-times the `rise` / `fall` / `land` thresholds

### LINKED ISSUE #2753: Locomotion Iteration #1: Avatar Capsule, Step Offset & Ground Contact v Unity
##  **⚠️ Problem:**
The player capsule does not reach the ground — it spans y 0.45 → 1.95 while the avatar's feet are at 0, so the volume that collides with scene geometry sits half a metre above the character and is narrower and shorter than the Unity reference. Wrong hitbox on every collision the player has. Drives **Playtime**.

## **🏁 Scope:**
Player collider shape and dimensions, plus the ground-contact set that has to land with it: step offset, floor snap / downslope stick, slope limit (audit §2 + §3.1). Jump and gravity are #1557, speed and acceleration are #2850, scene-collider generation is Iteration 2.

## **📝 Deliverables:**
- [ ] Player capsule height 1.6, radius 0.3, margin 0.08, center Y 0.8 — span 0.0 → 1.6
- [ ] `SeparationRayShape3D` re-evaluated against the new capsule position
- [ ] Duplicate capsules in `stuck_detector` and `camera_mode_area_detector` kept consistent
- [ ] Custom step offset 0.35 implemented — `CharacterBody3D` has no built-in (M1)
- [ ] Floor snap replaced by downslope raycast 0.45 jog / 0.55 run
- [ ] Slope limit set explicitly to 46°, not left at the engine default 45°
- [ ] Before/after climb + clearance comparison recorded in the test scene
- [ ] 0.35–0.5 m ledges walked up, not jumped — verified vs desktop

## Metrics
Playtime. No baseline for mobile climb-failure rate exists — needs one before/after. Measure via the #905 reproduction checklist (M1: walk into 0.35–0.5 m ledges; Unity steps up, Godot blocks). Measurement approach TBD with Data.

## Findings

| | Godot `main` | Unity |
|---|---|---|
| Height | 1.5 | 1.6 |
| Radius | 0.25 | 0.3 |
| Center Y | 1.2 | 0.8 |
| Vertical span | 0.45 → 1.95 | 0.0 → 1.6 |
| Margin vs skin width | 0.05 | 0.08 |
| Step offset | none — no built-in, no custom logic | 0.35, force-written every physics tick (`ApplySlopeModifier.cs:24`); prefab serializes 0.3 |
| Floor snap / downslope stick | `floor_snap_length = 0.2` (`player.gd:218`) | downslope raycast 0.45 jog / 0.55 run (`ApplySlopeModifier.cs:40,42`) |
| Slope limit | unset → engine default 45° | 46° (`CharacterObject.prefab:103`) |

Godot: `CharacterBody3D`, `collision_layer` 4 (`CL_PLAYER`), `collision_mask` 2 (`CL_PHYSICS`).
Unity: `CharacterController`, layer 16, collides with Default, Floor, InvisibleColliders, CharacterOnly, SDKAvatarTriggerArea, SDKEntityTriggerArea.

Constraints the fix has to account for:

- Godot carries a second shape Unity has no equivalent for — `SeparationRayShape3D`, length 1.0779, at y=1.1, basis rotates local +Z to world −Y, so it points down and ends at y≈0.022. It exists to bridge the 0.45m gap under the capsule. Its geometry is tuned to center Y = 1.2 and does not survive a change to that value untouched.
- That same 0.45m gap is currently the only thing letting the player clear low geometry: `CharacterBody3D` has no built-in step offset and the repo adds none, so sub-0.45m obstacles never touch the capsule. **Godot's de-facto "step clearance" is not step logic — it is the floating capsule rolling over low geometry.** Moving the capsule down removes that clearance, so step offset 0.35 has to ship in the same PR or climbing gets worse, not better.
- Two more capsules share the current dimensions at offset y=1.153: `stuck_detector.tscn` (Area3D, mask `4294967295`) and `camera_mode_area_detector.tscn`.
- PR #943 resized the capsule to 1.6 / 0.3 / margin 0.08 but left center Y at 1.2; PR #1417 did the same. Both closed unmerged 2026-04-19.

Prefab values are not the source of truth: `CharacterObject.prefab` serializes `StepOffset` 0.3, but `ApplySlopeModifier.cs:24` overwrites it with 0.35 every physics tick. Port 0.35. D3 (snap 0.2 vs raycast 0.45/0.55) and D4 (45° vs 46°) resolved above in favour of Unity.

## **🔗 References:**
- [decentraland/unity-explorer — `CharacterObject.prefab`](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/Character/CharacterObject/CharacterObject.prefab) — source of the Unity column: `m_Height` 1.6, `m_Radius` 0.3, `m_Center` (0, 0.8, 0), `m_SkinWidth` 0.08, slope limit 46°
- [`ApplySlopeModifier.cs`](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/Character/CharacterMotion/Velocity/ApplySlopeModifier.cs) — the live step offset 0.35 and the downslope raycast lengths
- [`godot/src/logic/player/player.tscn`](https://github.com/decentraland/godot-explorer/blob/main/godot/src/logic/player/player.tscn) — the two collision shapes to change
- [#905](https://github.com/decentraland/godot-explorer/issues/905) — parent audit, §2 + §3.1
- [#943](https://github.com/decentraland/godot-explorer/pull/943), [#1417](https://github.com/decentraland/godot-explorer/pull/1417) — closed unmerged; both attempted this resize, neither moved center Y

### LINKED ISSUE #2844: Locomotion Iteration #1: Migrate Godot Physics to Jolt
##  **⚠️ Problem:**
Godot Physics resolves capsule-vs-trimesh contacts inconsistently on thin and non-convex scene geometry — players tunnel through floors and stick on ramps that hold on desktop. Drives **Playtime**; no instrumented mobile baseline for fall-through rate exists today.

## **🏁 Scope:**
Swap the 3D physics server to Jolt on all build targets and revalidate character, scene-collider and raycast behavior. No locomotion tuning here — parameter parity ships in #1557, #2850, #2753.

## **📝 Deliverables:**
- [ ] Jolt enabled as 3D physics server; Godot Physics removed from build config
- [ ] iOS + Android + macOS + Windows + Linux builds green
- [ ] `SeparationRayShape3D` under the player verified against Jolt contact resolution
- [ ] Scene trimesh + primitive collider generation (`scene.rs`, `mesh_collider.rs`) regression-tested
- [ ] SDK raycast hit distances unchanged on flat + sloped geometry — guards #2830
- [ ] `stuck_detector` and `camera_mode_area_detector` Area3D overlaps unchanged
- [ ] Physics tick cost measured before/after on a mid-tier Android device
- [ ] Test scene `-98,103` walkthrough recorded before/after

## Notes
The #905 audit (§7) recommends against bundling the engine swap with parity behavior changes: all 6 parity bugs (B1–B6) are solver-independent script logic, and Jolt does not fix B2 — the one-way flag is set in `scene.rs`, not by the solver. Landing an engine swap and behavior changes in the same window makes regressions unattributable, the failure mode that killed #943 and #1417.

Product call: ship it in Iteration 1 anyway. Mitigation — keep Jolt in its own PR, separate from every parity fix, so a bisect stays possible.

## **🔗 References:**
- [#905](https://github.com/decentraland/godot-explorer/issues/905) — parent audit; §7 is the Jolt evaluation this issue overrides
- [#943](https://github.com/decentraland/godot-explorer/pull/943), [#1417](https://github.com/decentraland/godot-explorer/pull/1417) — both bundled Jolt with locomotion changes, both closed unmerged 2026-04-19
- [#1529](https://github.com/decentraland/godot-explorer/issues/1529) — one-way colliders act two-way (SPIKE, closed); root cause is our flag, not the solver — Jolt will not close it
- [#2830](https://github.com/decentraland/godot-explorer/issues/2830) — raycast `hitDistance` grows on sloped geometry, mobile only; the solver swap must not move it

### LINKED ISSUE #2850: Locomotion Iteration #1: Speed, Acceleration & Air Behavior v Unity
##  **⚠️ Problem:**
Godot assigns locomotion velocity directly — no acceleration, no air drag, full air control — so the avatar starts, stops and turns instantly where Unity carries weight. Movement reads "on rails" against the desktop client. Drives **Playtime**.

## **🏁 Scope:**
Speed constants, acceleration/deceleration curves, air control and air drag in `player.gd`. Jump arc and gravity stay in #1557. Slope speed modifier stays in Iteration 2.

## **📝 Deliverables:**
- [ ] Run speed 11 → 10; walk 1.5 and jog 8 confirmed unchanged
- [ ] Ground acceleration 20 → 25 over a 0.5 s curve
- [ ] Air acceleration 15 → 20 over a 0.5 s curve
- [ ] Deceleration dt-scaled — stopping distance independent of physics tick rate (B4)
- [ ] Quadratic horizontal air drag, coefficient 0.2
- [ ] Air control reduced vs grounded — air accel pair + MoveTowards, not full ground parity
- [ ] Start / stop / turn captured side-by-side vs desktop on test scene `-98,103`
- [ ] Stopping distance verified constant under forced frame drops

## Metrics
Playtime. No instrumented baseline for movement feel — measure via the #905 reproduction checklist (B4: release keys at full run under frame drops; Unity stops consistently, Godot varies with tick rate).

## Findings

| | Godot `main` | Unity |
|---|---|---|
| Walk / jog / run | 1.5 / 8 / 11 | 1.5 / 8 / 10 |
| Acceleration | none — velocity assigned directly (`player.gd:513-514`) | ground 20 → 25, air 15 → 20, 0.5 s curve (`ApplyCharacterMovementVelocity.cs:72-78`) |
| Deceleration | `move_toward(v, 0, walk_speed)` per tick, not dt-scaled ⇒ 90 m/s² at 60 Hz (`player.gd:520-521`) | SmoothDamp, `StopTimeSec = 0` ⇒ instant |
| Air drag | none on locomotion velocity | quadratic, 0.05 × 4 = 0.2 (`ApplyHorizontalAirDrag.cs`) |
| Air control | full, identical to grounded | reduced — air accel pair + MoveTowards |

B4 is a latent bug, not a tuning gap: the deceleration step is per-tick, not per-second. Correct at a stable 60 Hz, wrong under load — which is where mobile lives.

**Replicate behavior, not asset values.** Dead on the Unity side: `StopTimeSec = 0` (degenerate SmoothDamp = instant stop), `GroundDrag`, and the whole `ApplyAirDrag.cs` file — the live one is `ApplyHorizontalAirDrag.cs`.

D1 resolved: run 10, matching Unity. The reachable-geometry set changes with it — scenes are authored against desktop reachability.

## **🔗 References:**
- [`ApplyCharacterMovementVelocity.cs`](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/Character/CharacterMotion/Velocity/ApplyCharacterMovementVelocity.cs) — the acceleration pair and 0.5 s curve to port
- [`ApplyHorizontalAirDrag.cs`](https://github.com/decentraland/unity-explorer/blob/dev/Explorer/Assets/DCL/Character/CharacterMotion/Velocity/ApplyHorizontalAirDrag.cs) — live air drag; `ApplyAirDrag.cs` beside it is dead, do not port
- [#905](https://github.com/decentraland/godot-explorer/issues/905) — parent audit, §3.4
- [#943](https://github.com/decentraland/godot-explorer/pull/943) — closed unmerged; its body is an ADR-grade locomotion spec, reusable as the implementation checklist
- [#2753](https://github.com/decentraland/godot-explorer/issues/2753) — capsule fix; land it first, speed changes read differently on a capsule that reaches the ground

manuelmaceira:
## What
Player physics switches from Godot Physics to Jolt, and the avatar's locomotion moves to the Unity reference values: the capsule now reaches the ground (1.6/0.3, center 0.8), steps up to 0.35m are walked up with a predictive step-up (no blocked frame, no teleport), coyote time works, the jump arc uses asymmetric gravity (snappy up, floaty down + hold-to-float ≈ doubles the arc), and movement has real acceleration, reduced air control, air drag, and a 10 m/s run speed.

## Why
Sprint 25 locomotion/parity pass 1 (audit #905). The player capsule floated 0.45m above the feet with no step logic, coyote time was unreachable code, gravity was a single symmetric 10 m/s², and velocity was assigned directly with a per-tick deceleration that breaks under frame drops (B4). Each fix is its own commit so a regression bisects cleanly — the Jolt engine swap lands alone, per the #2844 mitigation (this is what killed #943 and #1417).

Deliberately NOT in this PR: scene-collider generation (thin-mesh one-way, B2 — that's Iteration 2, #2853), slope speed modifier (Iteration 2), jump animation re-timing (Iteration 3, #1553), and the scene-limits collider counting issue #2704 (unrelated subsystem).

Closes #2844, closes #2753, closes #1557, closes #2850

## Details

**Commit 1 — Jolt (#2844).** `physics/3d/physics_engine="Jolt Physics"`; Jolt is compiled into the fork's editor and all export templates (verified by binary inspection). Headless probe (capsule drop, `SeparationRayShape3D`, raycast) behaves identically between engines on the baseline case. Gotcha worth knowing: comments in `project.godot` must use `;` — a `#` comment silently breaks ConfigFile parsing of the section.

**Commit 2 — Capsule & ground contact (#2753).** Capsule becomes 1.6 high / 0.3 radius (margin 0.08, center 0.8) instead of 1.5/0.25 floating at 0.45. `stuck_detector` and `camera_mode_area_detector` duplicate capsules synced to the same span/center (the camera detector capsule also moves from y=1.153 to 0.8). `safe_margin` set explicitly to 0.01 (the Godot default, made explicit for review). Downslope stick uses dynamic `floor_snap_length` (0.45 jog / 0.55 run, Unity `DownwardsSlopeJogRaycastDistance`/`Run`) so `is_on_floor()` survives downhill moves. Slope limit explicit 46°.

**Commit 3 — Jump & gravity (#1557).** The elif chain put the ground-jump branch after `not on_floor`, so a press within 0.15s of leaving a ledge was dropped (B1) or eaten by the glide-open gate. Restructured so air branches exclude the coyote window and ground jump accepts it; the coyote branch additionally requires `velocity.y <= 0` so an ascending jump-pad velocity can't re-trigger a ground jump. Gravity is asymmetric (ascent ×4 = 40, descent 10) with the 0.5s hold window at ×0.5 (`_current_gravity()` — faithful port of `ApplyGravity.cs`: the modifier applies on ascent and early descent). Jump v0 comes from the exact `GetJumpHeight` port: height lerps 1.0 jog → 1.5 run by horizontal speed, v0 = √(2·h·g·ascentFactor). Air jump impulse `max(8, current speed)`, air jump height 2.0. Untouched on purpose (regression guard): jump buffer 0.15s, cooldown 0.3s, air jump delay 0.2s.

**Commit 4 — Speed & air behavior (#2850).** Faithful port of `ApplyCharacterMovementVelocity.cs`: target velocity comes from the raw input direction (not the smoothed heading) so the ADAD sign-flip preserves momentum into the new direction; acceleration follows the real 4-key curve (plateau 0→0.1, ramp to 0.9 over 0.5s) driving the accel pairs 20→25 ground / 15→20 air; deceleration is an instant stop (Unity `StopTimeSec = 0`); air with no input drifts toward zero at the air accel rate (reduced air control), plus quadratic air drag 0.2 (`ApplyHorizontalAirDrag.cs`; the sibling `ApplyAirDrag.cs` is dead in Unity and was not ported).

**Commits 5–8 — Step-up iteration & review fixes (#2753, #1557, #2850).** The original step probe (a single down-ray from +0.35) failed against real trimesh geometry; QA iteration replaced it with the Unity CharacterController cast sequence in `_step_up()`: (a) a low forward cast measures the real distance to the riser face (the resting gap varies), (b) an up cast measures headroom/lift, (c) a forward cast at the raised height detects "taller than a step" — with a bevel/rounded-edge tolerance that measures the surface just past the raised contact (PhysX reasons from the contact point, not from a free-space sweep), and (d) the landing height comes from a small margin-less sphere cast from a fixed height — a ray slips through trimesh tri edges, and a ray starting at feet+lift begins exactly on the surface for near-max steps. Shape tests use a margin-less clone of the capsule (the 0.08 skin width eats the clearance band). The **predictive pass** runs before `move_and_slide`: it casts the intended frame motion, and on a step it rises Y in place so the horizontal motion flows unobstructed — PhysX steps inside the Move, so there is no blocked frame and no stop/restart cadence; a post-move fallback covers direction changes mid-contact. Guards: no headroom, no floor below (gap), flat-or-lower, and >0.35+ε = wall. All four probe scenarios (0.3 step, 0.4 block, 4×0.15 staircase, 60° bevel) pass identically under Jolt and Godot Physics. `SeparationRayShape3D` is **removed** from the player (inert with a grounded capsule — it only ever corrected the floating-capsule case; no Unity equivalent; the camera/stuck detectors keep theirs).

**Commit 9 — Ramp re-trigger gate (#2753, review P2).** The predictive step-up re-evaluates every blocked tick, so a continuous slope steeper than the 46° limit (but not vertical) got stair-climbed in ~0.31m bands — one rise per tick — before wedging mid-face (probe: 50° ramp climbed 1.23m, 55° climbed 1.31m, identical on both engines). Fix: after a TALL rise (> 0.2m) the step-up disarms until the capsule rests — one grounded, wall-free frame. A ramp face never rests (constant wall contact / airborne), a staircase tread (0.15m) never disarms, and terrain bumps below 0.2m rise without arming side-effects. The floor snap is zeroed only for the rise frame's move so the lower floor cannot re-glue the capsule mid-crossing. "Grounded" follows the house idiom `is_on_floor() or y <= 0` (player.gd:553): the realm floor's y-clamp eats penetration, so `is_on_floor()` alone is false during normal walking — gating on it alone locked the step-up out permanently in the live client (caught via an instrumented QA walk). Probe: A/B/C/D unchanged, 50/55° ramps stop at one band (~0.31m pop) on both engines. Live QA on a dedicated geometry scene: A/C/D climb smoothly, B and both ramps block.
</details>

### Case 1 — Step up low ledges without jumping
**Steps:**
1. Open the app and enter the test scene (`decentraland://open?position=-98,103&realm=jezter.dcl.eth`).
2. Walk (don't jump) into a step/ledge about 0.3m high (knee-high). Try a staircase if available.

- [ ] **Expected:** the avatar walks up onto the ledge smoothly with no jump and no visible pop; on stairs it climbs continuously, step by step, without stuttering.
- [ ] **Regression — walls still block:** walk into a wall taller than ~0.35m (e.g. 0.5m chest-high). The avatar stays blocked, no climbing, no teleport.

### Case 1b — Sustained steep ramp must block (P2 regression)
**Steps:**
2. Find (or build) a continuous ramp steeper than 46° but not vertical (e.g. 50–55°, several meters long). Walk into it and keep pushing forward for ~3 seconds.

- [ ] **Expected:** the avatar blocks at the base — at most a single ~0.3m pop, then no further progress; it does not stair-climb the slope band by band.
- [ ] **Regression — stairs still chain:** a staircase of ≤0.3m treads is still climbed smoothly, step by step (the re-arm must fire on every tread).

### Case 2 — Coyote jump
**Steps:**
2. Walk off a ledge whose drop is at least ~0.6m deep (a shallow ledge would just snap the capsule back down), and press jump within a split second (~0.15s) of leaving the edge.

- [ ] **Expected:** the jump fires even though the avatar already left the floor.
- [ ] **Regression — glide unchanged:** walk off a HIGH ledge and press jump after falling visibly. Glider opens as before (no accidental ground jump).
- [ ] **Regression — jump pads:** if reachable, jump off an updraft/jump pad and confirm no double-jump re-trigger while still ascending.

### Case 3 — Jump arc: tap vs hold
**Steps:**
2. Tap jump quickly; then jump again holding the button the whole time. Repeat while sprinting.

- [ ] **Expected:** tap = short snappy hop (apex ≈ 1.0m jog / 1.5m run); hold = floaty descent and a noticeably higher arc (apex ≈ 2.0m jog / 2.8m run — the hold window halves gravity for the first 0.5s). Sprint-jump reaches higher than jog-jump.

### Case 4 — Acceleration and stopping
**Steps:**
2. Start running from standstill, then release the stick at full speed. Reverse direction at speed (ADAD). Repeat a few times.

- [ ] **Expected:** the avatar eases into the run (no longer instant full speed); on release it stops almost instantly (Unity `StopTimeSec = 0`) with no long skid; reversing at speed keeps momentum into the new direction instead of sliding sideways.
- [ ] **Regression — air:** jump and release the stick in the air; the avatar should bleed off horizontal speed slowly (reduced air control), not instantly.

### Case 5 — Engine swap regression (Jolt)
**Steps:**
1. Open the app and walk around Genesis Plaza and the test scene for a few minutes: ramps, stairs, thin floors, slopes.

- [ ] **Expected:** no falling through floors, no sticking on ramps, ground contact feels like before.
- [ ] **Regression — interaction raycasts:** hover scene items (buttons, doors). Highlight/tooltip distance behaves as before.


regenesis-claw (commented):
Reviewed against `REVIEW.md`. This is a genuinely well-structured PR — one commit per concern so the Jolt swap bisects cleanly, every parity constant cites its Unity source file, the `_current_gravity()` / step-offset comments explain *why*, and the coyote restructure (moving the ground-jump branch so the air branch excludes the window) is the correct fix for B1. Leaving this as a **comment, not a verdict**, because it's a draft and you say device QA is still pending — but flagging what would block it as-is.

## Pre-flight (REVIEW.md §0)

- **Branch is 57 commits behind `main` and `diverged`.** §0.1 says hold substantive review until it's updated — CI signals on this base aren't trustworthy, and this diff touches the same `_physics_process` chain that #2778 landed the `resolve_is_grounded` / `GROUNDED_GRACE_WINDOW` work into.
- **`📊 Static checks` is RED** — `gdformat -d` wants to rewrap `player.gd` (the two long lines are the `_accel_weight` `move_toward(...)` and the ternary `var accel := lerpf(...) if on_floor else lerpf(...)`). That reds the whole run: Linux, Clippy, Android and Docker all show `skipping`, so this PR currently has **zero build signal at all**.
- **No iOS/Android artifact, and this is an engine-level change.** §7 says a change to the Godot engine config escalates to a human-stakeholder call, and swapping the 3D physics server is in that class — it silently reaches scene colliders, `Area3D` detection (CameraModeArea, AvatarModifierArea, trigger areas) and every pointer/interaction raycast, not just the player. Needs the `build` label and a real device run on both platforms before it leaves draft. Agreed this is the point of the draft; noting it so it isn't skipped later.

## P1

**1. Test plan Case 1 contradicts the implementation — QA will mark it failed.**
Case 1 tells the tester to walk into a **"~0.35-0.5m"** ledge and expect it to be walked up. `STEP_OFFSET` is `0.35`, and your own commit message on the capsule commit records the probe result as *"steps onto 0.3m, blocked by 0.4m"*. A tester picking a 0.45m block from that stated range gets a blocked avatar and files a bug. Narrow Case 1 to `≤0.3m` and move `0.4-0.5m` into the regression line next to the chest-high wall.

**2. `camera_mode_area_detector.tscn` is an SDK-visible behaviour change that isn't in `## What` / `## Why`.**
That Area3D is what decides when a scene-authored camera-mode area takes effect, and the capsule goes from radius 0.25 / span 0.4-1.9 to radius 0.3 / span 0.0-1.6. Scenes with a camera-mode volume sitting near the ground or at head height now trigger at different positions than they do in production today. The description frames it as "duplicate capsules synced", which reads as a cosmetic tidy-up — per the AG rule in §4, a ride-along behaviour change outside the stated scope belongs in What/Why, not in Details. There's also no test-plan case that enters one. Same question for `stuck_detector.tscn`, though that one's blast radius is smaller.

**3. `in_coyote` isn't gated on falling, so it also swallows the first 0.15s of every *upward* ungrounded moment.**
`player.gd:423` — `var in_coyote := not on_floor and time_falling <= COYOTE_WINDOW`. `time_falling` counts any airborne time, including a jump-pad or scene-impulse launch. Two consequences inside that window:
- The `elif not on_floor and not in_coyote` branch is skipped, so `avatar.rise` / `avatar.fall`, the air-jump arm, the glide gates and the glide fall-speed clamp all stop updating for 150ms after launch.
- Worse, the branch that *is* reachable is the **ground**-jump branch (`on_floor or in_coyote`), whose only extra gate is `_time_since_last_jump >= JUMP_COOLDOWN` — and a jump pad never sets `_time_since_last_jump`. So a player who taps jump right after a pad launch gets a ground jump that overwrites `velocity.y` and cancels the launch.

Coyote time is a fall-only affordance in Unity too. One-line fix:

```suggestion
	var in_coyote := not on_floor and velocity.y <= 0.0 and time_falling <= COYOTE_WINDOW
```

## P2

**4. `_try_step_up` never collision-tests the destination, and accepts any surface as a step.** (`player.gd:687-714`)
`test_move(raised, probe)` validates the sweep with the capsule bottom at `+0.35`, but the body is then teleported to `floor_y`, which can be up to 0.30m *below* that — the band between `floor_y` and `+0.35` at the destination XZ is never tested, so an overhang or a thin slab in that band gets ignored. The ray hit's normal also isn't checked, so a narrow railing or fence rail top at 0.3m is a valid "step" and the player teleports onto it. Suggest `hit.normal.angle_to(Vector3.UP) <= floor_max_angle` before accepting, plus a `test_move` at the final transform.

**5. `safe_margin = 0.01` on the Player node isn't mentioned anywhere in the description.** The capsule `margin` 0.05→0.08 is explained; this one (10x the engine default) isn't. One line in Details on why it's needed under Jolt.

**6. Hold-to-float doubles the jump rather than softening the descent — worth stating explicitly.** With ascent gravity 40, time-to-apex at jog is ~0.22s, well inside the 0.5s hold window, so holding the button halves gravity for the *entire* ascent: apex goes 1.0→2.0m at jog and ~1.5→2.8m at run. The stated 1.0/1.5 parity numbers are tap-only, and the real ceiling is well above the old flat 1.8m. Design and QA should know that before it's tuned on-device.

**7. Case 2 may not be reproducible as written.** `floor_snap_length` goes from 0.2 to 0.45/0.55, so walking off any ledge shorter than the snap length now glues the avatar to the ground instead of leaving the floor — the coyote window never opens. Specify a ledge taller than 0.55m in the steps.

**8. "Jolt is compiled into all export templates (verified by binary inspection)" isn't verifiable from the diff or CI.** If a template *doesn't* carry the module, Godot falls back to Godot Physics with only a console warning — which would mean the whole parity pass silently behaves differently per platform. Worth either a line in Details saying how each target was checked, or a startup assertion that the active 3D server is the expected one.

## Verified as claimed

- `jump_velocity_0` has no remaining references outside `player.gd` — clean removal.
- Rust defaults in `dcl_locomotion_settings.rs` match the new GDScript defaults (10.0 / 1.0 / 1.5), and because `v0` is computed with the ascent gravity, a scene-authored `jump_height` still means "apex in metres" exactly as before.
- The new if/elif chain is a single chain — gravity is applied exactly once per tick on every path (I checked this specifically; the added trailing `else` is coyote-only, not a second integration).
- Sprint *is* reachable on mobile (`virtual_joystick.gd` drives `ia_sprint`), so the speed-lerped jump height isn't dead on touch.
- No new `push_error` / `printerr` / `tracing::error!`, no debug prints, no non-English comments, no `.claude/` or scratch files in the diff.

Happy to re-review once the branch is updated and static checks are green.

regenesis-claw (commented):
Delta re-review at head `1a06b5f` (this was a rebase, not an append — all four original commits got new SHAs plus 4 new fix commits and a merge from `main` — so I did a fresh full read of `player.gd`, not a diff-of-diff).

## Previous findings — status

| # | Finding | Status |
|---|---|---|
| Pre-flight | Branch 57 behind / diverged | ✅ now `ahead 9, behind 0` of `main` |
| Pre-flight | Static checks red (gdformat) | ✅ `📊 Static checks` passes now |
| Pre-flight | No iOS build for an engine-level change | ⏳ in progress — see CI section below |
| P1-1 | Test plan Case 1 contradicts `STEP_OFFSET`/commit evidence | ✅ narrowed to "~0.3m", 0.5m moved to the wall regression line |
| P1-2 | `camera_mode_area_detector.tscn` capsule resize undocumented | ✅ now called out in Details ("the camera detector capsule also moves from y=1.153 to 0.8") — still only in Details not top-level Why, minor, not blocking |
| P1-3 | `in_coyote` not gated on falling — swallows jump-pad launches | ✅ fixed exactly as suggested (`velocity.y <= 0.0` guard), plus a new jump-pad regression case added to Case 2 |
| P2-4 | `_try_step_up` never collision-tests the destination, accepts any surface | ✅ fully rewritten — predictive pre-move cast + post-move fallback, proper CCT-style sequence (face-distance cast, headroom cast, bevel-tolerant forward cast, margin-less sphere-cast landing), explicit height ceiling (`floor_y > global_position.y + STEP_OFFSET + 0.005` → treated as wall). Real upgrade, addresses the finding more thoroughly than I asked for |
| P2-5 | Undocumented `safe_margin = 0.01` | ✅ documented in Details |
| P2-6 | Hold-to-float doubles the jump, stated numbers were tap-only | ✅ `## What` now says "hold-to-float ≈ doubles the arc" and Test plan Case 3 states both apex ranges explicitly; verified the math again against the reordered `_current_gravity()` (hold multiplier now applies before the ascent multiplier) — still exactly 2x, checks out |
| P2-7 | Case 2 possibly unreproducible (bigger `floor_snap_length` glues short ledges) | ✅ Case 2 now specifies "drop at least ~0.6m deep" |
| P2-8 | "Jolt in all export templates, verified by binary inspection" unverifiable from here | unchanged — still asserted, not something I can check from the diff. Not blocking, just still unverified |

No new P1s from the rewrite. One new P2:

**New P2 — continuous slope-climbing isn't in the stated test matrix.** The four validated probe scenarios (0.3 step, 0.4 block, 4×0.15 staircase, 60° bevel) are all *discrete* geometry. `_step_up` is re-evaluated from the player's current position every tick it's blocked, with no cooldown between calls — on a continuous ramp steeper than `floor_max_angle` (46°) but shallower than vertical, each tick could in principle find a valid ≤0.35m rise just ahead and climb it, chaining into de-facto wall-climbing over several ticks even though a single tick's rise is correctly bounded. The 60° bevel test is reassuring (close to this case) but tests a single edge, not a sustained incline. Worth an explicit QA case: walk continuously into a ramp steeper than 46° for a few seconds and confirm you don't climb it.

## CI status (as of this review)

`📊 Static checks` and `🔧 Clippy & Build Checks` are green. `🐧 Linux` and `🤖 Android` builds are still running. The mobile distribution pipeline (iOS export → R2, Android APK) is also mid-run — build `2036` / `1.14.0.2036-1a06b5f-staging`. The sticky Build Report's "Android: ❌ Failed" line is from a *stale, cancelled* run (`35733028964`, superseded when the branch was pushed again) — not a live failure, ignore it.

Holding the formal verdict until the current Linux/Android builds and the iOS export finish — this is now a real (non-draft) engine-level physics change and I'd rather approve against a green build than a promise. Ping me once they're through and I'll close this out; if they're green this comes down to a straightforward approve given everything above is fixed. No P0/P1 blockers remain in the code itself.

manuelmaceira:
P2 addressed in `dbf1baae5b` — and the theoretical concern turned out to be real: a headless probe mirroring the shipped cast sequence (Jolt + Godot Physics, both) climbed a continuous 50° ramp by **1.23m** and a 55° ramp by **1.31m** in ~0.35m bands, one rise per tick, exactly the mechanism described.

**Fix:** after a *tall* rise (> 0.2m) the step-up disarms until the capsule rests — one grounded, wall-free frame. A ramp face never rests (constant wall contact / airborne), a staircase tread (0.15m) never disarms, and terrain bumps < 0.2m rise without arming side-effects. Floor snap is zeroed only for the rise frame's move so the lower floor can't re-glue the capsule mid-crossing. Probe post-fix: 50°/55° stop at one band (~0.31m pop) while the 0.3 step / 0.4 wall / 4×0.15 staircase / 60° bevel results are unchanged on both engines.

Two findings from live QA worth knowing:

1. `is_on_floor()` is **false during normal walking** in the live client — the game-wide grounded idiom is `is_on_floor() or position.y <= 0` (player.gd:553, 1294, 1329), because the realm floor's y-clamp eats the penetration that would register contact. An earlier iteration of the re-arm gate keyed on `is_on_floor()` alone and locked step-up out permanently; caught via an instrumented QA walk (per-frame floor/wall/y/vy logging).
2. The realm ground is bumpier than expected — the predictive pass legitimately rises over 0.05–0.27m terrain bumps on flat-looking ground. Gating disarm on *rise height* (> 0.2m) is what keeps those from producing choppy disarm windows before real steps.

Live QA on a dedicated geometry scene (0.3 step, 4×0.15 stairs, 60° bevel, 0.4 wall, 50° + 55° ramps): climbable set smooth, block set blocks. New explicit QA case added to the test plan (Case 1b — sustained steep ramp).

Builds: noted on the stale Android line; ping when Linux/Android/iOS come through and I'll keep an eye as well.

regenesis-claw (approved):
All findings from every review round are fixed, CI is fully green on the current head (Linux/Android/Docker/Static/Clippy), and this time the iOS/Android mobile-distribute build is correlated to the current head 5ff12ae (not the earlier stale dbf1baa run) — iOS export -> R2, TestFlight wait, and Android -> R2 all passed. Approving.

PR #2922, opened 2026-09-16
feat: add asset renderer mode for wearable and emote stills

LautaroPetaccio:
## What

A new headless CLI mode that renders still PNGs of individual wearables and emotes, consumed by the asset-vision pipeline (renders feed a vision model for metadata extraction: description, theme, tags, attributes, colors).

```
./decentraland.godot.client.x86_64 --rendering-driver opengl3 --asset-renderer --asset-input-file input.json
```

The input lists items (`wearable_standalone` / `wearable_on_avatar` / `emote`) with per-shot cameras (auto-fit orbit by yaw/pitch, or explicit position/target with the scene renderer's semantics; emote shots take a normalized `atTime`). Results are written to a report JSON; exit codes are 0 all ok / 1 bad input / 2 empty payload / 3 partial failure.

## How

- `wearable_standalone` synthesizes a neutral avatar with `showOnlyWearables`, so the only visible pixels belong to the item; `emote` uses a new `AvatarEmoteController.async_freeze_on_emote(urn, t)` (the `freeze_on_idle` technique at an arbitrary point of an arbitrary emote, prop included).
- Auto-fit frames the posed asset using `AvatarPreview`'s CPU-skinned AABBs, exposed via a new `compute_current_fit_aabb()`, so frozen emote poses are measured as posed. Captures go through a new `async_capture_current_view()` that leaves the camera entirely to the caller.
- Each item runs under a 60s watchdog: a hung content fetch errors that item in the report and the batch continues.
- Rust surface is limited to the two CLI args in `dcl_cli.rs`; everything else reuses the avatar renderer scene machinery.

## Testing

- CI (`docker_build_test.yml`) runs the mode against `tests/asset-renderer-test-input.json` (base wearable at 3 yaw angles, worn sunglasses, wave emote at 3 keyframes) and asserts every report entry is ok with `jq` — the container entrypoint ends in `|| true`, so the exit code cannot be trusted there.
- `cargo check` clean (note: `--no-default-features` alone is broken on main — `scene_manager.rs` uses `dcl::js` without the `use_deno` gate; unrelated to this change), `gdlint`/`gdformat` clean.
- Image snapshots can be added from this job's artifacts once reviewers are happy with the framing (same flow as the avatar test).


godot/src/decentraland_components/avatar/avatar_emote_controller.gd
```diff
@@ -1019,6 +1019,59 @@ func play_emote_audio(file_hash: String):
...
+
+	if emote_item_data.armature_prop != null and is_instance_valid(emote_item_data.armature_prop):
+		emote_item_data.armature_prop.show()
+
+	return true
+
+
+## Undoes a freeze so the avatar can play emotes again. freeze_on_idle and
+## async_freeze_on_emote both disable the tree, and nothing else re-enables it.
+func unfreeze() -> void:
```
regenesis-claw: [P2] `unfreeze()` has no callers — the asset renderer never calls it, and nothing else in the diff does. It is currently unreachable code added to a shared production file (REVIEW.md Tier 3 item 16).

It also does not appear to be needed: every item path re-disables the tree itself (`freeze_on_idle()` for the non-emote kinds, `async_freeze_on_emote()` for emotes), so consecutive items pose correctly without it. Either drop it, or call it at the end of the batch if the intent is to leave the avatar usable for a future in-client caller.

godot/src/tool/asset_renderer/asset_renderer_input_helper.gd
```diff
@@ -0,0 +1,161 @@
...
+			printerr("payload has to be an array")
+			return null
+
+		var ret := AssetInputFile.new()
+		ret.base_url = tmp_base_url
+		ret.output_json_path = tmp_output_json_path
+		ret.supersample = json_value.get("supersample", 1.0)
+		var index := 0
+		for maybe_entry in tmp_payload:
+			var parsed = AssetItem.from_dictionary(maybe_entry)
```
regenesis-claw, line 207: [P2] `maybe_entry` is not guaranteed to be a Dictionary, but `AssetItem.from_dictionary` takes a typed `Dictionary` parameter. A payload containing a string or a number raises a GDScript runtime type error at the call site instead of flowing into `invalid_items` — which defeats the per-item error reporting this whole path exists for. The `if maybe_entry is Dictionary` guard three lines below suggests that case was expected to reach it.

```suggestion
			if not maybe_entry is Dictionary:
				ret.invalid_items.push_back(
					{"id": "payload[%d]" % index, "error": "entry is not an object"}
				)
				index += 1
				continue
			var parsed = AssetItem.from_dictionary(maybe_entry)
```

Same applies to `Shot.from_dictionary(maybe_shot)` on line 101 and to `value.get("position", {}).get(...)` in `ShotCamera.from_dictionary`, which assumes `position`/`target` are objects.

godot/src/tool/asset_renderer/asset_renderer_input_helper.gd
```diff
@@ -0,0 +1,161 @@
...
+		if not ret.kind in VALID_KINDS:
+			return "invalid kind %s" % ret.kind
+
+		var shots_value = value.get("shots", [])
+		if not shots_value is Array:
+			return "shots has to be an array"
+		for maybe_shot in shots_value:
+			var shot: Shot = Shot.from_dictionary(maybe_shot)
+			if shot == null:
+				return "a shot is missing destPath"
```
regenesis-claw: [P2] This reports `"a shot is missing destPath"` for both `Shot.from_dictionary` failure modes — the other one is the explicit-camera `position == target` case, which already printed a different (correct) message to stderr. The report entry, which is what the pipeline consumes, ends up describing the wrong problem. Returning the reason string from `Shot.from_dictionary` instead of `null` would keep the two distinguishable.

Also, `ret.camera == null` on line 59 is unreachable: `ShotCamera.from_dictionary` never returns null.

godot/src/tool/asset_renderer/asset_renderer_standalone.gd
```diff
@@ -0,0 +1,400 @@
...
+			await get_tree().process_frame
+
+		_place_camera(shot.camera, float(shot.width) / float(maxi(shot.height, 1)))
+
+		var dest_path := _ensure_ends_with(shot.dest_path, ".png")
+		_ensure_base_dir_exists(dest_path)
+		# Explicitly typed: awaiting a coroutine yields an untyped value, so `:=`
+		# cannot infer here (gdlint accepts it, the Godot compiler does not).
+		var image: Image = await avatar_preview.async_capture_current_view(
+			Vector2i(shot.width, shot.height), 1
```
regenesis-claw, line 290: [P2] `shot.width` / `shot.height` go into `Vector2i` and then into `set_size` / `subviewport.set_size` unvalidated. `supersample` is clamped and `fov` is clamped, but the render size is not: `width: 0` or a mistyped `width: 32768` in the input file is a caller mistake that turns into a zero-size viewport (handled only by the `image == null` branch, 60s later in the worst case) or an out-of-memory allocation. A `clampi(..., 16, 4096)` at parse time in `Shot.from_dictionary` would keep it consistent with how the other numeric fields are treated.

regenesis-claw (approved):

**Pre-flight**
- Branch is 6 ahead / 0 behind `main` — not stale.
- No iOS build required: the change is GDScript under `godot/src/tool/` plus two CLI arg definitions in Rust. No iOS branch, no native plugin, no comms/av code.
- CI green, including the new `Run asset-renderer test` step in the Docker job — that is real signal that the mode produces 3 `ok` items and 7 non-empty PNGs end to end.

**Verdict: approve.** No P0/P1. The risky parts hold up under source-level checking:

- The auto-fit math is right: `avatar_preview.tscn`'s Camera3D leaves `keep_aspect` at its default (KEEP_HEIGHT), so `fov` really is vertical and widening the fit for portrait aspects is the correct compensation.
- `async_capture_current_view` sets `_lerp_paused = true` *before* `set_size`, so the `resized` -> `_on_resized` -> `_fit_camera_to_aabb(..., instant)` path cannot clobber the caller's camera mid-capture (the exact failure the comment at `avatar_preview.gd:858` documents). `fit_avatar = false` + `set_process(false)` cover the other two fit entry points.
- `clean_unused_emotes()` after every item is safe even on the first item's error path: `DclAvatar::init` seeds `avatar_data` with a default wire format, so it is never null and `get_emotes()` just returns empty.
- `_is_fallback_aabb` is a real check, not decoration: `_compute_avatar_aabbs` merges into `overall` only for `mi.visible`, so `showOnlyWearables` with no visible wearable mesh genuinely leaves the key absent and falls back.
- Watchdog: `_async_process_item` always reaches an `await` before it can commit, and the generation counter plus the batch abort on timeout are the right call given a detached coroutine can still rebuild the shared skeleton.

Findings below are all P2.

**Unverified (cannot be checked from the diff)**
- Framing quality of the produced stills. The CI assertions prove pixels exist, not that the asset is centred/uncropped — the snapshot follow-up you mention is the thing that would close this.
- Behaviour on a real (non-base) published wearable/emote: the test input only exercises off-chain base items.

**Nit on the description:** the repo's shape is `## What` / `## Why` / `## Test plan`; this one is What/How/Testing. Fine content-wise for a headless tool with no QA phone steps, just noting it.

godot/src/decentraland_components/avatar/avatar_emote_controller.gd
LautaroPetaccio: Dropped it. Nothing called it, and as you noted every item path re-disables the tree itself, so there is nothing for it to undo here.

godot/src/tool/asset_renderer/asset_renderer_input_helper.gd
LautaroPetaccio: Fixed, and taken wider than the three call sites. Every field read in the file now goes through a `Coerce` helper (`as_string`/`as_int`/`as_float`/`as_bool`/`as_vector3`), so a wrongly typed field falls back to its default instead of raising on assignment to a typed member. The three you named are guarded explicitly and each gets its own reason: `entry is not an object`, `a shot is not an object`, `shot <name> has a camera that is not an object`.

Reproduced the original failure first — a numeric shot gave `Invalid type in function 'from_dictionary' in base 'GDScript'. Cannot convert argument 1 from float to Dictionary`, which killed the batch rather than reaching `invalid_items`. That case is now a row in the new test.

godot/src/tool/asset_renderer/asset_renderer_input_helper.gd
LautaroPetaccio: Done — `Shot.from_dictionary` now returns the reason string instead of null, and `AssetItem.from_dictionary` passes it straight through. The two modes read as `shot y0 is missing destPath` and `shot y0 needs a distinct camera position and target`, so the report entry describes what actually happened. The `printerr` for the degenerate-camera case is gone, since the reason now travels in the report rather than only to stderr.

Also removed the unreachable `ret.camera == null`.

godot/src/tool/asset_renderer/asset_renderer_standalone.gd
LautaroPetaccio: Clamped at parse time as suggested: `clampi(..., MIN_SHOT_SIDE, MAX_SHOT_SIDE)` with 16 and 4096, next to the other numeric fields. A non-numeric width falls back to 1024 rather than clamping garbage, so `width: 0` becomes 16, `height: 999999` becomes 4096 and `width: "big"` becomes 1024. All four are asserted in the new test.

LautaroPetaccio:
Addressed the four P2 comments in bcb5308d (rebased onto the latest `main` merge).

The common thread in two of them was that a malformed input file could raise inside the parser instead of flowing into `invalid_items`, which is the whole point of that path — so rather than patch the three call sites, every field read now goes through a `Coerce` helper and each failure carries its own reason. `Shot.from_dictionary` returns that reason instead of null, so the report no longer describes every rejected shot as a missing `destPath`, and render sizes are clamped to [16, 4096] at parse time.

Added `cargo run -- test-asset-renderer` (`godot/src/test/asset_renderer/test_asset_renderer_input.gd`, wired into the Linux job) covering these paths: every rejection reason, one bad entry not taking its neighbours down with it, the size clamping and the wrongly-typed-field fallbacks. Confirmed it fails without the fix — dropping the shot type guard reproduces the original `Cannot convert argument 1 from float to Dictionary`.

One extra beyond the comments: an item whose `id` is not a usable string is now rejected as `item needs an id`. It previously became `""`, which produces a report entry the caller cannot match back to its request and reads as a missing result.

kuruk-mm (approved):
great!

PR #2930, opened 2026-09-17
Release Candidate 1.14.0

leanmendoza:
## Release Candidate 1.14.0

Promotes `main` (`54283d65`) into `release`. Clean promotion: 47 commits, no cherry-picks, nothing excluded — `release` is fully contained in this branch. Version `1.14.0` already bumped in #2818.

Two headlines. **Scenes load much faster and stop losing models**: the optimized-asset pipeline moves to v6, which shares textures across GLBs instead of embedding a copy in each — Genesis Plaza's GLB packs go from 164.6 MB to 45.0 MB and its loading screen from 12.2 s to 7.3 s on a Moto Edge 60 Pro. And **the app now speaks Spanish and Brazilian Portuguese**: 697 translated keys, a language picker in Settings, and a fresh install that follows the device locale. Alongside them, the Menu redesign lands across Social, Notifications and Settings, the HUD gets its second orb pass, pinch-to-zoom replaces the POV toggle, and SDK audio finally matches Unity.

- **Base:** `release` · **Head:** `release-1.14.0` (= `main` @ `54283d65`)
- 47 commits on top of `release`, including the back-merges of 1.13.0 / 1.13.1 / 1.13.3 (#2820, #2875, #2923) — already in `release`, so they promote nothing new

## What's included

**Features**

- **v6 optimized-asset pipeline** — shared textures across GLBs, plain `.scn`/`.res` files instead of per-asset zips, scene bundles and lazy colliders. Plaza's asset pipeline 8.6 s → 5.5 s cold and 9.0 s → 2.6 s warm, loading screen 12.2 s → 7.3 s, and the 28 GLBs v5 silently dropped in Plaza load again. Every phone re-downloads its optimized assets once after updating (#2764)
- **Spanish and Brazilian Portuguese** — 697 keys across the entry flow, menus and shared components; the picker lives in Settings → Gameplay → Language and a fresh install follows the device locale (#2754, #2849)
- **Menu redesign — Social** — Friends / Requests / Nearby tabs, friends in Worlds show their place name and support jump-in, plus the new navbar with a Discover section and a single moving highlight (#2792)
- **Menu redesign — Notifications** — the 4 notification types with per-type art and title colour, read/unread states, and tapping one routes to the right panel (#2826)
- **Menu redesign — Settings** — full-screen, list+detail in landscape and master-detail in portrait; Language split out, Report + Help merged into Help & Support (#2864)
- **HUD revamp, second iteration** — every standalone HUD action is now an orb button, the joypad arcs match the authored layout, and the virtual joystick was redrawn and made safe-area aware (#2723)
- **Interaction prompt redesign** — taller pill in the HUD's purple, no white keycap, and a much larger tap area; creator icons keep their own colours (#2874)
- **Pinch-to-zoom camera** — two fingers swap first/third person, the avatar is centered, and the crosshair sits above-right of the head with raycasts aimed where it is drawn (#2878, #2691)
- **SDK audio parity with Unity** — volume slider has a usable range, spatial sounds no longer play ~20 dB too loud from a metre out and keep their highs, `AudioStream` plays again after a year of silence, and `PBAudioEvent` / `PBAudioAnalysis` are implemented (#2801)
- **Graphic profile budgets** — view distance (60/100/200/300 m), particle quality and max active lights now follow the graphics profile, with individual controls under Custom (#2812)
- **Mobile Shop** — credits packs at the second price point, new pack and success art, purchase history, and the storefront routes per environment (#2608)
- **Google Play in-app review prompt (Android)** — Play's own rating card after a friend add, a favorite, a like or a wearable claim; at most three asks per install, ever (#2865)
- **Native bug reporting** — Settings → Report a Bug files an Intercom ticket in-app with up to 3 screenshots pre-filled from the game frame, a `Platform` attribute, and links to the Sentry event and feedback entry (#2779, #2906, #2916)
- **Ad-attributed installs boot into the scene the ad sold** — an install carrying a campaign token resolves its destination server-side and skips the FTUE; every failure path falls back to today's FTUE (#2788)
- **SDK: `teleportTo` can name a realm** — a scene can send a player to a place in another World or realm and the client changes realm first, then places them (#2897)
- **SDK: scenes read the player's language** — `getExplorerInformation().configurations.locale` plus a `localeChanged` event when the player switches language, no reload (#2905)
- **SDK: `TextEncoder` / `TextDecoder` globals** — native WHATWG codecs, so scenes and libraries that rely on them stop crashing at load or falling back to polyfills (#2542)
- **Scene stats: VFX emitter and light source rows** — four new per-scene rows (emitters, live particles, light sources, shadow lights) against the engine budgets; Genesis Plaza reads red on three of them (#2815)

**Fixes**

- **iOS froze for 5–130 s while shaders compiled at first draw** — every 1.13.1 iOS user hit it (Sentry App Hangs `2PK`/`2PH`/`2QH`/`2PS`, 23 of ~28 users, one watchdog kill). No store build ever shipped baked shaders: the baker silently does nothing under `--headless`, so the CI export now runs a windowed editor and fails unless the bake actually ran (#2886)
- **Half of all `DeleteComponent` messages were silently dropped** — the renderer declared 28 bytes while writing 20, so `@dcl/ecs` dropped the delete whenever it landed last in a chunk: tween removal not applied, avatars of players who left staying in scene player lists, Transform nulling lost (#2858)
- **Code-applied textures rendered ~2x too dark, and reloading a GLB lost them** — a double sRGB decode on every script-set albedo texture since June 2025, plus a spurious full reload on any GLB path with uppercase. Marsh Colony's creatures went near-black and lost their skins on every carry (#2866)
- **Scene-triggered swing animations left the avatar in a T-pose** — three root causes: the emote converter exported the wrong clip from a multi-animation GLB, clips with dots in their names were dropped silently, and the on-screen touch buttons ignored mouse clicks on desktop (#2902, #2766, #2618)
- **A scene whose raycast hit a freed collider froze forever** — `boedo.dcl.eth` (CS 1.6) loaded but never got past its second update; the scene thread also died on a backwards clock step (#2918)
- **Scene objects invisible until you interacted with them** — `VisibilityComponent` hid the whole subtree instead of only the entity's own visuals, so child entities marked visible stayed hidden (bloomgarden's plants) (#2653, #1888)
- **Nametags of avatars behind other avatars kept rendering** — occlusion was lost in #2321, so close-together tags painted over each other into an unreadable wall; tags now occlude and de-overlap, and `HideNameTags` areas survive a second overlapping modifier area (#2717, #2637, #2665, #2724)
- **Material `offset`/`tiling` moved textures the wrong way vertically** vs Unity whenever `offset.y ≠ 0` or `tiling.y ≠ 1` (#2813, #2719)
- **Android's memory-pressure monitor could never fire** — it measured our own footprint against a constant, so its thresholds sat *above* the point where Android kills the app; these deaths are `SIGKILL`, so Sentry has never recorded one and they look exactly like a crash. Now reads `MemAvailable`, and its logs reach logcat at all for the first time (#2721)
- **Signed requests rejected on a device with a drifting clock** — App Review hit exactly this on 2026-09-07: a device ~60 s fast got its IAP quote refused, StoreKit was never reached, and the app was rejected under Guideline 2.1(b). All six signing sites now use an NTP-corrected clock, falling back to today's behaviour if nothing answers (#2834)
- **Signed-fetch signed a different payload than it sent** — the metadata was folded before signing but delivered unfolded, so services authorized on bytes the signature never covered (#2762)
- **Preview hot-reload was silently dead after scanning a second QR** — the WebSocket stayed pointed at the previous server; touch-control icons also stayed stale after a reload (#2798, #2795, #2796)
- **Desktop browser sign-in dead-ended** on the auth site's "your app needs an update" screen after auth-site 5.0.0 retired `dcl_personal_sign`; those options are greyed out on desktop, leaving email OTP and guest. Mobile is unaffected (#2605)

**Technical** — no QA needed

- CI: the `build` label now keeps one live "Mobile build pipeline" card on the PR instead of three stale fragments (#2887)
- CI: headless Godot runs in the Linux build are judged by their success marker when the process is killed by a signal — 83% of the last 90 days' red jobs were this teardown crash, not a defect (#2870)
- Docs: `REVIEW.md` states Android and iOS as the only supported and QA'd targets; desktop/XR are legacy remnants (#2904)
- Docs & tooling: `pr-description` skill and PR description rules in `REVIEW.md` (#2863), `godot-issue-writer` skill (#2917), Sentry seeder breadcrumb titles recorded as not translatable (part of #2875)
- Chore: marketing version bumped to v1.14.0 (#2818); `release` 1.13.0 / 1.13.1 / 1.13.3 reconciled into `main` (#2820, #2875, #2923)

## Test plan

**Build:** `v1.14.0.<build>-54283d6-prod` · one Android + one iPhone.

- [ ] **Version** — login screen and **Settings → About** read `v1.14.0.<build>-54283d6-prod`
- [ ] **Asset pipeline** — cold-start into Genesis Plaza (-3,-2): it loads noticeably faster than 1.13.3, and nothing is missing — floor, buildings, trees, props all present, no untextured or black models. Revisit after walking away: it comes back faster still `#2764`
- [ ] **Spanish / Portuguese** — Settings → Gameplay → Language, switch to Español and to Português (Brasil) without restarting: the UI translates live, nothing is clipped or cut mid-sentence, and the choice survives a restart. On a fresh install with the phone set to Spanish, the app starts in Spanish `#2754` `#2849`
- [ ] **Social panel** — open the menu: the single highlight follows whichever section you tap. Friends / Requests / Nearby each list correctly; a request sent from a second account appears under SENT live; a friend inside a World shows that World's name and Jump In lands you at its spawn `#2792`
- [ ] **Notifications** — have a second account send a friend request, then accept one: each notification shows the sender's nickname as title and routes to the right tab when tapped; read ones are dimmed `#2826`
- [ ] **Settings** — open Settings in portrait and in landscape: portrait drills into each section with a ‹ back header, landscape shows list and content side by side. **Hide Player Names** now works on its own, without Hide Interface `#2864`
- [ ] **HUD orbs** — hold the joystick with one thumb and tap the emote orb with a second finger: the wheel opens. Rotate the device: the chat / flip / discover orbs resize and stay tappable on their visible disc `#2723`
- [ ] **Interaction prompt** — look at something interactable: the pill is taller, purple glyph on dark, no white keycap, and easy to hit with a thumb. In a scene that sets its own control icons, the icon keeps its own colours `#2874`
- [ ] **Pinch-to-zoom** — pinch in until the avatar hides (first person) and back out: the avatar stays horizontally centered, the crosshair glides between above-the-head and dead-center with no jump, and tapping a clickable under the raised crosshair responds `#2878`
- [ ] **SDK audio** — Settings → Audio, drag General to 50%: clearly audible at about half loudness (it was nearly silent before). In a scene with a spatial sound, walk away: it fades with distance and stays bright, not muffled `#2801`
- [ ] **Graphics profiles** — with Genesis Plaza loaded, switch profile in Settings → Graphics: view distance and particles change immediately, Very Low turns particles off and they stay off after a relaunch `#2812`
- [ ] **Shop & credits** (iPhone) — buy a credits pack: the whole card is tappable, the purchase appears under Purchases, and the balance updates. Open Settings → Shop and buy a wearable: it is equipped back in the app. On Android nothing IAP is visible `#2608`
- [ ] **Bug report** — Settings → Help & Support → Report a Bug: the first tile is pre-filled with the in-world frame, add two more from the gallery (no 4th slot appears), submit: "Report sent". The Intercom ticket carries `Platform`, all 3 images, and both Sentry links `#2779` `#2906` `#2916`
- [ ] **iOS shader freezes** — on an iPhone, first launch after install: enter Genesis Plaza, then teleport to a scene you have never visited. Nothing freezes for more than about a second at any point `#2886`
- [ ] **Deleted components** — in Genesis Plaza watch tweened objects for a minute: they keep cycling instead of sticking after their first move. Have a second user teleport away: their avatar disappears from the scene `#2858`
- [ ] **Code-applied textures** — go to Marsh Colony: the creatures look as bright as in the Unity client, and carrying a pet to the tub and putting it down keeps its skin the whole time `#2866`
- [ ] **Swing emote** — `pixelpirate.dcl.eth` (spawn -1,0): address the ball and lock power — the avatar plays a visible one-arm swing, not a T-pose. The waterfall on the course flows `#2902`
- [ ] **Raycast scene** — jump to `boedo.dcl.eth`: the CS 1.6 SELECT TEAM menu appears over the map within a few seconds and you can join a team `#2918`
- [ ] **Visibility** — `bloomgarden.dcl.eth`: the wilted roses and the water drops are visible right after spawn, without watering anything `#2653`
- [ ] **Nametags** — in Genesis Plaza stand so one avatar blocks another's tag: the blocked tag fades out and comes back when the line clears. In a cluster, tags and chat bubbles stack apart instead of overlapping `#2717`
- [ ] **Texture offset/tiling** — in a scene that sets a material `offset.y` or `tiling.y` (e.g. the repro linked in the PR), the texture moves the same vertical direction as the Unity client `#2813`
- [ ] **Low memory (Android, 4GB or less)** — on a Moto G30-class phone enter Genesis Plaza and stay while avatars appear: if it runs out of memory a warning dialog offers to continue or go back to Discover, instead of the app vanishing to the home screen. On a 6GB+ phone no warning ever appears `#2721`
- [ ] **Signed requests** — sign in, walk into a scene, change a wearable and save the profile, open the Friends panel, and buy a credits pack: none of them fail with a signature or authorization error `#2762`
- [ ] **Preview hot-reload** (creator check) — start two `sdk-commands start` previews on different ports, scan the first QR, then **without closing the app** scan the second and edit a file: the scene reloads. Edit an image used both in the scene UI and as a `PBTouchScreenControls` icon: both pick up the new image `#2798`
- [ ] **Realm teleport from a scene** — in a scene whose `teleportTo` names another realm, accept the travel modal: you change realm and land on the right parcel (or the realm's spawn when it gives no coordinates) `#2897`
- [ ] **Scene locale** — with the app in Spanish, enter a scene that reads the player language: it shows Spanish content, and switching language in Settings updates it without a reload `#2905`
- [ ] **Scene stats** — launch with the scene-stats overlay and open Scene Status in Genesis Plaza: VFX emitters, Live particles, Light sources and Shadow lights appear between Colliders and Content size, three of them red `#2815`
- [ ] **Play rating card (Android)** — add a friend, or favorite/like a place, on an install with 5+ sessions: Play's rating card appears shortly after, never during loading or over a modal `#2865`
- [ ] **Signing clock** — set the phone's clock ~60 s fast by hand, then buy a credits pack: the payment sheet still opens and the purchase completes `#2834`
- [ ] **Desktop sign-in** (desktop build, courtesy check) — Google / Apple / WalletConnect are greyed with a tooltip; Continue with Email and Play as Guest still work `#2605`
- [ ] **Regression** — enter a few scenes, chat, change a wearable, play an emote from the wheel, teleport twice, background and resume the app: no crashes. This also covers #2542, which has no surface of its own — a scene using `TextEncoder`/`TextDecoder` simply loads instead of crashing

## Known gaps

- **#2788 (ad-attributed FTUE) cannot be exercised on this build.** Its QA needs `?rotate-guest=true` to get back to the FTUE, and that deeplink is gated to non-production — an RC cut from `release-1.14.0` is built `--prod`, so it is inert. It has to be QA'd on a `main`/staging build. Separately, the attribution path itself has never run on a device: ad spend has been paused since late August, so the Play referrer and the GA4F deferred deep link are still unexercised in the wild.
- **#2865's `?review-debug=true` harness is also non-production-only**, so on this build the review prompt can only be observed at a real trigger, subject to Play's own quota.
- **#2812 shipped without its device validation** — the PR left scene pop-in per view distance, the per-profile GPU-ms/FPS A/B and the thermal baseline unchecked.
- **#2834's NTP sync depends on outbound UDP 123**, which some carriers and networks block. When that happens the fallback is silent and the fix simply does nothing, so the skewed-clock case is worth trying on cellular as well as Wi-Fi.
- **#2764 makes every phone re-download its optimized assets once** after updating (`LOCAL_ASSETS_CACHE_VERSION` 5 → 7). The first load after the update is expected to be slower than the numbers above; measure on the second.


kuruk-mm (approved):
looking great!

PR #2931, opened 2026-09-17
fix: implement stopEmote and tie masked emotes to their scene

### LINKED ISSUE #2859: [Bug] Avatar animation hangs holding invisible egg after hatching (mobile, masked emotes)
### Platform
Not specified (screenshot shows mobile client build `v1.13.0.1687-5d5da32-staging`)

### App Version
1.13.0.1687-5d5da32-staging

### Device Information
Not specified

### Issue Description
After hatching an egg (marsh-colony scene), the avatar's arm animation gets stuck mid-emote/gesture while holding an invisible egg — the held egg model isn't visible and the avatar animation hangs instead of completing/resetting. Suspected regression from the avatar masks / upper-body emote work in [#2690](https://github.com/decentraland/godot-explorer/pull/2690) (masked emotes are exempt from cancel-on-move by construction, and the terminal `EmoteState` reporting may not be firing for this playback path). Confirmed by Agustin: this does not reproduce on the Unity client for the same marsh-colony flow.

### Expected Behavior
The hatching animation/gesture should finish and clear normally, with the held egg model visible, matching Unity client behavior.

### Screenshots / Media
See attached screenshot in thread (avatar frozen with raised-hands gesture, holding invisible egg, "Sprout" pet visible nearby).

### Steps to Reproduce
1. Open the marsh-colony (My Dear Pet) scene on mobile.
2. Hatch an egg / trigger the hatch flow.
3. Observe the avatar's hands/animation after hatching completes.

### Occurrence
Not specified

### Context
- Reported by Manuel Maceira, flagged by Mateo as possibly related to the avatar masks PR ([#2690](https://github.com/decentraland/godot-explorer/pull/2690), closed [#2621](https://github.com/decentraland/godot-explorer/issues/2621) and [#2655](https://github.com/decentraland/godot-explorer/issues/2655)).
- Agustin confirmed this does not happen in Unity for marsh-colony, ruling out a scene-side/SDK bug and pointing at the mobile client's masked-emote/EmoteState handling.

### 🔗 References
- [decentraland/godot-explorer#2690](https://github.com/decentraland/godot-explorer/pull/2690) — PR that introduced avatar masks + EmoteState playback signals; masked emotes are deliberately exempt from cancel-on-move, a likely factor in the stuck animation.
- [decentraland/unity-explorer#8050](https://github.com/decentraland/unity-explorer/issues/8050) — related QA finding on Unity avatar masks ("User can emote while sostaining an object"), same feature area though not reproducing on Unity for this specific flow per Agustin.

EibrielInv:
## What
  Scenes can now stop an emote they started, through the SDK's `stopEmote` restricted action — it is
  specified in `@dcl/protocol` but was never implemented in this client. An upper-body emote a scene
  started now also lives and dies with that scene: it pauses when the player walks out, resumes when
  they walk back in, and is dropped for good when the scene unloads.

  ## Why
  A scene that started a *looping* upper-body emote had no way to end it, so the avatar was stuck
  with its arms raised forever — what players hit in Marsh Colony after hatching an egg (#2859).
  Marsh Colony ships a workaround that fires a second, non-looping emote on non-Unity clients only;
  after this it no longer needs to. Full-body emotes are unchanged, and this does not copy the other
  cases Unity suspends upper-body emotes for (a full-body emote playing, or gliding).

  Closes #2859

  **Heads-up:** needs a note in `#ext-foundation` before merge — it adds an SDK-visible restricted
  action and changes when a scene-triggered upper-body emote plays.

  ## Details
  <details>
  <summary>Expand</summary>

  **Root cause.** Three things had to line up: `RestrictedActions.js` never exported `stopEmote`, and
  scenes guard with `typeof stopEmote === 'function'`, so the call vanished silently; natural-end
  detection is gated on `not playing_loop`; and cancel-on-move is gated on `playing_single or
  playing_mixed`. Before the avatar-masks work (#2690) the same clip played full-body and died on the
  player's first step, so this is a regression from that PR.

  **Ported from the Unity Foundation Client** — `SceneMaskedEmoteSystem.UpdateMaskedEmoteVisibility`
  and `TryStopMaskedEmote`. Unity keeps masked-emote state inside the triggering scene's own ECS
  world, so ownership and unload teardown come for free. Our controller is per-avatar, so the owning
  scene id is threaded through `async_play_emote` and stored next to the emote URN instead. Unity's
  `permanent` distinction is kept: `stopEmote` also drops an emote parked at a scene boundary, so
  re-entry cannot resurrect something the scene explicitly ended.

  **No EmoteState at a scene boundary.** Crossing in or out emits neither `ES_STARTED` nor a terminal
  state, so scenes watching `AvatarEmoteCommand` see no churn as the player moves around. Safe
  because suspending does not clear `current_emote_urn`: the emote stays logically current and the
  single terminal state from #2655 is still reported by whatever really ends it. Called out because
  "silent" and "exactly-once" being compatible is not obvious from the code.

  **Verified by the author** against a running client, reading live controller state:
  - [x] owning scene recorded; leaving parks the emote (layer off, URN kept); re-entry resumes it
  - [x] `stopEmote` while parked is permanent — re-entry does not resurrect
  - [x] non-looping upper-body emote ends on leaving and never resumes
  - [x] unloading a different scene leaves it alone; unloading the owning scene clears it
  - [x] emote-wheel and full-body emotes unaffected by scene changes
  - [x] `cargo clippy -- -D warnings`, `cargo fmt`, `gdformat`, `gdlint`, `check-gdscript` green

  **Review fixes, verified the same way.** `stopEmote` is now scoped the way Unity scopes it —
  `TryStopEmote` stops the full-body emote globally but only touches its own scene world's masked
  component — because `_player_is_inside_scene` passes unconditionally for `SceneType::Global`,
  so a portable experience could otherwise end an emote it does not own. Not given a QA case: it
  needs a global/portable scene calling `stopEmote`, which no test scene provides.
  - [x] a non-owning scene does **not** stop a masked emote (playing or suspended at a boundary)
  - [x] the owning scene does stop it
  - [x] a non-owning scene **does** still stop a full-body emote, matching Unity
  - [x] the resume re-broadcast path runs and `send_emote` returns true for the parked urn/mask

  Not exercised by the author: the JS → op round trip, which needs the scene UI (Case 1 covers it).

  The resume re-broadcast **was** verified on the wire with two clients: the observer receives
  `PlayerEmote id=189232647 mask=Some(1)`, then a stop, then `PlayerEmote id=189685636 mask=Some(1)`
  on re-entry. It is not yet observable on screen — see **Known gap**.

  </details>

 ## Test plan

  **Setup:** the **Eibriel Tests** scene at **-31,19** → **Emote Mask** tab. The hold/release
  buttons are under the **#2859 — scene emote hold/release** header.

  **Case 1 — a scene can stop an emote**
  1. Open the app and go to **-31,19**.
  2. Open the scene's UI and select the **Emote Mask** tab.
  3. Tap **HOLD (loop)** — the avatar raises both arms as if holding something.
  4. Tap **RELEASE**.

  - [x] The arms come back down.
  - [x] The action log reads `RELEASE  stopEmote() called`, not `RELEASE  stopEmote ABSENT — no-op`.

  **Case 2 — the hold pauses outside its scene and resumes inside**
  1. Open the app and go to **-31,19**.
  2. Open the **Emote Mask** tab and tap **HOLD (loop)**.
  3. Walk around inside the scene and jump.
  4. Walk out into a neighbouring parcel, then walk back in.

  - [x] Inside, the arms stay raised through walking and jumping.
  - [x] On leaving, the arms drop and the avatar walks normally.
  - [x] On returning, the arms are raised again.

  **Case 3 — one-shot and stopped emotes do not come back**
  1. Open the app and go to **-31,19**.
  2. Open the **Emote Mask** tab, tap **HOLD (once)**, walk out of the scene and back in.
  3. Tap **HOLD (loop)**, then **RELEASE**, walk out and back in.

  - [x] After step 2 the emote plays once and does not restart on re-entry.
  - [x] After step 3 the arms stay down and the emote does not come back.

  **Regression** — shared emote code is touched. In the **Emote Mask** tab:
  - [x] Toggle to **MASKED: upper body**, tap **Wave**, then walk: the legs walk, the arms wave, and
        the emote is not cancelled.
  - [x] Toggle to **MASKED: off**, tap **Wave**, then walk: the emote cancels at once and
        `INTERRUPTED` appears under **AvatarEmoteCommand entries**.
  - [x] Tap **Clap (prop)**: the prop appears and its audio plays.
  - [x] Tap **Wave** and let it finish: the entry log shows `STARTED` then `FINISHED`.
  - [x] Open the emote wheel, play any emote, then walk: it cancels as before.

## Known gap — observer-side behaviour is blocked on #2943

A second player does **not** visibly see the emote resume, and this PR cannot fix that.
[#2943](https://github.com/decentraland/godot-explorer/issues/2943): `send_emote` also emits the
legacy `␐<urn>` chat encoding, which carries no mask, so every observer plays an upper-body emote
**full-body** — and full-body emotes are cancel-on-move, so the actor's movement kills the
observer's copy. The resume message is sent and received correctly (evidence in Details); it just
has no visible effect until #2943 lands.

QA: **do not run the two-client case yet.** Kept here to run once #2943 is fixed:

<details>
<summary>Two-client case, blocked on #2943</summary>

**Setup:** two devices, different accounts, both running this build, both in the **Eibriel Tests**
scene at **-31,19** and able to see each other. Device B must not be the Unity client — the stop at
step 3 rides Pulse, which Unity does not receive.

1. On device A, open the **Emote Mask** tab and tap **HOLD (loop)**.
2. On device B, look at player A.
3. On device A, walk out of the scene into a neighbouring parcel.
4. On device A, walk back into the scene.

- [ ] After step 1, B sees A's arms raised.
- [ ] After step 3, B sees A's arms drop.
- [ ] After step 4, B sees A's arms raised again.

</details>


godot/src/decentraland_components/avatar/avatar_emote_controller.gd
```diff
@@ -402,6 +423,77 @@ func _is_excluded_from_mask(track_path: NodePath) -> bool:
...
+		return
+	var emote_item_data: EmoteItemData = loaded_emotes_by_urn[current_emote_urn]
+	var anim_path := "emotes/" + emote_item_data.default_anim_name
+	if not animation_player.has_animation(anim_path):
+		stop_emote()
+		return
+	if not animation_tree.active:
+		animation_tree.active = true
+	if not _play_masked_emote(anim_path):
+		stop_emote()
```
regenesis-claw, line 485: [P1] The resume is silent locally, but the suspend was *not* silent on the wire: `_suspend_masked_emote()` drops `playing_masked`, so `avatar.gd::_process` reports `is_emoting = false` and `set_emoting(false)` sends a Pulse `EmoteStop` and clears `pending_pulse_emote_urn`. On resume `set_emoting(true)` finds nothing pending, so no `EmoteStart` goes out — remote players see the emote end at the boundary and never see it resume.

Unity re-broadcasts here (`ReplayMaskedEmote` adds a fresh `EmotePendingToBroadcast` with the same urn/mask). Suggest doing the same for the local player after a successful replay:

```suggestion
	if not _play_masked_emote(anim_path):
		stop_emote()
		return
	# Remote players got an EmoteStop when the suspend dropped `is_emoting`; announce the
	# emote again so they resume with us (Unity: ReplayMaskedEmote -> EmotePendingToBroadcast).
	if avatar != null and avatar.is_local_player:
		Global.comms.send_emote(current_emote_urn, current_emote_mask)
```

godot/src/decentraland_components/avatar/avatar_emote_controller.gd
```diff
@@ -223,6 +233,10 @@ func stop_emote():
 		animation_tree.set(MASKED_REQUEST_PARAM, AnimationNodeOneShot.ONE_SHOT_REQUEST_FADE_OUT)
 		playing_masked = false
 		_hide_all_props()
+	# A suspended masked emote is still "current" until something ends it for good;
+	# this is that something, so drop the parked state too or re-entering the owning
+	# scene would resurrect an emote the scene explicitly stopped.
+	masked_suspended = false
 	playing_single = false
```
regenesis-claw, line 240: [P2] `_emit_emote_finished(true)` on the first line of `stop_emote()` already clears `masked_suspended` along with the urn/mask/scene id, and the two are only ever set together — so this line is a no-op and the comment reads as if it were the one doing the work. Trim it, or reword to "belt and braces".

lib/src/scene_runner/rpc_calls/handle_restricted_actions.rs
```diff
@@ -424,3 +432,20 @@ pub fn trigger_scene_emote(
...
         .send_emote(scene_emote_urn.to_godot(), mask);
 }
+
+/// `stopEmote` — end whatever the local player is playing, on behalf of the scene.
+///
+/// Permanent by construction: unlike the scene-boundary suspend, this also drops a
+/// masked emote parked for replay, so walking back into the scene can't resurrect it
+/// (Unity does the same via `masked.EmoteUrn = default` in `TryStopEmote`). The stop
+/// broadcast to other players rides the usual `set_emoting(false)` edge in
+/// CommunicationManager, so there is nothing to send here.
```
regenesis-claw: [P2] This holds only under the `use_pulse` feature — `set_emoting(false)` sends `send_emote_stop()` inside a `#[cfg(feature = "use_pulse")]` block. On the rfc4/LiveKit path the client only ever sends `PlayerEmote { is_stopping: None }` at trigger time and has no stop message, so on a LiveKit-only realm remote viewers keep looping the emote after a scene calls `stopEmote`. Worth qualifying the comment (or filing the gap as a follow-up).

lib/src/scene_runner/rpc_calls/handle_restricted_actions.rs
```diff
@@ -424,3 +432,20 @@ pub fn trigger_scene_emote(
...
+pub fn stop_emote(scene: &Scene, current_parcel_scene_id: &SceneId) {
+    // Same gate as triggerEmote: only the scene the player is standing in may stop it.
+    if !_player_is_inside_scene(scene, current_parcel_scene_id) {
+        tracing::warn!("stopEmote failed: Primary Player is outside the scene");
+        return;
+    }
+
+    get_avatar_node(scene).call("stop_emote_from_scene", &[]);
+}
```
regenesis-claw, line 462: [P2] This stops whatever is current, regardless of owner. Unity's `TryStopEmote` stops the full-body emote globally but only touches **its own scene world's** masked component; and since `_player_is_inside_scene` is unconditionally `true` for `SceneType::Global`, a portable/global experience can drop a masked emote owned by the parcel scene the player is standing in. Consider gating the masked case on `current_emote_scene_id == scene.scene_id` for exact parity.

regenesis-claw (changes requested):
Reviewed against `REVIEW.md`.

**Pre-flight**
- Branch is up to date with `main` (merge commit `353e027`, today). ✅
- iOS: the 📱 Mobile build pipeline comment shows `🍏 iOS ✅ TestFlight build 2020` on `e15026e` (the merge commit adds no code of its own). ✅ The change is GDScript + Rust scene-runner with no platform branch, so an iOS build was not strictly required anyway.
- Description has What / Why / collapsed Details / executable per-case Test plan with a regression block, and the `#ext-foundation` heads-up is called out. ✅
- CI at review time: `📊 Static checks` green; Linux / Clippy / Android still running.

**Consumer impact.** `stopEmote` is purely additive on the SDK surface (new export in `RestrictedActions.js` + new `RpcCall` variant); scenes already guard with `typeof stopEmote === 'function'`, so nothing breaks for scenes on older clients. The behaviour change that *is* visible to existing scenes is the new scene-boundary suspend/resume for masked emotes — correctly flagged in the description as needing a heads-up.

**Security.** No findings. The new op carries no payload, reuses the same `_player_is_inside_scene` gate as `triggerEmote`, and introduces no user-controlled data path.

---

## Findings

### [P1] A resumed masked emote is never re-announced to other players

Suspend → resume is silent locally by design, but it is *not* silent on the wire, and the two halves are asymmetric:

- `_suspend_masked_emote()` sets `playing_masked = false`; the player has just walked across a boundary, so `self_idle` is false too → `avatar.gd::_process` computes `is_emoting = false` → `CommunicationManager::set_emoting(false)` takes the `true→false` edge and sends a Pulse `EmoteStop` (and clears `pending_pulse_emote_urn`).
- `_resume_masked_emote()` calls `set_emoting(true)` again on the next frame, but `pending_pulse_emote_urn` is `None` (consumed at the original start, cleared on the stop edge), so **no `EmoteStart` is sent**.

Net effect: everyone else sees the arms drop when the player steps out of the scene and never sees them come back up, while the local player's arms are raised again. The Unity system this is ported from handles both halves — `TryStopMaskedEmote` calls `messageBus.SendStop()` and `ReplayMaskedEmote` queues a fresh `EmotePendingToBroadcast` with the same urn/mask on every replay.

Fix: re-broadcast on a successful resume (local player only), e.g. `Global.comms.send_emote(current_emote_urn, current_emote_mask)` after `_play_masked_emote` returns true. `send_emote` defers the Pulse `EmoteStart` to the next `set_emoting(true)`, which the resume itself satisfies, so the ordering works out.

### [P2] The "the stop rides `set_emoting(false)`" comment holds only under `use_pulse`

That edge sends `pulse_room.send_emote_stop()` inside `#[cfg(feature = "use_pulse")]`. On the rfc4/LiveKit path the client only ever sends `PlayerEmote { is_stopping: None }` at trigger time and has no stop message at all — so on a LiveKit-only realm a scene's `stopEmote` stops the emote locally while remote viewers keep looping it (i.e. exactly the reported bug, as seen by everyone else). Worth either qualifying the comment or noting the gap as a follow-up; not blocking, since Pulse is the path this targets.

### [P2] The masked stop is not scoped to the calling scene

Unity's `TryStopEmote` stops the full-body emote globally, but sets `StopEmote`/clears `EmoteUrn` only on **its own scene world's** masked component. Here `stop_emote_from_scene()` ends whatever is current, whoever owns it — and `_player_is_inside_scene` returns `true` unconditionally for `SceneType::Global`, so a global/portable experience can drop a masked emote owned by the parcel scene the player is standing in (or a wheel emote). Mostly theoretical today, but gating the masked case on `current_emote_scene_id == scene.scene_id` would match Unity exactly.

### [P2] Redundant reset + slightly misleading comment in `stop_emote()`

`stop_emote()` opens with `_emit_emote_finished(true)`, which already clears `masked_suspended` (together with `current_emote_urn` / `current_emote_mask` / `current_emote_scene_id`), and the two can only ever be set together. The later `masked_suspended = false` plus its three-line comment reads as if it were the line doing the work. Defensive rather than wrong — trim it or reword.

---

The rest holds up on a close read: scene ids are monotonic (`Scene::new_id`), so the owner id cannot be reused by a later scene; suspension leaves `playing_loop` intact and nothing in `process()` touches a parked emote; emote audio rides the `_play_emote_audio` method track baked into the merged clip, so it replays with the resume; and `_deferred_play_emote` now correctly carries `mask`/`owner_scene_id` through the state-machine retry.

regenesis-claw (approved):
Re-reviewed the delta `353e027...9b081b3` (`review fixes`). All four findings are addressed, and the two substantive ones are fixed at the right level rather than patched locally.

**[P1] resume not re-announced — fixed.** `_resume_masked_emote` now calls `Global.comms.send_emote(current_emote_urn, current_emote_mask)` behind an `is_local_player` guard, after a successful replay only. Verified the ordering works: `_play_masked_emote` sets `playing_masked = true` first, so the next `_process` frame reports `is_emoting = true` and flushes the freshly stashed `pending_pulse_emote_urn` as an `EmoteStart`. `send_emote` has no local-playback side effects, and on a Pulse realm `emote_over_livekit` is false so nothing extra goes out on LiveKit. On a LiveKit-only realm the re-send lands as a new `PlayerEmote` with a bumped `incremental_id`, so remote viewers restart the clip from the top on each re-entry — strictly better than the current "no stop, no resume" behaviour there, just worth knowing.

**[P2] stop scoping — fixed, and better scoped than I suggested.** `AvatarEmoteController.stop_emote_from_scene(scene_id) -> bool` refuses only when what's current is masked (playing or parked) *and* owned by a different scene, so the full-body case still stops globally exactly like Unity's `TryStopEmote`; the boolean return lets `avatar.gd` leave the `_pending_network_emote` stash alone when nothing was stopped. The `SceneType::Global` caveat is now documented on both sides.

**[P2] Pulse-only stop comment — fixed.** The doc comment now states the rfc4/LiveKit gap explicitly and marks it pre-existing and out of scope.

**[P2] redundant reset — fixed.** Reworded to "belt and braces" with the invariant it guards.

**CI:** `📊 Static checks` green on `9b081b3`; Linux / Clippy / Android still running at review time. The `build`-label iOS artifact is from `e15026e` — this commit is GDScript + a one-line Rust arg change with no platform branch, so no new iOS build is required, but the TestFlight build QA runs predates the fixes.

Two non-blocking notes:
- The commit also adds three unrelated `.gd.uid` files under `tool/asset_renderer` and the asset-renderer test folder. They are editor-generated metadata for scripts that had none and are harmless to keep, but they are a ride-along the description does not mention.
- The test plan still covers only the single-client cases. The behaviour that changed in this round is what a *second* client sees — a case like "second account watches the first walk out of the scene and back in; the arms drop and come back up on their screen too" would cover the fix QA can actually observe.

A masked emote whose `current_emote_scene_id` is `-1` would now be un-stoppable by any scene, but the only local producer of masked emotes is a scene trigger, which always carries its id — noting it rather than flagging it.

JoanaDN (commented):
Test plan verified and approved

kuruk-mm (approved):
looking great!

PR #2950, opened 2026-09-21
chore(godot): pin the engine to the failed-setup2 Android fix (godotengine#21)

leanmendoza:
## What

The Godot editor and export templates now come from the branch of decentraland/godotengine#21 instead of the current `fix/android-sha1-build-id` pin. With that engine, an Android device that cannot create its rendering context shows the "Unable to initialize … video driver" alert and exits, instead of carrying on with no `DisplayServer` and crashing on the next focus event. The client itself is unchanged — this is one constant in `src/consts.rs`.

## Why

`Callable::is_valid()` on focus-out is ~8% of Android crash users (1.13.1 autopsy, Play cluster `f84e1eca…`; every device in it is a low-end Mali phone — Tecno Spark Go 1, Infinix HOT 50i/60i, Moto g35). The engine fix only reaches a phone through a pin, and this PR's Android CI build is the APK #2925 asks to verify on a Mali device.

It does **not** touch the freed-node GDScript crash family: that half of the engine PR was dropped on 2026-09-15, and the fix for it is architectural (#2925, #2910).

Part of #2925

## Details

<details>
<summary>Expand</summary>

**The pin**: `GODOT_USE_BRANCH = Some("fix/release-gdscript-object-validation-and-android-setup-failure")`. Set it back to `None` and bump `GODOT_BUILD_SHA` to the merge commit once #21 lands in `4.6.2`.

**Symbolication is not lost.** The previous pin was there for the sha1 build-id fix (decentraland/godotengine#20). That one merged into `4.6.2` as `c6be74855`, and #21's branch is `4.6.2` plus a single commit, so it carries the build-id fix too.

**Ride-along:** the pin swaps the engine for *every* platform, not only Android — iOS, macOS, Windows and Linux templates all come from the same branch build. After this merges, the next `cargo run -- install` re-downloads the editor and templates for everyone.

`GODOT_BUILD_SHA` stays at `6289a3b2b`, so `install` / `run` / `export` keep printing the non-fatal "Installed Godot is … expected …" mismatch warning — same as today's pinned-branch state.

- [x] Engine CI on #21 is green on all five platforms including *Process templates*, and the editor build is served under `branches/fix-release-gdscript-object-validation-and-android-setup-failure/`.
- [x] `cargo fmt --check` and `cargo check` pass.

</details>

## Test plan

**Merge gate:** hold until decentraland/godotengine#21 is approved — a pin ships an unmerged engine build to every dev and every CI run.

### Case 1 — Regression: Android on the new engine build
[...]

PR #2954, opened 2026-09-22
Release Candidate 1.14.0b

kuruk-mm:
## Release Candidate 1.14.0b

Re-cut of 1.14.0. #2930 merged into `release` on 2026-09-21 but was never published — three crash-and-build fixes landed on `main` afterwards, so this promotes `main` (`91a0ffb5`) again on top of it. Clean promotion so far: 9 commits, no cherry-picks, nothing excluded.

The three fixes are all the same story: **what a release build does when GDScript touches a freed node.** #2898 flips iOS store builds from the debug engine template to the release one — they had always shipped debug, which quietly turned those calls into logged errors; on release they are a SIGSEGV, exactly as on Android. #2910 removes the UI-side places that make those calls, and #2950 takes the engine off an unmerged branch build onto a merged one that also makes an Android phone with no rendering context exit cleanly instead of crashing on the next focus event.

**The marketing version stays `1.14.0` — deliberately not bumped.** The shipped version string is identical to the first 1.14.0 build, so QA cannot tell the two apart by version alone; the build number and short sha are the only difference. Cargo and `CFBundleShortVersionString` both reject a `1.14.0b` version string, so the `b` lives in the branch and PR name only.

- **Base:** `release` · **Head:** `release-1.14.0b` (= `main` @ `91a0ffb5`)
- 9 commits on top of `release`, from 4 PRs — everything QA'd for 1.14.0 (#2930) carries over unchanged; the list below is only what this RC adds

## What's included

**Fixes**

- **iOS App Store and TestFlight builds are finally release builds** — CI never passed `--release`, so every store binary since the beginning embedded the debug engine template (every iOS Sentry event reports `build_type: debug`). Store builds also stop embedding the dev-only plist keys (`NSLocalNetworkUsageDescription`, `NSAllowsLocalNetworking`) and the `--scene-inspector=ws://…` launch argument, and `MemoryDebugger` and the other `OS.is_debug_build()` gates turn off in production for the first time. **This is the change that exposes the freed-node crashes on iOS** — it was taken as a deliberate trade on 2026-09-15, with the GDScript fixed instead of the engine (#2898)
- **The menu, friend lists and the joypad no longer crash on a freed node** — every UI place that carried a node reference across an `await` now owns the node instead of guarding it: signing out seconds after closing a menu screen, a friends carousel rebuilding while its cards load, preset avatars swiped mid-fetch, a joypad button whose icon was still being fetched. This is the top Android crash family (`GODOT-EXPLORER-21E`, top frame `CanvasItem::show()`), and iOS joins it via #2898. **Behaviour change:** switching menu tabs no longer frees the previous screen after 5 s — it stays hidden until the menu closes, so switching back is instant and memory is released on close (#2910)
- **An Android phone that cannot create its rendering context now exits cleanly** instead of continuing with no `DisplayServer` and crashing on the next focus event — ~8% of Android crash users in the 1.13.1 autopsy (Play cluster `f84e1eca…`, every device a low-end Mali phone: Tecno Spark Go 1, Infinix HOT 50i/60i, Moto g35). It shows the "Unable to initialize … video driver" alert and quits (#2950, decentraland/godotengine#21)
- **The engine is a merged build again.** `release` today pulls an *unmerged* branch build (`fix/vertex-lighting-nan-mali`); this RC pins `4.6.2` stable at `2afb8592d`, which carries that Mali vertex-lighting NaN fix (#23), the Android setup2 fix (#21) and the sha1 build-id fix (#20). Everyone re-downloads the editor and templates once after this merges (#2950)

**Technical** — no QA needed

- Headless `--asset-renderer` CLI mode that renders still PNGs of wearables and emotes for the asset-vision pipeline, plus its CI job. Purely additive (1045 insertions, 0 deletions) and unreachable from the app; the two methods it adds to `avatar_emote_controller.gd` and `avatar_preview.gd` are new, not changed (#2922)

## Test plan

**Build:** `v1.14.0.<build>-<sha>-prod` · one Android + one iPhone.

Everything QA'd for 1.14.0 (#2930) carries over unchanged — the list below is only what this RC adds.

- [ ] **You are on the right build** — the version reads `v1.14.0.…` on both builds, same as before, so check the rest of the string: **Settings → About** must show a **higher build number** than the 1.14.0 build and a short sha that is **not** `54283d6`. If the build number is the same, stop and ask for a rebuild `#2930`

### Case 1 — iOS store build is a release build
**Platform:** iOS
**Steps:**
1. Open the app, sign in, and enter Genesis Plaza.
2. Teleport once to another scene, then open the Backpack.
- [ ] **Expected:** the app starts, loads the scene and opens the Backpack normally — no visible difference from 1.14.0.
- [ ] **Expected (Sentry, same day):** iOS events for this build carry `build_type: release` / `mode: export_release` and an empty `command_line_arguments` `#2898`

### Case 2 — Sign out right after closing a menu screen
**Steps:**
1. Open the app, sign in, open the Backpack from the bottom bar, switch to Discover, then close the menu with the X.
2. Within 3 seconds, open Settings and tap **Sign out**.
3. Sign in again as a guest and open the Backpack.
- [ ] **Expected:** no crash at any point; the Backpack opens normally after signing back in `#2910`

### Case 3 — Menu tabs survive switching and are rebuilt after close
**Steps:**
1. Open the app, sign in, open Discover, switch to Backpack, back to Discover, then Settings, then Backpack.
2. Close the menu with the X, reopen the Backpack from the bottom bar, then rotate the phone twice.
3. Open the Backpack, tap the X and immediately tap the Backpack button again; repeat 5 times.
- [ ] **Expected:** every screen shows its content on every switch (Discover lists, Backpack wearables) with no blank frame, and switching back to a tab you already visited is instant. After reopening, the Backpack loads again normally. No crash `#2910`

### Case 4 — Lists that rebuild while they are still loading
**Setup:** an account with at least 5 friends, some online.
**Steps:**
1. Open Discover; as the Friends carousel fills, pull to refresh twice quickly, then switch to Places and back.
2. Sign out, and on the guest lobby swipe the preset avatar carousel while its thumbnails are still loading.
- [ ] **Expected:** friend cards show name and place, presets show their thumbnails; no duplicate, blank or mismatched cards; no crash `#2910`

### Case 5 — Emote wheel into the Backpack
**Steps:**
1. Enter Genesis Plaza, open the emote wheel and tap **Customize**; repeat 3 times, alternating portrait and landscape.
- [ ] **Expected:** the Backpack opens on the Emotes tab every time, with the joypad icons drawn correctly underneath; no crash `#2910`

### Case 6 — Regression: the new engine build, both platforms
**Steps:**
1. Open the app, sign in and enter Genesis Plaza.
2. Equip a wearable from the Backpack, play an emote from the wheel, and send a chat message.
3. Press Home, return to the app, then swipe it away from Recents and open it again.
- [ ] **Expected:** startup, rendering, audio and backgrounding behave as on the 1.14.0 build; no crash on return or on swipe-away; no new rendering artefacts (this build changes the engine for **every** platform) `#2950`

### Case 7 — Regression: the flows that used to crash on a freed node
**Steps:**
1. Open the app, enter Genesis Plaza, open a friend's passport, then teleport to another scene while the passport is still open.
2. Open the Backpack, equip a wearable, and teleport again while the Backpack is still open.
3. Play an emote from the wheel and teleport mid-animation.
4. Open Settings, sign out, sign back in as a guest, open Settings again.
- [ ] **Expected:** no crash in any of these flows; the app lands in each scene normally `#2898` `#2910`

**Known gap — the Android no-rendering-context alert (#2950) cannot be checked on a normal phone.** It needs a device from the crash cluster (low-end Mali: Tecno Spark Go 1, Infinix HOT 50i/60i, Moto g35) or an emulator set to *Software – GLES 2.0*, on a release build; the expected result there is the "Unable to initialize … video driver" alert, a clean exit, and no native crash dialog when the app is backgrounded or swiped away. If no such device is available, this ships verified only by the engine PR's own testing — watch the Play cluster `f84e1eca…` after release.

**Heads-up — iOS debug tooling stops working on store builds.** The scene-inspector / debug-hub launch argument is no longer injected into App Store and TestFlight iOS builds (#2898), so anyone debugging on-device via TestFlight needs a local dev export instead.


rodrigo685 (approved):
Approved.

Jump In crash can no longer be replicated

rodrigo685 (approved):

PR #2955, opened 2026-09-22
feat(analytics): add utc_offset_minutes to every Segment event

### LINKED ISSUE #2763: [Data] Device Timezone in Analytics Event Payload
##  **⚠️ Problem:**
Timezone can already be calculated from the Firebase install-attribution table — a user's timezone is effectively static, so it's a reliable proxy. This adds capillarity: Firebase is per-install only and doesn't cover Segment events in `MARTS.TRACKS`, where retention cuts still rely on UTC midnight — 9.58% D1 for paid Android vs 4.71% rolling, a 5x artifact of the cut, not of players.

## **🏁 Scope:**
Add device timezone to common Segment event properties in `lib/src/analytics/data_definition.rs`, same mechanism as `client_timestamp` (#2440). Client only — no warehouse model changes, no backfill. Complements (does not replace) the existing Firebase-table-based timezone calculation.

## **📝 Deliverables:**
- [ ] `tz_id` IANA identifier (`America/Argentina/Buenos_Aires`) in common event properties
- [ ] `tz_offset_minutes` signed int, resolved at event creation — covers DST
- [ ] Android + iOS both populate; null when unavailable, no `unknown` sentinel
- [ ] Stamped at queue time alongside `client_timestamp`, not at flush
- [ ] Present on `install_attribution` — fires pre-login, needed for cohort day 0
- [ ] Fields land in Snowflake `MARTS.TRACKS.EXTRA_DATA`, verified

## **🔗 References:**
- [#2440](https://github.com/decentraland/godot-explorer/pull/2440) — added `client_timestamp` as custom property; same insertion point, and states `context` object skipped by design
- [#1782](https://github.com/decentraland/godot-explorer/pull/1782) — added `install_attribution` + Firebase; payload carries only `firebase_user_id`, no geo — this is the existing Firebase-based timezone source referenced above
- [#2664](https://github.com/decentraland/godot-explorer/issues/2664) — Push PRD needs timezone-local delivery + quiet hours; second consumer of the same field
- [#2659](https://github.com/decentraland/godot-explorer/issues/2659) — Retention/Playtime research; its D1 baseline is the UTC-cut number
- No prior art in unity-explorer or docs (searched `analytics timezone`, `segment analytics event`, 2026-08-24)

leanmendoza:
## What

Every Segment event now carries `utc_offset_minutes`: a signed integer, the device's offset from
UTC in minutes, where `local = utc + offset` (`-180` in Buenos Aires, `330` in India, `-210` in
Newfoundland). Resolved once at startup and merged into the common event properties, so no
per-event wiring was needed. No zone name is sent.

## Why

Segment stores every timestamp in UTC and recomputes its own `timestamp`, so the client's offset
never reaches the warehouse and retention can only be cut on UTC midnight. #2763 measured the
cost: D1 for the same cohort reads 24.75% for installs at 23:00 UTC and 4.20% at 01:00 UTC. The
local-day cut becomes `date(timestamp + utc_offset_minutes * interval '1 minute')`.

Nothing a player can see changes, and historical events cannot be backfilled.

Closes #2763

## Details

<details>
<summary>Expand</summary>

- **The field differs from the issue body.** `tz_id` + `tz_offset_minutes` was dropped in
  [popuz's comment](https://github.com/decentraland/godot-explorer/issues/2763#issuecomment-5759590672)
  and [unity-explorer#10168](https://github.com/decentraland/unity-explorer/pull/10168) in favour
  of one integer aligned across the three clients. The launcher's `fp_timezone` stays the only
  zone-name field.

- **Not `Time.get_time_zone_from_system()["bias"]`.** Its sign is right, but the minute
  arithmetic in `OS_Unix::get_time_zone_info()` double-negates the minutes part of negative
  offsets: against the 4.6.2 binary we ship it reports `-90` for Newfoundland (`-150`) and `-510`
  for the Marquesas (`-570`). Newfoundland is one of the zones the minutes-not-hours decision
  exists for, so the value is derived from the OS wall clock instead — read local as if it were
  UTC, diff against real UTC. `chrono::Local` is also unusable: it reads `/etc/localtime`, which
  Android does not maintain.

- **Placement — worth confirming before merge.** unity-explorer#10168's description says the
  trait lives in `context` and lands as `context_utc_offset_minutes`, but its diff writes into the
  same object as `renderer_version` and `operating_system`, which is `properties`. Here `context`
  is skipped by design (#2440), so this can only go in `properties` → `MARTS.TRACKS.EXTRA_DATA`.
  @popuz, which one does Unity ship? If they really differ, the retention cut has to union two
  columns.

Verified locally: `fmt`/`clippy` clean; a desktop run emits `utc_offset_minutes: -180` in
`properties` on every queued event, matching the machine's `-0300`, and `-150` under
`TZ=America/St_Johns`. `Install Attribution` is Android-only so it was not exercised, but it goes
through the same `build_segment_event_batch_item` choke point that merges `common` into
`properties` as every other event.

</details>

## Test plan

No QA needed — no behavior change. Analytics payload only.

**Known gap:** the issue's last deliverable — the field confirmed in Snowflake
`MARTS.TRACKS.EXTRA_DATA` — can't be ticked here. Segment→Snowflake runs ~2h behind and the events
have to come from a shipped build, so it stays open until the first release carrying this lands.


kuruk-mm (approved):
Great!

